PR #10, opened 2018-02-28
reduce the size of the type space

ChrisRackauckas:


AleMorales:
Why not allow passing `value` or `value_type` through the different wrappers around `Variable`? (i.e. `Parameters`, `DependentVariable`, etc)

ChrisRackauckas:
What do you mean?

AleMorales:
At the moment if I want to create a Parameter without deafult value I do `Parameter(:p)` but if I want it to have a default value I need to do `Variable(:p, :Parameter, 1.0, Float64)`. Not a big deal of course, just thinking whether this was on purpose and I am misinterpreting `value` and `value_type`

ChrisRackauckas:
oh yes, those just need to be functions with default values.

Issue #7, opened 2018-02-27
Handle literals


AleMorales:
Arithmetic involving literals (e.g. `2*v1` where `v1` is a `Variable`) does not work. I cannot imagine a convenient DSL without them. But we cannot extend the registered functions to deal with `Real`, so I am not sure at the moment how to handle this (using `Constant` is possible, but all the nice syntax goes away...).

ChrisRackauckas:
Yeah, there must be some way to do automatic promotions here.

AleMorales:
Is there a way to promote only one of the operands? That is, `promote_rule` will promote both arguments, but we only want to promote the literal to a `Constant`. For example, doing this `Base.promote_rule(::Type{<:Expression}, ::Type{<:Integer}) = Constant` does not work :(

ChrisRackauckas:
We can just have the registration macro automatically make `(Number,Expression)` dispatches add `Constant` on the numbers etc.

ChrisRackauckas:
Also, this can be handled in the `@diffeq` macro and things like that if it has to be.

AleMorales:
> We can just have the registration macro automatically make (Number,Expression) dispatches add Constant on the numbers etc.

Yes, I think that would be the best approach for now. And this could handle all the permutations of argument types.

ChrisRackauckas:
https://github.com/JuliaDiffEq/SciCompDSL.jl/pull/12

Issue #2, opened 2018-02-27
Function registration macro


ChrisRackauckas:
A function registration macro should be added that adds dispatches on `Expression` to create an `Operation`.

ChrisRackauckas:
https://github.com/JuliaDiffEq/SciCompDSL.jl/pull/12

PR #12, opened 2018-03-01
Function registration macros

AleMorales:
Add macros to register unary and binary operators/functions and handle literals. 
Registed methods for several operators and mathematical functions.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,1 +1,3 @@
 julia 0.6
+MacroTools
+IterTools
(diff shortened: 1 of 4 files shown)


ChrisRackauckas:
Instead of `@registerN f` for every amount of arguments, I think it should just be `@register f(x,y)` so that way it's a single macro for all of the cases.

AleMorales:
OK, done. One macro to rule them all.

I could not figure how to do it without relying on `MacroTools` and `IterTools`, but I am guessing these dependencies will not cause problems (we would end using `MacroTools` eventually).

The new `@register` macro will create methods for all combinations of `Number` and `Expression` except when they are all `Number`. Some of the combinations may not make sense (e.g. having a `Number` as first argument of `ifelse`), but I guess it is OK to generate the methods.

I think this PR would solve issues #7 and #2.

PR #14, opened 2018-03-01
RFC: add expandable derivative mechanism

ChrisRackauckas:
This adds an expandable derivative mechanism. Basically you overload

```julia
Derivative(::typeof(f),args,idx)
```

and return an `Operation` that corresponds to the derivative of `f` by the `idx` component. Is this nice, or should it be a dispatch on `Val{idx}` to make it easier to define different derivatives for different components?

@simonbyrne @AleMorales

### DIFF
--- a/src/SciCompDSL.jl
+++ b/src/SciCompDSL.jl
@@ -1,16 +1,14 @@
 module SciCompDSL
 
 abstract type AbstractSystem end
+abstract type AbstractOperator end
 
-# Parameterize by T so that way it can be Vector{Expression} which is defined after
-struct Operation{T}
-    op::Function
-    args::Vector{T}
-end
+include("operations.jl")
 include("variables.jl")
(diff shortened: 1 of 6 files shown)


AleMorales:
I am not sure I fully grasp the role of this mechanism in the DSL (second order derivatives?), but if this going to cover binary operators then behavior will be different per `idx` component (at least for  `/` and `^`).

ChrisRackauckas:
>I am not sure I fully grasp the role of this mechanism in the DSL (second order derivatives?)

It's to do symbolic generation of things like Jacobian functions and Hessians to speed up solvers.

I realized that the dispatch is a little funky. It should just give the args and that the derivative variable is.

ChrisRackauckas:
Let's bikeshed this in the issue and keep going.

simonbyrne:
Could do `Derivative{idx}(f, args...)`?

PR #23, opened 2018-03-02
Extract variable from equations

AleMorales:
Algorithms to walk the AST and accumulate `Variable`s of given subtypes.
Automatically extract parameters and different types of variables in `DiffEqSystem` and `NonlinearSystem`


ChrisRackauckas:
Thanks!

Issue #25, opened 2018-03-02
Simplification macros


ChrisRackauckas:
```julia
@DV x y z
@IV t
@Parameter σ,ρ,β
D = Derivative(t)
de = @diffeq begin
  D*x = σ*(y-x)
  D*y = x*(ρ-z)-y
  D*z = x*y - β*z
end
```

would be very nice. Basically it just assumes you want the name to match the variable name, which is very reasonable and probably recommended. It can also allow for having the value placed after the names, so like:

```julia
@DV x 2.0 y 3.0 z 4.0
```

AleMorales:
If you use `=` for the value assignment you could have expressions to calculate complex default values and easily mix variables with and without default values like
```
@DV x = 2.0 y = exp(3) z
```
and the macro will still capture three variables (x,y,z) and nicely associate each value (or expression that evaluates to value) to each name.

ChrisRackauckas:
That looks great! I think it would be nice to also parse the following:

```julia
@DV begin 
  x = 2.0
  y = exp(3)
  z
end
```

PR #30, opened 2018-03-04
Add some macros to make defining variables easier

YingboMa:
One can do stuff like
```julia
julia> @diffs D'''~t E'~t
((IndependentVariable(t),3), (IndependentVariable(t),1))

julia> @cons c=1 d=2
(Constant(1), Constant(2))

julia> @dvars x y z
(DependentVariable(x), DependentVariable(y), DependentVariable(z))

julia> @idvars t g
(IndependentVariable(t), IndependentVariable(g))

julia> @paras p b
(Parameter(p), Parameter(b))
```


ChrisRackauckas:
Just some bikeshedding: `@params` and `@consts`. I don't think `@consts` is necessary since the methods should up-promote constants automatically, but there's no reason to not have it.

src/utils.jl
```diff
@@ -1,5 +1,73 @@
+export @dvars, @idvars, @params, @diffs, @consts
+
 function expr_arr_to_block(exprs)
   block = :(begin end)
   foreach(expr -> push!(block.args, expr), exprs)
   block
 end
+
+# Build variables more easily
```
ChrisRackauckas: Move these to variables.jl

src/utils.jl
```diff
@@ -1,5 +1,73 @@
...
+    end
+    push!(ex.args, Expr(:tuple, lhss...))
+    ex
+end
+
+macro consts(x...)
+    esc(_const_assign(x))
+end
+
+function count_order(x)
```
ChrisRackauckas: move these to operators.jl

ChrisRackauckas:
We might as well make sure that the standard declarations can handle values like the comments: https://github.com/JuliaDiffEq/SciCompDSL.jl/issues/25#issuecomment-369975793 .

YingboMa:
Is it better rename the macros to `@DV`, `@IV`, `@Param`, `@Const` and `@D`?

AleMorales:
Not a big deal, but I would expand `@DV` and `@IV` to `@DVar` and `@IVar` and `@D` to `@Deriv`. Unless we want to really minimize typing, otherwise is confusing if some are very short and others not. And `@D` and `@DV` would be confusing for newcomers.

ChrisRackauckas:
I agree with @AleMorales. The names were what I was using when I proposed the idea in https://github.com/JuliaDiffEq/DifferentialEquations.jl/issues/261 , but the structure was much more thought out than the names. I think you improved them, and then @AleMorales improved it one step more.

AleMorales:
`@consts` can already be assigned values and expressions, you could use the same logic for the other inputs to implement #25

test/runtests.jl
```diff
@@ -1,6 +1,32 @@
 using SciCompDSL
 using Base.Test
 
+@testset "Parsing Test" begin
+    @DVar x y=1 z
+    x1 = DependentVariable(:x)
+    y1 = DependentVariable(:y, 1)
+    z1 = DependentVariable(:z)
+    @test x1 === x
```
ChrisRackauckas: use `isequal` for variable equality.

test/runtests.jl
```diff
@@ -1,6 +1,32 @@
 using SciCompDSL
 using Base.Test
 
+@testset "Parsing Test" begin
+    @DVar x y=1 z
```
AleMorales: Would be nice to use an expression like `z = 1 + div(5,9)` too and not only values in the tests

AleMorales: Or something more sensible than what I wrote there,...

PR #32, opened 2018-03-04
Use `DiffRules.jl` to build pre-defined derivatives

YingboMa:


src/operators.jl
```diff
@@ -45,9 +45,18 @@ function Derivative(O::Operation,idx)
     Derivative(O.op,O.args,Val{idx})
 end
 
-## Pre-defined derivatives
-function Derivative(::typeof(sin),args,::Type{Val{1}})
-    Operation(Base.cos,args)
+# Pre-defined derivatives
+import DiffRules, SpecialFunctions, NaNMath
+for (modu, fun, arity) in DiffRules.diffrules()
+    if arity ==  1
```
ChrisRackauckas: Instead, just loop over the `arity` and get the different diffrules. That's what the `::Type{Val{i}}` is for: defining the derivative by the different arguments, with `1` meaning the derivative by the first.

src/operators.jl
```diff
@@ -48,19 +48,35 @@ end
 # Pre-defined derivatives
 import DiffRules, SpecialFunctions, NaNMath
 for (modu, fun, arity) in DiffRules.diffrules()
-    if arity ==  1
+    if arity ==  1 && !(fun in (:-, :+)) # :+ and :- are both unary and binary operators
```
ChrisRackauckas: `:-`, `:*`, and `:+` should probably get special handling to allow for `n` many arguments. We can probably do this with a generated function. But I'll take this for now.

PR #35, opened 2018-03-05
Fix Internal Test

YingboMa:


test/internal.jl
```diff
@@ -2,14 +2,13 @@ using SciCompDSL
 using Base.Test
 
 # `Expr`, `Number` -> `Operation`
-@test isequal(Operation(2), 2)
-expr = :(-inv(2sqrt(+(a, b))))
+@IVar a b x y
```
ChrisRackauckas: Does it only work for independent variables?

YingboMa: It works for all kinds of variables.

ChrisRackauckas: Should test that then. Does it still work when they have values?

test/internal.jl
```diff
@@ -2,14 +2,15 @@ using SciCompDSL
 using Base.Test
 
 # `Expr`, `Number` -> `Operation`
-@test isequal(Operation(2), 2)
-expr = :(-inv(2sqrt(+(a, b))))
+@IVar a b
```
ChrisRackauckas: Parameters?

PR #40, opened 2018-03-06
WIP: Implement conversion to first-order ODEs

YingboMa:


src/variables.jl
```diff
@@ -52,6 +52,25 @@ function Base.show(io::IO, A::Variable)
     end
 end
 
+extract_idv(eq) = eq.args[1].diff.x
+function extract_symbol_order(eq)
```
ChrisRackauckas: This seems out of place since it's a function on equations instead of variables

YingboMa:
@ChrisRackauckas Ready for code review

src/systems.jl
```diff
@@ -4,14 +4,76 @@ struct DiffEqSystem <: AbstractSystem
     dvs::Vector{Variable}
     vs::Vector{Variable}
     ps::Vector{Variable}
+    function DiffEqSystem(eqs, idvs, dvs, vs, ps; lowered = false, kwargs...)
+        lowered_eqs = lowered ? eqs : ode_order_lowering(eqs; kwargs...)
```
ChrisRackauckas: I wouldn't try stuffing it into here. That will get messy really fast. Instead, just make it a function `new_sys = ode_order_lowering(sys)`

ChrisRackauckas:
Instead of being built into the `DiffEqSystem` call, this should be implemented as a transformation, so it should be a function which takes in a system and spits out a system. That makes the whole transformation system easier to extend.

test/system_construction.jl
```diff
@@ -38,71 +38,10 @@ test_vars_extraction(de, de2)
...
-# Define some variables
-@DVar x y z
-@IVar t
-@Deriv D'~t # Default of first derivative, Derivative(t,1)
-@Param σ ρ β
-@Const c=0
-@Var a
-
-# Define a differential equation
-eqs = [D*x ~ σ*(y-x),
```
ChrisRackauckas: where did all of this go?

src/systems.jl
```diff
@@ -4,18 +4,12 @@ struct DiffEqSystem <: AbstractSystem
     dvs::Vector{Variable}
     vs::Vector{Variable}
     ps::Vector{Variable}
-    function DiffEqSystem(eqs, idvs, dvs, vs, ps)
```
ChrisRackauckas: isn't this constructor missing now?

PR #51, opened 2018-03-17
Add a transformation for ODE order lowering

YingboMa:


test/system_construction.jl
```diff
@@ -69,7 +58,7 @@ function test_eqs(eqs1, eqs2)
     end
     @test_broken eq
```
ChrisRackauckas: the test is still broken

PR #59, opened 2018-03-22
Initial version of sensitivity analysis

Vaibhavdixit02:


ChrisRackauckas:
Sorry if it was unclear. The file should be in this folder: https://github.com/JuliaDiffEq/ModelingToolkit.jl/tree/master/src/systems/diffeqs

src/systems/sensitivityanalysis.jl
```diff
@@ -0,0 +1,28 @@
...
+            deriv_eq = de.eqs[j].args[2]
+            deriv_eq = D*deriv_eq
+            push!(deriv_param,expand_derivatives(deriv_eq))
+        end
+        push!(derivs,deriv_param)
+    end
+    derivs
+end
+
+function calculate_dvarderivs(de::DiffEqSystem)
```
ChrisRackauckas: I don't think this is needed since we already have `calculate_jacobian`?

ChrisRackauckas: https://github.com/JuliaDiffEq/ModelingToolkit.jl/blob/master/src/operators.jl#L107-L109

src/systems/sensitivityanalysis.jl
```diff
@@ -0,0 +1,28 @@
+function calculate_paramderivs(de::DiffEqSystem)
```
ChrisRackauckas: Should probably just create a generic gradient function, and then call that. See https://github.com/JuliaDiffEq/ModelingToolkit.jl/blob/master/src/operators.jl#L107-L109

src/systems/sensitivityanalysis.jl
Vaibhavdixit02: We need to change the dependent variables's `dependents` to our parameter, so just using `calculate_jacobian` will not be enough I think?

src/systems/sensitivityanalysis.jl
Vaibhavdixit02: I didn't quite get this, can you elaborate?

src/systems/sensitivityanalysis.jl
ChrisRackauckas: You get to pass in the `vars` to differentiate by. But for the Jacobian it's the derivative by the dependent variables anyways, so it's `calculate_jacobian(sys,sys.dvs)`. For the parameter Jacobian, it's `calculate_jacobian(sys,sys.ps)` etc.

src/systems/sensitivityanalysis.jl
ChrisRackauckas: you can make a `calculate_gradient` function like `calculate_jacobian` where it takes in equations and a variable, and it returns the derivative of each function by that variable. Or actually, in this case, the gradient vectors are simply the columns of the parameter Jacobian, so using the columns of `calculate_jacobian(sys,sys.ps)` would suffice as well.

src/systems/diffeqs/sensitivityanalysis.jl
```diff
@@ -0,0 +1,28 @@
...
+    derivs
+end
+
+function calculate_dvarderivs(de::DiffEqSystem)
+    derivs = []
+    for i in 1:length(de.ps)
+        deriv_param = []
+        @Deriv D'~de.ps[i]
+        for j in 1:length(de.dvs)
+            l = DependentVariable(de.dvs[j].name,dependents=[de.ps[i]])
```
Vaibhavdixit02: @ChrisRackauckas I understand that using `calculate_jacobian` simplifies the case for parameter gradients for the equation, but what I meant was that this function is required for this change in the `dependents` for the new `DependentVariable`s

ChrisRackauckas: There shouldn't be a change in the dependents. These partials should have `dx/dp = 0` etc.

src/systems/sensitivityanalysis.jl
Vaibhavdixit02: I get that for the parameter Jacobian it is that straightforward, but for the vector of sensitivities we need to re-define the dependent variable to have the parameter as he `dependents` I think?

src/systems/diffeqs/sensitivityanalysis.jl
Vaibhavdixit02: But the point is to create the new differential equation system with respect to the parameters to get the sensitivity, how does making `dx/dp= 0` help?

src/systems/sensitivityanalysis.jl
Vaibhavdixit02:
```function calculate_paramderivs(de::DiffEqSystem)
    eqns = []
    for i in de.eqs
        push!(eqns,i.args[2])
    end
    derivs = calculate_jacobian(eqns, de.ps)
    derivs
end```
You mean this?

File: src/systems/diffeqs/sensitivityanalysis.jl (hunk 3 above)
Comment (ChrisRackauckas) [reply src/systems/diffeqs/sensitivityanalysis.jl]: I see what you're saying, but I think it's much easier to just multiply by the vector of derivatives after the calculation of the Jacobian.

File: test/system_construction.jl [hunk 4]
```diff
@@ -33,6 +35,16 @@ function test_vars_extraction(de, de2)
 end
 test_vars_extraction(de, de2)
 
+#Lotka Volterra test for jacobians
+eqs = [D*x ~ a*x-b*x*y,
+       D*y ~c*x*y-d*y]
+sys = DiffEqSystem(eqs,[t],[x,y],Variable[],[a,b,c,d])
+param_jac = ModelingToolkit.calculate_jacobian(de,de.ps)
+dvars_jac = ModelingToolkit.calculate_jacobian(de,de.dvs)
+@test param_jac == [y - x  Constant(0)  Constant(0); Constant(0)  DependentVariable(x)  Constant(0);Constant(0) Constant(0) -1 * z]
```
ChrisRackauckas: use the multi-line matrix form. That's much easier to read.

src/systems/diffeqs/sensitivityanalysis.jl
```diff
@@ -0,0 +1,13 @@
+calculate_paramderivs(de::DiffEqSystem) = calculate_jacobian(de,de.ps)
+calculate_dvarderivs(de::DiffEqSystem) = calculate_jacobian(de.dvs,de.ps)
+
+function get_sensitivity(de::DiffEqSystem)
+    jac = calculate_jacobian(de)
+    param_jac = calculate_paramderivs(de)
+    dvar_jac = calculate_dvarderivs(de)
+    sensitivity = []
```
ChrisRackauckas: Type this to `Operation`

src/systems/diffeqs/sensitivityanalysis.jl
```diff
@@ -0,0 +1,13 @@
...
+    for i in 1:length(de.ps)
+        push!(sensitivity,jac*dvar_jac[i]+param_jac[i])
```
ChrisRackauckas: This doesn't make sense. You need to make a vector of new variables, and multiply that vector by the Jacobian to get the equations for the derivatives of the new variables.

src/systems/diffeqs/sensitivityanalysis.jl
```diff
@@ -1,11 +1,16 @@
 calculate_paramderivs(de::DiffEqSystem) = calculate_jacobian(de,de.ps)
-calculate_dvarderivs(de::DiffEqSystem) = calculate_jacobian(de.dvs,de.ps)
+function calculate_dvarderivs(de::DiffEqSystem)
+    params = de.ps
+    dvars = de.dvs
+    new_params = [DependentVariable(i.name,dependents=de.ivs) for i in de.ps]
```
ChrisRackauckas: This part isn't right. What's i? And you should never be creating same name variables

Vaibhavdixit02: `i `s are parameters of `de`

ChrisRackauckas: I see what you're doing now. This way doesn't allow you to name the derivatives or add proper context though

Vaibhavdixit02: we can name it once we get the final matrix and give separate names derived from the original variable names or any other name, what do you mean by `context`?

ChrisRackauckas: Context, as in domain, initial value, etc

jessebett:
I have a general performance question: I though `push!` was not performant, and that it's better to preallocate the array and set each index's element. 

i.e. `push!(sensitivity,jac*dvar_jac[i]+param_jac[i])` 
becomes `sensitivity[i]=jac*dvar_jac[i]+param_jac[i])`

Though we'd have to know the size of sensitivity and the types of its elements before hand, though isn't this the case here because we'll know it will have size `length(de.ps)` ?

Alternatively, am I wrong about `push!` being slow or avoidable here?

ChrisRackauckas:
Performance won't be an issue here since the `push!` doesn't exist at runtime.

jessebett:
Right, thanks!

ChrisRackauckas:
not in sync.

PR #60, opened 2018-03-26
Implementation of Leibnitz Transform

ashaywalke:
Leibnitz Transform to differentiate within integration sign. WIP


ChrisRackauckas:
There's a few steps to do here:

1) Define an integration operator. Look at derivatives for details.
2) Define a Leibniz transform function.
3) Write a test that the transform does what's expected.

Then the next steps are to use it

4) Define a transformation on IntegralSystem with a transform to a DiffEqSystem

PR #64, opened 2018-05-21
improved invW and calculate_jacobian

chakravala:
This pull-request introduces a new method called `expr2mk` which helps with converting an `Expr` back into a `ModelingToolkit.Operation` by assigning all the variables from `DiffEqSystem` and evaluating them.

This is then used along with `mat_jacobian` from `ReduceLinAlg` package to calculate the jacobian when the `Reduce` package is also loaded. With the `expr2mk` method this can be safely fed back into the original `DiffEqSystem` in the required format.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -6,4 +6,5 @@ DiffRules
 SpecialFunctions
 NaNMath
 Reduce
+ReduceLinAlg
 Requires
--- a/src/ModelingToolkit.jl
+++ b/src/ModelingToolkit.jl
@@ -9,7 +9,9 @@ import IterTools: product
 
 @require Reduce begin
     using Reduce
+    using ReduceLinAlg
     import Reduce: RExpr
+    Reduce.Rational(false)
     Algebra.operator(:\,:inv,:identity)
     Algebra.rlet(:(identity(~x))=>:x)
     Algebra.rlet(:(inv(~x))=>:(x^(-1)))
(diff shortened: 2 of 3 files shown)


REQUIRE
```diff
@@ -6,4 +6,5 @@ DiffRules
 SpecialFunctions
 NaNMath
 Reduce
```
ChrisRackauckas: Reduce and ReduceLinAlg can be test deps now.

chakravala: I wasn't sure how they should be handled in the REQUIRE file, since they are optional, so should they be removed from this file?

ChrisRackauckas: Yeah, they don't need to be here. But the weird thing is we also want to test with it, so I guess don't make it a test requirement either. You can split Travis environments like https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/.travis.yml and then have one of the environments install the right Reduce packages (though I would like to see a metapackage for it if it ends up being a lot of packages, otherwise maintenance of keeping up with which little packages to test with can be non-trivial)

chakravala: There is already a meta package, but it is not registered in Metadata yet https://github.com/JuliaReducePkg/ReduceAlgebra.jl

ChrisRackauckas:
I'll just merge. The test setup can be put together in a separate PR.

Issue #65, opened 2018-05-22
Warnings thrown by MacroTools


ChrisRackauckas:
This should be a high priority to fix:

```julia
julia> WARNING: Method definition max(ModelingToolkit.Expression, ModelingToolkit.Expression) in mod
ule ModelingToolkit at C:\Users\Chris\.julia\v0.6\MacroTools\src\utils.jl:260 overwritten at C:\User
s\Chris\.julia\v0.6\MacroTools\src\utils.jl:260.
WARNING: Method definition max(Number, ModelingToolkit.Expression) in module ModelingToolkit at C:\U
sers\Chris\.julia\v0.6\MacroTools\src\utils.jl:260 overwritten at C:\Users\Chris\.julia\v0.6\MacroTo
ols\src\utils.jl:260.
WARNING: Method definition max(ModelingToolkit.Expression, Number) in module ModelingToolkit at C:\U
sers\Chris\.julia\v0.6\MacroTools\src\utils.jl:260 overwritten at C:\Users\Chris\.julia\v0.6\MacroTo
ols\src\utils.jl:260.
WARNING: Method definition min(ModelingToolkit.Expression, ModelingToolkit.Expression) in module Mod
elingToolkit at C:\Users\Chris\.julia\v0.6\MacroTools\src\utils.jl:260 overwritten at C:\Users\Chris
\.julia\v0.6\MacroTools\src\utils.jl:260.
WARNING: Method definition min(Number, ModelingToolkit.Expression) in module ModelingToolkit at C:\U
sers\Chris\.julia\v0.6\MacroTools\src\utils.jl:260 overwritten at C:\Users\Chris\.julia\v0.6\MacroTo
ols\src\utils.jl:260.
WARNING: Method definition min(ModelingToolkit.Expression, Number) in module ModelingToolkit at C:\U
sers\Chris\.julia\v0.6\MacroTools\src\utils.jl:260 overwritten at C:\Users\Chris\.julia\v0.6\MacroTo
ols\src\utils.jl:260.
```

PR #67, opened 2018-06-11
Fixed warnings and power rule

### LINKED ISSUE #66: Ambiguity error with power of a variable
```jl
julia> t = IndependentVariable(:t)
IndependentVariable(t)

julia> t^4
ERROR: MethodError: ^(::ModelingToolkit.Variable, ::Int64) is ambiguous. Candidates:
  ^(x::ModelingToolkit.Expression, y::Number) in ModelingToolkit at /Users/dpsanders/.julia/v0.6/MacroTools/src/utils.jl:260
  ^(x::Number, p::Integer) in Base at intfuncs.jl:198
Possible fix, define
  ^(::ModelingToolkit.Expression, ::Integer)
Stacktrace:
 [1] literal_pow(::Base.#^, ::ModelingToolkit.Variable, ::Type{Val{4}}) at ./intfuncs.jl:208
```

chakravala:
resolves #66 and #65

### DIFF
--- a/src/function_registration.jl
+++ b/src/function_registration.jl
@@ -41,7 +41,7 @@ for (M, f, arity) in DiffRules.diffrules()
     @eval @register $sig
 end
 
-for fun = (:<, :>, :(==), :~, :!, :&, :|, :div, :max, :min)
+for fun = (:<, :>, :(==), :~, :!, :&, :|, :div)
     basefun = Expr(:., Base, QuoteNode(fun))
     sig = :($basefun(x,y))
     @eval @register $sig
@@ -50,4 +50,5 @@ end
 # ifelse
 @register Base.ifelse(cond,t,f)
 
-^(a::Expresion,b::T) where T <: Number = NaNMath.pow(a,b)
+# special cases
+Base.:^(x::Expression,y::T) where T <: Integer = Operation(Base.:^, Expression[x, y])
--- a/src/function_registration.jl
+++ b/src/function_registration.jl
@@ -49,3 +49,5 @@ end
 
 # ifelse
 @register Base.ifelse(cond,t,f)
+
+^(a::Expresion,b::T) where T <: Number = NaNMath.pow(a,b)


src/function_registration.jl
```diff
@@ -41,11 +41,14 @@ for (M, f, arity) in DiffRules.diffrules()
...
     basefun = Expr(:., Base, QuoteNode(fun))
     sig = :($basefun(x,y))
     @eval @register $sig
 end
 
 # ifelse
 @register Base.ifelse(cond,t,f)
+
+# special cases
+Base.:^(x::Expression,y::T) where T <: Integer = Operation(Base.:^, Expression[x, y])
```
ChrisRackauckas: oh I see. Yeah, the quick fix is to just special case the operators against literals. Do Integer and AbstractFloat. Are things like `O + 3` and stuff already handled?

src/function_registration.jl
```diff
@@ -41,11 +41,14 @@ for (M, f, arity) in DiffRules.diffrules()
     @eval @register $sig
 end
 
-for fun = (:<, :>, :(==), :~, :!, :&, :|, :div, :max, :min)
```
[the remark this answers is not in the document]
chakravala: the warnings were caused because these are extra definitions not needed. the functionality already works without it being defined here, don't ask me why though, since I havent figured out why that is yet.

src/function_registration.jl
chakravala: What is `O` here? a variable? It is not necessary to make one for `AbstractFloat`, all that is needed is `Integer`, this is a peculiarity with the `Base.:^` method.

ChrisRackauckas: Does `O^2.5` work already? `O` is just some `Expression`.

chakravala: Yes, that already works, the `Integer` method issue is an artifact due to raising matrices to integer powers in `Base`, I believe, so it is only necessary to add a special method for that. All the other extra methods in `Reduce` I needed to create were all for matrices.

src/function_registration.jl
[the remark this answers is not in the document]
chakravala: plus, the way it was defined here, the `max` and `min` method would be defined as Binary operators, when it is actually n-ary, so it wasn't a good definition anyway.

PR #75, opened 2019-01-02
Refactor simplification

HarrisonGrodin:


HarrisonGrodin:
Build fails on v0.7, since `Ref(::AbstractArray)` has a meaning in v1.0 that isn't backwards-compatible. Can we treat v0.7 as an allowed failure, or should I keep compatibility?

src/simplify.jl
```diff
@@ -1,92 +1,73 @@
...
+            O′ = Operation(O′.op, simplify_constants.(O′.args, shorten_tree))
         end
+        O == O′ && return O
+        O = O′
     end
-    _O
 end
 
-const TREE_SHRINK_OPS = [:*,:+]
+const AC_OPERATORS = [*, +]
```
ChrisRackauckas: This should be a tuple so it infers.

PR #76, opened 2019-01-09
[WIP/RFC] Refactor IR around Terms.jl

HarrisonGrodin:
[Terms.jl](https://github.com/HarrisonGrodin/Terms.jl) will mostly replace the IR *structure* (e.g. `Operation`), leaving other data structures (e.g. `Variable`) the same until future PRs.

**TODO**:
- [X] Swap `Operation` for `Term`
- [X] Swap `Constant(x)` for `@term(x)`
- [ ] Update algorithms
- [ ] Update README


HarrisonGrodin:
Closing this PR and bringing back changes in smarter, more targeted PRs. Although it initially seemed that exposing the use of Terms.jl would be ideal, it soon became clear that this is hard to use and develop with.

After making some of the more structural changes that accidentally became part of this, I'll open a PR which only changes the *internals* of `Operation`, ideally without changing the API. This leaves us the benefit of speed and compatibility with other Terms-based packages without requiring direct use of the abstract viewpoint taken in Terms.

PR #80, opened 2019-01-11
Update IR for equations

HarrisonGrodin:


src/function_registration.jl
```diff
@@ -32,7 +32,7 @@ for (M, f, arity) in DiffRules.diffrules()
     @eval @register $sig
 end
 
-for fun = (:<, :>, :(==), :~, :!, :&, :|, :div)
```
ChrisRackauckas: ahh yes, I could see that causing issues.

src/systems/diffeqs/diffeqsystem.jl
```diff
@@ -41,71 +41,64 @@ function generate_ode_function(sys::DiffEqSystem;version = ArrayFunction)
...
+
+    return :($lhs = $(convert(Expr, eq.rhs)))
 end
 
-function calculate_jacobian(sys::DiffEqSystem,simplify=true)
-    diff_idxs = map(eq->eq.args[1].diff !=nothing,sys.eqs)
-    diff_exprs = sys.eqs[diff_idxs]
-    rhs = [eq.args[2] for eq in diff_exprs]
+function calculate_jacobian(sys::DiffEqSystem, simplify=true)
+    calcs, diff_exprs = partition(isintermediate, sys.eqs)
```
ChrisRackauckas: what's this `partition` function?

src/utils.jl
```diff
@@ -33,6 +33,12 @@ end
 
 toexpr(ex) = MacroTools.postwalk(x -> isa(x, Expression) ? convert(Expr, x) : x, ex)
 
+function partition(f, xs)
```
ChrisRackauckas: found it.

src/systems/diffeqs/diffeqsystem.jl
```diff
@@ -144,5 +137,44 @@ function DiffEqBase.ODEFunction(sys::DiffEqSystem;version = ArrayFunction,kwargs
     end
 end
 
+function extract_elements(eqs, targetmap, default = nothing)
```
ChrisRackauckas: why was this moved here?

HarrisonGrodin: At one point, it wouldn't precompile because I was dispatching on `Equation` there before it was defined. This shouldn't be an issue anymore, though, since I switched to `isa`. Would you like me to move it back?

HarrisonGrodin: Ah, good point. Done.

PR #81, opened 2019-01-11
Remove unused Variable information

HarrisonGrodin:
This PR removes variable fields (and related functionalities) which are currently unused, serving as a stepping-stone towards refactoring the Variable IR. Ideally, most/all fields will be reintroduced at a later date, but possibly in a different form/location. Thus, it seems productive to remove the client-side expectation that these attributes will be stored/used until we have the functionalities to back this up and the implementation-side requirement to store the data before it's clear where/how it should be used.


HarrisonGrodin:
TODO: Remove `AbstractOperation` (in favor of `Operation <: Expression`). Depends on #80 (moving `extract_elements!`).

Issue #82, opened 2019-01-11
Segfault in Derivative


jpfairbanks:
This package looks so cool. I was trying to use it based on the readme and came across this error.

It is probably user error, but why does this segfault?

```julia
# This causes a seg fault.
using ModelingToolkit

# Define some variables
@IVar t
@DVar x(t) y(t) z(t)
@Param σ ρ β

f(x,y,z) = z*x+y
Derivative(f, x, 1)
```

```
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.0.2 (2018-11-08)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

julia> include("test/derivatives_fail.jl")
Segmentation fault: 11
```

HarrisonGrodin:
This seems like a combination of user and internal error. Hopefully the internal error is solved in #84.

To get the desired behavior, I believe you should be able to use the following:
```julia
@Deriv D'~x

expand_derivatives(D(f(x,y,z)))
```

PR #84, opened 2019-01-11
Reduce potential for nontermination in Derivative

HarrisonGrodin:
Re: #82.


HarrisonGrodin:
Also, do we want to expose the `Derivative` function, or should it be kept internal (only for method extension)?

ChrisRackauckas:
>Also, do we want to expose the Derivative function, or should it be kept internal (only for method extension)?

It should probably just be kept as an internal detail.

PR #85, opened 2019-01-12
Refactor :DependentVariable into :Unknown

HarrisonGrodin:
Rethinks `DependentVariable` as a generic unknown function (subtype `:Unknown`).

Note that zero-argument `Unknown`s are legal, representing an unknown function with zero arguments (e.g. `x()` in `x() ~ 2x() - σ`). This is commonly written without the explicit function call notation: `@Unknown x` (equivalent to `@Unknown x()`). These are used, for example, in the representation of nonlinear systems.

In the process, the `:Variable` subtype was removed. Existing `@Var` variables (e.g. `@Var a; a ~ y-x`) are replaced with standard Julia variables (`a = y-x`) or zero-argument `@Unknown`s.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ to manipulate.
 ### Example: ODE
 
 Let's build an ODE. First we define some variables. In a differential equation
-system, we need to differentiate between our dependent variables, independent
+system, we need to differentiate between our unknown variables, independent
 variables, and parameters. Therefore we label them as follows:
 
 ```julia
@@ -29,7 +29,7 @@ using ModelingToolkit
 
 # Define some variables
 @IVar t
-@DVar x(t) y(t) z(t)
+@Unknown x(t) y(t) z(t)
 @Deriv D'~t
 @Param σ ρ β
 ```
@@ -87,12 +87,12 @@ f = ODEFunction(de)
 
 We can also build nonlinear systems. Let's say we wanted to solve for the steady
 state of the previous ODE. This is the nonlinear system defined by where the
-derivatives are zero. We could use dependent variables for our nonlinear system
+derivatives are zero. We could use unknown variables for our nonlinear system
 (for direct compatibility with the above ODE code), or we can use non-tagged
 variables. Here we will show the latter. We write:
 
 ```julia
-@DVar x y z
+@Unknown x y z
 @Param σ ρ β
 
 # Define a nonlinear system
@@ -191,7 +191,7 @@ structure is as follows:
   the system of equations.
 - Name to subtype mappings: these describe how variable `subtype`s are mapped
   to the contexts of the system. For example, for a differential equation,
-  the dependent variable corresponds to given subtypes and then the `eqs` can
+  the unknown variable corresponds to given subtypes and then the `eqs` can
   be analyzed knowing what the state variables are.
 - Variable names which do not fall into one of the system's core subtypes are
   treated as intermediates which can be used for holding subcalculations and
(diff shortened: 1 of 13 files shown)


src/systems/diffeqs/diffeqsystem.jl
```diff
@@ -96,12 +93,12 @@ function generate_ode_jacobian(sys::DiffEqSystem, simplify=true)
...
 function generate_ode_iW(sys::DiffEqSystem, simplify=true)
-    var_exprs = [:($(sys.dvs[i].name) = u[$i]) for i in 1:length(sys.dvs)]
-    param_exprs = [:($(sys.ps[i].name) = p[$i]) for i in 1:length(sys.ps)]
+    var_exprs = [:($(sys.dvs[i].name) = u[$i]) for i in eachindex(sys.dvs)]
+    param_exprs = [:($(sys.ps[i].name) = p[$i]) for i in eachindex(sys.ps)]
     diff_exprs = filter(!isintermediate, sys.eqs)
     jac = sys.jac
 
-    gam = Variable(:gam)
+    gam = Unknown(:gam)
```
HarrisonGrodin: Does this seem correct?

README.md
```diff
@@ -21,15 +21,15 @@ to manipulate.
 ### Example: ODE
 
 Let's build an ODE. First we define some variables. In a differential equation
-system, we need to differentiate between our dependent variables, independent
+system, we need to differentiate between our unknown variables, independent
```
ChrisRackauckas: put (dependent) afterwards, since they are never really referred to as the "unknown variables". I guess this is more general since "unknowns" apply to non-diffeq systems.

HarrisonGrodin: :+1: - adding now. (Since nonlinear systems now use zero-argument unknowns (see edited initial comment), this seemed like a natural terminology transition.)

src/equations.jl
```diff
@@ -12,40 +12,36 @@ Base.:~(lhs::Expression, rhs::Number    ) = Equation(lhs, rhs)
...
-        targets = [unique(collect(values(targetmap))), default]
-    end
-    for target in targets
-        elems[target] = Vector{Variable}()
-        names[target] = Set{Symbol}()
-    end
-    for eq in eqs
-        extract_elements!(eq, elems, names, targetmap, default)
+_is_derivative(x::Variable) = x.diff !== nothing
+_is_dependent(x::Variable) = x.subtype === :Unknown && !isempty(x.dependents)
```
ChrisRackauckas: Why would it have to be an unknown to be a dependent variable? This is now disambiguate that there are dependent variables which are known. For example, `w(t) = 2t` is a dependent variable that is known before ever integrating whatever ODE it might show up in, so then it would be some time-dependent parameter (exactly what we need for PuMaS!)

test/system_construction.jl
```diff
@@ -62,11 +61,11 @@ end
 @test_broken test_eqs(de1.eqs, lowered_eqs)
```
ChrisRackauckas: did this get fixed?

src/equations.jl
HarrisonGrodin: The `_is_dependent` function is only used in extraction, taking the place of the `targetmap` approach which looked for the previously-named `:DependentVariable`. Perhaps there's a better name for this, though. (To comment on your specific example, I imagined this being represented as `w = 2t`, defining `w` precisely as a programmatic variable equivalent to `2t`. )

test/system_construction.jl
HarrisonGrodin: I don't believe so. It's something I could work on in an upcoming PR, thogh.

ChrisRackauckas: Okay, no worries. I just thought you mentioned the `subtype` bug fixed it. Since it is a transform that exists though, it would be good to get it working and keep it working as an example of the kinds of calculations we want to do on the IR.

PR #87, opened 2019-01-13
Remove diff field from Variable

HarrisonGrodin:
TODO:
- [X] Remove `.diff` field
- [x] Fix first-order transform
- [x] Update README

### DIFF
--- a/src/systems/diffeqs/diffeqsystem.jl
+++ b/src/systems/diffeqs/diffeqsystem.jl
@@ -45,16 +45,9 @@ end
 isintermediate(eq::Equation) = !(isa(eq.lhs, Operation) && isa(eq.lhs.op, Differential))
 
 function build_equals_expr(eq::Equation)
-    @assert !isa(eq.lhs, Constant)
+    @assert !isintermediate(eq)
 
-    if isintermediate(eq)
-        @assert isa(eq.lhs, Variable)
-        lhs = eq.lhs.name
-    else
-        @assert isa(eq.lhs, Operation) && isa(eq.lhs.op, Differential)
-        lhs = Symbol(eq.lhs.args[1].name, :_, eq.lhs.op.x.name)
-    end
-
+    lhs = Symbol(eq.lhs.args[1].name, :_, eq.lhs.op.x.name)
     return :($lhs = $(convert(Expr, eq.rhs)))
 end
 
--- a/src/systems/diffeqs/diffeqsystem.jl
+++ b/src/systems/diffeqs/diffeqsystem.jl
@@ -4,6 +4,12 @@ mutable struct DiffEqSystem <: AbstractSystem
     dvs::Vector{Variable}
     ps::Vector{Variable}
     jac::Matrix{Expression}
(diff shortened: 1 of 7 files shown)


src/systems/diffeqs/diffeqsystem.jl
```diff
@@ -44,13 +42,18 @@ function generate_ode_function(sys::DiffEqSystem;version = ArrayFunction)
...
+isintermediate(eq::Equation) = !(isa(eq.lhs, Operation) && isa(eq.lhs.op, Differential))
 
 function build_equals_expr(eq::Equation)
-    @assert typeof(eq.lhs) <: Variable
-
-    lhs = eq.lhs.name
-    isintermediate(eq) || (lhs = Symbol(lhs, :_, "$(eq.lhs.diff.x.name)"))
+    @assert !isa(eq.lhs, Constant)
+
+    if isintermediate(eq)
```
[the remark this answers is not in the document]
HarrisonGrodin: Ah, good point! I guess not.

test/system_construction.jl
```diff
@@ -57,7 +57,7 @@ function test_eqs(eqs1, eqs2)
     end
     eq
 end
-@test_broken test_eqs(de1.eqs, lowered_eqs)
+@test test_eqs(de1.eqs, lowered_eqs)
```
HarrisonGrodin: :tada:

src/systems/diffeqs/diffeqsystem.jl
```diff
@@ -4,24 +4,31 @@ mutable struct DiffEqSystem <: AbstractSystem
     dvs::Vector{Variable}
     ps::Vector{Variable}
     jac::Matrix{Expression}
+    function DiffEqSystem(eqs, ivs, dvs, ps, jac)
+        all(!isintermediate, eqs) ||
```
ChrisRackauckas: hmm, if we don't allow intermediates, are we putting a lot of pressure on CSE optimizations? Will we be boosting compile times?

src/variables.jl
```diff
@@ -1,22 +1,15 @@
-mutable struct Variable <: Expression
+struct Variable <: Expression
```
ChrisRackauckas: so Variables are uniquely identified by name? We need a good naming system to make sure things like first order transform don't overlap.

src/systems/diffeqs/diffeqsystem.jl
HarrisonGrodin: Not sure. I'm very open to changing this, but for now it seems like it makes correctness easier to guarantee, since we can assume the format of each equation in analysis/transformation.

src/variables.jl
HarrisonGrodin: Due to the definition of `Base.:(==)(::Variable, ::Variable)` we already have, we don't discern between variables other than name/subtype equality anyways. For now, this removes an unnecessary source of mutability (and in some sense, feels more honest/compatible with `==`). In the near future, I'm hoping to put together a more clever naming scheme for variables (with the need to generate new names in places like first-order transform in mind).

PR #90, opened 2019-01-24
Clean up derivatives

HarrisonGrodin:
- Remove `Derivative` export, securing it as an internal function
- Fix differentiation on non-`Operation`


HarrisonGrodin:
Build failure seems unrelated (due to Coveralls submission issue).

PR #92, opened 2019-01-31
[WIP] Refactor system storage

HarrisonGrodin:


HarrisonGrodin:
Build failure unrelated (re: JuliaWeb/MbedTLS.jl#193).

README.md
```diff
@@ -46,7 +46,7 @@ Each operation builds an `Operation` type, and thus `eqs` is an array of
 analyzed by other programs. We can turn this into a `DiffEqSystem` via:
 
 ```julia
-de = DiffEqSystem(eqs,[t],[x,y,z],[σ,ρ,β])
+de = DiffEqSystem(eqs,t,[x,y,z],[σ,ρ,β])
```
ChrisRackauckas: Rename this ODE system, since something like a PDE system would have two independent variables.

ChrisRackauckas:
We can always rename later. I'm not convinced that splitting out ODEs to their own form is actually going to not lead to a ton of code duplication, but for now it's fine.

PR #97, opened 2019-02-03
[WIP] General cleanup

HarrisonGrodin:


src/equations.jl
```diff
@@ -5,17 +5,17 @@ struct Equation
...
 Base.:~(lhs::Expression, rhs::Expression) = Equation(lhs, rhs)
 Base.:~(lhs::Expression, rhs::Number    ) = Equation(lhs, rhs)
 Base.:~(lhs::Number    , rhs::Expression) = Equation(lhs, rhs)
 
 
-_is_dependent(x::Variable) = x.subtype === :Unknown && !isempty(x.dependents)
-_is_parameter(iv) = x -> x.subtype === :Parameter && x ≠ iv
-_subtype(subtype::Symbol) = x -> x.subtype === subtype
+_is_dependent(x::Variable) = !x.known && !isempty(x.dependents)
+_is_parameter(iv) = x -> x.known && x ≠ iv
```
ChrisRackauckas: why return a function here?

HarrisonGrodin: `_is_parameter` is a higher-order function used in system information extraction:
https://github.com/JuliaDiffEq/ModelingToolkit.jl/blob/7dcd418ccd5f64372bba7b0be30d712b32eef676/src/systems/diffeqs/diffeqsystem.jl#L38

Once we know the independent variable `iv`, we find parameters by checking that the parameter is both known and not the independent variable. This function is generated by `_is_parameter(iv)`.

HarrisonGrodin:
Ready to merge.

PR #103, opened 2019-02-14
Minor cleanup

### LINKED ISSUE #99: Use "variable" instead of "unknown"
I suggest to use the word "variable" instead of "unknown".

### LINKED ISSUE #100: Don't use capitals for macros
Is there a particular reason for `@Param` instead of `@parameter`?

### LINKED ISSUE #101: Parsing of `<=`
Instead of
```jl
julia> @Unknown x y
(x, y)

julia> ex = x^2 + y^2 <= 1
(x ^ 2 + y ^ 2 < 1) | false

julia> dump(ex)
Operation
  op: | (function of type typeof(|))
  args: Array{Expression}((2,)) Expression[x ^ 2 + y ^ 2 < 1, Constant(false)]
```
could this be parsed as something like 
```jl
Condition(<, x^2 + y^2, 1)
```
with a new `Condition` type?

HarrisonGrodin:
Closes #99, closes #100, closes #101.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,9 +28,9 @@ and parameters. Therefore we label them as follows:
 using ModelingToolkit
 
 # Define some variables
-@param t σ ρ β
-@variable x(t) y(t) z(t)
-@deriv D'~t
+@parameters t σ ρ β
+@variables x(t) y(t) z(t)
+@derivatives D'~t
 ```
 
 Then we build the system:
@@ -81,8 +81,8 @@ state of the previous ODE. This is the nonlinear system defined by where the
 derivatives are zero. We use (unknown) variables for our nonlinear system.
 
 ```julia
-@variable x y z
-@param σ ρ β
+@variables x y z
+@parameters σ ρ β
 
 # Define a nonlinear system
 eqs = [0 ~ σ*(y-x),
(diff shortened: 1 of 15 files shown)


HarrisonGrodin:
It seems like we should be somewhat consistent with how we shorten macro names. Some ideas:

```julia
# Option A
@param t σ
@var x(t) y(t)
@deriv D'~t
```
```julia
# Option B
@parameter t σ
@variable x(t) y(t)
@derivative D'~t D''~t
```

We could also pluralize. Thoughts? @ChrisRackauckas @dpsanders

ChrisRackauckas:
I think Option A is fine, but I think @dpsanders likes the full names?

HarrisonGrodin:
Whoa, this has a version-based bug - works on 1.1/nightly, but not on 1.0.

HarrisonGrodin:
The more I look at the options, I like the full names (Option B) - since these lines will only be typed a few times at the beginning of each model, it seems reasonable to make the names slightly more verbose/clear. I also think pluralizing will help with this goal.

```julia
@parameters t σ
@variables x(t) y(t)
@derivatives D'~t D''~t
```

@ChrisRackauckas Am I good to change this?

ChrisRackauckas:
Yeah that looks good.

HarrisonGrodin:
Update: the reason 1.1 works (while 1.0 fails) seems to be due to JuliaLang/julia#29668. This does beg the question of whether or not we should have comparison functions return non-booleans (outside of an explicit context), though.

ChrisRackauckas:
What booleans would they return?

PR #111, opened 2019-03-25
Make <= and >= behave like < and >

dpsanders:


ChrisRackauckas:
Is there a reason to not do this?

dpsanders:
I guess this is a bad idea if the variables are discrete, but a good idea if they're continuous.

ChrisRackauckas:
Why would it be a bad idea for discrete? That's not a case I was thinking of. I was thinking of how it currently splits it into a larger boolean. I wonder if that is useful, or just a distraction and adding this simplifies a bit.

dpsanders:
I was thinking that for discrete it's wrong to conflate <= and < as makes sense for continuous.

But that's orthogonal to this so I don't see any reason not to do this.

HarrisonGrodin:
The potential argument against this is that it allows for two distinct representations of the same notion (i.e. `a <= b` and `a < b | a == b`), which could make condition manipulation more complex.

For example, simplifying `(a < b | a == b) & (a > b | a == b)` to `a == b` can be based around fairly fundamental axioms, but dealing with additional combinations involving explicit `<=`/`>=` operators makes this process prone to edge cases.

Since it's not clear at this point whether or not this will pose an issue in the future, it might be worth erring on the side of caution and keeping the behavior as it is now. I'm happy to move forward with this PR, though, but ideally under the assumption that the behavior would only be internally visible (i.e. leaving it as an implementation detail that client libraries don't rely on).

ChrisRackauckas:
This should move to Symbolics.jl, if at all.

PR #117, opened 2019-03-31
Add the matching (`augment_path`) for the Pantelides algorithm

JunpengGao233:
https://epubs.siam.org/doi/abs/10.1137/0909014 

I have implemented the augment path algorithm in the paper.  I will open follow-up PRs to complete the entire Pantelides algorithm.


ChrisRackauckas:
What's the plan for integrating this with the MT expression trees? I don't quite see it yet.

PR #118, opened 2019-04-01
`generate_ode_iW` -> `generate_factorized_W`

YingboMa:
The generated factorized W matrix is in the bare factors form. So, one can use it by

```julia
julia> factors = zeros(3, 3); F = LU(factors, collect(axes(factors, 1)), 0);

julia> fw(factors, u, p, 1, 1); F
LU{Float64,Array{Float64,2}}
L factor:
3×3 Array{Float64,2}:
  1.0   0.0  0.0
 -0.4   1.0  0.0
 -0.4  -6.5  1.0
U factor:
3×3 Array{Float64,2}:
 5.0  -4.0   0.0
 0.0   0.4   1.0
 0.0   0.0  13.5
```

### DIFF
--- a/src/systems/diffeqs/diffeqsystem.jl
+++ b/src/systems/diffeqs/diffeqsystem.jl
@@ -76,31 +76,29 @@ function generate_function(sys::DiffEqSystem; version::FunctionVersion = ArrayFu
 end
 
 
-function generate_ode_iW(sys::DiffEqSystem, simplify=true; version::FunctionVersion = ArrayFunction)
+function generate_factorized_W(sys::DiffEqSystem, simplify=true; version::FunctionVersion = ArrayFunction)
     jac = calculate_jacobian(sys)
 
     gam = Variable(:gam; known = true)
 
     W = LinearAlgebra.I - gam*jac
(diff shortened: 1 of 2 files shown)


src/systems/diffeqs/diffeqsystem.jl
```diff
@@ -76,22 +76,20 @@ function generate_function(sys::DiffEqSystem; version::FunctionVersion = ArrayFu
...
-function generate_ode_iW(sys::DiffEqSystem, simplify=true; version::FunctionVersion = ArrayFunction)
+function generate_factorized_W(sys::DiffEqSystem, simplify=true; version::FunctionVersion = ArrayFunction)
     jac = calculate_jacobian(sys)
 
     gam = Variable(:gam; known = true)
 
     W = LinearAlgebra.I - gam*jac
-    W = SMatrix{size(W,1),size(W,2)}(W)
-    iW = inv(W)
+    iW = lu(W, Val(false), check=false).factors
```
ChrisRackauckas: if it's not iW we should change the name. Wfact?

PR #119, opened 2019-04-07
[WIP] Refactor variables as functions

HarrisonGrodin:
Allows for full TVCOV. :tada:

### DIFF
--- a/src/differentials.jl
+++ b/src/differentials.jl
@@ -4,27 +4,27 @@ export Differential, expand_derivatives, @derivatives
 struct Differential <: Function
     x::Expression
 end
+(D::Differential)(x) = Operation(D, Expression[x])
 
 Base.show(io::IO, D::Differential) = print(io, "(D'~", D.x, ")")
 Base.convert(::Type{Expr}, D::Differential) = D
 
-(D::Differential)(x::Operation) = Operation(D, Expression[x])
-function (D::Differential)(x::Variable)
-    D.x === x             && return Constant(1)
-    has_dependent(x, D.x) || return Constant(0)
-    return Operation(D, Expression[x])
-end
-(::Differential)(::Any) = Constant(0)
 Base.:(==)(D1::Differential, D2::Differential) = isequal(D1.x, D2.x)
 
 function expand_derivatives(O::Operation)
     @. O.args = expand_derivatives(O.args)
 
-    if O.op isa Differential
-        D = O.op
-        o = O.args[1]
-        isa(o, Operation) || return O
-        return simplify_constants(sum(i->derivative(o,i)*expand_derivatives(D(o.args[i])),1:length(o.args)))
+    if isa(O.op, Differential)
+        (D, o) = (O.op, O.args[1])
+
+        isequal(o, D.x)     && return Constant(1)
+        occursin(D.x, o)    || return Constant(0)
+        isa(o, Operation)   || return O
+        isa(o.op, Variable) && return O
+
+        return sum(1:length(o.args)) do i
+            derivative(o, i) * expand_derivatives(D(o.args[i]))
+        end |> simplify_constants
     end
 
     return O
@@ -79,6 +79,6 @@ macro derivatives(x...)
     esc(_differential_macro(x))
 end
 
-function calculate_jacobian(eqs,vars)
-    Expression[Differential(vars[j])(eqs[i]) for i in 1:length(eqs), j in 1:length(vars)]
+function calculate_jacobian(eqs, dvs, iv)
+    Expression[Differential(dv(iv()))(eq) for eq ∈ eqs, dv ∈ dvs]
 end
--- a/src/systems/diffeqs/diffeqsystem.jl
+++ b/src/systems/diffeqs/diffeqsystem.jl
@@ -16,18 +16,20 @@ end
 
 
 struct DiffEq  # dⁿx/dtⁿ = rhs
-    x::Expression
-    t::Variable
+    x::Variable
     n::Int
     rhs::Expression
 end
-function Base.convert(::Type{DiffEq}, eq::Equation)
+function to_diffeq(eq::Equation)
     isintermediate(eq) && throw(ArgumentError("intermediate equation received"))
     (x, t, n) = flatten_differential(eq.lhs)
-    return DiffEq(x, t, n, eq.rhs)
+    (isa(t, Operation) && isa(t.op, Variable) && isempty(t.args)) ||
+        throw(ArgumentError("invalid independent variable $t"))
+    (isa(x, Operation) && isa(x.op, Variable) && length(x.args) == 1 && isequal(first(x.args), t)) ||
+        throw(ArgumentError("invalid dependent variable $x"))
+    return t.op, DiffEq(x.op, n, eq.rhs)
 end
 Base.:(==)(a::DiffEq, b::DiffEq) = isequal((a.x, a.t, a.n, a.rhs), (b.x, b.t, b.n, b.rhs))
-get_args(eq::DiffEq) = Expression[eq.x, eq.t, eq.rhs]
 
 struct DiffEqSystem <: AbstractSystem
     eqs::Vector{DiffEq}
@@ -35,32 +37,32 @@ struct DiffEqSystem <: AbstractSystem
     dvs::Vector{Variable}
     ps::Vector{Variable}
     jac::RefValue{Matrix{Expression}}
-    function DiffEqSystem(eqs, iv, dvs, ps)
+    function DiffEqSystem(eqs)
+        reformatted = to_diffeq.(eqs)
+
+        ivs = unique(r[1] for r ∈ reformatted)
+        length(ivs) == 1 || throw(ArgumentError("one independent variable currently supported"))
+        iv = first(ivs)
+
+        deqs = [r[2] for r ∈ reformatted]
+
+        dvs = [deq.x for deq ∈ deqs]
+        ps = filter(vars(deq.rhs for deq ∈ deqs)) do x
+            x.known & !isequal(x, iv)
+        end |> collect
+
         jac = RefValue(Matrix{Expression}(undef, 0, 0))
-        new(eqs, iv, dvs, ps, jac)
+
+        new(deqs, iv, dvs, ps, jac)
     end
-end
-
-function DiffEqSystem(eqs)
-    dvs, = extract_elements(eqs, [_is_dependent])
-    ivs = unique(vcat((dv.dependents for dv ∈ dvs)...))
-    length(ivs) == 1 || throw(ArgumentError("one independent variable currently supported"))
-    iv = first(ivs)
-    ps, = extract_elements(eqs, [_is_parameter(iv)])
-    DiffEqSystem(eqs, iv, dvs, ps)
-end
-
-function DiffEqSystem(eqs, iv)
-    dvs, ps = extract_elements(eqs, [_is_dependent, _is_parameter(iv)])
-    DiffEqSystem(eqs, iv, dvs, ps)
 end
 
 
 function calculate_jacobian(sys::DiffEqSystem)
     isempty(sys.jac[]) || return sys.jac[]  # use cached Jacobian, if possible
-    rhs = [eq.rhs for eq in sys.eqs]
+    rhs = [eq.rhs for eq ∈ sys.eqs]
 
-    jac = expand_derivatives.(calculate_jacobian(rhs, sys.dvs))
+    jac = expand_derivatives.(calculate_jacobian(rhs, sys.dvs, sys.iv))
     sys.jac[] = jac  # cache Jacobian
     return jac
 end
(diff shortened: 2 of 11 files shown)


src/systems/diffeqs/diffeqsystem.jl
```diff
@@ -16,51 +16,53 @@ end
 
 
 struct DiffEq  # dⁿx/dtⁿ = rhs
-    x::Expression
-    t::Variable
+    x::Variable
```
HarrisonGrodin: This makes invalid states unrepresentable, leading to less bugs/confusion. In `to_diffeq`, we enforce the necessary invariants about each `DiffEq` and store the independent variable `t` externally (in the `DiffEqSystem`), guaranteeing that all `t`s are equal after construction.

src/systems/diffeqs/diffeqsystem.jl
```diff
@@ -16,51 +16,53 @@ end
...
     iv::Variable
     dvs::Vector{Variable}
     ps::Vector{Variable}
     jac::RefValue{Matrix{Expression}}
-    function DiffEqSystem(eqs, iv, dvs, ps)
-        jac = RefValue(Matrix{Expression}(undef, 0, 0))
-        new(eqs, iv, dvs, ps, jac)
-    end
-end
+    function DiffEqSystem(eqs)
```
HarrisonGrodin: Since the equations provide enough information to create a valid `DiffEqSystem`, it seemed unnecessary to allow manual construction (which is prone to contradictions). Also, by making use of the structure of a system of ODEs, the complexity involving `extract_elements` was removed, instead getting the variables from the LHS and the parameters from a `filter` on `vars` of RHS.

However, this has the unintended side effect of breaking tests involving `generate_function(::DiffEqSystem)`, since we as the user don't know which order the dependent variables and parameters are stored in. I'd like to maintain this level of abstraction if possible, but I'm not familiar enough with the solvers generated functions are fed into to be convinced that this isn't an issue. @ChrisRackauckas?

src/systems/diffeqs/diffeqsystem.jl
```diff
@@ -70,16 +72,30 @@ function generate_jacobian(sys::DiffEqSystem; version::FunctionVersion = ArrayFu
     return build_function(jac, sys.dvs, sys.ps, (sys.iv.name,); version = version)
 end
 
+struct DiffEqToExpr
```
HarrisonGrodin: This is a somewhat odd hack, in an attempt to remove the requirement that all parameters be given as functions (e.g. want `2`, not `t -> 2`, in `ps` vector). We pass a custom conversion function to `build_function` represents variables and 0-ary parameters as variables in the `Expr` tree.

It may be worthwhile to modify `convert(::Type{Expr}, ::Operation)` to account for the unknowns, although this isn't yet clear. I'm not sure if there's a cleaner way of dealing with 0-ary parameters, though, other than requiring that parameters be functions of the independent variable.

src/variables.jl
```diff
@@ -1,13 +1,12 @@
 export Variable, @variables, @parameters
 
 
-struct Variable <: Expression
+struct Variable <: Function
```
HarrisonGrodin: The primary change behind this PR: variables are now *functions* (tree nodes), not *expressions* (tree leaves).

test/derivatives.jl
```diff
@@ -2,9 +2,9 @@ using ModelingToolkit
 using Test
 
 # Derivatives
-@parameters t σ ρ β
+@parameters t() σ() ρ() β()
```
HarrisonGrodin: Variables/parameters are parsed differently based on whether or not they appear as calls. Calls (such as these) are parsed as having implicit arguments, whereas symbols are parsed as callable functions.

```julia
t = Variable(:t; known=true)()
u = Variable(:u; known=true)

t + u(t - 1)
```

test/system_construction.jl
```diff
@@ -2,49 +2,60 @@ using ModelingToolkit
...
+@parameters t() σ() ρ() β()
 @variables x(t) y(t) z(t)
 @derivatives D'~t
 
+function test_diffeq_inference(name, de, iv, dvs, ps)
+    @testset "DiffEqSystem construction: $name" begin
+        @test de.iv.name == :t
+        @test Set([dv.name for dv ∈ de.dvs]) == Set(dvs)
+        @test Set([p.name  for p  ∈ de.ps ]) == Set(ps)
+    end
```
HarrisonGrodin: It's unclear if we should be testing by field names, since that could be considered violating the layer of abstraction around `DiffEqSystem`.

test/system_construction.jl
```diff
@@ -2,49 +2,60 @@ using ModelingToolkit
...
+    test_diffeq_inference("global iv-varying", de, :t, (:x, :y, :z), (:σ′, :ρ, :β))
     @test begin
         f = eval(generate_function(de))
         du = [0.0,0.0,0.0]
         f(du, [1.0,2.0,3.0], [x->x+7,2,3], 5.0)
         du ≈ [12, -3, -7]
     end
+
+    @parameters σ
+    eqs = [D(x) ~ σ(t-1)*(y-x),
```
HarrisonGrodin: This is now representable! :tada:

src/systems/diffeqs/diffeqsystem.jl
HarrisonGrodin: Upon further thought, I think we could just make `generate_function` take in vectors of dependent variables/parameters, generating functions which handle the given shape of I/O.
```julia
f = eval(generate_function(de, [x,y,z], [σ,ρ,β]))
```

I like how this pushes the variable ordering to the only part of the system that needs it. Does this seem reasonable?

HarrisonGrodin: Update: this has been implemented, at least temporarily. The primary issue it causes is the inability to directly convert to an `ODEFunction` without the additional information about ordering.

Since all variables are required anyways, maybe it would be worthwhile to provide accessor functions which show the user stored variables, in the order that they will be assumed to be within the vector.

src/differentials.jl
```diff
@@ -79,6 +79,6 @@ macro derivatives(x...)
     esc(_differential_macro(x))
 end
 
-function calculate_jacobian(eqs,vars)
-    Expression[Differential(vars[j])(eqs[i]) for i in 1:length(eqs), j in 1:length(vars)]
+function calculate_jacobian(eqs, dvs, iv)
```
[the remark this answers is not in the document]
ChrisRackauckas: you shouldn't need the independent variable to calculate the Jacobian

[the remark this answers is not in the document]
ChrisRackauckas: you don't need to know the independent variable to calculate the Jacobian

src/systems/diffeqs/diffeqsystem.jl
ChrisRackauckas: the LHS isn't necessarily a single variable. If we want to enforce this, we should be strict with the naming and call it a mass-matrix free ODE.

src/systems/diffeqs/diffeqsystem.jl
```diff
@@ -16,51 +16,53 @@ end
 
 
 struct DiffEq  # dⁿx/dtⁿ = rhs
-    x::Expression
-    t::Variable
+    x::Variable
     n::Int
     rhs::Expression
 end
-function Base.convert(::Type{DiffEq}, eq::Equation)
```
ChrisRackauckas: why not convert?

src/systems/diffeqs/diffeqsystem.jl
ChrisRackauckas: >I like how this pushes the variable ordering to the only part of the system that needs it. Does this seem reasonable?

Yeah, that seems reasonable. 

>Since all variables are required anyways, maybe it would be worthwhile to provide accessor functions which show the user stored variables, in the order that they will be assumed to be within the vector.

Yes, this is especially needed after transforms, since in that case we will have programmatically created new variables.

src/systems/diffeqs/diffeqsystem.jl
ChrisRackauckas: This looks fine. In fact, I think that's how it has to be. Each function construction will likely need its own idea for how to convert variables to expressions. For example, DDEs will need to specialize on the delayed variables and change to `h(tau;idxs=...)` and stuff like that, while DAEs will need `du` terms all over.

src/variables.jl
ChrisRackauckas: what are the leaves then?

src/variables.jl
```diff
@@ -30,11 +29,7 @@ Base.isequal(c::Constant, n::Number) = c.value == n
 Base.isequal(n::Number, c::Constant) = c.value == n
 Base.isequal(a::Constant, b::Constant) = a.value == b.value
 
-function Base.convert(::Type{Expr}, x::Variable)
-    x.known               || return x.name
-    isempty(x.dependents) && return x.name
-    return :($(x.name)($(convert.(Expr, x.dependents)...)))
-end
+Base.convert(::Type{Expr}, x::Variable) = x
```
ChrisRackauckas: hmm yeah, that does seem odd.

test/derivatives.jl
ChrisRackauckas: While it makes sense when you think about it, it feels very wrong from an API standpoint. Why can't the omission of () work here? It seems that if we did `convert(::Type{Expression},x::Variable) = x()` then the parenless one would act as we need it to?

test/system_construction.jl
ChrisRackauckas: yeah we should have an accessor function for this

test/system_construction.jl
```diff
@@ -59,17 +82,21 @@ lowered_eqs = [D(u_tt) ~ 2u_tt + u_t + x_t + 1
...
        D(y) ~ x*(ρ-z)-y,
        D(z) ~ x*y - β*z]
-de = DiffEqSystem(eqs,t,[x,y,z],[σ,ρ,β])
-generate_function(de)
+de = DiffEqSystem(eqs)
+generate_function(de, [x,y,z], [σ,ρ,β])
 jac = calculate_jacobian(de)
-f = ODEFunction(de)
+@test_broken begin
+    f = ODEFunction(de)
```
ChrisRackauckas: why is this broken?

src/differentials.jl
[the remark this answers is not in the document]
HarrisonGrodin: Since the `dvs` no longer implicitly store the independent variable (stored as function `x`, not `x(t())`), we need to inject the `iv` somewhere. Would it be better prior to this `calculate_jacobian` call (i.e. in `calculate_jacobian(::DiffEqSystem)`)?

src/systems/diffeqs/diffeqsystem.jl
HarrisonGrodin: Hmm. Does this mean that our existing functionalities on `DiffEqSystem`s support things like `D(sin(x)) ~ ...`? And what about `2D(sin(x)) + D(y)^2 ~ ...`? I assumed we wanted things of the form `[D(x) ~ ..., D(y) ~ ...]` based on the test cases, and thus tried to mirror that with the shape of the data itself, but if that's not the case I can easily switch it.

The purpose behind this change was to ensure that the invariants implicitly assumed in other functions (e.g. `first_order_transform`) hold for any constructed system, rather than having that assumption be false and throwing a runtime error.

src/systems/diffeqs/diffeqsystem.jl
HarrisonGrodin: > Yes, this is especially needed after transforms, since in that case we will have programmatically created new variables.

 Ah, right - good call.

src/systems/diffeqs/diffeqsystem.jl
HarrisonGrodin: Oh, true - I guess let's keep it, then. I'm not a fan of the current implementation/strategy, but that's an internal feature which we can change down the road when it's clearer what we need.

src/variables.jl
HarrisonGrodin: Either constants (e.g. `Constant(2)`) or "0-ary variables" (e.g. `t()`).

(Interestingly enough, term rewriting literature doesn't even have the former - constants are all treated as 0-ary symbolic functions.)

test/derivatives.jl
HarrisonGrodin: Oh, right - should've seen that. :) I'll implement it now.

test/system_construction.jl
HarrisonGrodin: What should we have? `independent_variable`, dependent_variable`, and `parameters`?

src/systems/diffeqs/diffeqsystem.jl
HarrisonGrodin: The function now returns a 2-tuple of `(t::Variable, ::DiffEq)`, so it felt odd to continue calling it `convert`. And all of this is purely internal, anyways.

test/system_construction.jl
HarrisonGrodin: In `ODEFunction`, we have:
```julia
expr = generate_function(sys; version = version)
```

This no longer works, since `generate_function` requires the dvs/ps as a parameter. We have a couple of ways we could fix it:
1. `ODEFunction` could take in the dvs/ps ordering, as well.
2. Now that we're adding accessor functions, we should be able to just provide the internal dvs/ps and let the user deal with ordering.

Which sounds better?

test/derivatives.jl
HarrisonGrodin: Actually, wait - this does have some drawbacks. If the types force it to be an `Expression` (e.g. `x(t)`), this should work. However, we can't just treat it like an `Expression` everywhere - regardless of whether or not we have conversion, `t - 1` won't work, unless we define all functions which take in `Expression`s on `Variable`s, as well (including helper functions and such). This seems rather cumbersome...

Although it does add a few characters to the setup, I think it might make sense to keep this, since it explicitly notes that the parameters are constants (take in zero arguments). It does make sense only after you think about it, but I'd rather that than allow for undesired edge cases.

src/systems/diffeqs/diffeqsystem.jl
ChrisRackauckas: >And what about 2D(sin(x)) + D(y)^2 ~ ...?

That's a DAE, and if it's linear in the derivatives, it's a mass matrix ODE. We want to go there next. Since we're going to be hitting this issue soon, we should start thinking about how to cleanly handle the differences between them.

test/system_construction.jl
ChrisRackauckas: `independent_variables`, `dependent_variables`, `parameters`.

test/system_construction.jl
ChrisRackauckas: I think in every case with dvs/ps kind of inputs, we should have default values be from the accessor functions, and then let the user override with their own ordering.

src/ModelingToolkit.jl
```diff
@@ -2,6 +2,7 @@ module ModelingToolkit
 
 export Operation, Expression
 export calculate_jacobian, generate_jacobian, generate_function
+export independent_variables, dependent_variables, parameters
```
ChrisRackauckas: why do you have to export them as functions?

ChrisRackauckas:
I still don't like that `@parameters x()` issue, and don't know about those stub functions, but we can merge this and handle the various pieces next. Altogether, this is well done and complete.

ChrisRackauckas:
Merge conflicts with the change from iW to factorization.

PR #123, opened 2019-04-11
simplified interface for declaring parameters

ChrisRackauckas:
This makes `@parameter a` a constant again, but makes `@dependent_parameters` exist so the additional functionality is still possible. The underlying IR is untouched. This makes it look nice again


src/variables.jl
```diff
@@ -50,7 +50,11 @@ function _parse_vars(macroname, known, x)
             expr = :($var_name = $Variable($(Meta.quot(var_name)); known = $known)($(_var.args[2:end]...)))
         else
             var_name = _var
-            expr = :($var_name = $Variable($(Meta.quot(var_name)); known = $known))
+            if add_call
```
HarrisonGrodin: This means that we can't express variable functions with differing inputs (using the macro API, at least). For example, we couldn't produce the `x` in:
```julia
eqs = [x(t) ~ 2x(t-1)]
```

ChrisRackauckas:
Anything else @HarrisonGrodin ?

PR #124, opened 2019-04-20
Say ModelingToolkit is ready to use

ChrisRackauckas:
Basic syntax should change all too much?

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,8 +14,6 @@ nicely with user-defined routines. Together, this is an abstract form of a
 scientific model that is easy for humans to generate but also easy for programs
 to manipulate.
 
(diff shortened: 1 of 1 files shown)


HarrisonGrodin:
It is still somewhat WIP, but I guess that's more wrt future versions (with new version numbers).

ChrisRackauckas:
Yeah, there will be changes of course, but I think we are safe to say that the library is no longer a hell-hole of broken madness but something that is a usable piece of software that is now being actively maintained. At least the user-facing API should be mostly solid, maybe with ODEFunction dispatching being replaced with ODEProblem down the line but that's something that still needs to be discussed.

PR #125, opened 2019-04-26
Start on API documentation

jlumpe:
Added basic docstrings to most of the core types and a few functions. Created documentation skeleton with API page showing docstrings.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -6,6 +6,7 @@ version = "0.3.0"
 [deps]
 DiffEqBase = "2b5f629d-d688-5b77-993f-72d75c75574e"
 DiffRules = "b552c78f-8df3-52c6-915a-8e097449b14b"
+DocStringExtensions = "ffbed154-4ef7-542d-bbb7-c09d3a79fcae"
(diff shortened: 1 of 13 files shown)


jlumpe:
@HarrisonGrodin makes sense, I've removed the documentation for all fields which seem to have accessors. Those were `ODESystem` and `NonlinearSystem` (except for `.exprs`), and `Constant.value` (which is available through `Base.get()`).

jlumpe:
@HarrisonGrodin the [most recent build](https://travis-ci.org/JuliaDiffEq/ModelingToolkit.jl/builds/525139452) passed but it didn't seem to update here. This should be ready to merge.

HarrisonGrodin:
It seems like there are still some TODOs?

jlumpe:
I actually intentionally left those in for other developers who know more about how the package works than I do. I can remove them if you don't want them in master.

jlumpe:
@HarrisonGrodin merge conflicts resolved, should I remove the TODOs or leave them in?
\

ChrisRackauckas:
The hold up is really that we don't know what the public API really should be yet... but screw it. Let's merge and keep refining it.

PR #142, opened 2019-05-30
Add static version of Wfact and negate the W operator

YingboMa:


src/utils.jl
```diff
@@ -48,7 +48,7 @@ function build_function(rhss, vs, ps, args = (), conv = rhs -> convert(Expr, rhs
         let_expr = Expr(:let, var_eqs, sys_expr)
         :((u,p,$(args...)) -> begin
             X = $let_expr
-            T = StaticArrays.similar_type(typeof(u), eltype(X))
+            T = $(constructor === nothing ? :(StaticArrays.similar_type(typeof(u), eltype(X))) : constructor)
```
ChrisRackauckas: Maybe this is cleaner just as a default for the constructor.

PR #144, opened 2019-07-06
Include Jacobian in generated ODEFunction

ChrisRackauckas:


src/systems/diffeqs/diffeqsystem.jl
```diff
@@ -196,10 +196,11 @@ are used to set the order of the dependent variable and parameter vectors,
 respectively.
 """
 function DiffEqBase.ODEFunction(sys::ODESystem, dvs, ps; version::FunctionVersion = ArrayFunction)
-    expr = generate_function(sys, dvs, ps; version = version)
+    expr = eval(generate_function(sys, dvs, ps; version = version))
+    jac_expr = isempty(sys.jac[]) ? nothing : eval(generate_jacobian(sys))
```
HarrisonGrodin: This means that unless the Jacobian has already been computed, it won't be included. Is that the desired behavior?

(Note that `jac_expr = eval(generate_jacobian(sys))` would use the cached version if one exists, computing it otherwise.)

ChrisRackauckas: Yes, this gives an opt-in mechanism: call `calculate_jacobian` if you want it.

HarrisonGrodin: That seems like a rather strange/unintuitive way of opting in (which also makes the fact that we currently cache Jacobians user-facing)... couldn't we just have a kwarg for `jac::Bool = false`?

ChrisRackauckas: Yes, we can do it with a bool as well.

PR #146, opened 2019-07-06
Add modelingtoolkitize

YingboMa:


src/utils.jl
```diff
@@ -84,3 +84,27 @@ function vars!(vars, O)
 
     return vars
 end
+
```
ChrisRackauckas: This should probably go with the ODESystem stuff.

HarrisonGrodin:
Can we name this differently, please? Haha. Maybe `symbolic` or something?

ChrisRackauckas:
`symbolify`? `symbolicitizationifyification`

ChrisRackauckas:
ehh I actually like the name, since most others will clash with libraries like SymPy and whatnot, so we might as well say how we are symbolicing it.

PR #149, opened 2019-07-08
Use dispatch for in-place out-of-place function versions

ChrisRackauckas:
Since anonymous functions cannot dispatch, this creates a gensym'd generic function with the two dispatches, making it simultaniously holding the oop and iip versions. Then ODEFunction just specifies which path to take. Defaults to saying it should be ran in-place.


ChrisRackauckas:
This should fix the FIRK tests since it's matching what ParameterizedFunctions.jl did here in terms of outputs, but doing it all semi-anonymous.

src/utils.jl
```diff
@@ -31,26 +31,31 @@ function flatten_expr!(x)
...
-        :((u,p,$(args...)) -> begin
+    fname = gensym()
+
+    X = gensym()
+    ip_sys_exprs = [:($X[$i] = $(conv(rhs))) for (i, rhs) ∈ enumerate(rhss)]
+    ip_let_expr = Expr(:let, var_eqs, build_expr(:block, ip_sys_exprs))
+
+    sys_expr = build_expr(:tuple, [conv(rhs) for rhs ∈ rhss])
+    let_expr = Expr(:let, var_eqs, sys_expr)
+    quote
```
HarrisonGrodin: Hmm, cool. I was never a fan of `version` anyway; this is a cute solution.

The near-overlapping dispatch is slightly concerning, but given that this code is unlikely to be touched frequently, it'll probably be fine.

ChrisRackauckas:
>Any particular reason not to remove version immediately? 

It's breaking in a way that throws an error if an unknown keyword argument is passed, so we'll get errors in Pumas and DiffEqProblemLibrary. There's no reason to leave this for more than like a month though.

PR #154, opened 2019-07-27
move indices to subscripts

ChrisRackauckas:


ChrisRackauckas:
@saschatimme @HarrisonGrodin I think we should instead do indexed variables as one object with indexing. However, I am at least going to here since it gets my newest coolest prototype working.

chakravala:
The `DirectSum` package provides extended indexing for full-alphanumeric indices up to 64 indices

https://github.com/chakravala/DirectSum.jl/blob/master/src/indices.jl

Highly recommend using it for printing indices, especially if you need a dual space also and are interested in working with conformal projective geometry (e.g. point at infinity, origin)

It is designed for my own purposes, but I can accommodate your requirements.

ChrisRackauckas:
I think we can just pull apart the symbol and map the characters directly. The only peculiarity is how to group the subscripts.

chakravala:
Also, my package provides some special caching for indices so that they can be obtained at precompilation time for maximum performance of printing the indices.

ChrisRackauckas:
Index limit fixed by https://github.com/JuliaDiffEq/ModelingToolkit.jl/pull/155/commits/9ab9df34da260309363b77d892919978f29d96e5

Issue #109, opened 2019-03-12
Add gradient, jacobian and hessian functions


dpsanders:
Given an `Operation`, it would be great to be able to calculate symbolically its `gradient`, `jacobian` and `hessian` as e.g.
`gradient(ex, [x, y])` or similar.

I believe this is currently possible but could be usefully streamlined.

ChrisRackauckas:
Completed. We should get full documentation soon.

PR #156, opened 2019-07-28
Improved direct usage

ChrisRackauckas:
For things that @dpsanders wants to do, a simplified form of expression building is quite handy.


ChrisRackauckas:
Direct usage is usage outside of an `AbstractSystem`. This PR adds calculus (gradients, Jacobians, and Hessians) to generic systems of equations, has a way of displaying the functions in a simplified way (no `()`), has helpers for building functions with and without parameters, and tests the world-age fix.

ChrisRackauckas:
Fixes #109

src/direct.jl
```diff
@@ -0,0 +1,43 @@
...
+function simplified_expr(c::Constant)
+    c.value
+end
+
+function simplified_expr(eq::Equation)
+    Expr(:(=), simplified_expr(eq.lhs), simplified_expr(eq.rhs))
+end
+
+macro I(ex)
+    name = :ICompile
```
ChrisRackauckas: Is there a nicer way to handle this than returning the macro and then requiring the macro to be called? If we do this with a gensym we can write a macro which calls this macro to return a macro and then calls the returned macro, which is a crazy definition of "identity"

test/direct.jl
```diff
@@ -0,0 +1,127 @@
...
+f(out,[1.0,2,3],[1.0,2,3])
+@test all(o1 .== out)
+
+function test_worldage()
+   @parameters σ() ρ() β()
+   @variables x y z
+   eqs = [σ*(y-x),
+          x*(ρ-z)-y,
+          x*y - β*z]
+   _f = eval(ModelingToolkit.build_function(eqs,[x,y,z],[σ,ρ,β]))
```
ChrisRackauckas: eval on the user side makes it easier to pick the appropriate module, which is important for newly registered function calls

ChrisRackauckas:
I there's a lot of other fixes in here, so I'm going to merge and tag and we can continue refining this, especially the `@I` macro which is a little peculiar. All of this will remain undocumented for now.

PR #160, opened 2019-07-30
Fix the order of params in a `ODESystem`

YingboMa:
Unfortunately, we cannot assume the order of params in an ODE system. For instance, the Lotka-Volterra is

```julia
function lotka(du,u,p,t)
  x = u[1]
  y = u[2]
  du[1] = p[1]*x - p[2]*x*y
  du[2] = -p[3]*y + p[4]*x*y
end
```

and generated Jacobian function is

```julia
    function ##405(##406, u, p, t)
        #= /Users/yingboma/.julia/dev/ModelingToolkit/src/utils.jl:55 =#
        let (x₁, x₂, α₁, α₃, α₄, α₂) = (u[1], u[2], p[1], p[2], p[3], p[4])
            ##406[1] = (+)(α₁, (*)(x₂, α₂, -1))
            ##406[2] = (*)(x₂, α₄)
            ##406[3] = (*)(α₂, x₁, -1)
            ##406[4] = (+)((*)(α₃, -1), (*)(α₄, x₁))
        end
        #= /Users/yingboma/.julia/dev/ModelingToolkit/src/utils.jl:56 =#
        nothing
    end
```
.

Note that it generates
```julia
(x₁, x₂, α₁, α₃, α₄, α₂) = (u[1], u[2], p[1], p[2], p[3], p[4])
```
which is incorrect. The order which params appear is not necessarily the same with the order which the user supplies it. Hence, we have to have an optional argument for params that retains the order.

### DIFF
--- a/src/systems/diffeqs/diffeqsystem.jl
+++ b/src/systems/diffeqs/diffeqsystem.jl
@@ -81,7 +81,7 @@ struct ODESystem <: AbstractSystem
     Wfact_t::RefValue{Matrix{Expression}}
 end
 
-function ODESystem(eqs)
+function ODESystem(eqs::AbstractArray{<:Equation}, ps=nothing)
     reformatted = to_diffeq.(eqs)
 
     ivs = unique(r[1] for r ∈ reformatted)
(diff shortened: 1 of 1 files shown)


YingboMa:
Ah, we could do it just by
```julia
de = modelingtoolkitize(prob)
jac = eval(ModelingToolkit.generate_jacobian(de...))
```

PR #171, opened 2019-08-28
Fix incorrect syntax in README

non-Jedi:


HarrisonGrodin:
Oh, I guess that works. I would have preferred updating Line 312 to this, though:
```
@parameters t σ(..) ρ β[1:3]
```

PR #173, opened 2019-09-02
compatibility with distributed

ChrisRackauckas:
Uses `@everywhere` with a MT-specific tag to make the generic functions safe for distributed use.


ChrisRackauckas:
Fixes https://github.com/JuliaDiffEq/ModelingToolkit.jl/issues/174 . Requires GG.jl

Issue #152, opened 2019-07-16
Inconsistent Differential across Julia versions.


korsbo:
I ran across a problem when I tried to add support for ModelingToolkits to Latexify.jl: when constructing an Equation, I get different results when using different Julia versions, even though the ModelingToolkit version is the same.

From the first example in the readme:
```julia
using ModelingToolkit

# Define some variables
@parameters t σ ρ β
@variables x(t) y(t) z(t)
@derivatives D'~t

eqs = [D(x) ~ σ*(y-x),
       D(y) ~ x*(ρ-z)-y,
       D(z) ~ x*y - β*z]
```

Results using julia 1.1 and 1.2-rc2:

![image](https://user-images.githubusercontent.com/16306680/61290682-5c91d300-a7c4-11e9-8abb-15e92baf4714.png)

note the first part of each `Equation`.

korsbo:
I just updated everything on Julia 1.1 but the discrepancy persisted.

ChrisRackauckas:
We don't support the earlier Julias now which fixes this in some sense.

PR #190, opened 2019-10-02
Add latexify support.

korsbo:
This PR takes a first step in providing Latexify.jl support for ModelingToolkit.jl (see also https://github.com/korsbo/Latexify.jl/pull/64).

I'm defining a basic `@latexrecipe` which adds support for arrays of Equations. 

```julia
using Latexify
using ModelingToolkit

@parameters t σ ρ β
@variables x(t) y(t) z(t)
@derivatives D'~t

eqs = [D(x) ~ σ*(y-x)*D(x-y)/D(z),
       0 ~ σ*x*(ρ-z)/10-y,
       D(z) ~ x*y - β*z]

latexify(eqs, iv=:t)
```
image

This can be merged as-is and let further support come later, or it could be extended immediately to support `ODESystems` and such.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -9,6 +9,7 @@ DiffRules = "b552c78f-8df3-52c6-915a-8e097449b14b"
 Distributed = "8ba89e20-285c-5b6f-9357-94700520ee1b"
 DocStringExtensions = "ffbed154-4ef7-542d-bbb7-c09d3a79fcae"
 GeneralizedGenerated = "6b9d7cbe-bcb9-11e9-073f-15a7a543e2eb"
+Latexify = "23fbe1c1-3f47-55db-b15f-69d7ec21a316"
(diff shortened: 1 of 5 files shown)


korsbo:
This works on Julia 1.2. Earlier versions run in to errors described in #152.

korsbo:
Earlier julia versions still produce an output. It is less nice, but not fatal. The inconsistency does however make testing difficult.

A potential solution is to just not do testing.

I'm already a bit ambivalent about testing. The current way it is done is to check that the latexify output matches an output that I eyeballed to be correct. This is however sensitive to any small changes in latexify's output formatting. 

The exact output formatting of latexify is usually not critical for any real application and I would thus like to be able to tweak it without major releases. However, testing in external packages would break from such tweaks. 

The obvious downside to not do automated testing of the latexrecipe is of course that internal breaking changes would likely go unnoticed.

ChrisRackauckas:
I'll just comment out the test for now.

PR #191, opened 2019-10-03
Add methods to convert Operation to Expr

jamesjscully:


ChrisRackauckas:
I guess this is fine. It may not be the most useful though since lowering depends on context, and if it's no context then people might want `simplified_expr` instead.

jamesjscully:
aside from changing variables like y(t) to y, this is a subset. I didn't know simplify_expr existed. simplify_expr is essentially what i was going for. Looking at it, this one doesn't handle constants or differentials correctly.  When would it be useful to have y(t) in the expression instead of y though?`eval(:(y(t)))#error`

jamesjscully:
nevermind just saw you already explained this in your comment :1st_place_medal:

ChrisRackauckas:
>When would it be useful to have y(t) in the expression instead of y though?eval(:(y(t)))#error

When it's actually a function, for example time-varying covariate, or the solution to a differential equation.

So should this just be replaced with simplify_expr? That might make it easier for most use cases.

jamesjscully:
Yes simplify_expr is probably better

ChrisRackauckas:
Thanks!

PR #205, opened 2019-11-28
No bounds check in generated functions

ChrisRackauckas:


ChrisRackauckas:
Not sufficient to fix the Pumas tests, but probably good to do anyways? Branch prediction should essentially make this free.

Issue #214, opened 2019-12-22
Naming, Laziness, and Ordering


ChrisRackauckas:
There are three issues that are hitting the same fundamental problem. Here's a few questions to get started:

1. When doing ODE order lowering, what should we name the outputs? The user has to know and understand the semantics in order to know how to set the inputs to the correct values, so this is not an internal detail.
2. When doing component-based systems, how do we name things? How does the user know how to set the parameters of component 1 vs component 2?
3. When building a simple ODE system with automated parameter and variable capturing, how can the user sanely set the parameters?
4. When doing index reduction, same thing.
5. Etc. etc. etc. this comes up with almost every transform.

These are all the same underlying issue that, when we generate code:

1. We need some compiler-y way to make sure we don't override namings.
2. It needs to be user-facing, because the user must understand the resulting semantics to know how to set values.

How can we solve it? @HarrisonGrodin and I talked in detail in a few different ways. One thing we can do is give up trying to come up with sane names and try to keep things lazy for as long as possible. `D(x)` makes sense, so try keeping it. The only place where names are then required would be at the very end when we flatten the system to compile it, in which case hopefully the user doesn't have to interact with it anymore.

However, the issue that came up with that is that, the user still has to understand the `f` that you built because they need to give `u0` and `p`. Do we then move `u0` and `p` into the `ODESystem` and directly generate `ODEProblem`? What about `tspan`? This is the line of thought I was going down before, and it doesn't seem so generalizable: we'd end up special casing a lot of things specifically for DESystem types, and then not a lot of reusability could be had to other domains. Also, it just feels wrong.

But recently an idea that I had was just about building "map vector" functions for the user. For example, if we know internally the order of the variables we are using, then:

```julia
map_vector_state(odesys,[x=>1.5,y=>2.0,D(x)=>5.0])
```

can be used to spit out a `u0` vector that matches the system we are generating. The parameters can be handled similarly, and this idea then applies even to NonlinearSystem, optimization, and beyond. It can allow for handling the lazy stuff in a nice way as well. Also, this would be a good place to throw errors, asking for values. For example, if `daesys` comes from an index reduction, the user might not know up front what values they need to define, and the context of the system itself will be necessary for giving the user a clue. From this we can say `Error: D(y) required for the system`, and that is easily interpretable by the user.

This still needs more thinking, but it could possibly work. In order to do such matching though, we need to enforce a normal form, which is not something we do well with right now.

@ranjanan @YingboMa @shashi

[...]

PR #220, opened 2020-01-03
correct ordering of sys.dvs in ode_order_lowering

### LINKED ISSUE #219: Wrong solution for simple second order ODE
The following problem gives a plainly wrong solution:
```julia
using DifferentialEquations
using ModelingToolkit

@parameters x E
@variables ψ(x)
@derivatives ∂²''~x

V(x) = .5x^2
eqs = [∂²(ψ) ~ 2(V(x) - E) * ψ]

de = ode_order_lowering(ODESystem(eqs))

prob = ODEProblem(ODEFunction(de), [.2, .1], (-2., 5.), [.5])

sol = solve(prob)
```
`de.eqs` looks fine:
```
julia> de.eqs
2-element Array{ModelingToolkit.DiffEq,1}:
 ModelingToolkit.DiffEq(ψ_x, 1, (2 * (0.5 * x ^ 2 - E)) * ψ(x))
 ModelingToolkit.DiffEq(ψ, 1, ψ_x(x))
```
The following gives the correct solution:
```julia
function f(du,u,p,x)
    du[1] = 2*(V(x) - 0.5) * u[2]
    du[2] = u[1]
end
prob = ODEProblem(f, [.2, .1], (-2., 4.), [.5])
sol = solve(prob,Tsit5(),abstol=1e-8,reltol=1e-8)
using Plots; plot(sol)
```

simeonschaub:
This should fix #219.

### DIFF
--- a/src/systems/diffeqs/first_order_transform.jl
+++ b/src/systems/diffeqs/first_order_transform.jl
@@ -8,8 +8,8 @@ function lower_varname(var::Variable, idv, order)
 end
 
 function ode_order_lowering(sys::ODESystem)
-    (eqs_lowered, new_vars) = ode_order_lowering(sys.eqs, sys.iv)
-    ODESystem(eqs_lowered, sys.iv, [sys.dvs; new_vars], sys.ps)
+    eqs_lowered, _ = ode_order_lowering(sys.eqs, sys.iv)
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
Ahh, it's a case of https://github.com/JuliaDiffEq/ModelingToolkit.jl/issues/214. While this isn't a good solution in general, it's reasonable enough to understand for now, at least until we get a better #214 solution. Thanks!

PR #222, opened 2020-01-13
Add `conj` as a special case and add promotion rule

YingboMa:


src/function_registration.jl
```diff
@@ -55,6 +55,7 @@ end
 # special cases
 Base.:^(x::Expression,y::T) where T <: Integer = Operation(Base.:^, Expression[x, y])
 Base.:^(x::Expression,y::T) where T <: Rational = Operation(Base.:^, Expression[x, y])
+Base.conj(x::Expression) = Operation(Base.conj, [x])
 Base.muladd(a::Expression, b::Expression, c::Expression) = a * b + c
```
ChrisRackauckas: why not make this actually a `muladd` call?

YingboMa: `muladd` has a promotion call, and that is not defined on `Expression`.

ChrisRackauckas: is there a way to bypass that?

YingboMa: Right, a better fix is to define a promotion rule.

PR #233, opened 2020-02-06
fix array fast path on matrices, do N-tensor generation, and fast sparse functions

ChrisRackauckas:
Fixes https://github.com/JuliaDiffEq/ModelingToolkit.jl/issues/231


ChrisRackauckas:
Implements https://github.com/JuliaDiffEq/ModelingToolkit.jl/issues/133

PR #238, opened 2020-02-07
Add n-ary *, + rules

YingboMa:


src/differentials.jl
```diff
@@ -109,9 +109,8 @@ for (modu, fun, arity) ∈ DiffRules.diffrules()
     end
 end
 
-derivative(::typeof(+), args::NTuple{T,Operation}, ::Val{i}) where {T,i} = args[i]
-derivative(::typeof(+), args::Tuple{Operation,Operation}, ::Val{1}) = args[1]
-derivative(::typeof(+), args::Tuple{Operation,Operation}, ::Val{2}) = args[2]
```
ChrisRackauckas: Why delete the ambiguity fixes? Is it good now?

YingboMa: Ah, I didn't realize it.

ChrisRackauckas: It's because DiffRules defines the 2-ary version and we need to disambiguate from that. Probably the same for `*`.

PR #246, opened 2020-02-19
Add Latexify for ODESystem

ChrisRackauckas:


ChrisRackauckas:
@korsbo can you help with figuring out why the recipe for the Vector{DiffEq} throws an error that it's a malformed LaTeX expression?

korsbo:
Sure, I just pulled the pr and I'll give it a go now.

src/latexify_recipes.jl
```diff
@@ -21,3 +21,27 @@
...
+    rhs = getfield.(eqs, :rhs)
+    rhs = convert.(Expr, rhs)
+    rhs = [postwalk(x -> x isa ModelingToolkit.Constant ? x.value : x, eq) for eq in rhs]
+    rhs = [postwalk(x -> x isa Expr && length(x.args) == 1 ? x.args[1] : x, eq) for eq in rhs]
+    rhs = [postwalk(x -> x isa Expr && x.args[1] == :Differential && length(x.args[2].args) == 2 ? :($(Symbol(:d, x.args[2]))/($(Symbol(:d, x.args[2].args[2])))) : x, eq) for eq in rhs]
+    rhs = [postwalk(x -> x isa Expr && x.args[1] == :Differential ? "\\frac{d\\left($(Latexify.latexraw(x.args[2]))\\right)}{d$iv}" : x, eq) for eq in rhs]
+
+    var = getfield.(getfield.(eqs, :x),:name)
+    ns  = getfield.(eqs, :n)
+    lhs = [ns[i] == 1 ? "\\frac{d$(Latexify.latexraw(var[i]))}{d$iv}" : "\\frac{d^{$(ns[i])}$(Latexify.latexraw(var[i]))}{d$iv^{$(ns[i])}}" for i in 1:length(var)]
```
korsbo: strings will be parsed which is not what you want. You want these strings to be LaTeXStrings which are passed through verbatim. 
```julia
    lhs = [ns[i] == 1 ? LaTeXString("\\frac{d$(Latexify.latexraw(var[i]))}{d$iv}") : LaTeXString("\\frac{d^{$(ns[i])}$(Latexify.latexraw(var[i]))}{d$iv^{$(ns[i])}}") for i in 1:length(var)]

```

src/latexify_recipes.jl
```diff
@@ -21,3 +21,27 @@
...
+    return lhs, rhs
+end
+
+@latexrecipe function f(sys::ModelingToolkit.ODESystem; iv=:t)
+    latexify(sys.eqs)
```
korsbo: replace with 
```julia
return sys.eqs
```

korsbo:
There was one error here, but I also think that there is a bug in LaTeXify itself. 
If you fix the comment above, you should get that `latexify(sys.eqs)` works but `latexify(sys)` does not. 

That, however, must be a Latexify.jl bug. 

Tomorrow though, I just learnt that my kid is hurt, got to go.

ChrisRackauckas:
You forgot the comment, but all is good. Take your time!

src/latexify_recipes.jl
ChrisRackauckas: Looks like that's not recursing: is that the bug?

korsbo: Indeed

src/latexify_recipes.jl
```diff
@@ -21,3 +21,27 @@
...
+    rhs = [postwalk(x -> x isa Expr && x.args[1] == :Differential ? "\\frac{d\\left($(Latexify.latexraw(x.args[2]))\\right)}{d$iv}" : x, eq) for eq in rhs]
+
+    var = getfield.(getfield.(eqs, :x),:name)
+    ns  = getfield.(eqs, :n)
+    lhs = [ns[i] == 1 ? Latexify.LaTeXString("\\frac{d$(Latexify.latexraw(var[i]))}{d$iv}") : LaTeXString("\\frac{d^{$(ns[i])}$(Latexify.latexraw(var[i]))}{d$iv^{$(ns[i])}}") for i in 1:length(var)]
+    return lhs, rhs
+end
+
+@latexrecipe function f(sys::ModelingToolkit.ODESystem; iv=:t)
+    sys.eqs
```
korsbo: It works if you replace this with 
```julia
return latexify(sys.eqs; iv=:t)
```

I have been thinking about a neater solution from within Latexify itself but I have not come up with anything. Since I found a "working" solution, I'll drop it for now and return to revising my papers/writing my PhD thesis/finding a new job. I opened an issue in Latexify where any updates can be tracked https://github.com/korsbo/Latexify.jl/issues/92

PR #251, opened 2020-03-13
Add new systems types in support of DiffEqBiological.jl

ChrisRackauckas:
add ReactionSystem and start SDESystem components

convert of ReactionSystem to ODEsystem is not quite complete yet... but it's a start for people to work from.


ChrisRackauckas:
@isaacsas @TorkelE @korsbo I'll try to get basic building blocks done ASAP and we can keep iterating from there.

PR #255, opened 2020-03-21
Add Recursive Variable Extraction and Substitution

arnavs:
e.g. 

```
julia> expr
(((1 / β - 1) + δ) / α) ^ (1 / (α - 1))

julia> get_variables(expr)
3-element Array{Any,1}:
 β
 δ
 α
```

and 

```
julia> s = δ => γ
δ => γ

julia> ModelingToolkit.substitute_expr!(expr, s)

julia> expr
(((1 / β - 1) + γ) / α) ^ (1 / (α - 1))
```

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -149,10 +149,9 @@ end
 is_singleton(e) = false
 is_singleton(e::Operation) = e.op isa Variable
 
+get_variables(e::ModelingToolkit.Constant, vars) = nothing
 function get_variables(e::Expression, vars = Operation[])
-    if e isa ModelingToolkit.Constant
-        # do nothing
-    elseif is_singleton(e)
+    if is_singleton(e)
         push!(vars, e)
     else
         foreach(x -> get_variables(x, vars), e.args)
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -165,5 +165,5 @@ function substitute_expr!(expr::Expression, s::Pair{Operation, Operation})
     end
     good_args = filter(x -> hasproperty(x, :args) && !(is_singleton(x) || x isa ModelingToolkit.Constant), expr.args)
     [substitute_expr!(arg, s) for arg in good_args] # iterate where there is more to go
-    return nothing 
+    return nothing
 end
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -154,5 +154,5 @@ function get_variables(e::Expression, vars = nothing)
   elseif hasproperty(e, :args) # because it can sometimes be 0
     foreach(x -> get_variables(x, vars), e.args)
   end
-  return vars
+  return unique(vars)
 end
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -149,7 +149,7 @@ end
 is_singleton(e) = hasproperty(e, :name) || hasproperty(e, :op) && hasproperty(e.op, :name)
 
 function get_variables(e::Expression, vars = nothing)
-  vars = isnothing(vars) ? [] : vars
+  vars = isnothing(vars) ? Operation[] : vars
   if is_singleton(e)
     push!(vars, e)
   elseif hasproperty(e, :args) # because it can sometimes be 0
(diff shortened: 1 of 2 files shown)


arnavs:
Also added a substitution thing.

src/utils.jl
```diff
@@ -144,3 +144,26 @@ function vars!(vars, O)
 
     return vars
 end
+
+# variable extraction
+is_singleton(e) = hasproperty(e, :name) || hasproperty(e, :op) && hasproperty(e.op, :name)
+
+function get_variables(e::Expression, vars = nothing)
+  vars = isnothing(vars) ? [] : vars
```
ChrisRackauckas:
```suggestion
  vars = isnothing(vars) ? Variable[] : vars
```

ChrisRackauckas: Or do you want to extract the Operation of the variable? That may be more useful?

src/utils.jl
```diff
@@ -144,3 +144,26 @@ function vars!(vars, O)
 
     return vars
 end
+
+# variable extraction
+is_singleton(e) = hasproperty(e, :name) || hasproperty(e, :op) && hasproperty(e.op, :name)
```
ChrisRackauckas:
```suggestion
is_singleton(e::Operation) = e.op isa Variable
```

src/utils.jl
ChrisRackauckas:
```suggestion
  vars = isnothing(vars) ? Operation[] : vars
```

src/utils.jl
ChrisRackauckas: then add `is_singleton(e) = false`

src/utils.jl
```diff
@@ -144,3 +144,26 @@ function vars!(vars, O)
...
+  if is_singleton(e)
+    push!(vars, e)
+  elseif hasproperty(e, :args) # because it can sometimes be 0
+    foreach(x -> get_variables(x, vars), e.args)
+  end
+  return unique(vars)
+end
+
+# variable substitution
+function substitute_expr!(expr::Expression, s::Pair{Operation, Operation})
```
ChrisRackauckas: make a dispatch on Operation and then a dispatch on etc. and that would make this be a lot safer.

src/utils.jl
arnavs: @ChrisRackauckas The issue with this is that `Variable`s don't have `Op`s, right?

```
julia> x = Variable(:x, known = true)
x (callable known variable)

julia> dump(x)
x (function of type Variable)
  name: Symbol x
  known: Bool true
```

arnavs: oh, I see what you mean. I wanted to leave it open to the possibility of non-Operation variables, but this works too.

arnavs: Done.

src/utils.jl
arnavs: Done.

src/utils.jl
```diff
@@ -144,3 +144,30 @@ function vars!(vars, O)
 
     return vars
 end
+
+# variable extraction
+is_singleton(e) = false
+is_singleton(e::Operation) = e.op isa Variable
+
+function get_variables(e::Expression, vars = nothing)
+    vars = isnothing(vars) ? Operation[] : vars
```
ChrisRackauckas:
```suggestion
function get_variables(e::Expression, vars = Operation[])
```

src/utils.jl
```diff
@@ -144,3 +144,30 @@ function vars!(vars, O)
...
+    if e isa ModelingToolkit.Constant
```
ChrisRackauckas: different dispatch

arnavs: Done.

ChrisRackauckas:
There was too much to do in this PR so I just finished it. Superseded by the squash merge https://github.com/JuliaDiffEq/ModelingToolkit.jl/commit/8c274a76ca027d83df6a9196a58416093a872ab5 . See if that works for you.

arnavs:
Looks good to me.

PR #285, opened 2020-04-12
Pantelides algorithm

YingboMa:


test/index_reduction.jl
```diff
@@ -0,0 +1,63 @@
...
+        0 ~ x^2 + y^2 - L^2]
+pendulum2 = ODESystem(eqs2, t, [x, y, T], [L, g], name=:pendulum)
+lowered_sys = ModelingToolkit.ode_order_lowering(pendulum2)
+
+lowered_eqs = [D(x_t) ~ T*x,
+               D(y_t) ~ T*y - g,
+               0 ~ x^2 + y^2 - L^2,
+               D(x) ~ x_t,
+               D(y) ~ x_t]
+@test_skip ODESystem(lowered_eqs) == lowered_sys # not gonna work
```
ChrisRackauckas, line 22: why?

src/systems/diffeqs/index_reduction.jl
```diff
@@ -0,0 +1,162 @@
...
+            end
+        end
+    end
+
+    fullvars = [xvars; dxvars; algvars] # full list of variables
+    vars_asso = [(1:xvar_offset) .+ xvar_offset; zeros(Int, length(fullvars) - xvar_offset)] # variable association list
+    return edges, fullvars, vars_asso
+end
+
+print_bigraph(sys, vars, edges) = print_bigraph(stdout, sys, vars, edges)
```
ChrisRackauckas, line 54: would there be an advantage to using LightGraphs in here?

YingboMa: I don't think so. The graph algorithms here are pretty self-contained. Also, `print_bigraph` is only used for debugging.

test/index_reduction.jl
YingboMa: We cannot differentiate between time-dependent parameters and algebraic variables. So, `ODESystem(lowered_eqs)` construction is not going to work.

ChrisRackauckas: You shouldn't be using the DSL here.

ChrisRackauckas, line 22:
```suggestion
@test ODESystem(lowered_eqs,[x,y,x_t,y_t,T],[L, g]) == lowered_sys
```

test/index_reduction.jl
```diff
@@ -0,0 +1,63 @@
...
+@variables x(t) y(t) w(t) z(t) T(t) x_t(t) y_t(t)
+@derivatives D'~t
+
+eqs2 = [D(D(x)) ~ T*x,
+        D(D(y)) ~ T*y - g,
+        0 ~ x^2 + y^2 - L^2]
+pendulum2 = ODESystem(eqs2, t, [x, y, T], [L, g], name=:pendulum)
+lowered_sys = ModelingToolkit.ode_order_lowering(pendulum2)
+
+lowered_eqs = [D(x_t) ~ T*x,
```
ChrisRackauckas, line 17: that's the wrong ˍ

PR #289, opened 2020-04-13
add ODE ratelaw expressions

isaacsas:
@TorkelE @ChrisRackauckas feel free to send comments. We can use this branch to get the ODE rate law expressions in / designed. I'll try to add tests tomorrow to get this tested/finished. Some issues to handle / think about in the future:

1. One thing to think about / optimize now or later -- should we precalculate the net stoichiometry and reaction depedencies, and where should they be stored (`ReactionSystem`?) Net stoichiometry is a useful optimization even for ODEs (i.e. with reactions like A+B -> A+D, otherwise one gets redundant terms in the ODEs) Dependencies aren't needed here, but are needed for jumps...
2. We probably want a way to mimic the DEBio arrows that only use the rate (and ignore multiplying by the normal mass action terms), i.e. for users to specify that their rate is actually the full rate law they want to use and we shouldn't modify it at all. The issue is where this information should be provided (in the `Reaction`, in the `ReactionSystem`, etc).


src/systems/reaction/reactionsystem.jl
```diff
@@ -1,71 +1,97 @@
...
+    substrates::Vector{Operation}
     products::Vector{Operation}
+    substoich::Vector{T}
+    prodstoich::Vector{T}
 end
 
 struct ReactionSystem <: AbstractSystem
     eqs::Vector{Reaction}
     iv::Variable
-    states::Vector{Variable}
```
ChrisRackauckas, line 10: having these as states and ps will make it compatible with the other abstract system parameters and make components work

ChrisRackauckas: I'll get to documenting the AbstractSystem interface soon to clarify this.

ChrisRackauckas:
>One thing to think about / optimize now or later -- should we precalculate the net stoichiometry and reaction depedencies, and where should they be stored (ReactionSystem?) Net stoichiometry is a useful optimization even for ODEs (i.e. with reactions like A+B -> A+D, otherwise one gets redundant terms in the ODEs) Dependencies aren't needed here, but are needed for jumps...

I think we should do it when we construct the reactions. I think of it like this. `Reaction` should be about representations of the reactions, and `ReactionSystem` should be a designation of context over a set of reactions, like what are the states and parameters. 

>We probably want a way to mimic the DEBio arrows that only use the rate (and ignore multiplying by the normal mass action terms), i.e. for users to specify that their rate is actually the full rate law they want to use and we shouldn't modify it at all. The issue is where this information should be provided (in the Reaction, in the ReactionSystem, etc).

I am thinking we can do this by keeping a 3-argument constructor to Reaction, and that can probably be the one that most things interact with, while the many argument one is generally called from the 3-arg one that calculates the extra values.

src/systems/reaction/reactionsystem.jl
isaacsas: Ahh, ok I'll change that back tomorrow. Good to know.

isaacsas:
Are the `essemble_*` functions also part of the general `AbstractSystem` API? I only ask as I would change them to `assemble_*` instead.

ChrisRackauckas:
What are those ones?

ChrisRackauckas:
oh, oh haha. Those are not. Feel free to change those.

TorkelE:
> One thing to think about / optimize now or later -- should we precalculate the net stoichiometry and reaction depedencies, and where should they be stored (ReactionSystem?) Net stoichiometry is a useful optimization even for ODEs (i.e. with reactions like A+B -> A+D, otherwise one gets redundant terms in the ODEs) Dependencies aren't needed here, but are needed for jumps...

`
Given that Net stochiometry and dependencies fulfil the criteria

- Being a fundamental property of a reaction.
- Being useful in more than one downstream node (that is, both ODEs and Jumps will use the information).

I think they merit inclusion in the Reactant structure. But let us be vary of adding too much here, especially things which only makes sense in systems biology (but wouldn't for someone, say, using this to model infectious diseases).

> We probably want a way to mimic the DEBio arrows that only use the rate (and ignore multiplying by the normal mass action terms), i.e. for users to specify that their rate is actually the full rate law they want to use and we shouldn't modify it at all. The issue is where this information should be provided (in the Reaction, in the ReactionSystem, etc).

It makes sense to continue enabling both mass action and non-mass action kinetics (or, it makes even more sense than it used to do; since we are aiming to generalise even further). However, I see several ways to do this, which I will attempt to break down. Essentially it comes to two decisions:

- At which point in the process [Input System -> Reaction -> Equations] should this information be given.
- Should the information about this choice be stored somewhere?

So given that we have the reaction `a, x -->y` information might flow something like this:

1. A DSL or macro calls everything: `r = @reaction a, x --> y`. Here different types of arrows could indicate different options.
2. This would create a reaction by calling `Reaction(a,[x],[y],args...)`. Here args could be things such whenever to use mass action logics or not. This function could also be called directly by the user.
3. Finally, to use this, we need to create an ODEs or similar; `sys = convert(ODESystem,r,args...)`. Again, whenever this should use mass action kinetics or not could be stored as a field in `r`, or given in `args` when called.

(To simplify, this example ignores the distinction between reaction systems and reactions)

So we would have alternatives:

- When a reaction is created the choice is made (in the DSL, by choice of the arrow, in the direct call, by passing the information into `args`. The `rate` field would then be set to the corresponding rate (either `a` or `a*x`.). There is also the option to also store the information that this choice has been made in an additional field (for use by further algorithms). If so, one could let `rate=a`, even if mass action is used.
- Nothing is done at the level of a reaction (so `rate` is `a`). When, however, one does step 3 (to generate something useful) one would pass the information whenever to create ODEs from the reaction by using mass action or not. E.g. using `sys = convert(ODESystem,r,mass_action=true)`

I think the disadvantage with the second alternative is that it might be difficult to use if one wants the option only for some reactions, but not all. Then one would need a way to successively build up an `ODEsystem` by adding one reaction at a time. 

I would further argue, that if possible, we avoid creating a dedicated field in the `Reaction` struct, storing a boolean on whenever the reaction is a mass action type or not. If we need it, instead we should introduce a more general system with "flags", or similar, which could be given to a reaction with various pieces of information. Again, that would be of more general use. Unless one makes an argument that this is a fundamental property of these kinds of structures across all disciplines. I would accept that argument.

One thing, which I would like to achieve, and which relates to this, would be to have noise only present in parts of the system. E.g. to have a deterministic system, driven by a noisy one. One would first declare a reaction network. Then one would manually add a couple of reactions with noise. Again, unless we define a way to build SDEs in parts, this information would need to be passed to the reaction (and stored there). It doesn't feel like a fundamental enough property to merit its own field. But there should be some way to do this.

On the topic of SDEs, there should also be options for `sys = convert(SDESystem,r,args...)` to determine which rules to use to generate the SDEs. In systems biology, it would be CLE, but other disciplines might have their own rules (and even we would want an option for affine noise).

Sorry for the long post, reiterating a few things that have been said already, but to make my answer short: The information (on whenever to use mass action or not) should be given when a reaction is created, and store there in some very general form. We should also have some performant way of storing other pieces of niche information in the reaction, to be used later on (unless we want to enable people to build ODEs and similar piece by piece). 

If it would be performant to guarantee that all reactions in a system use mass action (or a similar property), we could have an option when a `ReactionSystem` is created to guarantee that a selected property is fulfilled for all reactions.

isaacsas:
Thanks for the thoughts. I did end up going with a boolean for the mass action property -- this seems like an intrinsic property of a reaction, and the general problem is we can't tell the user to give the full ODE rate law, which might include mass action already, since the rate laws for ODEs vs. jumps (as an example) are not the same. You can see how I handled it in the `oderatelaw` function, basically for any reaction when creating the full rate law, if it isn't mass action there is nothing done in the function.

I'm not sure how to allow reactions to have arbitrary information associated with them, beyond having them each contain a dictionary for stashing "all other info". Maybe Chris has suggestions. Perhaps we just require users to handle storing information beyond the basic set of fields we currently have? Or perhaps we can make an abstract reaction supertype and require that subtypes have at least the same fields as the concrete `Reaction` type we currently have. Users could then extend reactions with more fields as needed for their applications.

The species dependencies of a given reaction (i.e. those the full rate law depends on), are needed for jump methods, but this is essentially a type of graph (reaction -> dependent species). So perhaps that is better stored and calculated as a system property. It is really more a question of whether we want to calculate and store such info upon reaction construction, or defer till it is needed (say in jump creation).

ChrisRackauckas:
>The species dependencies of a given reaction (i.e. those the full rate law depends on), are needed for jump methods, but this is essentially a type of graph (reaction -> dependent species). So perhaps that is better stored and calculated as a system property. It is really more a question of whether we want to calculate and store such info upon reaction construction, or defer till it is needed (say in jump creation).

I think the graph to me is clear system property. The other stuff is a bunch of winging it until we get it right.

>On the topic of SDEs, there should also be options for sys = convert(SDESystem,r,args...) to determine which rules to use to generate the SDEs. In systems biology, it would be CLE, but other disciplines might have their own rules (and even we would want an option for affine noise).

Yeah, this will get documented as one of many transformations on a ReactionSystem, and I don't assume that the conversion to an SDE will be uniquely one method.

isaacsas:
For now it is CLE, but we can add options over time. This just needs tests, which I'll try to do later today, and I think it will be good to go as a first pass. It is a bit messier than it would otherwise be as I tried to minimize things like x^1 and -1*x. Later on as simplification tools improve we can get rid of this code and just (hopefully) call simplify on the expressions.

isaacsas:
@ChrisRackauckas Think this is ok now, but let me know if you have comments.

isaacsas:
One thing to mention. The reaction syntax needed a way to handle reactions with no substrates or no products. I set things up to assume the user passes `nothing` in that case, but we could change to taking an empty array instead. When `nothing` is passed for the substrate or product vector, I set the stoichiometry to be an empty vector of `Int`.

Project.toml
```diff
@@ -13,6 +13,7 @@ Latexify = "23fbe1c1-3f47-55db-b15f-69d7ec21a316"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 MacroTools = "1914dd2f-81c6-5fcd-8719-6d5c9610ff09"
 NaNMath = "77ba4419-2d1f-58cd-9bb1-8ffee604a2e3"
+Parameters = "d96e819e-fc66-5662-9728-84c9c7592b0a"
```
ChrisRackauckas, line 16: Directly use Unpack.jl

test/reactionsystem.jl
```diff
@@ -1,10 +1,68 @@
...
+sdesys = convert(SDESystem,rs)
+
+# hard coded ODE rhs
+function oderhs(u,k,t)
+       A = u[1]; B = u[2]; C = u[3]; D = u[4];
+       du = zeros(eltype(u),4)
+       du[1] = k[1] - k[3]*A + k[4]*C + 2*k[5]*C - k[6]*A*B + k[7]*B^2/2 - k[9]*A*B - k[10]*A^2 - k[11]*A^2/2 - k[12]*A*B^3*C^4/144 - k[13]*A/(2+A)
+       du[2] = -k[2]*B + k[4]*C - k[6]*A*B - k[7]*B^2 - k[8]*A*B - k[9]*A*B + k[11]*A^2/2 - 3*k[12]*A*B^3*C^4/144
+       du[3] = k[3]*A - k[4]*C - k[5]*C  + k[6]*A*B + k[8]*A*B + k[9]*A*B + k[10]*A^2/2 - 2*k[12]*A*B^3*C^4/144
+       du[4] = k[9]*A*B + k[10]*A^2/2 + 3*k[12]*A*B^3*C^4/144
```
ChrisRackauckas, line 30: awesome test

isaacsas: Sucked to write out though :)

Project.toml
isaacsas: I'll switch to Unpack and update.

test/reactionsystem.jl
ChrisRackauckas: that's why I'm so grateful you did it haha

ChrisRackauckas:
I see what you were saying about the reactions. I had `2x-> y` too directly, and that made it hard to actually use. Splitting the stoichiometry out makes a whole lot of sense, and it really is the job of a front end to get it there.

isaacsas:
Ok, UnPack should be fixed.

PR #294, opened 2020-04-13
More reasonable transformation variable names

ChrisRackauckas:
The recent changes that added a problem mapping interface really opened up our opportunities for backend variable naming options, because now we do not expect the users to ever have to know the names that we are creating. In order for sane debugging, it's good to create variable names that look nice and close enough, but the user doesn't interact directly with said symbols, so we are free to choose unicode characters. However, since Julia doesn't allow punctuation characters, we do have to make some concessions.

## Subsystem Referencing

Looking at http://xahlee.info/comp/unicode_math_operators.html, the ₊ looked the closest to a period, since we aren't allowed to use punctuation. Also it kind of makes sense because it's a form of unioning two symbols, so I think that works.

## Order Lowering

For lowering, I chose https://www.fileformat.info/info/unicode/char/02cd/index.htm since ˍ looks fairly close to _ but is weird enough that we have a vanishingly low probability of collisions.

### DIFF
--- a/src/ModelingToolkit.jl
+++ b/src/ModelingToolkit.jl
@@ -81,6 +81,18 @@ include("variables.jl")
 include("context_dsl.jl")
 include("operations.jl")
 include("differentials.jl")
+
+function Base.convert(::Type{Variable},x::Operation)
+    if x.op isa Variable
+        x.op
+    elseif x.op isa Differential
+        var = x.args[1].op
+        rename(var,Symbol(var.name,:ˍ,x.args[1].args[1].op.name))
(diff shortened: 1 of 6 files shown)


ChrisRackauckas:
@YingboMa

asinghvi17:
This makes 2nd order ODEs really annoying to give initial conditions for (since you do have to provide the initial value of the first derivative).

ChrisRackauckas:
Users shouldn't be directly specifying the names. `D(x)` should just work instead.

PR #304, opened 2020-04-14
treat symbols symbolically when converting from expressions

adediego:
Before when calling e.g. convert(Expression, :(x+y)), the values of the variables x and y were interpolated into the resulting expression from global scope. With this PR, they are treated as Variables.


ChrisRackauckas:
I remember now that the issue with this is that functions will be treated like variables if you do it like this.

adediego:
Aren't functions always the first element of the `args` field of an expression when they are called? Above in line 5 and 6, the convert function only descents into `args[2:end]` and `eval`s the first argument. Or am I missing a case where you want to treat a symbol that is not the first argument as a function?

ChrisRackauckas:
Oh I see, yes.

Issue #311, opened 2020-04-16
Allow providing parameters to ODEProblem as a Dict


asinghvi17:
Instead of a Vector{Pair}.  This makes working with Observables much easier, as one need only mutate the Dict and remake the problem instead of replacing an element of an array (and storing indices, etc, etc).

ChrisRackauckas:
Yes, that's a nice suggestion. We generally create the output based on the type of the input here, but in this case we could just assume the user wants an Array. The piece of code in this spot is:

https://github.com/SciML/ModelingToolkit.jl/blob/master/src/variables.jl#L242-L262

I think it would be enough to add a dispatch on Dict that creates the array and loops through the variables picking out the values.

ChrisRackauckas:
This is implemented.

PR #312, opened 2020-04-16
allow dict value maps

ChrisRackauckas:
Fixes https://github.com/SciML/ModelingToolkit.jl/issues/311

### DIFF
--- a/src/variables.jl
+++ b/src/variables.jl
@@ -101,7 +101,7 @@ Base.convert(::Type{Expr}, c::Constant) = c.value
 
 # For dictionary lookup
 Base.hash(v::Variable,h) = hash(Symbol(v),h)
-
+Base.hash(v::Variable, u::UInt64) = hash(Symbol(v), u)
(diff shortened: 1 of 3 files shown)


src/variables.jl
```diff
@@ -99,6 +99,10 @@ Base.isequal(a::Constant, b::Constant) = a.value == b.value
 
 Base.convert(::Type{Expr}, c::Constant) = c.value
 
+# For dictionary lookup
+Base.hash(v::Variable,h) = hash(Symbol(v),h)
+
```
asinghvi17, line 104:
```suggestion
Base.hash(v::Variable, u::UInt64) = hash(Symbol(v), u)
```
seems to avoid the method ambiguity, though there really should be a better solution.

ChrisRackauckas: yeah, that's why I stopped this WIP. There's a real design decision hiding in here for handling some of the cases.

PR #316, opened 2020-04-17
Fix typo

ChrisDeGrendele:
generate_jacobian takes an ODE and sys is a ::Tuple{ODESystem,Array{Operation,1},Array{Operation,1}}
=> pass sys[1].


ChrisRackauckas:
We should probably change modelingtoolkitize to just return the `sys`: it doesn't make sense to return the other stuff anymore since they are fields of the sys. Any objection @YingboMa ?

YingboMa:
Yes, I agree. I will open a PR to fix this.

YingboMa:
I think this has been fixed.

PR #326, opened 2020-04-25
SymbolicUtils integration

shashi:
Tests pass locally. Needs a release of the SymbolicUtils package (might rename), but putting it out there for comments and testing!


src/direct.jl
```diff
@@ -50,6 +50,7 @@ function simplified_expr(O::Operation)
 end
 
 simplified_expr(c::Constant) = c.value
+simplified_expr(c) = c
```
ChrisRackauckas, line 53: what's this needed for?

src/utils.jl
```diff
@@ -114,3 +114,6 @@ function substitute_expr!(expr::Operation, s::Pair{Operation, Operation})
     end
     return nothing
 end
+
+to_Expr(x::Expression) = convert(Expr, x)
+to_Expr(x::Number) = to_Expr(Constant(x))
```
ChrisRackauckas, line 119: that shouldn't be needed? An overload of `convert(Expr,x::Constant)` handles it?

test/simplify.jl
```diff
@@ -14,17 +14,18 @@ identity_op = Operation(identity,[x])
...
+@test simplified_expr(simplify_constants(x+2+3)) == :(5 + x)
 
-d1 = Differential(x)((x-2)^2)
+d1 = Differential(x)((-2 + x)^2)
 d2 = Differential(x)(d1)
 d3 = Differential(x)(d2)
+
 @test simplified_expr(expand_derivatives(d3)) == :(0)
-@test simplified_expr(simplify_constants(x^0)) == :(1)
+#@test simplify_constants(x^0) == 1
```
ChrisRackauckas, line 31: please don't comment out tests. First of all, why would this break? But secondly, mark it `@test_broken` instead.

shashi: it's not broken. it's just the wrong test for this file.

src/utils.jl
shashi: I tried returning only `Constant`s from `simplify_constants` but a number of things don't work. For example line 13 in test/mass_matrix.jl. It seems you weren't doing `simplify_constants` on the final expression in `expand_derivatives` and that's why it happened to work. So now since sometimes it returns numbers, I had to handle all these little cases.

src/direct.jl
shashi: Sometimes `c` is a number.

src/utils.jl
ChrisRackauckas: but is that just a missing method `convert(::Type{Expression},x::Number) = Constant(x)`?

shashi: No, actually you are converting to `Expr` not `Expression`.

shashi: I will try once more to return `Constant` instead of numbers when the result of simplification is a Constant.

ChrisRackauckas: I see, but what's wrong with the default of `convert(Expr,::Number)`?

shashi: problems come from `==` being an Operation if an argument is a Constant.

shashi: There isn't a default.

src/ModelingToolkit.jl
```diff
@@ -58,6 +58,7 @@ include("variables.jl")
 include("context_dsl.jl")
 include("operations.jl")
 include("differentials.jl")
+include("symutils.jl")
```
YingboMa, line 61: I didn't find the `symutils.jl` file in the diff. Did I miss it?

shashi: removed.

test/simplify.jl
YingboMa: Where should we add this test?

shashi: I'm still messing with the return type being `Constant`, this test will change, I'll move it.

src/utils.jl
shashi: This is not needed anymore. :+1:

PR #328, opened 2020-04-30
Enable reaction without substrates or products

TorkelE:
After getting some weird versioning issues out of the way I started looking at integrating ModelingToolkit into DiffEqBio. However, there seemed to be some basic things that didn't work yet when creating ODE models, so figured we should ensure that worked in MTK first.

It is not unlikely that I have no idea what I am doing here, but when trying I was unable to create reactions without either substrates or products, these lines was problematic:
```
    subsv  = isnothing(subs) ? Vector{Operation}() : subs
    prodsv = isnothing(prods) ? Vector{Operation}() : prods
```
When the input became `[]` it was not nothing, and was passed on and caused an error. I changed it to
```
    subsv  = (isempty(subs)||isnothing(subs)) ? Vector{Operation}() : subs
    prodsv = (isempty(prods)||isnothing(prods)) ? Vector{Operation}() : prods
```
and now there is at least no error here.
(Maybe `isnothing(...)` should also be removed, but I am afraid to make to much change in something I do not understand)

Since I still do not know exactly what is going on here this might not actually be an error, and might all be working as intended. If so feel free to scrap this without mercy. At least it has served the purpose of setting up so I can make pull requests to MTK from my computer vie GitKraken, something I had mostly forgotten.

### DIFF
--- a/src/systems/reaction/reactionsystem.jl
+++ b/src/systems/reaction/reactionsystem.jl
@@ -3,22 +3,22 @@ struct Reaction{S <: Variable, T <: Number}
     substrates::Vector{Operation}
     products::Vector{Operation}
     substoich::Vector{T}
-    prodstoich::Vector{T}    
+    prodstoich::Vector{T}
     netstoich::Vector{Pair{S,T}}
     only_use_rate::Bool
 end
 
-function Reaction(rate, subs, prods, substoich, prodstoich; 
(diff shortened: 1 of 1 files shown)


TorkelE:
Apparently the syntax is `nothing` instead of `[]`.

PR #330, opened 2020-05-02
Allow reactions with: Multiple copies of one components and either no substrates or no products

TorkelE:
Was working with stuff on the DiffEqBio side of things, and it seems like certain combinations of empty reactions, with more than one of some components don't work, example:
```julia
using ModelingToolkit
@parameters t k
@variables X(t) Y(t)
Reaction(k, [X], nothing)                       #  Works
Reaction(k, nothing, [X])                       # Works
Reaction(k, [X], [Y], [2], [1])                  # Works
Reaction(k, [X], nothing, [2])                 # Don't work work.
Reaction(k, [X], nothing, [2], [])             # Don't work work.
Reaction(k, [X], nothing, [2], nothing)    # Don't work work.
Reaction(k, nothing, [X], [2])                  # Don't work work.
Reaction(k, nothing, [X], [], [2])              # Don't work work.
Reaction(k, nothing, [X], nothing, [2])    # Don't work work.
```
Possibly it is again me who aren't familiar with the syntax yet. But if not I guess we should fix it. As a started I made some small changes so that
```
Reaction(k, [X], nothing, [2], nothing)
Reaction(k, nothing, [X], nothing, [2]) 
```
should work. In addition,
```
Reaction(k, [X], nothing, [2], [])    
Reaction(k, nothing, [X], [], [2])     
```
works as a by-product. Finally, 
```
Reaction(k, [X], nothing, [2])       
Reaction(k, nothing, [X], [2])        
```
is not allowed.

If things make sense, feel free to change the coding to something you prefer. If it does make sense, I am happy to add it to tests as well.

### DIFF
--- a/src/systems/reaction/reactionsystem.jl
+++ b/src/systems/reaction/reactionsystem.jl
@@ -3,22 +3,22 @@ struct Reaction{S <: Variable, T <: Number}
     substrates::Vector{Operation}
     products::Vector{Operation}
     substoich::Vector{T}
-    prodstoich::Vector{T}    
+    prodstoich::Vector{T}
     netstoich::Vector{Pair{S,T}}
     only_use_rate::Bool
 end
 
-function Reaction(rate, subs, prods, substoich, prodstoich; 
+function Reaction(rate, subs, prods, substoich, prodstoich;
                   netstoich=nothing, only_use_rate=false, kwargs...)
 
-    subsv  = isnothing(subs) ? Vector{Operation}() : subs
-    prodsv = isnothing(prods) ? Vector{Operation}() : prods
-    ns = isnothing(netstoich) ? get_netstoich(subsv, prodsv, substoich, prodstoich) : netstoich
-    Reaction(rate, subsv, prodsv, substoich, prodstoich, ns, only_use_rate)
+    subsv,substoichv  = isnothing(subs) ? (Vector{Operation}(),Int[]) : (subs,substoich)
+    prodsv,prodstoichv = isnothing(prods) ? (Vector{Operation}(),Int[]) : (prods,prodstoich)
+    ns = isnothing(netstoich) ? get_netstoich(subsv, prodsv, substoichv, prodstoichv) : netstoich
+    Reaction(rate, subsv, prodsv, substoichv, prodstoichv, ns, only_use_rate)
 end
 
 # three argument constructor assumes stoichiometric coefs are one and integers
-function Reaction(rate, subs, prods; kwargs...) 
+function Reaction(rate, subs, prods; kwargs...)
 
     sstoich = isnothing(subs) ? Int[] : ones(Int,length(subs))
     pstoich = isnothing(prods) ? Int[] : ones(Int,length(prods))
@@ -54,18 +54,18 @@ end
 function ReactionSystem(eqs, iv, species, params; systems = ReactionSystem[],
                                                   name = gensym(:ReactionSystem))
 
-    ReactionSystem(eqs, iv, convert.(Variable,species), convert.(Variable,params), 
+    ReactionSystem(eqs, iv, convert.(Variable,species), convert.(Variable,params),
                    name, systems)
 end
 
 # Calculate the ODE rate law
 function oderatelaw(rx)
-    @unpack rate, substrates, substoich, only_use_rate = rx    
+    @unpack rate, substrates, substoich, only_use_rate = rx
     rl = rate
     if !only_use_rate
         coef = one(eltype(substoich))
         for (i,stoich) in enumerate(substoich)
-            coef *= factorial(stoich)        
+            coef *= factorial(stoich)
             rl   *= isone(stoich) ? substrates[i] : substrates[i]^stoich
         end
         (!isone(coef)) && (rl /= coef)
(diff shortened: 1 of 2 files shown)


TorkelE:
@isaacsas @ChrisRackauckas

src/systems/reaction/reactionsystem.jl
```diff
@@ -3,22 +3,22 @@ struct Reaction{S <: Variable, T <: Number}
...
 
-function Reaction(rate, subs, prods, substoich, prodstoich; 
+function Reaction(rate, subs, prods, substoich, prodstoich;
                   netstoich=nothing, only_use_rate=false, kwargs...)
 
-    subsv  = isnothing(subs) ? Vector{Operation}() : subs
-    prodsv = isnothing(prods) ? Vector{Operation}() : prods
-    ns = isnothing(netstoich) ? get_netstoich(subsv, prodsv, substoich, prodstoich) : netstoich
-    Reaction(rate, subsv, prodsv, substoich, prodstoich, ns, only_use_rate)
+    subsv,substoichv  = isnothing(subs) ? (Vector{Operation}(),Int[]) : (subs,substoich)
```
isaacsas, line 14: Should we add a bit more logic to set the `eltype` of `substoichv` to be that of `prodstoichv` instead of defaulting to integers (and vice-versa if `prodstoichv` is the one that is `nothing`)? If both are `nothing` we could give an error. (I'm trying to think ahead to allow non-integer stoichiometry as was previously requested in DEBio.)

isaacsas:
Aside from my comment about types it looks good to me. I was imagining only having the 3 and 5 argument versions. Adding tests sounds good. You could just add some more reactions onto the example I made.

src/systems/reaction/reactionsystem.jl
TorkelE: That is a good point.

TorkelE:
Some changes.

As a consequence of the update, this will not work anymore:
```
Reaction(k[1], [X], nothing, [2], [])    
Reaction(k[1], nothing, [X], [], [2])  
```

I have also added a warning for when:
- Both the prod and subs vector is nothing.
- Both stoichiometry vectors are nothing.
- When a non-nothing stochiometry vector is accompanied by a corresponding nothing substrate or product vector.

Finally, if one of the stoichiometry vectors are nothing, and the other one is of Float64 (or whatever), the nothing one is converted to an empty vector of the same type.

Finally: To simplify the code with the changes I skipped the steps in between where a `subsv` and `prodsv` vector was created (as well as the `substoichv` and `prodstoichv` I created). These are no longer made, I don't think it should affect things, although you should be aware to ensure that I am not messing something up. Admittedly
```
du2 = sf.f(u,p,t)
@test norm(du-du2) < 100*eps()
G2 = sf.g(u,p,t)
@test norm(G-G2) < 100*eps()
```
now starts to generate errors, however, reverting to using `subsv` and `prodsv` does not seem to change this, so I do not think this is it.

TorkelE:
Ok, the tests should hopefully be fixed now. I have updated them, and the non-reaction functions with these two reactions:
```
Reaction(k[13], [A,B], nothing, [3,1], nothing), # 3A+B -> 0
Reaction(k[14], nothing, [A], nothing, [2]), # 0 -> 2A
```

src/systems/reaction/reactionsystem.jl
```diff
@@ -3,25 +3,38 @@ struct Reaction{S <: Variable, T <: Number}
...
+        prods = Vector{Operation}()
+        (prodstoich!=nothing) && error("If products are nothing, product stiocihometries have to be so too.")
+        prodstoich = (substoich == nothing) ? nothing : typeof(substoich)()
+    end
+    ns = isnothing(netstoich) ? get_netstoich(subs, prods, substoich, prodstoich) : netstoich
+    Reaction(rate, subs, prods, substoich, prodstoich, ns, only_use_rate)
 end
 
+a =[1.,2.3,4.]
+typeof(a)()
```
ChrisRackauckas, line 31:
```suggestion
```

test/reactionsystem.jl
```diff
@@ -1,11 +1,11 @@
 using ModelingToolkit, LinearAlgebra, Test
 
-@parameters t k[1:13]
+@parameters t k[1:15]
```
ChrisRackauckas, line 3: was this a specific example @isaacsas or was it just trying to cover all of the cases?

isaacsas: Just covering as many standard cases as I could think of

src/systems/reaction/reactionsystem.jl
```diff
@@ -3,25 +3,36 @@ struct Reaction{S <: Variable, T <: Number}
...
 end
 
+
 # three argument constructor assumes stoichiometric coefs are one and integers
-function Reaction(rate, subs, prods; kwargs...) 
+function Reaction(rate, subs, prods; kwargs...)
 
-    sstoich = isnothing(subs) ? Int[] : ones(Int,length(subs))
-    pstoich = isnothing(prods) ? Int[] : ones(Int,length(prods))
+    sstoich = isnothing(subs) ? nothing : ones(Int,length(subs))
```
isaacsas, line 34: @ChrisRackauckas whst do you think is the better default in this case? Using `nothing’s or an empty array of `ints`?

ChrisRackauckas: `Int[]` is a little bit cleaner, but this is probably going to be a lot safer in the end. I'd prefer `nothing`.

src/systems/reaction/reactionsystem.jl
```diff
@@ -3,25 +3,36 @@ struct Reaction{S <: Variable, T <: Number}
...
-    subsv  = isnothing(subs) ? Vector{Operation}() : subs
-    prodsv = isnothing(prods) ? Vector{Operation}() : prods
-    ns = isnothing(netstoich) ? get_netstoich(subsv, prodsv, substoich, prodstoich) : netstoich
-    Reaction(rate, subsv, prodsv, substoich, prodstoich, ns, only_use_rate)
+      (isnothing(prods)&&isnothing(subs)) && error("A reaction requires a non-nothing substrate or product vector.")
+      (isnothing(prodstoich)&&isnothing(substoich)) && error("Both substrate and product stochiometry inputs cannot be nothing.")
+      if isnothing(subs)
+        subs = Vector{Operation}()
+        (substoich!=nothing) && error("If substrates are nothing, substrate stiocihometries have to be so too.")
+        substoich = (prodstoich == nothing) ? nothing : typeof(prodstoich)()
```
isaacsas, line 19: Why do you need this check `prodstoich == nothing`? Hasn't that already been ruled by here?

src/systems/reaction/reactionsystem.jl
```diff
@@ -3,25 +3,36 @@ struct Reaction{S <: Variable, T <: Number}
...
+    end
+    if isnothing(prods)
+        prods = Vector{Operation}()
+        (prodstoich!=nothing) && error("If products are nothing, product stiocihometries have to be so too.")
+        prodstoich = (substoich == nothing) ? nothing : typeof(substoich)()
```
isaacsas, line 24: Similar comment to above.

TorkelE: Good point (on both), fixed.

isaacsas:
Looks good to me now. Thanks!

PR #332, opened 2020-05-03
SDEProblem from SDESystem fix

isaacsas:
Seemed to be a typo in `SDEProblem` construction.


isaacsas:
@ChrisRackauckas  Would it make sense to also add:
```julia
function DiffEqBase.SDEProblem(sys::AbstractODESystem, args...; kwargs...)
    SDEProblem{true}(sys, args...; kwargs...)
end
```
similar to what you have for `ODEProblem`s?

isaacsas:
Sorry, I mean
```julia
function DiffEqBase.SDEProblem(sys::AbstractSDESystem, args...; kwargs...)
    SDEProblem{true}(sys, args...; kwargs...)
end
```

ChrisRackauckas:
Oh yes, that would make sense.

isaacsas:
Will submit shortly

PR #336, opened 2020-05-05
add mass action jump support

isaacsas:
~Still needs tests for zero order, second order and decay reactions.~ Should be set now.

### DIFF
--- a/src/systems/jumps/jumpsystem.jl
+++ b/src/systems/jumps/jumpsystem.jl
@@ -44,8 +44,13 @@ end
 
 function assemble_maj(js, maj::MassActionJump{U,Vector{Pair{V,W}},Vector{Pair{V2,W2}}}, 
                       statetoid, ptoid, p, pcontext) where {U,V,W,V2,W2}
-    pval = Base.eval(pcontext, Expr(maj.scaled_rates))
-
+    sr = maj.scaled_rates
+    if sr isa Operation || sr isa Variable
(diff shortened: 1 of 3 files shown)


isaacsas:
Also needs a new release of DiffEqJump with the iszero fix.

src/systems/jumps/jumpsystem.jl
```diff
@@ -42,6 +42,29 @@ function assemble_crj(js, crj, statetoid)
     ConstantRateJump(rate, affect)
 end
 
+function assemble_maj(js, maj::MassActionJump{U,Vector{Pair{V,W}},Vector{Pair{V2,W2}}}, 
+                      statetoid, ptoid, p, pcontext) where {U,V,W,V2,W2}
+    pval = Base.eval(pcontext, Expr(maj.scaled_rates))
```
ChrisRackauckas, line 47: I assume this is the evaluation?

isaacsas: Yeah

isaacsas: If you have suggestions on a better way to do this please do let me know...

ChrisRackauckas: We can go with it for now while we work on a better solution.

isaacsas:
@ChrisRackauckas Assuming tests pass I think this is good to go.

Issue #339, opened 2020-05-06
Link to SymbolicUtils Substitution


ChrisRackauckas:
When https://github.com/JuliaSymbolics/SymbolicUtils.jl/issues/59 is complete, we should make our substitution tools all make use of it. This includes the substitutions in the jumps (https://github.com/SciML/ModelingToolkit.jl/pull/336/files#diff-4991951911e82642fe59ec98a772f888R47).

@shashi @isaacsas

isaacsas:
Does MT allow Dicts now for the u0 and param maps?

ChrisRackauckas:
Not yet. I need to figure out how to do the checking/hashing: https://github.com/SciML/ModelingToolkit.jl/pull/312

shashi:
```julia
function substitute(x, dict)
    dict1 = Dict(map(Pair,
                       map(to_symbolic, keys(dict)),
                       map(to_symbolic, values(dict))))
    RuleSet([@rule(~x::(x->haskey(dict, x)) => dict[~x])])(to_symbolic(x)) |> to_mtk
end
```

Can be used now.

ChrisRackauckas:
Is that how we should do it? Like, should we just get a PR of this and replace https://github.com/SciML/ModelingToolkit.jl/blob/master/src/utils.jl#L101-L116 ?

shashi:
Yes that would be the way to do it. And you won't even need #312 since I think Symutils will hash converted MTK IR correctly.

And you may want to throw in a `simplify` before `to_mtk` so you can simplify before converting.

PR #342, opened 2020-05-06
substitute using SymbolicUtils

ChrisRackauckas:
Fixes https://github.com/SciML/ModelingToolkit.jl/issues/339

MWE:

```julia
using ModelingToolkit
@parameters α β γ δ
s = α => γ
expr = (((1 / β - 1) + δ) / α) ^ (1 / (α - 1))
ModelingToolkit.substitute(expr, s)
new  = (((1 / β - 1) + δ) / γ) ^ (1 / (γ - 1))
isequal(expr, new)
```


src/direct.jl
```diff
@@ -34,7 +34,7 @@ an array of variable expressions.
 """
 function hessian(O::Expression, vars::AbstractVector{<:Expression}; simplify = true)
     out = [expand_derivatives(Differential(v2)(Differential(v1)(O))) for v1 in vars, v2 in vars]
-    simplify ? simplify_constants.(out) : out
+    simplify ? simplify.(out) : out
```
shashi, line 37: Don't think this will work :)

src/simplify.jl
```diff
@@ -38,10 +35,12 @@ SymbolicUtils.symtype(x::Expression) = Number
 
 # SymbolicUtils -> ModelingToolkit
 
-function simplify_constants(expr)
+function SymbolicUtils.simplify(expr::Expression)
```
shashi, line 38: Nice!

src/direct.jl
ChrisRackauckas: why not?

ChrisRackauckas:
@isaacsas this'll handle the maj eval

shashi:
https://github.com/JuliaRegistries/General/pull/14285 has and `substitute` that gets used here. We can add more smarts/perf improvements over at SymUtils.

Project.toml
```diff
@@ -37,7 +37,7 @@ NaNMath = "0.3"
 SafeTestsets = "0.0.1"
 SpecialFunctions = "0.7, 0.8, 0.9, 0.10"
 StaticArrays = "0.10, 0.11, 0.12"
-SymbolicUtils = "0.1.1, 0.2"
+SymbolicUtils = "0.2.1, 0.2"
```
shashi, line 40: not sure if I did this right, so weird travis has no logs for the failure.

ChrisRackauckas:
@shashi is this supposed to work with SymbolicUtils master?

```julia
using ModelingToolkit
@parameters α β γ δ
s = α => γ
expr = (((1 / β - 1) + δ) / α) ^ (1 / (α - 1))
ModelingToolkit.substitute(expr, s)
new  = (((1 / β - 1) + δ) / γ) ^ (1 / (γ - 1))
isequal(expr, new)
```

still fails for me.

shashi:
No that's not supposed to work... `isequal` only checks the structure, it doesn't know that the two names are equal. That would require contextual simplification something like:

`simplify(expr == new, Constraints([α == γ]))`

isaacsas:
@ChrisRackauckas @shashi This PR seems to have broken MassActionJumps. It appears test passed as I forgot to add the jump tests to the runtest file... Sorry about that.

There seem to be two issues: first the number of arguments in the call to `assemble_maj` is wrong as `ptoid` was not deleted from the function signature. Even fixing this though I'm still having an issue. In `assemble_maj` `pval` needs to be a numeric value (i.e. Float64 or such). The previous code was evaluating the full rate expression to generate this numeric value from the parameter vector. On the latest master and after removing the extra `ptoid` argument, for me the new code seems to return an MT expression for `pval`, instead of substituting in the numeric values from the parameter vector and then evaluating the numeric value of `sr`.

shashi:
@isaacsas oops, sorry about that! do you think you can give me a small test that I can try to fix?

shashi:
Ah ok, I just saw your comment on that PR, thanks!

isaacsas:
Awesome! If you just go from that PR you can run the JumpSystem tests to check what’s going on.

PR #346, opened 2020-05-08
Enable creation of JumpSystems from ReactionSystems

TorkelE:
- convert(JumpSystem,rs::ReactionSystem) should now work.
- Type of Jump (Variable Rate, ConstantRate, MassAction) should now be correctly identified.

### DIFF
--- a/src/systems/reaction/reactionsystem.jl
+++ b/src/systems/reaction/reactionsystem.jl
@@ -147,7 +147,7 @@ function assemble_jumps(rs)
 end
 
 # Calculate the Jump rate law (like ODE, but uses X instead of X(t).
-# The former geenrates a "MethodError: objects of type Int64 are not callable" when trying to solve the problem.
+# The former generates a "MethodError: objects of type Int64 are not callable" when trying to solve the problem.
 function jumpratelaw(rx)
     @unpack rate, substrates, substoich, only_use_rate = rx
     rl = deepcopy(rate)
--- a/src/systems/reaction/reactionsystem.jl
+++ b/src/systems/reaction/reactionsystem.jl
@@ -121,6 +121,10 @@ function assemble_diffusion(rs)
     eqs
 end
 
+function assemble_jumps(rs)
+
+end
+
 function Base.convert(::Type{<:ODESystem},rs::ReactionSystem)
     eqs = assemble_drift(rs)
     ODESystem(eqs,rs.iv,rs.states,rs.ps,name=rs.name,
@@ -132,4 +136,10 @@ function Base.convert(::Type{<:SDESystem},rs::ReactionSystem)
     noiseeqs = assemble_diffusion(rs)
     SDESystem(eqs,noiseeqs,rs.iv,rs.states,rs.ps,
               name=rs.name,systems=convert.(SDESystem,rs.systems))
+end
+
+function Base.convert(::Type{<:JumpSystem},rs::ReactionSystem)
+    eqs = assemble_jumps(rs)
+    JumpSystem(eqs,rs.iv,rs.states,rs.ps,name=rs.name,
+              systems=convert.(JumpSystem,rs.systems))
 end
--- a/src/systems/reaction/reactionsystem.jl
+++ b/src/systems/reaction/reactionsystem.jl
@@ -135,8 +135,8 @@ function assemble_jumps(rs)
         elseif rx.only_use_rate || any([isequal(state,r_op) for state in rs.states, r_op in getfield.(get_variables(rx.rate),:op)])
             push!(eqs,ConstantRateJump(rl,affect))
         else
-            reactant_stoch = Pair.(var2op.(getfield.(rx.substrates,:op)),rx.substoich)
-            net_stoch = isempty(rx.substoich) ? [0 => 1]  : map(p -> Pair(var2op(p[1]),p[2]),rx.netstoich)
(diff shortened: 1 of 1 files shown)


TorkelE:
Not done, but should be in a workable state now where one can try out the code. I haven't started on tests yet.

Some irregularities:
- I had to create a separate `jumpratelaw()`, instead of using the `oderatelaw()`. The reason for this was that the later generated rates on the form `A*X(t)*Y(t)`, these generated `MethodError: objects of type Int64 are not callable` when trying to solve the problem. If I replaced it instead with  `A*X*Y` things looked like in the normal `JumpSystems`, and the solving worked. I am not sure what exactly is going on here, and I might have taken the wrong approach (and if not, there is likely a prettier way to solve it, but I didn't want to intrude into other function just yet).
- Relating to this, at several places I had to go from a `Variable` to an `Operation`, and so created a `var2op()` function (didn't find one already present). Again, I am very uncertain whenever I here have used the best approach but will leave it to someone more involved to decide.

TorkelE:
A simple example of usage
```julia
using DiffEqJump, ModelingToolkit
@parameters t p d
@variables X(t)
rxs = [Reaction(p, nothing, [X]),
       Reaction(d, [X], nothing)
       ]
rs = ReactionSystem(rxs,t,[X],[p,d])

js = convert(JumpSystem,rs)
dprob = DiscreteProblem(js, [X => 10], (0,10.), [p =>20., d=>1.])
sol = solve(jprob, SSAStepper())
using Plots; plot(sol)
```

ChrisRackauckas:
>I had to create a separate jumpratelaw(), instead of using the oderatelaw(). The reason for this was that the later generated rates on the form A*X(t)*Y(t), these generated MethodError: objects of type Int64 are not callable when trying to solve the problem. If I replaced it instead with A*X*Y things looked like in the normal JumpSystems, and the solving worked. I am not sure what exactly is going on here, and I might have taken the wrong approach (and if not, there is likely a prettier way to solve it, but I didn't want to intrude into other function just yet).

It would be nice to figure out why that's necessary. Something probably isn't using the right expression generation from variable setup: https://github.com/SciML/ModelingToolkit.jl/blob/master/src/systems/diffeqs/abstractodesystem.jl#L23-L37

ChrisRackauckas:
I think we can make this better, but let's iteratively improve it.

TorkelE:
There seems to be something fishy going on now. When I try the code again
```julia
using DiffEqJump, ModelingToolkit
@parameters t p d
@variables X(t)
rxs = [Reaction(p, nothing, [X]),
       Reaction(d, [X], nothing)
       ]
rs = ReactionSystem(rxs,t,[X],[p,d])

js = convert(JumpSystem,rs)
dprob = DiscreteProblem(js, [X => 10], (0,10.), [p =>20., d=>1.])
sol = solve(jprob, SSAStepper())
using Plots; plot(sol)
```
I get an 
```
MethodError: no method matching assemble_maj(::JumpSystem, ::MassActionJump{Operation,Array{Pair{Int64,Int64},1},Array{Pair{Operation,Int64},1}}, ::Dict{Variable{Number},Int64}, ::Array{Pair{Variable{ModelingToolkit.Parameter{Number}},Float64},1})
```
when trying to create the `JumpProblem`. If I go back to the branch, before it was merged though, it is all fine. I don't know if it was something that was added between the fork and the merge which changed things up.

In general, there seems to be a problem creating creating `JumpProblem`s  from `MassActionJump`s now.

ChrisRackauckas:
Yeah, something happened with the `substitute` stuff. @shashi could we get a SymbolicUtils.jl tag in the next hour or so?

ChrisRackauckas:
It's UnPack.jl bounds. Things need to propagate through.

isaacsas:
@ChrisRackauckas The JumpSystem crash seems to still be an issue.

Also, @TorkelE the SSA rate law should be different from the ODE rate law; the former should use binomials (see `mass_rate_SSA` in DEBio or the docs). (If I've misunderstood your code I apologize.)

TorkelE:
Oh yeah, totally forgot about those.

TorkelE:
Ok, I will wait with correcting until the JumpSystems run, to avoid the risk of adding errors when errors already present. 

Also, @isaacsas for `MassActionJumps`, what should the rate be (the actual rate, or just the constant,)? That is, for `A, 2X + Y --> ...` the actual rate is `A*binomial(X,2)*Y`, but am I correct that `MassActionJump` deduces that by itself, and only `A` should be passed into it?

isaacsas:
It should be the parameter that represents the rate constant, or any expression involving just parameters (i.e. `5*k1*k2+k3` would be fine). It just can't involve `t` or any other species variables. This should match the current DEBio functionality too.

TorkelE:
Got, thanks.
(Currently when I created a MassActionJump with the input rate `A` for the reaction `A, 2X --> ` the rate recorded in the structure seemed to be `A/2`, which got be start thinking about it, and I wanted to make sure.)

isaacsas:
That’s what it should be!

isaacsas:
The full rate expression will be A*X*(X-1)/2.

isaacsas:
@TorkelE Just in case my answer wasn't completely clear, since I think I missed what you were actually getting at, the `MassActionJump` constructor, by default, adds the combinatoric scaling to the rate, see:
https://github.com/SciML/DiffEqJump.jl/blob/6f41387f472999deb54a1a6695847f19a90f11df/src/jumps.jl#L70

This is set by the constructor's `scale_rates` kwarg, which is `true` by default. I think this is the version we use through DEBio too.

PR #347, opened 2020-05-08
remove to_mtk in substitute

shashi:
you don't need it if you're calling ModelingToolkit.substitute.

this was causing results to be wrapped in a `Constant` because `Expression <: Number` and there's a `to_mtk(x::Number) = Constant(x)` method.

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -116,7 +116,7 @@ function _substitute(expr, ks, vs)
 end
 
 function _substitute(expr, dict::Dict)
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
What problem does this solve?

shashi:
before:
```julia
julia> substitute(a+b, b=>1)
Constant(a+1)
```
after:
```julia
julia> substitute(a+b, b=>1)
a+1
```

PR #349, opened 2020-05-08
Let SymbolicUtils figure out types of expressions

shashi:
This allows

```julia
julia> expr = (((1 / β - 1) + δ) / α) ^ (1 / (α - 1))
(((1 / β - 1) + δ) / α) ^ (1 / (α - 1))

julia> expr2  = (((1 / β - 1) + δ) / γ) ^ (1 / (γ - 1))
(((1 / β - 1) + δ) / γ) ^ (1 / (γ - 1))

julia> substitute(expr == expr2, α => γ)
ModelingToolkit.Constant(true)
```

Because symutils applies bool rules only on expressions of type `Bool`.

### DIFF
--- a/src/simplify.jl
+++ b/src/simplify.jl
@@ -29,16 +29,14 @@ SymbolicUtils.symtype(x::Variable) = _vartype(x) # needed for a()
 SymbolicUtils.symtype(x::SymbolicUtils.Sym{<:Parameter{T}}) where {T} = T
 
 # returning Any causes SymbolicUtils to infer the type using `promote_symtype`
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
Is this failure understood?

shashi:
Yes https://github.com/JuliaSymbolics/SymbolicUtils.jl/issues/85

I worked on a total order, but it's kinda slow so I'm trying to make it fast or alternatively will provide a way to test equality despite ordering like you suggested.

shashi:
closed in favor of #518

PR #354, opened 2020-05-10
Fix MassActionJump parameter evaluation

isaacsas:
@ChrisRackauckas @shashi This fixes the function argument error to calling `assemble_maj` and adds the `JumpSystem` tests, but the issue with not generating a numerical value for `pval` remains.


ChrisRackauckas:
@TorkelE you should be good with this.

isaacsas:
@ChrisRackauckas thanks!

PR #355, opened 2020-05-11
keep the same result, do just 1 simplification

ChrisRackauckas:


ChrisRackauckas:
@shashi any reason this wasn't a PR?

shashi:
hah, just wanted to run CI first, you found it before I got back to it.

PR #357, opened 2020-05-11
Fix to making JumpSystems from ReactionSystems

TorkelE:
- `convert(JumpSystem,rn:ReactionSystem)` should now work properly. (`var2op()` still required).
- Binomials should be used correctly when `ConstantRateJump`s and `VariableRateJump`s are created.

### DIFF
--- a/src/systems/reaction/reactionsystem.jl
+++ b/src/systems/reaction/reactionsystem.jl
@@ -148,14 +148,13 @@ end
 function jumpratelaw(rx)
     @unpack rate, substrates, substoich, only_use_rate = rx
     rl = deepcopy(rate)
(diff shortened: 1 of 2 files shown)


isaacsas:
If you don't mind, maybe add a couple tests with non-trivial stoichiometry (say one that will generate a `MassActionJump` and one generating a `ConstantRate` or `VariableRate` jump), and check `Direct` gives the right answer when compared to a hand-written `JumpSystem`?

TorkelE:
Probably should just add a full test set of the entire `ReactionSystem` -> `JumpsSystem` functionality, better to have it in place earlier rather than late. Will probably be ready later today.

PR #369, opened 2020-05-17
add ismassaction and get rid of some temp arrays

isaacsas:
@TorkelE Mind taking a look that this looks ok to you? `ismassaction` is so useful I've found that I wanted a separate routine for it; we can drop it from DEBio now. This also re-arranges a bit so that the ratelaw calculation is only called for non-mass action jumps. On a big system that should help system construction time.

### DIFF
--- a/src/ModelingToolkit.jl
+++ b/src/ModelingToolkit.jl
@@ -116,7 +116,7 @@ export JumpProblem, DiscreteProblem
 export NonlinearSystem, OptimizationSystem
 export ode_order_lowering
 export PDESystem
-export Reaction, ReactionSystem
+export Reaction, ReactionSystem, ismassaction
 export Differential, expand_derivatives, @derivatives
 export IntervalDomain, ProductDomain, ⊗, CircleDomain
 export Equation, ConstrainedEquation
--- a/src/systems/reaction/reactionsystem.jl
+++ b/src/systems/reaction/reactionsystem.jl
@@ -52,7 +52,6 @@ function get_netstoich(subs, prods, sstoich, pstoich)
     ns
 end
 
-
 struct ReactionSystem <: AbstractSystem
     eqs::Vector{Reaction}
     iv::Variable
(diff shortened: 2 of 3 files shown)


isaacsas:
@ChrisRackauckas I think with this we are done with `ReactionSystem`s and their conversions over here, modulo we should add a RegularJump system and conversion with the new leaping code coming online.

isaacsas:
but let me know if you see anything we've missed.

ChrisRackauckas:
Looks good. Could a follow up PR flesh out the docs for JumpSystem and ReactionSystem?

We need to think about how we want to do the conversions there to regular jump problems. I think `full_regular_jump` or something like that would be to get to standard tau-leaping, and we want to leave open things like `hybrid_regular` that have double ups with extra parameters etc.

isaacsas:
Yup, doc updates are on the TODO list next over here. I didn't want to advertise all this stuff till we had it working / somewhat finalized, but we are there now. 

Hybrid systems will need a bit of thought, but I can go ahead with the basic RegularJumpSystem to start at least. If we later allow RegularJumpSystems to be hybrids instead of making a separate HybridJumpSystem we can update over here as necessary.

isaacsas:
Should hybrid be working at the DiffEqJump / StochasticDiffEq level already?

ChrisRackauckas:
>Should hybrid be working at the DiffEqJump / StochasticDiffEq level already?

Yes, you just can't mix that hybrid with actual SDEs/ODEs yet, which is my next goal.

>Hybrid systems will need a bit of thought, but I can go ahead with the basic RegularJumpSystem to start at least. If we later allow RegularJumpSystems to be hybrids instead of making a separate HybridJumpSystem we can update over here as necessary.

It would be good to keep hybrid separate for SimpleTauLeaping. 

One thing that would be interesting too would be a static splitting, but it might be difficult to choose which jumps go where. 

Then dynamic would be a bigger goal down the line, which is possible by making every jump part of both and adding parameters, but we'd need solvers to "know about it" in a sensible way.

ChrisRackauckas:
https://travis-ci.org/github/SciML/ModelingToolkit.jl/builds/688124601#L570

That tolerance should be made higher since it seems reliant on compiler optimization levels.

isaacsas:
I'm away from my computer the next few hours, but will bump that up (and double check the generated code) later tonight.

TorkelE:
Good stuff.

PR #371, opened 2020-05-17
Update DiffEqJump version for recent bug fixes

isaacsas:
The recent bugs are serious enough we should manually update the DiffEqJump version.


isaacsas:
@ChrisRackauckas I don't have merge permissions on MT, so you'll have to merge :)

PR #377, opened 2020-05-18
update get_variables! for Equations

isaacsas:
Notes; `get_variables!` returns `Operations`, but we also have `vars` which returns `Variables`, so we have some redundancy here.

I still need tests for ODEs here, but local testing looks ok.

### DIFF
--- a/test/dep_graphs.jl
+++ b/test/dep_graphs.jl
@@ -79,7 +79,7 @@ deps = equation_dependencies(os)
 eq_sdeps  = [[S,I], [S,I], [S,I,R]]
 @test all(i -> isequal(Set(eq_sdeps[i]),Set(deps[i])), 1:length(deps))
 
-sdes = convert(ODESystem, rs)
+sdes = convert(SDESystem, rs)
 deps = equation_dependencies(sdes)
 @test all(i -> isequal(Set(eq_sdeps[i]),Set(deps[i])), 1:length(deps))
 
--- a/test/dep_graphs.jl
+++ b/test/dep_graphs.jl
@@ -83,6 +83,10 @@ sdes = convert(ODESystem, rs)
 deps = equation_dependencies(sdes)
 @test all(i -> isequal(Set(eq_sdeps[i]),Set(deps[i])), 1:length(deps))
 
+deps = variable_dependencies(os)
+s_eqdeps = [[1],[2],[3]]
+@test deps.fadjlist == s_eqdeps
+
 #####################################
 #       testing for nonlin sys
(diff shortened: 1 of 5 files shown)


ChrisRackauckas:
Looks great. Is there anything else needed to get the dependency graph? Or can I now plot the direct connections say between variables today?

isaacsas:
You can get a LightGraph directed graph as: `asdigraph(asgraph(sys),sys)` for an ODE, SDE (but not including noise) or Jump system. The vertices are ordered first by equation and then state variable. Probably we would want some type of plot recipe or metagraph addtion on top to label nodes as Eq_i or with the corresponding state variable symbol.

ChrisRackauckas:
Is there an issue with NonlinearSystem?

isaacsas:
As long as `eq.rhs` has all the equation it should work. I can add a test for that too.

isaacsas:
Something like `x^2 + y^2 ~ x + y` isn't going to work currently.

ChrisRackauckas:
It's all in `0 ~ ...` form. We should enforce that more in the construction though.

isaacsas:
@ChrisRackauckas Assuming tests pass this should be good now.

isaacsas:
Oh, I might have misunderstood your earlier comment. This will do graphs showing which states an equation depends on currently, and all of eq-eq, eq-state, state-eq, state-state for jumps. State to state for ODEs needs a version of `modified_states!` too. I can write a simple version of that which would work for ODE/SDEs if one can assume the LHS of the equation only has one unique variable. Anything more complicated would require dispatches for the dg routines and some more work.

isaacsas:
Added a version that can make all four dg types for ODEs/SDEs, correctly if the LHS of equations only involve one variable, which can be assumed to be the one modified by the equation.

test/dep_graphs.jl
```diff
@@ -71,4 +71,32 @@ end
...
\ No newline at end of file
+#####################################
+#       testing for ODE/SDEs
+#####################################
+os = convert(ODESystem, rs)
+deps = equation_dependencies(os)
+eq_sdeps  = [[S,I], [S,I], [S,I,R]]
+@test all(i -> isequal(Set(eq_sdeps[i]),Set(deps[i])), 1:length(deps))
+
+sdes = convert(ODESystem, rs)
```
ChrisRackauckas, line 82:
```suggestion
sdes = convert(SDESystem, rs)
```

isaacsas: Thanks!

PR #380, opened 2020-05-19
update docstring example for ODESystem

caryan:
The constructor takes independent variable as second argument. We could also use the `ODESystem(eqs)` constructor if preferable.


ChrisRackauckas:
yes thanks!

PR #382, opened 2020-05-19
[WIP] automated linearity checks

ChrisRackauckas:
Currently blocked by https://github.com/JuliaSymbolics/SymbolicUtils.jl/issues/94

### DIFF
--- a/src/direct.jl
+++ b/src/direct.jl
@@ -100,41 +100,57 @@ end
 
 isidx(x) = x isa TermCombination
 
-"""
-```julia
-jacobian_sparsity(ops::AbstractVector{<:Expression}, vars::AbstractVector{<:Expression})
-```
+let
+    _scalar = one(TermCombination)
 
-Return the sparsity pattern of the Hessian of an array of expressions with respect to
-an array of variable expressions.
-"""
-function hessian_sparsity(f, u)
-    idx(i) = TermCombination(Set([Dict(i=>1)]))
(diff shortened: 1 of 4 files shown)


shashi:
@ChrisRackauckas could you add tests for the other 2 systems?

Right now the `@btime` for https://github.com/SciML/ModelingToolkit.jl/pull/382/files#diff-976b86e531528452332bc85d74807c8dR17 is ~100ms, it could be faster I guess. but we can look at it if it becomes a bottleneck.

shashi:
This can be done using `hessian_sparsity` now....

ChrisRackauckas:
Oh yes, hadn't thought of that! Could you whip up a version and we can just quickly merge then?

src/direct.jl
```diff
@@ -100,41 +100,57 @@ end
...
+                  b = isidx(~y) ? ~y : _scalar
                   combine_terms(linearity(~f, Val{2}()), a, b)
               else
                   error("Function of unknown linearity used: ", ~f)
               end
-          end]
+          end] |> Rewriters.Chain |> Rewriters.Postwalk |> Rewriters.Fixpoint
+
+    # we do this in a let block so that Revise works on the list of rules
+    global hessian_sparsity
```
ChrisRackauckas, line 127: what's this for?

ChrisRackauckas: oh because of the let block?

shashi: yeah.

PR #400, opened 2020-05-25
[WIP] Causal Input/Output variables, explicit algebraic variables, aliases, and open connections

ChrisRackauckas:


src/systems/abstractsystem.jl
```diff
@@ -135,12 +137,21 @@ function Base.getproperty(sys::AbstractSystem, name::Symbol)
...
         end
     end
+
+    if :outputs ∈ fieldnames(typeof(sys))
+        i = findfirst(x->convert(Variable,x.lhs).name==name,sys.outputs)
+        if i !== nothing
+            return rename(convert(Variable,sys.outputs[i].lhs),renamespace(sys.name,name))(getfield(sys,:iv)())
+        end
+    end
+
```
shashi, line 154: lol impressive.

shashi:
Ok this is basically ready as a low level infrastructure.

ChrisRackauckas:
🎉 looks good. If it passes let's merge.

shashi:
Some things to review:

- `observed(sys)` will return both the observed of `sys` and the namespaced obsreved of `sys.systems` in a vector of equations
- `equations(sys)` will try to do the simple `lhs = rhs` substitution of the observed variables in all the equations of `sys` and `sys.systems` (you can disable this with `remove_aliases=false` kwarg)

src/systems/reaction/reactionsystem.jl
```diff
@@ -294,6 +299,7 @@ explicitly on the independent variable (usually time).
 function ismassaction(rx, rs; rxvars = get_variables(rx.rate),
                               haveivdep = any(var -> isequal(rs.iv,convert(Variable,var)), rxvars),
                               stateset = Set(states(rs)))
+    return !(haveivdep || rx.only_use_rate || any(convert(Variable,rxv) in states(rs) for rxv in rxvars))
```
isaacsas, line 302: @ChrisRackauckas Any reason you cut off the code path here? This was the old version of `ismassaction`, which was having issues when profiling.

PR #402, opened 2020-05-26
[WIP] Utility to move all derivative terms to LHS 

ranjanan:
So far, works on: 
```julia
julia> eq = x ~ D(x) + Constant(0.5)
Equation(x, derivative(x, t) + 0.5)

julia> ModelingToolkit.movediffvars(eq)
Equation(-1 * derivative(x, t) + x, Constant(0.5))

julia> eq = x ~ D(x) * Constant(0.5)
Equation(x, derivative(x, t) * 0.5)

julia> ModelingToolkit.movediffvars(eq)
Equation(-0.5 * derivative(x, t) + x, Constant(0.0))

```

Doesn't work on: 
```julia
eq = x ~ Constant(5.0) * (D(x) + Constant(0.5))
```
Would be nice to have simplification based on distributive rule (https://github.com/JuliaSymbolics/SymbolicUtils.jl/pull/95)

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -100,7 +100,7 @@ end
 function movediffvars(eqn)
 	ops = Vector{Any}()
 	movediffvars(eqn.rhs, ops)
-	Equation(simplify(eqn.lhs + reduce(+, ops)), simplify(eqn.rhs + reduce(+, ops)))
+	Equation(simplify(eqn.lhs + sum(ops)), simplify(eqn.rhs + sum(ops)))
 end
 
 # variable extraction
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -80,27 +80,53 @@ function vars!(vars, O)
 end
 
 # function that moves all diff variable terms to the lhs
-function movediffvars(expr, ops)
-	for arg in expr.args
+function movevars(expr, terms; diff = true)
+    
+    iters = expr.args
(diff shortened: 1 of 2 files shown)


src/utils.jl
```diff
@@ -79,6 +79,30 @@ function vars!(vars, O)
...
+			end
+		elseif isa(arg, Operation)
+			movediffvars(arg, ops)
+		end
+	end
+end
+function movediffvars(eqn)
+	ops = Vector{Any}()
+	movediffvars(eqn.rhs, ops)
+	Equation(simplify(eqn.lhs + reduce(+, ops)), simplify(eqn.rhs + reduce(+, ops)))
```
YingboMa, line 103:
```suggestion
	Equation(simplify(eqn.lhs + sum(ops)), simplify(eqn.rhs + sum(ops)))
```

ChrisRackauckas:
Shouldn't we move all terms to the RHS for consistency?

YingboMa:
Yeah, we need another pass (or to modify this pass) to move non-differential terms to the RHS, too.

ranjanan:
Yes, I'll write another pass to move the non-differential terms to the RHS.

ChrisRackauckas:
This code is old enough that it's better to start from scratch.

PR #404, opened 2020-05-26
provide iszero for Operation to help sparse matrix addition and multiplication

### LINKED ISSUE #381: Operation times a sparse array errors
Using ModelingToolkit v3.4.0 and julia 1.4.0 I'm trying to multiply an operation by a sparse array and not having any luck:

```julia
julia> using ModelingToolkit

julia> using SparseArrays: sparse

julia> @parameters ω
(ω,)

julia> X = Float64[[0 1]; [1 0]]
2×2 Array{Float64,2}:
 0.0  1.0
 1.0  0.0

julia> ω.*sparse(X)
ERROR: TypeError: non-boolean (Operation) used in boolean context
Stacktrace:
 [1] _map_notzeropres!(::SparseArrays.HigherOrderFns.var"#3#4"{typeof(*),SparseArrays.HigherOrderFns.var"#17#20"{Operation,SparseArrays.HigherOrderFns.var"#21#24"}}, ::Operation, ::SparseArrays.SparseMatrixCSC{Operation,Int64}, ::SparseArrays.SparseMatrixCSC{Float64,Int64}) at /home/cryan/repos/julia/usr/share/julia/stdlib/v1.4/SparseArrays/src/higherorderfns.jl:271
 [2] _noshapecheck_map(::SparseArrays.HigherOrderFns.var"#3#4"{typeof(*),SparseArrays.HigherOrderFns.var"#17#20"{Operation,SparseArrays.HigherOrderFns.var"#21#24"}}, ::SparseArrays.SparseMatrixCSC{Float64,Int64}) at /home/cryan/repos/julia/usr/share/julia/stdlib/v1.4/SparseArrays/src/higherorderfns.jl:165
 [3] _shapecheckbc at /home/cryan/repos/julia/usr/share/julia/stdlib/v1.4/SparseArrays/src/higherorderfns.jl:1025 [inlined]
 [4] _copy at /home/cryan/repos/julia/usr/share/julia/stdlib/v1.4/SparseArrays/src/higherorderfns.jl:1015 [inlined]
 [5] _copy at /home/cryan/repos/julia/usr/share/julia/stdlib/v1.4/SparseArrays/src/higherorderfns.jl:1020 [inlined]
 [6] copy at /home/cryan/repos/julia/usr/share/julia/stdlib/v1.4/SparseArrays/src/higherorderfns.jl:1011 [inlined]
 [7] materialize(::Base.Broadcast.Broadcasted{SparseArrays.HigherOrderFns.SparseMatStyle,Nothing,typeof(*),Tuple{Operation,SparseArrays.SparseMatrixCSC{Float64,Int64}}}) at ./broadcast.jl:820
 [8] top-level scope at REPL[12]:1

julia> 
```

caryan:
I was poking around at adding supporting for `*(Operation,Sparse)` for #381 and I think we can easily support everything if we just provide `iszero(O::Operation)`.  In the `SparseArrays` stdlib [_noshapecheck_map](https://github.com/JuliaLang/julia/blob/3995a02e14db043ee337676cca00c9f3b55e1518/stdlib/SparseArrays/src/higherorderfns.jl#L158) is trying to figure out whether the function applied to zeros of the element types will also give a zero. If it does then it can return another sparse array. If we provide an `iszero(O::Operation)` then it will just do the right thing. Although I'm probably missing what other damage this will cause. 

E.g.
```julia 
@parameters Δ
Hdrift = zeros(Operation, 3,3)
Hdrift[3,3] = Δ
sparse(Hdrift)
3×3 SparseArrays.SparseMatrixCSC{Operation,Int64} with 1 stored entry:
  [3, 3]  =  Δ
```

or
```julia
@parameters ω
X = Float64[[0 1]; [1 0]]

Htot = ω * sparse(X)
display(Htot)

@parameters U[1:2,1:2]

display(-1im*Htot*U)
```

gives

```
2×2 SparseArrays.SparseMatrixCSC{Operation,Int64} with 2 stored entries:
  [2, 1]  =  ω * 1.0
  [1, 2]  =  ω * 1.0
2×2 Array{Operation,2}:
 identity(0) + ((0-1im) * (ω * 1.0)) * (U₂ˏ₁ * true)  …  identity(0) + ((0-1im) * (ω * 1.0)) * (U₂ˏ₂ * true)
 identity(0) + ((0-1im) * (ω * 1.0)) * (U₁ˏ₁ * true)     identity(0) + ((0-1im) * (ω * 1.0)) * (U₁ˏ₂ * true)
```

Fixes #381

### DIFF
--- a/src/operations.jl
+++ b/src/operations.jl
@@ -54,6 +54,11 @@ Base.isequal(::Variable , ::Operation) = false
 Base.isequal(::Operation, ::Constant ) = false
 Base.isequal(::Constant , ::Operation) = false
 
+# provide iszero for Operations to help sparse addition and multiplication
+Base.iszero(O::Operation) = ((O.op == identity) && iszero(O.args[1])) ||
+                            ((O.op == +) && all(iszero(arg) for arg in O.args)) ||
+                            ((O.op == *) && any(iszero(arg) for arg in O.args))
+
 Base.show(io::IO, O::Operation) = print(io, convert(Expr, O))
 
 # For inv
@@ -72,13 +77,5 @@ Base.convert(::Type{Expr},x::Operation) = Expr(x)
 # promotion
 Base.promote_rule(::Type{<:Constant}, ::Type{<:Operation}) = Operation
 Base.promote_rule(::Type{<:Operation}, ::Type{<:Constant}) = Operation
-
-# Fix Sparse MatMul
-Base.:*(A::SparseMatrixCSC{Operation,S}, x::StridedVector{Operation}) where {S} =
-    (T = Operation; mul!(similar(x, T, A.m), A, x, true, false))
-Base.:*(A::SparseMatrixCSC{Tx,S}, x::StridedVector{Operation}) where {Tx,S} =
-    (T = LinearAlgebra.promote_op(LinearAlgebra.matprod, Operation, Tx); mul!(similar(x, T, A.m), A, x, true, false))
-Base.:*(A::SparseMatrixCSC{Operation,S}, x::StridedVector{Tx}) where {Tx,S} =
-    (T = LinearAlgebra.promote_op(LinearAlgebra.matprod, Operation, Tx); mul!(similar(x, T, A.m), A, x, true, false))
 
 LinearAlgebra.lu(O::AbstractMatrix{<:Operation};kwargs...) = lu(O,Val(false);kwargs...)
--- a/src/operations.jl
+++ b/src/operations.jl
@@ -55,9 +55,8 @@ Base.isequal(::Operation, ::Constant ) = false
 Base.isequal(::Constant , ::Operation) = false
 
 # provide iszero for Operations to help sparse addition and multiplication
-Base.iszero(O::Operation) = ((O.op == identity) && iszero(O.args[1])) ||
-                            ((O.op == +) && all(iszero(arg) for arg in O.args)) ||
-                            ((O.op == *) && any(iszero(arg) for arg in O.args))
+# e.g. we want to tell the sparse library that iszero(zero(Operation) + zero(Operation)) == true
+Base.iszero(x::Operation) = (_x = simplify(x); _x isa Constant && iszero(_x.value))
 
 Base.show(io::IO, O::Operation) = print(io, convert(Expr, O))
 
--- a/src/variables.jl
+++ b/src/variables.jl
@@ -89,7 +89,8 @@ Get the value of a [`ModelingToolkit.Constant`](@ref).
 """
 Base.get(c::Constant) = c.value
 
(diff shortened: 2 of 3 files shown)


src/operations.jl
```diff
@@ -54,6 +54,11 @@ Base.isequal(::Variable , ::Operation) = false
 Base.isequal(::Operation, ::Constant ) = false
 Base.isequal(::Constant , ::Operation) = false
 
+# provide iszero for Operations to help sparse addition and multiplication
+Base.iszero(O::Operation) = ((O.op == identity) && iszero(O.args[1])) ||
```
YingboMa, line 58: I am not sure if `iszero` should be defined on an Operation, since sparsity is often structural information. I can definitely see where you are coming from though.

What about
```julia
Base.iszero(x::Operation) = (_x = simplify(x); _x isa Constant && iszero(_x.value))
```

YingboMa:
Could you also add tests?

src/operations.jl
caryan: Oh nice I hadn't discovered `simplify.jl` and I don't see it being too expensive. It should only be called occasionally on small `Operation` expressions and the time spent constructing the system using ModellingToolkit is usually immaterial compared to the time spent solving it . E.g. when trying to add two `SparseArrays.SparseMatrixCSC{Operation,Int64} ` it will check `iszero(zero(Operation) + zero(Operation))` and using your suggestion:

```julia
@btime iszero(zero(Operation) + zero(Operation))
  23.781 μs (179 allocations: 5.84 KiB)
```

I don't think `Base.iszero(x::Operation) = false` does what we want because it will again assume that it needs to densify sparse operations which leads to the `TypeError: non-boolean (Operation) used in boolean context` error in some cases and densified sparse matrices in others.

```
@parameters Δ
Hdrift = zeros(Operation, 3,3)
Hdrift[3,3] = Δ
sparse(Hdrift)
```

gives

```3×3 SparseArrays.SparseMatrixCSC{Operation,Int64} with 9 stored entries:
  [1, 1]  =  identity(0)
  [2, 1]  =  identity(0)
  [3, 1]  =  identity(0)
  [1, 2]  =  identity(0)
  [2, 2]  =  identity(0)
  [3, 2]  =  identity(0)
  [1, 3]  =  identity(0)
  [2, 3]  =  identity(0)
  [3, 3]  =  Δ
```

shashi: https://github.com/SciML/ModelingToolkit.jl/commit/c42b3e408903a4ce00c48e9a33e860e09eb4b299#r39710820

caryan: :grimacing:  oof! I guess my flippant dismissal of "too expensive" was off base

PR #410, opened 2020-05-29
Update LICENSE.md

ViralBShah:


ViralBShah:
I wonder why github can't auto-detect this file as the MIT license.

ChrisRackauckas:
Probably the quoting of it. It's an old v0.6 Pkg.dev thing.

Issue #409, opened 2020-05-29
Converting Derivative to Variable does not depend on type of Derivative


ArnoStrouwen:
```jl
using ModelingToolkit
@parameters t k
@variables x(t)
@derivatives D'~k
convert(Variable,D(x))
julia>xˍt
```

YingboMa:
Yes, that's expected.

ChrisRackauckas:
no, that's an issue. It should be _k

YingboMa:
Oh yeah, I didn't notice.

PR #412, opened 2020-05-29
fix convert(Variable, D(x))

YingboMa:
Fix https://github.com/SciML/ModelingToolkit.jl/issues/409


ChrisRackauckas:
For consistency, we should probably make second derivatives x_t_t instead of x_tt, since otherwise D(D(x)) will have two forms depending on how you `convert(Variable`

PR #413, opened 2020-05-29
Fix False Negative in Sparse Tests

arnavs:
The arrays of arrays of matrices examples was actually dense in both the sparse and dense tests.


ChrisRackauckas:
rebase

arnavs:
Done

PR #417, opened 2020-06-01
Add convert from ReactionSystem to NonlinearSystem

TorkelE:
This adds a function `Base.convert(::Type{<:NonlinearSystem},rs::ReactionSystem)`. Essentially it uses `assemble_drift(rs)` to generate the equations, and then recursive replacement to replace things like `X(t)` with `X`. It probably should use a version of `substitute()` to do this, but initially could not get it to work and wrote my own naive version.

I tried something like
```julia
swaps = Dict(zip(map(state -> Operation(state,[var2op(rs.iv,)]), rs.states),map(state -> var2op(state), rs.states)))
substitute(eq,swaps)
```
but that generated an error along the lines of
```error

map is not defined on sets
in top-level scope at test.jl:28
in substitute at dev/ModelingToolkit/src/utils.jl:121
in _substitute at dev/ModelingToolkit/src/utils.jl:125
in map at base/abstractarray.jl:2101
in error at base/error.jl:33
```

Tests not added yet.


TorkelE:
Ok, there seems to snuck in a weird commit `Merge pull request` here as well. I think it was from when I tried to synch my fork of MTK with the main one (which I might not have needed to do, maybe, I realised later).

src/systems/reaction/reactionsystem.jl
```diff
@@ -300,3 +300,30 @@ function Base.convert(::Type{<:JumpSystem},rs::ReactionSystem)
...
+"""
+```julia
+Base.convert(::Type{<:NonlinearSystem},rs::ReactionSystem)
+```
+
+Convert a ReactionSystem to a  NonlinearSystem.
+"""
+function Base.convert(::Type{<:NonlinearSystem},rs::ReactionSystem)
+    states_swaps = map(states -> Operation(states,[var2op(rs.iv)]), rs.states)
+    eqs = map(eq -> 0 ~ make_sub!(eq,states_swaps),getproperty.(assemble_drift(rs),:rhs))
```
ChrisRackauckas, line 314: should this just be using `substitute`?

TorkelE: Yes it should but didn't get it to work, I got a `map is not defined on sets` error when trying. It would be good if it could be sorted out before merging (see opening post for details)

ChrisRackauckas: Let's just get an issue open on that and merge. I don't want something like this to be the blocker of DEBio

TorkelE: Good, I've added the issue.

TorkelE:
I have written some basic code for doing bifurcation analysis. I have only checked it on the reaction networks we used for testing in DiffEqBio, so will need some more checking. It should contain most of the features we already have (finding the initial steady states could be done both better and faster though).

In principle, it should take any `ODESystem`, and analyze it, but I have only really kept biochemical reaction networks in mind while doing this. Periodic orbits currently not supported.

I have also made a new plot receipt for the output. This one could use some improvement, especially by building on the receipts already present in PALC (I ran into some troubles when doing so initially, and just wanted to get something out. Will try to make an issue on it later).

I am a little bit unsure though exactly where this code belongs. If we aim to remove most things no directly related to the DSL from DiffEqBio, then maybe not here?

While it is nice to have something now, I am again unsure how much effort we want to put into it. This considering that both HC and PALC have updated lined up which might make things a lot easier.

I am attaching the code. The basic syntax should be:
```julia
using ModelingToolkit, DiffEqBiological, Plots

v0 = 0.005; v = 0.1; K = 2.8; n = 4;
kD = 10; kB = 100; kC = 0.1;
deg = 0.01; L = 0.;
#Normal σV model, without inducable production.
σV_p = [v0, v, K, n, kD, kB, kC, deg, L];
σV_network = @MT_reaction_network begin
    v0 + hill(σV,v,K,n), ∅ → (σV+A)
    deg, (σV,A,AσV) → ∅
    (kB,kD), A + σV ↔ AσV
    L*kC, AσV → σV
end v0 v K n kD kB kC deg L;
ode_sys = convert(ODESystem,σV_network.reaction_system)
bif = MT_bifurcations(ode_sys,σV_p,:L,(0.,1.5))
plot(bif)
```

Cannot attach files in GitHub, so here is the code:
```julia
using DifferentialEquations, ModelingToolkit, DiffEqBiological, PseudoArcLengthContinuation, Setfield, LinearAlgebra, Plots, RecipesBase

function MT_bifurcations(os::ODESystem,p_vals,p,pspan,args...;
		steadystateAlg=defaultSteadyStatesAuto,
		steaty_state_args=(), steaty_state_kwargs=(),
		init_guess_min = [ones(length(os.states))],
		init_guess_max = [ones(length(os.states))],
                verbosity = 1, tangentAlgo = BorderedPred(),
                theta=0.5,
                contParams=ContinuationPar(
                        dsmax = (pspan[2]-pspan[1])/100.,
                        dsmin = min((pspan[2]-pspan[1])/5000.,1e-5),
                        ds = min((pspan[2]-pspan[1])/500., 0.001),
                        maxSteps = 10000,
                        pMin = pspan[1],
                        pMax = pspan[2],
                        computeEigenValues = true,
			newtonOptions = NewtonPar(tol = 1e-8, verbose = true, maxIter = 15),
                        theta = theta,
			saveSolEveryNsteps = 1
                        ),
                kwargs...)
		contParamsBackwards = ContinuationPar(contParams,ds=-contParams.ds)
                p_idx = findfirst(p .== Symbol.(os.ps))
                ss_pmin = steadystateAlg(os,setindex!(copy(p_vals),pspan[1],p_idx),init_guess_min,steaty_state_args...;steaty_state_kwargs...)
                ss_pmax = steadystateAlg(os,setindex!(copy(p_vals),pspan[2],p_idx),init_guess_max,steaty_state_args...;steaty_state_kwargs...)
		ode_function = ODEFunction(os,jac=true)
		F = (x,p) -> ode_function.f(x,p,0.)
		J = (x,p) -> ode_function.jac(x,p,0.)
		fps = [ss_pmin...,ss_pmax...]; ignore = fill(false,length(fps));
		bifurcation_paths = Vector{ContResult}()
		for (i,ss) in enumerate(fps)
			ignore[i] && continue
			p_vals_here = setindex!(copy(p_vals),(i<=length(ss_pmin)) ? pspan[1] : pspan[2],p_idx)
			new_path = continuation(F, J, ss, p_vals_here, (@lens _[p_idx]), (i<=length(ss_pmin)) ? contParams  : contParamsBackwards ; tangentAlgo=tangentAlgo)
			push!(bifurcation_paths,new_path[1])
			ignore[argmin(map(s -> norm(s .- new_path[2].u),fps))] = true
		end
		return bifurcation_paths
end

function defaultSteadyStatesAuto(os::ODESystem,p_vals,init_guesses,args...;kwargs...)
	output = []
	for init_guess in init_guesses
		for factor in [1e-3,1e-2,1e-1,1,1e1,1e2,1e3]
			ss_prob = SteadyStateProblem(ODEProblem(os,Pair.(os.states,factor*init_guess),(0.,100.),Pair.(os.ps,p_vals)))
			sol = solve(ss_prob,args...;kwargs...).u
			(minimum([1.,map(o -> norm(o .- sol)/norm(o), output)...]) > 0.001) && push!(output,sol)
		end
	end
	return output
end

@recipe function f(contreses::Vector{ContResult}; var = 1,linewidth=3)
	label --> ""
	linewidth --> linewidth
	for (i, contres) in enumerate(contreses)
	    @series begin
	        primary --> i==1
			stabilities = maximum.(real.(getproperty.(contres.eig,:eigenvals))).<0
			color --> map(b -> b ? :blue : :red, stabilities)
			linestyle --> map(b -> b ? :solid : :dot ,stabilities)
	        getproperty.(contres.sol,:p),getindex.(getproperty.(contres.sol,:x),var)
	    end
	end
end

```

PR #421, opened 2020-06-02
Update utils.jl

arnavs:
As discussed on Slack, to get rid of the `identity` vs `Constant` issue. 

So far have only seen this for constants, so I only did it for `identity(x::Int64)`. 

(And because I don't want it to have things like `ModelingToolkit.Constant(x)`)

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -15,6 +15,7 @@ Base.convert(::Type{Expression}, x::Variable) = convert(Operation,x)
 Base.convert(::Type{Expression}, x::Operation) = x
 Base.convert(::Type{Expression}, x::Symbol) = Operation(Variable(x),[])
 Expression(x::Bool) = Constant(x)
+identity(x::Int64) = Constant(x)
 
 function build_expr(head::Symbol, args)
     ex = Expr(head)


src/utils.jl
```diff
@@ -15,6 +15,7 @@ Base.convert(::Type{Expression}, x::Variable) = convert(Operation,x)
 Base.convert(::Type{Expression}, x::Operation) = x
 Base.convert(::Type{Expression}, x::Symbol) = Operation(Variable(x),[])
 Expression(x::Bool) = Constant(x)
+identity(x::Int64) = Constant(x)
```
ChrisRackauckas, line 18: oh nononononononoono.

YingboMa: lol, what?!

arnavs: Chris suggested in Slack I add `identity(x::Expression) = Constant(x)`. But that led to bizarre outcomes like `Constant(x)` where `x` was a variable. 

Not sure where `lol` etc. comes in, but OK.

ChrisRackauckas: This is scary amounts of no. Like, I'm surprised this overload doesn't instantly crash Julia.

ChrisRackauckas: Basically, we don't want to type-pirate `identity` on `Int`. We can do as much as we want on standard MTK types, but we shouldn't make it so that something as standard as `identity(::Int)` turns things into MTK: that's crazy.

PR #422, opened 2020-06-02
Update build_function_arrayofarray.jl

arnavs:
Disappeared for some reason.


ChrisRackauckas:
The test failure looks real.

PR #423, opened 2020-06-02
Fix MTK.build_function on "arrays of sparse matrices" and "arrays of arrays of sparse matrices"

### LINKED ISSUE #407: Arrays of sparse matrices not sparse
It appears that the zeros are being written out for sparse arrays of arrays so that the number of `nzval` wouldn't match

```julia
using ModelingToolkit, SparseArrays
@variables a b c
h_1 = sparse([a 0; b c])
h_2 = sparse.([[a 0; b c], [1 0; c 0]])
display(build_function(h_1, [a, b, c], linenumbers = false)[2])
display(build_function(h_2, [a, b, c], linenumbers = false)[2])
```
Shows the sparse matrix first, which is correct,
```julia
:((var"##MTIIPVar#311", var"##MTKArg#309")->begin
          @inbounds begin
                  let (a, b, c) = (var"##MTKArg#309"[1], var"##MTKArg#309"[2], var"##MTKArg#309"[3])
                      (var"##MTIIPVar#311").nzval[1] = a
                      (var"##MTIIPVar#311").nzval[2] = b
                      (var"##MTIIPVar#311").nzval[3] = c
                  end
              end
          nothing
      end)
```

But note what happens with the array of arrays
```julia
:((var"##MTIIPVar#315", var"##MTKArg#313")->begin
          @inbounds begin
                  let (a, b, c) = (var"##MTKArg#313"[1], var"##MTKArg#313"[2], var"##MTKArg#313"[3])
                      (var"##MTIIPVar#315"[1]).nzval[1] = a
                      (var"##MTIIPVar#315"[1]).nzval[2] = b
                      (var"##MTIIPVar#315"[1]).nzval[3] = 0
                      (var"##MTIIPVar#315"[1]).nzval[4] = c
                      (var"##MTIIPVar#315"[2]).nzval[1] = 1
                      (var"##MTIIPVar#315"[2]).nzval[2] = c
                      (var"##MTIIPVar#315"[2]).nzval[3] = 0
                      (var"##MTIIPVar#315"[2]).nzval[4] = 0
                  end
              end
          nothing
      end)
```

wupeifan:
The original codes loop over all the elements in a sparse array instead of the non-zero elements.
Fixes #407 .


wupeifan:
Yep, later today

wupeifan:
It turns out that we only need to revive one of the previous broken tests. @YingboMa

ChrisRackauckas:
even better!

wupeifan:
Great, it seems that the tests pass.

PR #424, opened 2020-06-03
Fix array nested sparse cases

wupeifan:
Two changes:
1. Adjusted the order in `if...else` in `build_function`. This is because `SparseMatrixCSC <: AbstractArray` is true. Therefore we need to adjust the order to trigger sparse builds before it goes through dense builds. I also added some `Init=Expr[]` to prevent empty calls on `reduce`.
2. Changed the related test case. The previous one was a dense matrix therefore it didn't test the related codes.

This PR can replace #422 .


wupeifan:
OK, it looks like all tests passed.

PR #425, opened 2020-06-03
Enable the creation of Problems from ReactionSystems

TorkelE:
Enable the creation of ODE, SDE, Discrete, and Jump Problems from a `ReactionSystem`s. The input u0/p arrays can either be of the form `[x=10.]` or simply `[10.]`.

Example code
```julia
using ModelingToolkit
using DifferentialEquations
using Plots

@parameters t p d
@variables X(t)
rxs = [Reaction(p, nothing, [X])
       Reaction(d, [X], nothing)]
rs  = ReactionSystem(rxs, t, [X], [p,d])

o_prob = ODEProblem(rs,[10.],(0.,10.),[100.,5.])
sol = solve(o_prob)
plot(sol)

s_prob = SDEProblem(rs,[10.],(0.,10.),[100.,5.])
sol = solve(s_prob)
plot(sol)

d_prob = DiscreteProblem(rs, [20],(0.,10.),[100.,5.])
j_prob = JumpProblem(d_prob,Direct(),rs)
sol = solve(j_prob,SSAStepper())

o_prob = ODEProblem(rs,[X => 10.],(0.,10.),[p => 100., d => 5.])
sol = solve(o_prob)
plot(sol)

s_prob = SDEProblem(rs,[10.],(0.,10.),[p => 100., d =>5.])
sol = solve(s_prob)
plot(sol)

d_prob = DiscreteProblem(rs, [X => 20],(0.,10.),[100.,5.])
j_prob = JumpProblem(d_prob,Direct(),rs)
sol = solve(j_prob,SSAStepper())
```

### DIFF
--- a/src/systems/reaction/reactionsystem.jl
+++ b/src/systems/reaction/reactionsystem.jl
@@ -353,6 +353,6 @@ function DiffEqBase.DiscreteProblem(rs::ReactionSystem, u0::Union{AbstractArray,
 end
 
 # JumpProblem from AbstractReactionNetwork
-function DiffEqJump.JumpProblem(prob, aggregator, rs::ReactionSystem, args...; kwargs...)
+function DiffEqJump.JumpProblem(rs::ReactionSystem, prob, aggregator, args...; kwargs...)
     return JumpProblem(convert(JumpSystem,rs), prob, aggregator, args...; kwargs...)
 end
--- a/src/systems/reaction/reactionsystem.jl
+++ b/src/systems/reaction/reactionsystem.jl
@@ -327,3 +327,26 @@ function make_sub!(eq,states_swaps)
 	end
 	return eq
 end
+
+### Converts a reaxction system to ODe or SDE problems ###
+
+# ODEProblem from AbstractReactionNetwork
+function DiffEqBase.ODEProblem(rs::ReactionSystem, u0::Union{AbstractArray, Number}, tspan, p, args...; kwargs...)
+    return ODEProblem(convert(ODESystem,rs),Pair.(rs.states,u0),tspan,Pair.(rs.ps,p), args...; kwargs...)
+end
+
+# SDEProblem from AbstractReactionNetwork
+function DiffEqBase.SDEProblem(rs::ReactionSystem, u0::Union{AbstractArray, Number}, tspan, p, args...; kwargs...)
+    p_matrix = zeros(length(rs.states), length(rs.eqs))
+    return SDEProblem(convert(SDESystem,rs),Pair.(rs.states,u0),tspan,Pair.(rs.ps,p),args...; noise_rate_prototype=p_matrix,kwargs...)
+end
+
+# DiscreteProblem from AbstractReactionNetwork
+function DiffEqBase.DiscreteProblem(rs::ReactionSystem, u0, tspan::Tuple, p=nothing, args...; kwargs...)
+    return DiscreteProblem(convert(JumpSystem,rs), Pair.(rs.states,u0),tspan,Pair.(rs.ps,p), args...; kwargs...)
+end
+
+# JumpProblem from AbstractReactionNetwork
+function DiffEqJump.JumpProblem(prob, aggregator, rs::ReactionSystem, args...; kwargs...)
+    return JumpProblem(convert(JumpSystem,rs), prob, aggregator, args...; kwargs...)
+end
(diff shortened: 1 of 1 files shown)


src/systems/reaction/reactionsystem.jl
```diff
@@ -327,3 +327,32 @@ function make_sub!(eq,states_swaps)
 	end
 	return eq
 end
+
+### Converts a reaxction system to ODe or SDE problems ###
+
+# ODEProblem from AbstractReactionNetwork
+function DiffEqBase.ODEProblem(rs::ReactionSystem, u0::Union{AbstractArray, Number}, tspan, p, args...; kwargs...)
+    u0 = typeof(u0) <: Array{<:Pair} ? u0 : Pair.(rs.states,u0)
```
ChrisRackauckas, line 335: this isn't something we allow on any of the other dispatches, so we should either change the others to allow that via a `varmap_to_vars` dispatch, or we should remove it here.

src/systems/reaction/reactionsystem.jl
```diff
@@ -327,3 +327,32 @@ function make_sub!(eq,states_swaps)
 	end
 	return eq
 end
+
+### Converts a reaxction system to ODe or SDE problems ###
```
ChrisRackauckas, line 331:
```suggestion
### Converts a reaction system to ODE or SDE problems ###

```

src/systems/reaction/reactionsystem.jl
isaacsas: I do think it would be nice if one could just pass numeric initial conditions and parameters as an option for any problem (not just those from `ReactionSystem`s). Making the mapping vectors is an additional thing to do, and often one knows the orderings. (Though one could argue having to make the mapping vectors does force the user to check what ordering is being used, which may help avoid bugs.)

isaacsas:
If these overloads are kept, I'd suggest also having the `JumpProblem` take `rs` as its first argument. Though this is breaking with DEBio it is consistent with all problems and easier to remember for users.

ChrisRackauckas:
Agreed, I think we should swap the order there.

src/systems/reaction/reactionsystem.jl
TorkelE: My main reason for this is that when creating reaction systems through the DSL, variables and parameters are typically unavailable to the user, e.g.
```julia
rn = @reaction_network begin
      p, 0 --> X
      d, X --> 0
end p d
[X => 1.]
[p=> 2., d=>0.1]
```
wouldn't work. Hence for this case, I think we would need this is an alternative.

ChrisRackauckas: We can merge for now, but this should get an issue.

src/systems/reaction/reactionsystem.jl
```diff
@@ -327,3 +327,32 @@ function make_sub!(eq,states_swaps)
...
+# SDEProblem from AbstractReactionNetwork
+function DiffEqBase.SDEProblem(rs::ReactionSystem, u0::Union{AbstractArray, Number}, tspan, p, args...; kwargs...)
+    u0 = typeof(u0) <: Array{<:Pair} ? u0 : Pair.(rs.states,u0)
+    p = typeof(p) <: Array{<:Pair} ? p : Pair.(rs.ps,p)
+    p_matrix = zeros(length(rs.states), length(rs.eqs))
+    return SDEProblem(convert(SDESystem,rs),u0,tspan,p,args...; noise_rate_prototype=p_matrix,kwargs...)
+end
+
+# DiscreteProblem from AbstractReactionNetwork
+function DiffEqBase.DiscreteProblem(rs::ReactionSystem, u0::Union{AbstractArray, Number}, tspan::Tuple, p=nothing, args...; kwargs...)
```
ChrisRackauckas, line 349: `p = nothing`?

TorkelE: Not really sure honestly! Just copied from DEB :P

isaacsas: Should `p=nothing` be allowed in all the problem types?

ChrisRackauckas: no, it should never be `nothing`. We have `DiffEqBase.NullDEParameters()` for this, which gives a nice error message if you try to index it which catches a common bug where people forget parameters. But I think the default here should just be removed since we know that if people have parameters in their models that they will need to be included, so we might as well error check right there.

PR #427, opened 2020-06-04
revert back to using a parameter eval module

isaacsas:
I think there is room to do better still, for example, I think we can get another ~20% speed up by having the `ReactionSystem` calculate the combinatoric stoichiometry factor for the rescaled rate. I'll work on that tomorrow or Friday if I have some time. Still, this gives a noticeable improvement.

### DIFF
--- a/src/systems/jumps/jumpsystem.jl
+++ b/src/systems/jumps/jumpsystem.jl
@@ -91,13 +91,18 @@ function assemble_crj(js, crj, statetoid)
 end
 
 function assemble_maj(js, maj::MassActionJump{U,Vector{Pair{V,W}},Vector{Pair{V2,W2}}},
-                      statetoid, parammap) where {U,V,W,V2,W2}
+                      statetoid, parammap, pcontext) where {U,V,W,V2,W2}
+    
     sr = maj.scaled_rates
-    if sr isa Operation
-        pval = simplify(substitute(sr,parammap)).value
+    if sr isa Operation 
+        if isempty(sr.args)
+            pval = parammap[sr.op]
+        else
+            pval = Base.eval(pcontext, Expr(maj.scaled_rates))
+        end
     elseif sr isa Variable
-        pval = Dict(parammap)[sr()]
-    else
+        pval = parammap[sr]
+    else   
         pval = maj.scaled_rates
     end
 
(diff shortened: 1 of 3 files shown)


src/systems/jumps/jumpsystem.jl
```diff
@@ -91,13 +91,18 @@ function assemble_crj(js, crj, statetoid)
...
+                      statetoid, parammap, pcontext) where {U,V,W,V2,W2}
+    
     sr = maj.scaled_rates
-    if sr isa Operation
-        pval = simplify(substitute(sr,parammap)).value
+    if sr isa Operation 
+        if isempty(sr.args)
+            pval = parammap[sr.op]
+        else
+            pval = Base.eval(pcontext, Expr(maj.scaled_rates))
```
ChrisRackauckas, line 101: I don't think we need to eval though: that's unsafe. Didn't we have another option here?

isaacsas: We used `eval` in DEBio, and here before the switch to `substitute`. I'd certainly prefer not to use `eval`, but I thought `substitute` was the only other option?

isaacsas: The rate can be a general `Operation`, i.e. `k^2 / (1 + k*a)`, etc.

ChrisRackauckas: We can go with eval for now, but there should be a faster way to do this doing direct substitution on the Operation itself. I think there's a few ways to break this eval usage, like distributed usage, so it's worth opening an issue over and cleaning up in the future, but at least it's a major improvement.

test/reactionsystem.jl
```diff
@@ -115,9 +115,15 @@ jumps[19] = VariableRateJump((u,p,t) -> p[19]*u[1]*t, integrator -> (integrator.
 jumps[20] = VariableRateJump((u,p,t) -> p[20]*t*u[1]*binomial(u[2],2)*u[3], integrator -> (integrator.u[2] -= 2; integrator.u[3] -= 1; integrator.u[4] += 2))
 
 statetoid = Dict(convert(Variable,state) => i for (i,state) in enumerate(states(js)))
-parammap = map((x,y)->Pair(x(),y),parameters(js),pars)
+parammap  = Dict(convert(Variable,param) => pars[i] for (i,param) in enumerate(parameters(js)))
+pvars = parameters(js)
+param_context = Module()
```
ChrisRackauckas, line 120: the downside is that I think these models won't precompile.

isaacsas: We should definitely look for another solution long term, but this gives us a short-term fix at least...

isaacsas:
Not sure why tests didn't run. I'll close and re-open. Assuming they pass this should be good to merge.

isaacsas:
Hmm, for some reason it seem the Travis tests aren't being queued...

PR #431, opened 2020-06-04
Add DaggerForm

jpsamaroo:
@ChrisRackauckas is there a defined code style for MTK? It seems like things are a mix of 4 spaces and tabs, which messes up the view of the diff.

Todo:
- [x] Tests
- [ ] Docs


ChrisRackauckas:
yeah we need to fix it

ChrisRackauckas:
is this the right level to chop up the computation? Will dagger be able to find common subexpressions between the nodes?

jpsamaroo:
Not yet, which is part of why this is still a draft PR; the next step is to recurse through expressions and also wrap those with `Dagger.delayed` calls.

jpsamaroo:
Not yet, which is part of why this is still a draft PR; the next step is to recurse through expressions and also wrap those with `Dagger.delayed` calls.

Issue #426, opened 2020-06-04
Slow jump construction likely due to substitute


isaacsas:
On Discourse, [here](https://discourse.julialang.org/t/lightgraph-modeltoolkit-optimization/40696), a graph SIR model was reported to take an exceptionally long time to construct the JumpProblem (18.55s out of a total time to run a simulation of 18.85s).

On running the profiler it seems the overwhelming majority of the time is being taken in the `substitute` call within `assemble_maj` at https://github.com/SciML/ModelingToolkit.jl/blob/6afdbd252096c05e7fd03a9a05e60456da43332f/src/systems/jumps/jumpsystem.jl#L97

I'll look into whether we can detect that the `rate` is just a symbol and skip this call, which should be the case for this example, but more generally is there anything that can be done about this?

ChrisRackauckas:
@shashi could you take a look at this? `substitute` shouldn't take that much time here.

isaacsas:
FWIW it is just simplifying essentially `p/1` for `p` different parameters, so nothing complicated.

ChrisRackauckas:
Yeah, so if we need an alternative implementation for this kind of thing that's fine, but I'm curious how any library can take 18 seconds on that.

erlebach:
You were able to duplicate my timings?

isaacsas:
Yes. and the issue is clear. We can fix it easily, it is just a matter of trying to avoid a temp fix for a more robust solution.

@ChrisRackauckas is that same section of code a `Dict` is being built that is keyed by `Operation`s. Is that a good idea? I usually use `Variable`s as keys.

ChrisRackauckas:
It's probably not a good idea to hit a dictionary here, though hashing may be useful when there's tons of variables? There just has to be a better way.

isaacsas:
But is hashing on an `Operation` ok to do?

ChrisRackauckas:
Maybe

isaacsas:
I think I had originally written this to take a premade `Dict` mapping `Variable`s to numeric values when it used evaluation in a Module context. That's basically what we used in DEBio with `Symbol`s and it had good performance (note the Dict is only built once in the `JumpProblem` call, not constructed each time within the jump call).

ChrisRackauckas:
We might need to revert back to that.

isaacsas:
I'm happy to revert it back if @shashi thinks getting a more permanent fix will take a while. Just let me know.

[...]

PR #432, opened 2020-06-05
`substituter` to avoid Dict re-construction

shashi:
`substituter(ks, vs)`

returns a function which takes an MTK expression, substitutes ks with vs and simplifies it.

This brings down the example in #426 from 15 seconds to 0.3 seconds in my test,

@isaacsas  Any idea what the test failure is about? Could it be because `1/x` is rewritten as `x^-1` for simplification purposes?

### DIFF
--- a/src/direct.jl
+++ b/src/direct.jl
@@ -43,6 +43,11 @@ function simplified_expr(O::Operation)
     isempty(O.args) && return O.op.name
     return Expr(:call, Symbol(O.op), simplified_expr.(O.args)...)
   end
+  if O.op === (^)
+      if length(O.args) > 1  && O.args[2] isa Constant && O.args[2].value < 0
+          return Expr(:call, :^, Expr(:call, :inv, simplified_expr(O.args[1])), -(O.args[2].value))
+      end
+  end
   return Expr(:call, Symbol(O.op), simplified_expr.(O.args)...)
 end
 
--- a/src/systems/jumps/jumpsystem.jl
+++ b/src/systems/jumps/jumpsystem.jl
@@ -91,12 +91,12 @@ function assemble_crj(js, crj, statetoid)
 end
 
 function assemble_maj(js, maj::MassActionJump{U,Vector{Pair{V,W}},Vector{Pair{V2,W2}}},
-                      statetoid, parammap) where {U,V,W,V2,W2}
+                      statetoid, subber) where {U,V,W,V2,W2}
     sr = maj.scaled_rates
     if sr isa Operation
-        pval = simplify(substitute(sr,parammap)).value
+        pval = subber(sr).value
     elseif sr isa Variable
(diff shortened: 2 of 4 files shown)


isaacsas:
State variables in JumpProblems may be integer valued, but I'm not sure why you are getting that error. I've never encountered it in applications before, even when taking powers of integers within  rate laws (which would be

shashi:
I see that the generated function refers to ModelingToolkit.:(^) which is sidestepping the syntax-based lowering of `x^-1` which actually works correctly.

I think then then internals should be escaped?

cc @ChrisRackauckas 

```julia
        begin
            $(Expr(:inbounds, true))
            local var"#8#val" = begin
                        let (a,) = (var"##MTKArg#293"[1],)
                            (ModelingToolkit).:^((ModelingToolkit).:+(2, a), -1)
                        end
                    end
            $(Expr(:inbounds, :((ModelingToolkit).pop)))
            var"#8#val"
        end
```

shashi:
What about the second failure though? It's just inaccurate. I am still unsure why this change makes causes these two failures :o

ChrisRackauckas:
Or we need to overload Base.literal_pow?

shashi:
Well the built function only deals with numbers, so I don't know if you want to do that... If the Expr is an honest to goodness `Expr(:call, [:^, 1, :x])` it should work.

Maybe we should special case for ^ and replace it with :^

shashi:
I see that you're actually using the symbol in place of the function, but I think GeneralizedGenerated is qualifying every symbol to be within ModelingToolkit.

isaacsas:
When this is done, before merging this can we test performance and memory use vs the current release on the SIR example of the issue? Hopefully it will be as fast or faster; I'd like to avoid a performance regression.

isaacsas:
Preferably for a larger number of vertices or edges where the JumpProblem construction timing is more noticeable.

shashi:
I'm about to do that now. Thanks!

shashi:
(the same params as the original discourse post)

This branch
```julia
  682.803 ms (2074579 allocations: 55.53 MiB)
```
Master branch
```julia
  968.292 ms (2014685 allocations: 58.28 MiB)
```

I am working on speeding this up even further, let's make it so fast that it's not going to impact anything.

shashi:
Ok my attempt to make it faster didn't work out very well. Was basically trying to add a rule which checks that all arguments are numbers and then applies the function to them, but the check ends up becoming expensive enough that it's not effective.

shashi:
I tried this example for 100k edges and 20k vertices, master 196 seconds, this branch 177

I think we should merge it?

isaacsas:
In meetings this afternoon, but will take a look this evening. Numbers look great @shashi!

isaacsas:
Is Travis not running it again?

PR #434, opened 2020-06-05
Fix issues on heterogeneous element types inside an array

wupeifan:
This PR fixes an issue when `build_function` from "arrays of arrays" and "arrays of arrays of matrices". The previous version of `build_function` detects an "array of arrays of matrices" object with nested `eltype` calls, which critically hinges on homogeneity across different elements under the outermost array. However, in practice, we have examples that break this homogeneity: For example, an array can consist of two array elements where the first is an `MTK.constant` and the second is an `Expression` matrix. In this case, the outermost `type` will be `Array{Array{T, 1} where T, 1}`, and it fails to detect "arrays of arrays of matrices".

To fix this issue, and to further make it more precise, we add a bunch of functions to replace those `eltype` calls but rather to check the type of each element of the outermost array.

I updated the functions in `build_function`, and added related tests. (I also deleted the oop functions that we are not going to use in the test file).

Tests passed.

I do have a question: I am not sure whether adding those functions under 'build_function.jl' is a great idea, and am open to discussions/modifications.

### DIFF
--- a/src/build_function.jl
+++ b/src/build_function.jl
@@ -131,6 +131,14 @@ function _build_function(target::JuliaTarget, op::Operation, args...;
     end
 end
 
+# Detect heterogeneous element types of "arrays of arrays of matrices/sparce matrices"
+function is_array_array_matrix(F)
+    return isa(F, AbstractVector) && all(isa.(F, AbstractArray{<:AbstractMatrix}))
+end
+function is_array_array_sparse_matrix(F)
+    return isa(F, AbstractVector) && all(isa.(F, AbstractArray{<:AbstractSparseMatrix}))
+end
+
 function _build_function(target::JuliaTarget, rhss, args...;
                          conv = simplified_expr, expression = Val{true},
                          checkbounds = false, constructor=nothing,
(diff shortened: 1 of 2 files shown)


src/build_function.jl
```diff
@@ -131,6 +131,21 @@ function _build_function(target::JuliaTarget, op::Operation, args...;
     end
 end
 
+# Detect heterogeneous element types of "arrays of matrices/sparce matrices"
+function is_array_matrix(F)
+    return isa(F, AbstractVector) && all(isa.(F, AbstractArray))
```
YingboMa, line 136:
```suggestion
    return isa(F, AbstractVector) && all(x->isa(x, AbstractArray), F)
```

wupeifan: I see. I'll change those two broadcasts

YingboMa:
```julia
julia> using BenchmarkTools

julia> foo(F) = isa(F, AbstractVector) && all(x->isa(x, AbstractArray), F)

julia> goo(F) = isa(F, AbstractVector) && all(isa.(F, AbstractArray))
goo (generic function with 1 method)

julia> F = map(_->rand(5), 1:3);

julia> @btime foo($F);
  2.934 ns (0 allocations: 0 bytes)

julia> @btime goo($F);
  48.860 ns (2 allocations: 128 bytes)
```

wupeifan: Cool. I changed all of those.

wupeifan:
OK, I think as tests passed after I incorporated Yingbo's suggestions, this one is ready to merge.

Issue #436, opened 2020-06-06
No support for integer or float rates for JumpSystems.


TorkelE:
This piece of code
```julia
reaction_network = @reaction_network begin
    A, ∅ → X
    1, 2X + Y → 3X
    B, X → Y
    1, X → ∅
end A B;
u0 = rand(1:Int64(10),length(reaction_network.states))
p = factor*rand(length(reaction_network.ps))
prob1 = JumpProblem(reaction_network,DiscreteProblem(reaction_network,u0,(0.,1.),p),Direct())
```
generates a 
```
MethodError: no method matching get_variables(::Int64)
```
error. Similar with `(::Float64)` for `1, 2X + Y → 3X`. This seems like something that should not happen?

isaacsas:
So there are two issues I'm seeing (so far). The first is easy; when converting `ReactionSystem`s we need to check if the rate is not an operation, and if so we don't call `get_variables` on it.

@ChrisRackauckas The second I need some advice. DiffEqJump requires the numeric reaction rate to be the same type for all MassActionJumps since it merges them into a single jump with vectors. With an integer type being passed in above for the rate it causes an error. Is it reasonable in the MT JumpProblem construction to assume the MassActionJump rate that is sent to DiffEqJump has the same type as the time variable? The only issue I can see here is if one is using units the rate should actually have inverse units to time, in which case maybe I should use the type of 1/tspan[2]?

PR #438, opened 2020-06-06
fix https://github.com/SciML/ModelingToolkit.jl/issues/436

isaacsas:
Fixes https://github.com/SciML/ModelingToolkit.jl/issues/436.


isaacsas:
Should be good if tests pass.

isaacsas:
@ChrisRackauckas I think this is good to merge now. Thanks!

PR #442, opened 2020-06-07
Enable creation of Steady State Problems from Reaction Systems.

TorkelE:
As the title says, enables the following code:
```julia
using Revise, DifferentialEquations, DiffEqBiological
rn = @reaction_network begin
    p, 0 --> X
    d, X --> 0
end p d

prob = SteadyStateProblem(rn,[10.],[1.,2.])
sol = solve(prob)
```

It should be noted that unlike the previous `ODEProblem`s and similar, SteadyStateProblem does not transform the `u0` and `p` vectors into a set of pairs, but simply passes on the input `u0` and `p`. 

If I were to try the following code:
```julia

@parameters p d
@variables X
prob = SteadyStateProblem(rn,[X => 10.],[p => 1.,d => 2.])
sol = solve(prob)
```
I get an error
```error

MethodError: no method matching Pair{Operation,Float64}(::Float64)
Closest candidates are:
  Pair{Operation,Float64}(::Any, !Matched::Any) where {A, B} at pair.jl:7
in top-level scope at test.jl:44
```
which may or may not be of interest to someone.

### DIFF
--- a/test/steadystatesystems.jl
+++ b/test/steadystatesystems.jl
@@ -1,5 +1,5 @@
 using ModelingToolkit
-using DifferentialEquations
+using SteadyStateDiffEq
 using Test
 
 @parameters t r
@@ -12,6 +12,6 @@ for factor in [1e-1, 1e0, 1e10], u0_p in [(2.34,2.676),(22.34,1.632),(.3,15.676)
     u0 = [x => factor*u0_p[1]]
     p = [r => factor*u0_p[2]]
     ss_prob = SteadyStateProblem(de,u0,p)
-    sol = solve(ss_prob).u[1]
+    sol = solve(ss_prob,SSRootfind()).u[1]
     @test abs(sol^2 - factor*u0_p[2]) < 1e-8
 end
--- a/test/steadystatesystems.jl
+++ b/test/steadystatesystems.jl
@@ -1,0 +1,17 @@
+using ModelingToolkit
+using DifferentialEquations
+using Test
+
+@parameters t r
+@variables x(t)
+@derivatives D'~t
+eqs = [D(x) ~ x^2-r]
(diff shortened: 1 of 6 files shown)


isaacsas:
Is the error just indicative that no `SteadyStateProblem` has been defined within MT yet to handle that case?

TorkelE:
That's probably it. Checking, and ModelingToolkit doesn't actually export `SteadyStateProblem`s. I guess that would have to fix, but maybe we would want to wait until when/if ModelingToolkit more generally supports these problems?

isaacsas:
The `ODEProblem` code is minimal, just 
https://github.com/SciML/ModelingToolkit.jl/blob/794e0e302c6663054338338299a9ef694ed4c937/src/systems/diffeqs/abstractodesystem.jl#L192
to the end of the file. So you could just add a similar `SteadyStateProblem` wrapper.

ChrisRackauckas:
it's also going to need something like: https://github.com/SciML/ModelingToolkit.jl/blob/56b4bbfda2b032cfee73c5ab6dc834fb6b391c51/src/systems/diffeqs/abstractodesystem.jl#L200-L229 . However, since there is a constructor for SteadyStateProblems from ODEProblems, you may just be able to run it through that existing one.

TorkelE:
Not sure I got it right, but maybe like this?

ChrisRackauckas:
that should work. Build a test case? Or at least extend an ODESystem test to call this and then run it.

test/steadystatesystems.jl
```diff
@@ -0,0 +1,17 @@
+using ModelingToolkit
+using DifferentialEquations
```
ChrisRackauckas, line 2: never use the full DifferentialEquations in a test

PR #444, opened 2020-06-07
Speed up JumpSystem generation from ReactionSystems

isaacsas:
To further improve large system performance, motivated by 
https://github.com/SciML/ModelingToolkit.jl/issues/426

and based on profiling the conversion and jump generation parts.


src/systems/jumps/jumpsystem.jl
```diff
@@ -141,8 +141,12 @@ dprob = DiscreteProblem(js, u₀map, tspan, parammap)
 """
 function DiffEqBase.DiscreteProblem(sys::AbstractSystem, u0map, tspan::Tuple,
                                     parammap=DiffEqBase.NullParameters(); kwargs...)
-    u0 = varmap_to_vars(u0map, states(sys))
-    p  = varmap_to_vars(parammap, parameters(sys))
+    #u0 = varmap_to_vars(u0map, states(sys))
```
isaacsas, line 144: This isn't final, just testing. It seems using a `Dict` does help a lot on large systems, so perhaps we should have a `varmap_to_vars` dispatch that takes a `Dict`. I found about a 20x speedup for a system with 25000 jumps.

isaacsas:
This speeds up the creation of `ReactionSystem`s and conversion to `JumpSystem`s a fair amount. I believe they are now allocation limited in performance, so we would require a more substantive alteration for further significant performance improvements. (Such as directly generating `JumpProblem`s, or at least the `MassActionJump`s, bypassing conversion.) 

The generation of `JumpProblem`s from a `JumpSystem` is the last piece to optimize.

ChrisRackauckas:
What's the status here? Is this something that should get updated, merged, and continued?

isaacsas:
I'm still not happy with where we are at, but probably this work should get merged. If I can get some time this week I'll try to cleanup and submit what is done -- the past few weeks have just been super busy... I've gotten very good speed ups on generating `ReactionSystem`s and `DiscreteProblem`s, and good speed ups on conversion to a `JumpSystem`. I haven't had a chance to work much on the `JumpSystem` to `JumpProblem` conversion, which is now the slowest component. 

More generally, we may want to rethink the internal design here somewhat. There are a lot of (intentional) allocations going on with having three levels of representations. This is one of two broad issues, the other being dynamic dispatch and instabilities due to using non-concrete types. For example, I believe conversion to `JumpSystem`s is now allocation limited in this PR. Profiling has been a pain here though, with many (to me) non-intuitive results. For example, in `ismassaction` I was seeing dynamic dispatch when using `any` (I think I've now eliminated that by manually searching), and significant time being spent in calls to `isempty` on an array. The latter is better when explicitly checking if the length is zero, but still a non-trivial amount of time, which just seems perplexing...

Ultimately, we may want to change JumpSystems to store `MassActionJumps` using integer identifiers for species, so we can reuse stoichiometry vectors in the `JumProblem`s, or perhaps avoid converting reaction systems to a `JumpSystem` intermediate. These would entail some larger changes though.

isaacsas:
Along those lines; `Operation`s are not concretely typed right (given that they store abstract types, `Function`s and `Expression`s?)

ChrisRackauckas:
>More generally, we may want to rethink the internal design here somewhat. There are a lot of (intentional) allocations going on with having three levels of representations. This is one of two broad issues, the other being dynamic dispatch and instabilities due to using non-concrete types. For example, I believe conversion to JumpSystems is now allocation limited in this PR. Profiling has been a pain here though, with many (to me) non-intuitive results. For example, in ismassaction I was seeing dynamic dispatch when using any (I think I've now eliminated that by manually searching), and significant time being spent in calls to isempty on an array. The latter is better when explicitly checking if the length is zero, but still a non-trivial amount of time, which just seems perplexing...

Yeah in terms of speed there's a lot to discuss. I've been discussing it with @YingboMa and @shashi in some other channels, but let's start taking it to #diffeq-bridged. There's a lot going on here. Essentially, it doesn't make sense to strictly type these because the structure of the equations is different every single time. If we concretely type the operations, then the type tree would encode the structure of the equations and therefore the type-stable functions for building the output would have to recompile for every new equation. So you'd make the compiled code fast but then you'd have to compile every time, which defeats the purpose of making it fast. So these pieces are kept abstractly typed so that it never has to recompile (I noticed this might be an issue in this PR). This BTW is the reason why most compilers, including Julia's, are not type-stable.

Thus making this code fast needs a different focus from standard Julia code. Specifically:

- One of the things that we need to think about for example is avoiding multiple dispatch since that's more expensive than a branch, so things like `assemble_crj` and `assemble_maj` are correctly doing the if statements to then call single dispatch functions. We need to find other dispatch hotspots and make them do case handling.
- We might want to collapse the types that we support to increase the number of single dispatch functions. Do we need `Constant(1)`? That's equivalent to `Operation(identity,Expression[1])`, so could we always use `Operation`, or is it lighter to not have that array? 
- Decrease the number of abstract types around. `Expression[x,y,z]` tends to give the compiler more trouble than just `Any[x,y,z]`, so we might want to just change `Operation` to the latter and then we might have more things resolve correctly in the compiler instead of relying on unions.
- Possibly use `(module,name)` (both as symbols) instead of `Function` inside of `Operation`. 

Etc. We should probably get an issue going to track updates in this as well.

isaacsas:
Yeah, it would be great to be able to see that conversation through the Slack or issues. Probably much of what you discuss could be helpful for this PR too...

As I said, I'll try to cleanup and then update to master this week to merge what I have so far.

isaacsas:
Timings on the SIR network model with 20000 edges:
```
PR:
julia> @time rs = ReactionSystem(rxs, t, vars, params);
  0.008907 seconds (14 allocations: 3.663 MiB)

julia> @time rs = ReactionSystem(rxs, t, vars, params);
  0.011149 seconds (14 allocations: 3.663 MiB)

Master: 
julia> @time rs = ReactionSystem(rxs, t, vars, params);
  0.302977 seconds (604.29 k allocations: 29.668 MiB)

julia> @time rs = ReactionSystem(rxs, t, vars, params);
  0.048388 seconds (179.00 k allocations: 6.394 MiB)

————————————————————————

PR:
julia> @time js = convert(JumpSystem, rs);
  0.371522 seconds (2.94 M allocations: 130.833 MiB)

julia> @time js = convert(JumpSystem, rs);
  0.411264 seconds (2.94 M allocations: 130.833 MiB)

Master:
julia> @time js = convert(JumpSystem, rs);
 60.578009 seconds (9.02 M allocations: 383.179 MiB, 0.27% gc time)

julia> @time js = convert(JumpSystem, rs);
 59.042692 seconds (5.68 M allocations: 213.643 MiB, 0.44% gc time)

————————————————————————

PR:
julia> @time dprob = DiscreteProblem(js, initial_state, tspan, initial_params)
  0.591422 seconds (1.29 M allocations: 67.144 MiB)

julia> @time dprob = DiscreteProblem(js, initial_state, tspan, initial_params)
  0.202933 seconds (597.07 k allocations: 30.512 MiB)

Master:
julia> @time dprob = DiscreteProblem(js, initial_state, tspan, initial_params)
 44.106209 seconds (1.52 M allocations: 86.365 MiB)

julia> @time dprob = DiscreteProblem(js, initial_state, tspan, initial_params)
 43.070374 seconds (239.78 k allocations: 13.723 MiB)

————————————————————————
PR:
julia> @time jprob = JumpProblem(js, dprob, Direct(), save_positions=(false,false))
  3.000220 seconds (9.75 M allocations: 277.510 MiB, 23.84% gc time)

julia> @time jprob = JumpProblem(js, dprob, Direct(), save_positions=(false,false))
  2.185403 seconds (9.35 M allocations: 255.276 MiB, 6.34% gc time)

Master:
julia> @time jprob = JumpProblem(js, dprob, Direct(), save_positions=(false,false))
 12.597332 seconds (56.60 M allocations: 1.456 GiB, 11.01% gc time)

julia> @time jprob = JumpProblem(js, dprob, Direct(), save_positions=(false,false))
  9.869804 seconds (50.79 M allocations: 1.159 GiB, 13.19% gc time)
```

isaacsas:
@ChrisRackauckas Let me know if you have any comments. Otherwise, if tests pass I'd say to merge and I'll keep working on this as I have time. There is definitely more to do, especially with reducing how many times `MassActionJump`s get created going between `ReactionSystem`s, `JumpSystem`s and the final `MassActionJump` for the solvers.

ChrisRackauckas:
>There is definitely more to do, especially with reducing how many times MassActionJumps get created going between ReactionSystems, JumpSystems and the final MassActionJump for the solvers.

I think we should definitely do some caching like we do with the Jacobians, i.e. when we create it, keep a reference in the system so we only create them once. Stuff like that, but yes, I agree: let's get this merged and keep improving.

isaacsas:
One thing the `DiscreteProblem` changes show is that using a `Dict` for the variable to value mappings can give a big speedup for large systems. It might be good to generally allow users to pass `Dict`s instead of vectors in the various problems.

ChrisRackauckas:
>It might be good to generally allow users to pass Dicts instead of vectors in the various problems.

Yes definitely, and there's an issue open for that. We need to save all variables as operations to do that well though, but that's a chance I really want to make.

PR #445, opened 2020-06-08
Track external registrations of Operations, so they can be injected into generated functions.

dpad:
This solves the issue where an external package may wish to use a function it defines "my_func" as an Operation, and so calls "@register my_func(x)" on it. However, when ModelingToolkit eventually generates a function including that Operation, GeneralizedGenerated renames all calls to the function to "(ModelingToolkit).my_func", which does not exist because that function was not defined in the ModelingToolkit namespace. Instead, we track all the functions that are registered and which module they come from, and use that to re-write the generated functions with the appropriate module.

I'm not sure if this is the best way to do it, but since at least it works for me, I'm offering it just in case it helps. (Note: I'm quite new to Julia so please forgive me for any silly mistakes!)


ChrisRackauckas:
Very interesting. Works. Want a second opinion on how robust this is.

PR #446, opened 2020-06-09
Number -> Real

YingboMa:
Much code uses `Real` as type constraints so maybe we should use `Real` instead of `Number` to be able to trace through more code.

CC: @jlperla

### DIFF
--- a/src/systems/jumps/jumpsystem.jl
+++ b/src/systems/jumps/jumpsystem.jl
@@ -195,7 +195,7 @@ tspan = (0.0, 250.0)
 dprob = DiscreteProblem(js, u₀map, tspan, parammap)
 ```
 """
-function DiffEqBase.DiscreteProblemExpr(sys::JumpSystem, u0map, tspan::Tuple,
+function DiscreteProblemExpr(sys::JumpSystem, u0map, tspan::Tuple,
                                     parammap=DiffEqBase.NullParameters(); kwargs...)
     u0 = varmap_to_vars(u0map, states(sys))
     p  = varmap_to_vars(parammap, parameters(sys))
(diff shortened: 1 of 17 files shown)


ChrisRackauckas:
What about going crazy and using AbstractFloat?

isaacsas:
Would this break things like
```julia
@parameters n
b = binomial(n,2)
```

ChrisRackauckas:
Nope, it'll only superset the number of functions we can trace into (since being a stricter subtype makes you still a subtype of the supertype). So in a way, AbstractFloat might just be better than Real and better than Number if the only goal is to just trace whatever we can.

jlperla:
Yeah, AbstractFloat might be better.

YingboMa:
> What about going crazy and using AbstractFloat?

I don't think there is much code uses `::AbstractFloat` signature that's not an elementary/special function, so using `AbstractFloat` won't help us much.

ChrisRackauckas:
But if it allows strictly more code to trace, is it harmful? I think it might as well be done?

src/systems/jumps/jumpsystem.jl
```diff
@@ -154,6 +176,40 @@ function DiffEqBase.DiscreteProblem(sys::JumpSystem, u0map, tspan::Tuple,
     DiscreteProblem(df, u0, tspan, p; kwargs...)
 end
 
+"""
+```julia
+function DiffEqBase.DiscreteProblemExpr(sys::JumpSystem, u0map, tspan,
+                                    parammap=DiffEqBase.NullParameters; kwargs...)
+```
+
+Generates a black DiscreteProblem for a JumpSystem to utilize as its
```
isaacsas, line 185: We should fix that typo "black" -> "blank".

PR #447, opened 2020-06-10
bump SymbolicUtils version

### LINKED ISSUE #437: subber not evaluating numerical expressions
@shashi it seems like the substitution in `assemble_maj` is still not always returning numeric values:

```julia
@parameters t k
@variables S I
rxs = Reaction(exp(k),[S],[I])
rs = ReactionSystem([rxs], t, [S,I], [k])
js = convert(JumpSystem, rs)
dprob = DiscreteProblem(js, [S => 1, I => 1], (0.0,10.0), [k => 1])
```
leads to an error at
https://github.com/SciML/ModelingToolkit.jl/blob/71bc681125dc7fd80440e1922d947ca912b3b0eb/src/systems/jumps/jumpsystem.jl#L97
as there is no value field... `subber(sr)` is returning `exp(k)`, an `Operation`... This method should always return a numerically evaluated result though.

shashi:
fixes #437

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "ModelingToolkit"
 uuid = "961ee093-0014-501f-94e3-6117800e7a78"
 authors = ["Chris Rackauckas <accounts@chrisrackauckas.com>"]
-version = "3.8.0"
+version = "3.8.1"
 
 [deps]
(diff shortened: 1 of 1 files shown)


isaacsas:
Thanks @shashi!

PR #448, opened 2020-06-10
Add skipzeros

YingboMa:


src/build_function.jl
```diff
@@ -202,18 +202,55 @@ function _build_function(target::JuliaTarget, rhss, args...;
...
+        end
     elseif rhss isa SparseMatrixCSC
-        ip_sys_exprs = [:($X.nzval[$i] = $(conv(rhs))) for (i, rhs) ∈ enumerate(_rhss)]
+        for (i, rhs) ∈ enumerate(_rhss)
+            rhs′ = conv(rhs)
+            (skipzeros && rhs′ isa Number && iszero(rhs′)) && continue
+            push!(ip_sys_exprs, :($X.nzval[$i] = $rhs′))
+        end
     else
-        ip_sys_exprs = [:($X[$(oidx(i))] = $(conv(rhs))) for (i, rhs) ∈ enumerate(_rhss)]
```
isaacsas, line 216: I think the test error is from dropping `oidx` in the new code.

YingboMa: Ah thanks!

Issue #450, opened 2020-06-11
Crash on solving an identical ODEProblem a large number of times


TorkelE:
I ran a couple of tests where I solved a large number of ODE problems, and got weird errors. In the end, I managed to narrow it down to this minimal example:
```julia
using ModelingToolkit, OrdinaryDiffEq

@parameters t d
@variables X(t)
@derivatives D'~t
eqs = [D(X) ~ 1 - d*X]
sys = ODESystem(eqs)

for i = 1:500
    print("\n",i) #Prints the number of iteration. The crash should happen at the same point every time.
    prob = ODEProblem(sys,[X => 1.0],(0.,1.),[d => 2.0])
    sol = solve(prob,Rosenbrock23())
    #print("          ",sol.u[end]) #This prints the same value each time.
end
```
every time I run this, I get a crash at iteration 217. The error message follows at the end. Things I have already tested (with no change) is:
- Using the startup option `--check-bounds=yes`.
- Replacing line `147` in the file `abstractodesystem.jl`, this one:
```
    f_oop,f_iip = generate_function(sys, dvs, ps; expression=Val{false}, kwargs...)
```
with
```
    f_oop,f_iip = eval(generate_function(sys, dvs, ps; expression=Val{false}, kwargs...))
```

Finally, the error message I get:
```
julia: /buildworker/worker/package_linux64/build/src/codegen.cpp:3322: jl_cgval_t emit_invoke(jl_codectx_t&, jl_expr_t*, jl_value_t*): Assertion `(((jl_value_t*)(((jl_taggedvalue_t*)((char*)(mi) - sizeof(jl_taggedvalue_t)))->header & ~(uintptr_t)15))==(jl_value_t*)(jl_method_instance_type))' failed.

signal (6): Aborted
in expression starting at /home/username/.julia/dev/DiffEqBiological/test/solve_ODEs.jl:9
gsignal at /lib/x86_64-linux-gnu/libc.so.6 (unknown line)
abort at /lib/x86_64-linux-gnu/libc.so.6 (unknown line)
unknown function (ip: 0x7f41cbb58399)
__assert_fail at /lib/x86_64-linux-gnu/libc.so.6 (unknown line)
emit_invoke at /buildworker/worker/package_linux64/build/src/codegen.cpp:3322
emit_expr at /buildworker/worker/package_linux64/build/src/codegen.cpp:4139
[...]

PR #451, opened 2020-06-13
Avoid GG via EvalFunc for standard diffeq usage

ChrisRackauckas:
Requires https://github.com/SciML/DiffEqBase.jl/pull/534

Fixes https://github.com/SciML/ModelingToolkit.jl/issues/450

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -30,7 +30,7 @@ Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
 [compat]
 ArrayInterface = "2.8"
 DataStructures = "0.17"
-DiffEqBase = "6.28"
+DiffEqBase = "6.38"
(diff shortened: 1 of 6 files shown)


ChrisRackauckas:
@dpad your fix in https://github.com/SciML/ModelingToolkit.jl/pull/445 no longer works in the ODE case because it's not going through GeneralizedGenerated. I think the only way to do a long term fix here is to have `@register` evaluate into a module so that all of the "known" functions have the same definition space, and then `eval` on that module will always work. @YingboMa what do you think?

Since this fixes more than a few bugs, I'm going to go ahead and merge this, but we should keep that register issue in mind.

dpad:
@ChrisRackauckas I believe I have fixed the regression in #463.

Issue #452, opened 2020-06-14
Crash on solving an identical JumpProblem a large number of times


TorkelE:
While https://github.com/SciML/ModelingToolkit.jl/pull/451 solved the problems outlied in https://github.com/SciML/ModelingToolkit.jl/issues/450 for ODE and SDE Problems, I still encounter the same problem for Jump Problems. Checking the changes made in https://github.com/SciML/ModelingToolkit.jl/pull/451 it only seems to touch ODE/SDe stuff. My guess is that the same changes to the Jump problems would solve this.
(I'd do it myself, but it will probably go smoother if someone who knows what is going on does it)

Only making a discrete problem does not seem to crash it. This works fine (but not sure what would happen for more iterations).
```julia

@parameters β γ t
@variables S I R
rate₁   = β*S*I
affect₁ = [S ~ S - 1, I ~ I + 1]
rate₂   = γ*I
affect₂ = [I ~ I - 1, R ~ R + 1]
j₁      = ConstantRateJump(rate₁,affect₁)
j₂      = ConstantRateJump(rate₂,affect₂)
j₃      = MassActionJump(2*β+γ, [R => 1], [S => 1, R => -1])
js      = JumpSystem([j₁,j₂,j₃], t, [S,I,R], [β,γ])
u₀map = [S => 999, I => 1, R => 0]
parammap = [β => .1/1000, γ => .01]
tspan = (0.0, 250.0)

for i = 1:500
    println(i)
    dprob = DiscreteProblem(js, u₀map, tspan, parammap)
end
```
this, however, causes a crash at `i=60`
```julia

@parameters β γ t
@variables S I R
rate₁   = β*S*I
affect₁ = [S ~ S - 1, I ~ I + 1]
rate₂   = γ*I
affect₂ = [I ~ I - 1, R ~ R + 1]
j₁      = ConstantRateJump(rate₁,affect₁)
j₂      = ConstantRateJump(rate₂,affect₂)
j₃      = MassActionJump(2*β+γ, [R => 1], [S => 1, R => -1])
js      = JumpSystem([j₁,j₂,j₃], t, [S,I,R], [β,γ])
u₀map = [S => 999, I => 1, R => 0]
parammap = [β => .1/1000, γ => .01]
tspan = (0.0, 250.0)

for i = 1:500
    println(i)
    dprob = DiscreteProblem(js, u₀map, tspan, parammap)
    jprob = JumpProblem(js, dprob, Direct())
[...]

Issue #288, opened 2020-04-12
Latexify for arrays of chemical reactions


ChrisRackauckas:
It would be nice to get a special one here for generating the chemical syntax. @korsbo did you have previous code for this?

korsbo:
We have the latex recipe for DiffEqBiological.
https://github.com/SciML/DiffEqBiological.jl/blob/master/src/latexify_recipes.jl

korsbo:
actually that only does the equation bit, not the reaction arrows.

korsbo:
The reaction arrow bit is done here:
https://github.com/korsbo/Latexify.jl/blob/fb778e5e1527a584407fd19ad3ceabe4f83ccd66/src/plugins/DiffEqBiological.jl#L60-L116

PR #453, opened 2020-06-14
Generate JumpProblem with EvalFunc to avoid GG

ChrisRackauckas:
Does the same change as https://github.com/SciML/ModelingToolkit.jl/pull/451 except on JumpSystem to fix https://github.com/SciML/ModelingToolkit.jl/issues/452

### DIFF
--- a/src/systems/jumps/jumpsystem.jl
+++ b/src/systems/jumps/jumpsystem.jl
@@ -27,7 +27,7 @@ js      = JumpSystem([j₁,j₂,j₃], t, [S,I,R], [β,γ])
 """
 struct JumpSystem{U <: ArrayPartition} <: AbstractSystem
     """
-    The jumps of the system. Allowable types are `ConstantRateJump`, 
+    The jumps of the system. Allowable types are `ConstantRateJump`,
     `VariableRateJump`, `MassActionJump`.
     """
     eqs::U
@@ -48,7 +48,7 @@ function JumpSystem(eqs, iv, states, ps; systems = JumpSystem[],
 
     ap = ArrayPartition(MassActionJump[], ConstantRateJump[], VariableRateJump[])
     for eq in eqs
-        if eq isa MassActionJump 
+        if eq isa MassActionJump
             push!(ap.x[1], eq)
         elseif eq isa ConstantRateJump
             push!(ap.x[2], eq)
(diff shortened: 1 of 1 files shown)


src/systems/jumps/jumpsystem.jl
```diff
@@ -139,11 +143,13 @@ tspan = (0.0, 250.0)
 dprob = DiscreteProblem(js, u₀map, tspan, parammap)
 ```
 """
-function DiffEqBase.DiscreteProblem(sys::AbstractSystem, u0map, tspan::Tuple,
+function DiffEqBase.DiscreteProblem(sys::JumpSystem, u0map, tspan::Tuple,
```
ChrisRackauckas, line 146: @isaacsas generally we should leave DiscreteProblem open for discrete dynamical systems. I want to add a differencing operator and handle building these, but with JumpProblem we know this is always identity so we can have this special dispatch.

src/systems/jumps/jumpsystem.jl
```diff
@@ -139,11 +143,13 @@ tspan = (0.0, 250.0)
...
                                     parammap=DiffEqBase.NullParameters(); kwargs...)
     u0 = varmap_to_vars(u0map, states(sys))
     p  = varmap_to_vars(parammap, parameters(sys))
-    f  = (du,u,p,t) -> du.=u    # identity function to make syms works
+    # identity function to make syms works
+    # EvalFunc because we know that the jump functions are generated via eval
+    f  = DiffEqBase.EvalFunc(DiffEqBase.DISCRETE_INPLACE_DEFAULT)
```
ChrisRackauckas, line 152: This is just a special trick so if you use `FunctionMap` it knows to ignore handling the discrete function map part.

ChrisRackauckas:
@TorkelE was this the last issue for getting DiffEqBio working?

TorkelE:
Yes. This should make all my tests succeed.

There are some tangential things which are not implemented yet:
- Network property query. We already get everything that works on `ReactionSystem`s in MTK, so not sure what more we would want to add here.
- Not sure what we decided on a `merge!(rs1,rs2)` function was, but if that gets implemented in MTK, a way for calling that from DiffEqBio needs implementing (likely trivial).
- Latexify support.
- Scaling the noise in the chemical Langevin equation. This might actually need some decision on how to parse additional input, so not yet supported.

But if we want to, we could probably merge without these points fixed.

ChrisRackauckas:
>Network property query. We already get everything that works on ReactionSystems in MTK, so not sure what more we would want to add here.

I think we're good here then.

>Not sure what we decided on a merge!(rs1,rs2) function was, but if that gets implemented in MTK, a way for calling that from DiffEqBio needs implementing (likely trivial).

That's through the same system as components. I think I need to talk to @isaacsas about it, but don't worry about it in DEBio quite yet.

>Latexify support.

@korsbo we should add a special Latexify recipe for ReactionSystem. Latexify recipes already exist on other systems, but the arrow notation is nice. This is https://github.com/SciML/ModelingToolkit.jl/issues/288 , so we'd just need to update https://github.com/SciML/DiffEqBiological.jl/blob/master/src/latexify_recipes.jl

>Scaling the noise in the chemical Langevin equation. This might actually need some decision on how to parse additional input, so not yet supported.

I am not sure if we should generally do this. We should discuss other transformations, like LNA and such, but these are transforms on the ReactionSystem.

ChrisRackauckas:
So I think we're good then. @isaacsas and I should have a chat about composing though.

TorkelE:
Sounds good. I agree there is some dubiousness with scaling noise in CLA, which would require further discussion.

Regarding system components, when I had a look at it earlier I allowed the macro without input (`rn = @reaction_network` to create an empty `ReactionSystem`, which then could be built upon. This worked fine, but it seemed that someway along with the development of MTK that was disallowed, and creating empty `ReactionSystem`s now yields an error. Depending on how this goes this might be something to consider.

isaacsas:
I’d like to get the relevant subset of the API updated before we release so if @TorkelE is done I can take over the PR.

isaacsas:
Alternatively we could merge it to master and then I can update and release.

ChrisRackauckas:
Which PR?

isaacsas:
His DEBio one. Not here.

ChrisRackauckas:
I think it could be good to just tidy it up and merge to master, but not release it yet. Then do some updates via other PRs before releasing. PRs to PRs can go too deep and just cause more of a mess IMO.

TorkelE:
Yes, let's merge and then @isaacsas can fix the last few parts. There has to be a new release of this for Travis to load the latest version, right? When the tests have let's merge.

ChrisRackauckas:
It's tagged.

PR #463, opened 2020-06-17
Inject registered functions into expressions used to build EvalFunc.

dpad:
Fixes the regression of #445 caused by #451.

### DIFF
--- a/test/function_registration.jl
+++ b/test/function_registration.jl
@@ -1,3 +1,8 @@
+# Each test here builds an ODEFunction including some user-registered
+# Operations. The test simply checks that calling the ODEFunction
+# appropriately calls the registered functions, whether the call is
+# qualified (with a module name) or not.
+
 # TEST: Function registration in a module.
 # ------------------------------------------------
 module MyModule
@@ -11,11 +16,11 @@ module MyModule
     end
     @register do_something(a)
 
-    eq  = Dt(u) ~ do_something(x)
+    eq  = Dt(u) ~ do_something(x) + MyModule.do_something(x)
     sys = ODESystem([eq], t, [u], [x])
     fun = ODEFunction(sys)
 
-    @test_broken fun([0.5], [5.0], 0.) == [15.0]
+    @test fun([0.5], [5.0], 0.) == [30.0]
 end
 
 # TEST: Function registration in a nested module.
@@ -32,11 +37,11 @@ module MyModule2
         end
         @register do_something_2(a)
 
-        eq  = Dt(u) ~ do_something_2(x)
+        eq  = Dt(u) ~ do_something_2(x) + MyNestedModule.do_something_2(x)
         sys = ODESystem([eq], t, [u], [x])
         fun = ODEFunction(sys)
 
-        @test_broken fun([0.5], [3.0], 0.) == [23.0]
+        @test fun([0.5], [3.0], 0.) == [46.0]
     end
 end
 
@@ -52,8 +57,8 @@ function do_something_3(a)
 end
 @register do_something_3(a)
 
-eq  = Dt(u) ~ do_something_3(x)
+eq  = Dt(u) ~ do_something_3(x) + (@__MODULE__).do_something_3(x)
 sys = ODESystem([eq], t, [u], [x])
 fun = ODEFunction(sys)
 
-@test_broken fun([0.5], [7.0], 0.) == [37.0]
+@test fun([0.5], [7.0], 0.) == [74.0]
--- a/src/build_function.jl
+++ b/src/build_function.jl
@@ -125,7 +125,7 @@ function _build_function(target::JuliaTarget, op::Operation, args...;
     end
 
     if expression == Val{true}
-        return oop_ex
+        return ModelingToolkit.inject_registered_module_functions(oop_ex)
     else
         _build_and_inject_function(@__MODULE__, oop_ex)
     end
@@ -133,7 +133,7 @@ end
 
 function _build_and_inject_function(mod::Module, ex)
     # Generate the function, which will process the expression
-    runtimefn = GeneralizedGenerated.mk_function(@__MODULE__, ex)
+    runtimefn = GeneralizedGenerated.mk_function(mod, ex)
 
     # Extract the processed expression of the function body
     params = typeof(runtimefn).parameters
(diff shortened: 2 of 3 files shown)


test/function_registration.jl
```diff
@@ -11,11 +16,11 @@ module MyModule
     end
     @register do_something(a)
 
-    eq  = Dt(u) ~ do_something(x)
+    eq  = Dt(u) ~ do_something(x) + MyModule.do_something(x)
```
ChrisRackauckas, line 19: nice extension to the test

ChrisRackauckas:
Want a second check from @YingboMa , but I think it's good. Thanks!

dpad:
If you don't mind, I'd like to hold off on getting this merged. I've run into a seemingly related issue that I'd like to look into, it seems that the solution isn't quite working on my external package.

dpad:
I haven't had time to look into this in detail yet but here's the issue, in case you guys have any insights.

The changes in #451 makes it so that my external package building an ODEFunction at precompile time now gets a bunch of `WARNING: eval into closed module ModelingToolkit: a_really_big_long_Expr... ** incremental compilation may be fatally broken for this module **` warnings during precompilation. That's fine, I suppose, but not ideal.

The main issue now though is that I can't call the `my_ode_function.jac(dvs, ps, iv)` at precompile time (I was using it to build another ODEFunction based on the results of the automatically computed Jacobian), as it errors out with a world error:

```julia
ERROR: LoadError: MethodError: no method matching (::ModelingToolkit.var"#310#312")(::Array{ModelingToolkit.Operation,1}, ::Array{ModelingToolkit.Operation,1}, ::ModelingToolkit.Operation)
The applicable method may be too new: running in world age 27499, while current world is 27505.
Closest candidates are:
  #310(::Any, ::Any, ::Any) at none:0 (method too new to be called from this world context.)
Stacktrace:
 [1] (::ModelingToolkit.var"#_jac#88"{ModelingToolkit.var"#310#312",ModelingToolkit.var"#314#315"})(::Array{ModelingToolkit.Operation,1}, ::Array{ModelingToolkit.Operation,1}, ::ModelingToolkit.Operation) at C:\Users\DPad\.julia\dev\ModelingToolkit\src\systems\diffeqs\abstractodesystem.jl:163
 [2] (::DiffEqBase.EvalFunc{ModelingToolkit.var"#_jac#88"{ModelingToolkit.var"#310#312",ModelingToolkit.var"#314#315"}})(::Array{ModelingToolkit.Operation,1}, ::Vararg{Any,N} where N) at C:\Users\DPad\.julia\packages\DiffEqBase\TXlS1\src\solve.jl:8
```

Is there any scope for ModelingToolkit to be used at precompile time of other packages? I was previously able to use `MyModule.PREBUILT_ODE_FUNCTION_CONST`, instead of having to call `MyModule.build_ode_function_with_ModelingToolkit()` at runtime.

ChrisRackauckas:
>Is there any scope for ModelingToolkit to be used at precompile time of other packages? I was previously able to use MyModule.PREBUILT_ODE_FUNCTION_CONST, instead of having to call MyModule.build_ode_function_with_ModelingToolkit() at runtime.

There isn't a reason to run MTK at precompile time if you've already generated the function. I think we should just have better ways to save the generated code to a file for users to slap into modules.

dpad:
Ok, I was seemingly able to get thinks precompiling again by calling `Base.invokelast(my_ode_function.jac, dvs, ps, iv)` instead (keeping all the warnings). So I think it wasn't related to this PR, I'll reopen it for review.

For running at precompile time, I guess my motivation for this is that I have some ODEFunctions being built based on other generated ODEFunctions (e.g. their automatically computed Jacobian), all specified nicely with ModelingToolkit's beautiful syntax. I quite liked this because my package only included these clean models plus some simple to understand "recipes" for building ODEFunctions out of combinations of them. 

I'm not sure what the other workflow would like like, I guess I'd have to manually run my recipes, export the resulting generated ODEFunction to a file, and save that file somewhere that gets included in my package? Or maybe such a file would be generated as a part of "build.jl", for example as an artifact? I'm not really sure how it would work.

dpad:
I was seemingly able to get some things precompiling again by calling `Base.invokelast(my_ode_function.jac, dvs, ps, iv)` instead (EDIT: at least it seems to get past the issue before, but eventually something else seems to crash Julia completely, so I'll have to keep digging). Either way I think it might not be related to this PR, so I'll reopen it for review.

For running at precompile time, I guess my motivation for this is that I have some ODEFunctions being built based on other generated ODEFunctions (e.g. their automatically computed Jacobian), all specified nicely with ModelingToolkit's beautiful syntax. I quite liked this because my package only included these clean models plus some simple to understand "recipes" for building ODEFunctions out of combinations of them. 

I'm not sure what the other workflow would like like, I guess I'd have to manually run my recipes, export the resulting generated ODEFunction to a file, and save that file somewhere that gets included in my package? Or maybe such a file would be generated as a part of "build.jl", for example as an artifact? I'm not really sure how it would work.

ChrisRackauckas:
>For running at precompile time, I guess my motivation for this is that I have some ODEFunctions being built based on other generated ODEFunctions (e.g. their automatically computed Jacobian), all specified nicely with ModelingToolkit's beautiful syntax. I quite liked this because my package only included these clean models plus some simple to understand "recipes" for building ODEFunctions out of combinations of them.

>I'm not sure what the other workflow would like like, I guess I'd have to manually run my recipes, export the resulting generated ODEFunction to a file, and save that file somewhere that gets included in my package? Or maybe such a file would be generated as a part of "build.jl", for example as an artifact? I'm not really sure how it would work.

Yeah it really depends on how often you're generating new versions of the equations. Stick with what you were before for your case. We might want to add an option in `ODEFunction` etc. to use GeneralizedGenerated, specifically for your kinds of cases, while defaulting to straight `eval` in the case of more interactive function definition (since otherwise you can kill Julia by building millions of types!)

dpad:
> We might want to add an option in `ODEFunction` etc. to use GeneralizedGenerated, specifically for your kinds of cases

Great idea, I'll look into doing that in a separate PR when I get the chance.
Thanks for all your comments! I'll leave this PR open in case you want to merge it (it seems to work at least, not sure how useful it is outside my use case).

ChrisRackauckas:
Yeah, this is good for anything that's not in a precompiled module, so I think I'll merge it anyways. But we should get you a solution because that's an important case. https://github.com/SciML/ModelingToolkit.jl/pull/466 might be a good one. You can then either save it out to a file or eval it.

PR #466, opened 2020-06-17
Function and Expr outs

ChrisRackauckas:


ChrisRackauckas:
@dpad this might be what you need, since then you could just `eval` at the top level which should be safe.

@YingboMa @ranjan this should be what you need

ChrisRackauckas:
@isaacsas I didn't finish up JumpProblem here since I assume you'll want to change a lot in https://github.com/SciML/ModelingToolkit.jl/pull/444

dpad:
> @dpad this might be what you need, since then you could just eval at the top level which should be safe.

Thanks a lot, will look into it now that it's merged :)

PR #476, opened 2020-06-23
Update substitute test

mforets:
The file `test/variable_utils.jl` doesn't seem to be tested. If this was not on purpose, there is a test that needs update because in master gives

```julia
julia> @test isequal(expr, new)
Test Failed at REPL[13]:1
  Expression: isequal(expr, new)
   Evaluated: isequal((((1 / β - 1) + δ) / α) ^ (1 / (α - 1)), (((1 / β - 1) + δ) / γ) ^ (1 / (γ - 1)))
```


ChrisRackauckas:
Thanks! This will help. @shashi I assume that this and the last ping are related/the same.

PR #477, opened 2020-06-23
Provide as_expression=false option to ODEs/SDEs, for specific use cases that need to avoid world age issues (generally not needed)

dpad:
New option that restores the GeneralizedGenerated functions in ODEFunction, for some very specific use cases:

* Building an ODEFunction during precompilation (albeit still with long "evaluating into a closed module/incremental compilation may be fatally broken" warnings).
* Building an ODEFunction and calling it from within the same non-global scope.


ChrisRackauckas:
IMO, this should be applied uniformly, i.e. ODEProblem, SDEProblem/Function, Nonlinear, Optimization

dpad:
I've added ODEProblem, SDEProblem, and SDEFunction. However, Nonlinear and Optimization seem to already be split according to `expression=Val{false}` and `expression=Val{true}` for the Expr version, so I wasn't too sure what I should change there.

PR #481, opened 2020-06-25
don't call get_variables! on non-Operations for jumps

isaacsas:


ChrisRackauckas:
just an optimization?

isaacsas:
No, it was crashing since the method is only defined on a few types. (Specifically, if the jump rate is just number it crashes.)

The alternative would be to add some generic versions of `get_variables!` that just return the input. I can do that instead if you think that might be useful elsewhere.

ChrisRackauckas:
avoiding multiple dispatch is better for performance anyways, so this is good

PR #489, opened 2020-07-04
Restructure tutorial section of the documentation

ChrisRackauckas:
@dpsanders I'd like you to take a look after this goes through for some thoughts and opinions


dpsanders:
Personally I would start with the most basic level, something like

```jl
using ModelingToolkit

f(x, y, z) = sin(x + y) * z

@variables a, b, c

2 * f(a, b, c)
```

to show that in the end you have a symbolic manipulation tool.

PR #497, opened 2020-07-06
force_SA and document new options

ChrisRackauckas:


ChrisRackauckas:
@dcabecinhas is this what you need?

dcabecinhas:
> @dcabecinhas is this what you need?

Yes, this works. Thanks.

PR #498, opened 2020-07-07
Fast Jacobian Sparsity

shashi:
From 121 seconds to:
```
julia> @time jac = ModelingToolkit.jacobian_sparsity(vec(du),vec(u))
  0.401049 seconds (2.62 M allocations: 61.215 MiB, 2.42% gc time)
3072×3072 SparseMatrixCSC{Bool,Int64} with 13184 stored entries:
```
on https://github.com/JuliaSymbolics/SymbolicUtils.jl/issues/61#issuecomment-653410335

### DIFF
--- a/src/direct.jl
+++ b/src/direct.jl
@@ -24,6 +24,55 @@ end
 
 """
 ```julia
+sparsejacobian(ops::AbstractVector{<:Expression}, vars::AbstractVector{<:Expression}; simplify = true)
+```
+
+A helper function for computing the sparse Jacobian of an array of expressions with respect to
+an array of variable expressions.
+"""
+function sparsejacobian(ops::AbstractVector{<:Expression}, vars::AbstractVector{<:Expression}; simplify = true)
+    I = Int[]
+    J = Int[]
+    du = Expression[]
+
+    sp = jacobian_sparsity(ops, vars)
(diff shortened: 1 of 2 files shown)


src/direct.jl
```diff
@@ -22,6 +22,55 @@ function jacobian(ops::AbstractVector{<:Expression}, vars::AbstractVector{<:Expr
...
+sparsejacobian(ops::AbstractVector{<:Expression}, vars::AbstractVector{<:Expression}; simplify = true)
+```
+
+A helper function for computing the sparse Jacobian of an array of expressions with respect to
+an array of variable expressions.
+"""
+function sparsejacobian(ops::AbstractVector{<:Expression}, vars::AbstractVector{<:Expression}; simplify = true)
+    I = Int[]
+    J = Int[]
+    du = Expression[]
```
ChrisRackauckas, line 36:
```suggestion
```

src/direct.jl
```diff
@@ -22,6 +22,55 @@ function jacobian(ops::AbstractVector{<:Expression}, vars::AbstractVector{<:Expr
...
+
+    sp = jacobian_sparsity(ops, vars)
```
ChrisRackauckas, line 38: actually, this was meant to be du?

PR #499, opened 2020-07-07
jacobian performance improvements

shashi:
tests coming.

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -1,3 +1,10 @@
+function make_operation(op, args...)
+    if all(x-> x isa Constant || !(x isa Expression), args)
+        return op(map(x->x isa Constant ? x.value : x, args)...)
+    else
+        return op(args...)
+    end
+end
 Base.convert(::Type{Expression}, ex::Expr) = Expression(ex)
 function Expression(ex;mod=Main)
     ex.head === :if && (ex = Expr(:call, ifelse, ex.args...))
@@ -5,11 +12,7 @@ function Expression(ex;mod=Main)
 
     op = getproperty(mod,Symbol(ex.args[1]))
     args = convert.(Expression, ex.args[2:end])
-    if all(x-> x isa Constant || !(x isa Expression), args)
-        return op(map(x->x isa Constant ? x.value : x, args)...)
-    else
-        return op(args...)
-    end
+    make_operation(op, args...)
 end
 Base.convert(::Type{Expression}, x::Expression) = x
 Base.convert(::Type{Expression}, x::Number) = Constant(x)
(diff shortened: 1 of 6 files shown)


shashi:
From not being able to finish to 2.5 seconds with simplification ;)

```julia
julia> @time jac = ModelingToolkit.sparsejacobian(vec(du),vec(u),simplify=true)
  2.503883 seconds (20.93 M allocations: 599.834 MiB, 4.96% gc time)
3072×3072 SparseMatrixCSC{Expression,Int64} with 13184 stored entries:
  [1   ,    1]  =  -1u₁ˏ₁ˏ₂
  [2   ,    1]  =  Constant(100.0)
  [33  ,    1]  =  Constant(100.0)
  [1025,    1]  =  -1u₁ˏ₁ˏ₂
  [2049,    1]  =  u₁ˏ₁ˏ₂
  [1   ,    2]  =  Constant(200.0)
  [2   ,    2]  =  -1u₂ˏ₁ˏ₂
  [3   ,    2]  =  Constant(100.0)
  [34  ,    2]  =  Constant(100.0)
  [1026,    2]  =  -1u₂ˏ₁ˏ₂
  ⋮
  [3069, 3069]  =  Constant(-2.0)
  [1022, 3070]  =  Constant(1.0)
  [2046, 3070]  =  Constant(1.0)
  [3070, 3070]  =  Constant(-2.0)
  [1023, 3071]  =  Constant(1.0)
  [2047, 3071]  =  Constant(1.0)
  [3071, 3071]  =  Constant(-2.0)
  [1024, 3072]  =  Constant(1.0)
  [2048, 3072]  =  Constant(1.0)
  [3072, 3072]  =  Constant(-2.0)
```

test/variable_parsing.jl
```diff
@@ -35,7 +35,7 @@ D1 = Differential(t)
 @test @macroexpand(@parameters x, y, z(t)) == @macroexpand(@parameters x y z(t))
 @test @macroexpand(@variables x, y, z(t)) == @macroexpand(@variables x y z(t))
 
-convert(Expression, :($x == 0 ? $y : $x))
+@test_broken convert(Expression, :($x == 0 ? $y : $x))
```
ChrisRackauckas, line 38: that never really worked before so it's fine.

src/utils.jl
```diff
@@ -1,12 +1,29 @@
+function make_operation(op, args)
```
ChrisRackauckas, line 1: looks like this is where all of the complexity will go... which is fine, because if we just make this complex a lot of things will get faster and stay nice, so that's fine with me. I think this is a case where we should investigate `@nospecailize(op)` and `op ==` since replying on multiple dispatch + compilation will be slower than a branch.

ChrisRackauckas: I think it's fine to merge without doing this, but @shashi could you look into timing that?

shashi: Ok yeah.

PR #501, opened 2020-07-08
pervasive use of sparsejacobian

ChrisRackauckas:


ChrisRackauckas:
@shashi thanks!

@isaacsas I don't know if you saw `sparsejacobian`, but it's orders of magnitude faster for generating sparse Jacobians than what we had before, so we should probably start doing a lot of testing on systems like BCR.

isaacsas:
I'm working on finishing up the DEBio conversion, hopefully I'll be done soon, then I'll work on getting the importers converted to MT. At that point we can test performance of BCR and how it has changed.

PR #504, opened 2020-07-09
skipzeros by default in diffeq calls

ChrisRackauckas:
DiffEq is safe with this, and this will speed up codegen quite a bit.


YingboMa:
Superseded by #508

PR #508, opened 2020-07-09
Add fillzeros kwarg to build_function

YingboMa:
Add `fillzeros`

### DIFF
--- a/src/build_function.jl
+++ b/src/build_function.jl
@@ -170,7 +170,7 @@ function is_array_array_sparse_matrix(F)
 end
 
 function fill_array_with_zero!(x::AbstractArray)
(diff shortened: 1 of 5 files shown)


src/build_function.jl
```diff
@@ -215,6 +224,10 @@ function _build_function(target::JuliaTarget, rhss, args...;
 	end
 
     ip_sys_exprs = Expr[]
+    # we cannot reliably fill the array with the presence of index translation
+    if fillzeros && fillzeros === nothing
```
isaacsas, line 228: Should this be `outputidxs===nothing`?

YingboMa: Yeah, typed it wrong :p

PR #512, opened 2020-07-10
WIP: Really complicated occursin optimization

shashi:
We first crawl the expression and mark which subtrees have the independent variable.

### DIFF
--- a/src/differentials.jl
+++ b/src/differentials.jl
@@ -62,15 +62,15 @@ TODO
 function expand_derivatives(O::Operation, simplify=true; occurances=nothing)
     if isa(O.op, Differential)
         @assert length(O.args) == 1
+        arg = expand_derivatives(O.args[1], false)
 
         if occurances == nothing
-            occurances = occursin_info(O.op.x, O.args[1])
+            occurances = occursin_info(O.op.x, arg)
         end
 
         _isfalse(occurances) && return Constant(0)
         occurances isa Constant && return Constant(1) # means it's a Constant(true)
 
-        arg = expand_derivatives(O.args[1], false)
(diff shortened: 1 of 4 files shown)


shashi:
```
julia> @time ModelingToolkit.sparsehessian(rosenbrock(x), x, simplify=false);
  6.096420 seconds (24.14 M allocations: 1.197 GiB, 4.94% gc time)

julia> @time ModelingToolkit.sparsehessian(rosenbrock(x), x, simplify=false);
  7.673513 seconds (26.64 M allocations: 1.263 GiB, 4.47% gc time)
```
there is a difference, I guess.

ChrisRackauckas:
that's a pretty big difference!

shashi:
ok with only the last commit

```julia
julia> @time ModelingToolkit.sparsehessian(rosenbrock(x), x, simplify=true);
  2.823342 seconds (4.16 M allocations: 94.528 MiB, 0.92% gc time)
```

With this whole PR:
```julia
julia> @time ModelingToolkit.sparsehessian(rosenbrock(x), x, simplify=true);
  1.405966 seconds (2.76 M allocations: 85.100 MiB, 1.94% gc time)
```

So the occursin optimization is within the `O.op isa Differential` branch. The `hasderiv` optimization is on the other.

shashi:
```julia
julia> @time ModelingToolkit.sparsejacobian(ModelingToolkit.sparsejacobian([rosenbrock(x)], x)[:], x);
  1.452917 seconds (3.48 M allocations: 105.766 MiB, 1.50% gc time)
```

Dammit this is still pretty fast.

PR #513, opened 2020-07-10
define modelingtoolkitize for SDESystem with conversion tests

frankschae:
This PR adds a dispatch for `modelingtoolkitize` of an SDESystem and allows us to convert the drift term from Ito to Stratonovich sense and vice versa.  This should be useful to support also SDE adjoints for Ito SDEs, see https://github.com/SciML/DiffEqSensitivity.jl/issues/279 .

The conversion factor of 1//2 is not strictly fixed, as in the adjoints, one needs to apply the correction term two times. Therefore, this could now be handled by 

`sys = modelingtoolkitize(prob,correction_factor=-1)`

### DIFF
--- a/src/systems/diffeqs/modelingtoolkitize.jl
+++ b/src/systems/diffeqs/modelingtoolkitize.jl
@@ -26,3 +26,79 @@ function modelingtoolkitize(prob::DiffEqBase.ODEProblem)
 
     de
 end
+
+
+
+"""
+$(TYPEDSIGNATURES)
+
+Generate `SDESystem`, dependent variables, and parameters from an `SDEProblem`.
+Choose correction_factor=-1//2 (1//2) to converte Ito -> Stratonovich (Stratonovich->Ito).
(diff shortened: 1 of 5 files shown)


src/systems/diffeqs/modelingtoolkitize.jl
```diff
@@ -26,3 +26,79 @@ function modelingtoolkitize(prob::DiffEqBase.ODEProblem)
...
+    else
+        lhs = prob.f(vars, params, t)
+        if DiffEqBase.is_diagonal_noise(prob)
+            neqs = prob.g(vars, params, t)
+        else
+            neqs = prob.g(vars, params, t)
+        end
+    end
+
+    if correction_factor!=nothing
```
ChrisRackauckas, line 70: for modularity, I think it would be better to just return the SDESystem as it's defined, and then have a transformation function `stochastic_integral_transform(sys,correctionfactor)` that takes an Ito system to Stratanovich and vice versa.

PR #515, opened 2020-07-12
update ReactionSystem docs and add reaction dep graphs

isaacsas:
@ChrisRackauckas Alternatively I could move `ReactionSystem` over now to DEB and change this PR to take that functionality out of MT.

### DIFF
--- a/docs/src/systems/ReactionSystem.md
+++ b/docs/src/systems/ReactionSystem.md
@@ -49,6 +49,8 @@ ReactionSystem
 
 ## Query Functions
 ```@docs
+oderatelaw
+jumpratelaw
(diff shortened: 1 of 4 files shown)


isaacsas:
OK, should be good to go unless it would be better to move everything to DEBio now.

ChrisRackauckas:
>Alternatively I could move ReactionSystem over now to DEB and change this PR to take that functionality out of MT.

Let's give it about a month. Once the causal system is finalized, the basic `AbstractSystem` structure should be much more fixed and then it will make sense on its own. Until then I think it may have to move in tandem in those coming changes.

PR #517, opened 2020-07-13
Enable noise scaling in convert(SDESystem,::ReactionSystem)

TorkelE:
This adds a kwarg to the function:
```julia
convert(::Type{<:SDESystem},rs::ReactionSystem;noise_scaling_parameter=:internal___no___noise___scalling::Symbol)
```
if the argument is set to anything but its default, a variable with that name will be created and added to the parameter vector of the outputted SDESystem. This parameter will linearly scale the noise in the created chemical Langevin equation.

Can also use it as input when creating `SDEProblems`
```julia
SDEProblem(rs::ReactionSystem, u0::Union{AbstractArray, Number}, tspan, p, args... ;noise_scaling_parameter=:internal___no___noise___scalling::Symbol, kwargs...)
```

### DIFF
--- a/src/systems/reaction/reactionsystem.jl
+++ b/src/systems/reaction/reactionsystem.jl
@@ -418,11 +418,12 @@ function DiffEqBase.ODEProblem(rs::ReactionSystem, u0::Union{AbstractArray, Numb
 end
 
 # SDEProblem from AbstractReactionNetwork
-function DiffEqBase.SDEProblem(rs::ReactionSystem, u0::Union{AbstractArray, Number}, tspan, p, args...; kwargs...)
-    u0 = typeof(u0) <: Array{<:Pair} ? u0 : Pair.(rs.states,u0)
-    p = typeof(p) <: Array{<:Pair} ? p : Pair.(rs.ps,p)
-    p_matrix = zeros(length(rs.states), length(rs.eqs))
-    return SDEProblem(convert(SDESystem,rs),u0,tspan,p,args...; noise_rate_prototype=p_matrix,kwargs...)
+function DiffEqBase.SDEProblem(rs::ReactionSystem, u0::Union{AbstractArray, Number}, tspan, p, args... ;noise_scaling_parameter=:internal___no___noise___scalling::Symbol, kwargs...)
+    sde_sys = convert(SDESystem,rs,noise_scaling_parameter=noise_scaling_parameter)
+    u0 = typeof(u0) <: Array{<:Pair} ? u0 : Pair.(sde_sys.states,u0)
+    p = typeof(p) <: Array{<:Pair} ? p : Pair.(sde_sys.ps,p)
+    p_matrix = zeros(length(sde_sys.states), length(sde_sys.eqs))
+    return SDEProblem(sde_sys,u0,tspan,p,args...; noise_rate_prototype=p_matrix,kwargs...)
 end
 
 # DiscreteProblem from AbstractReactionNetwork
--- a/src/systems/reaction/reactionsystem.jl
+++ b/src/systems/reaction/reactionsystem.jl
@@ -154,15 +154,15 @@ generated ODEs for the reaction. Note, for a reaction defined by
 `k*X*Y, X+Z --> 2X + Y`
 
 the expression that is returned will be `k*X(t)^2*Y(t)*Z(t)`. For a reaction of the
-form 
+form
 
 `k, 2X+3Y --> Z`
 
-the `Operation` that is returned will be `k * (X(t)^2/2) * (Y(t)^3/6)`. 
+the `Operation` that is returned will be `k * (X(t)^2/2) * (Y(t)^3/6)`.
 
 Notes:
 - Allocates
-""" 
+"""
 function oderatelaw(rx)
     @unpack rate, substrates, substoich, only_use_rate = rx
     rl = rate
@@ -199,12 +199,13 @@ function assemble_drift(rs)
     eqs
 end
 
-function assemble_diffusion(rs)
+function assemble_diffusion(rs,updated_ps)
     eqs = Expression[Constant(0) for x in rs.states, y in rs.eqs]
     species_to_idx = Dict((x => i for (i,x) in enumerate(rs.states)))
 
     for (j,rx) in enumerate(rs.eqs)
         rlsqrt = sqrt(oderatelaw(rx))
+        (length(updated_ps)!=length(rs.ps)) && (rlsqrt *=var2op(updated_ps[end]))
         for (spec,stoich) in rx.netstoich
             i            = species_to_idx[spec]
             signedrlsqrt = (stoich > zero(stoich)) ? rlsqrt : -rlsqrt
@@ -230,7 +231,7 @@ for a reaction defined by
 `k*X*Y, X+Z --> 2X + Y`
 
 the expression that is returned will be `k*X^2*Y*Z`. For a reaction of
-the form 
+the form
 
 `k, 2X+3Y --> Z`
 
@@ -240,7 +241,7 @@ binomial(Y,3)`.
 Notes:
 - `rxvars` should give the `Variable`s, i.e. species and parameters, the rate depends on.
 - Allocates
-""" 
+"""
 function jumpratelaw(rx; rxvars=get_variables(rx.rate))
     @unpack rate, substrates, substoich, only_use_rate = rx
     rl = rate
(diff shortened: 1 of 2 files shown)


src/systems/reaction/reactionsystem.jl
```diff
@@ -358,10 +359,11 @@ Base.convert(::Type{<:SDESystem},rs::ReactionSystem)
 
 Convert a ReactionSystem to a SDESystem.
 """
-function Base.convert(::Type{<:SDESystem},rs::ReactionSystem)
+function Base.convert(::Type{<:SDESystem},rs::ReactionSystem;noise_scaling_parameter=:internal___no___noise___scalling::Symbol)
```
ChrisRackauckas, line 362: why not a gensym?

src/systems/reaction/reactionsystem.jl
```diff
@@ -358,10 +359,11 @@ Base.convert(::Type{<:SDESystem},rs::ReactionSystem)
 
 Convert a ReactionSystem to a SDESystem.
 """
-function Base.convert(::Type{<:SDESystem},rs::ReactionSystem)
+function Base.convert(::Type{<:SDESystem},rs::ReactionSystem;noise_scaling_parameter=:internal___no___noise___scalling::Symbol)
     eqs = assemble_drift(rs)
-    noiseeqs = assemble_diffusion(rs)
-    SDESystem(eqs,noiseeqs,rs.iv,rs.states,rs.ps,
+    new_ps = (noise_scaling_parameter == :internal___no___noise___scalling) ? rs.ps : [rs.ps..., Variable(noise_scaling_parameter)]
```
ChrisRackauckas, line 364: vcat

And only one noise scaling parameter?

src/systems/reaction/reactionsystem.jl
ChrisRackauckas: Shouldn't it be an `Operation`?

TorkelE: You mean like `gensym(:no_noise_scalling)`?

In principle, I guess it ends up becoming an operation, but I figured the user doesn't really care about what it is and just want to set the name, and an `Operation` is created from it down the code.

In a large portion of the cases, I think a symbol isn't even needed, and `use_noise_scalling=true` would be enough of an argument. The only time one really encounters the symbol set as the noise scaling parameter is if one looks at the actual content of the `SDESystem`, of if one prints the equation with Latexify.

src/systems/reaction/reactionsystem.jl
TorkelE: Right now only one for all of the reactions (if that is what you mean). I guess it would be easy to also add the opportunity to add a full vector, one for each reaction.

src/systems/reaction/reactionsystem.jl
TorkelE: Actually, if I make the default argument `gensym(:no_nois_scalling)`, how do I later check that the value of the variable is equal to that default. Calling `gensym(:no_nois_scalling)` again gives me another value.

TorkelE: `nothing` seems like a good alternative, let's go with that?

src/systems/reaction/reactionsystem.jl
```diff
@@ -199,12 +199,13 @@ function assemble_drift(rs)
...
 end
 
-function assemble_diffusion(rs)
+function assemble_diffusion(rs,updated_ps)
     eqs = Expression[Constant(0) for x in rs.states, y in rs.eqs]
     species_to_idx = Dict((x => i for (i,x) in enumerate(rs.states)))
 
     for (j,rx) in enumerate(rs.eqs)
         rlsqrt = sqrt(oderatelaw(rx))
+        (length(updated_ps)!=length(rs.ps)) && (rlsqrt *=var2op(updated_ps[end]))
```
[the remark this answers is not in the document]
TorkelE: My Git said you suggested changes, but was unsure how to make them happen. I simply wrote it up and pushed it.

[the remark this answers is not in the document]
isaacsas: No worries!

src/systems/reaction/reactionsystem.jl
```diff
@@ -358,10 +359,10 @@ Base.convert(::Type{<:SDESystem},rs::ReactionSystem)
 
 Convert a ReactionSystem to a SDESystem.
 """
-function Base.convert(::Type{<:SDESystem},rs::ReactionSystem)
+function Base.convert(::Type{<:SDESystem},rs::ReactionSystem;noise_scaling_parameter=nothing::Union{Symbol,Nothing})
     eqs = assemble_drift(rs)
-    noiseeqs = assemble_diffusion(rs)
-    SDESystem(eqs,noiseeqs,rs.iv,rs.states,rs.ps,
+    noiseeqs = assemble_diffusion(rs,noise_scaling_parameter)
+    SDESystem(eqs,noiseeqs,rs.iv,rs.states,(noise_scaling_parameter==nothing) ? rs.ps : vcat(rs.ps,Variable(noise_scaling_parameter)),
```
isaacsas, line 365: `===`?

src/systems/reaction/reactionsystem.jl
```diff
@@ -199,12 +199,13 @@ function assemble_drift(rs)
...
 end
 
-function assemble_diffusion(rs)
+function assemble_diffusion(rs,noise_scaling_parameter::Union{Symbol,Nothing})
     eqs = Expression[Constant(0) for x in rs.states, y in rs.eqs]
     species_to_idx = Dict((x => i for (i,x) in enumerate(rs.states)))
 
     for (j,rx) in enumerate(rs.eqs)
         rlsqrt = sqrt(oderatelaw(rx))
+        (noise_scaling_parameter!=nothing) && (rlsqrt *= var2op(Variable(noise_scaling_parameter)))
```
isaacsas, line 208: `!==`?

src/systems/reaction/reactionsystem.jl
```diff
@@ -414,13 +415,14 @@ function DiffEqBase.ODEProblem(rs::ReactionSystem, u0::Union{AbstractArray, Numb
     p = typeof(p) <: Array{<:Pair} ? p : Pair.(rs.ps,p)
     return ODEProblem(convert(ODESystem,rs),u0,tspan,p, args...; kwargs...)
 end
-
+gensym(:no_noise_scalling)
```
isaacsas, line 418: What’s this for?

src/systems/reaction/reactionsystem.jl
TorkelE: If the noise scaling parameter is nothing, use the unchanged parameters of rs.
If the noise scaling parameter is not nothing, append it to the parameters of rs, and use that as parameters for the `SDESystem`.

src/systems/reaction/reactionsystem.jl
TorkelE: Nothing, was testing how the gensym function worked, removed it now.

src/systems/reaction/reactionsystem.jl
TorkelE: If the noise scaling parameter is not nothing, multiply the noise term by it.

isaacsas: Right, but doesn’t `===` and `!==` allow the compiler to potentially evaluate the branch at compile time instead of runtime? That was my whole reason for suggesting using `nothing`, trying the write the code in a way that it would be optimized away when the scaling parameter is not used.

TorkelE: Ahh, you mean to replace `==` and `!=` by `===` and `!==`? Not too familiar with these, but could definitely do that, good point.

isaacsas: Sorry if I was confusing! I was multitasking (poorly) when commenting on this PR...

TorkelE:
Small update. The keyword is now the slightly more ambiguous `noise_scaling_param`, which can be either
- `nothing` (no noise scaling, no parameter added)
- A `Vector{Symbol}` of the length of the number of reactions. This allows for individual scaling of each reaction. The number of parameters added is the same as the number of unique symbols in the input vector.
- A `Symbol`, which then triggers `(typeof(noise_scaling_param) <: Symbol) && (noise_scaling_param = fill(noise_scaling_param,length(rs.eqs)))` and the above case is used, but with all elements being the same.
I also added an appropriate test for this.

src/systems/reaction/reactionsystem.jl
```diff
@@ -358,10 +359,12 @@ Base.convert(::Type{<:SDESystem},rs::ReactionSystem)
 
 Convert a ReactionSystem to a SDESystem.
 """
-function Base.convert(::Type{<:SDESystem},rs::ReactionSystem)
+function Base.convert(::Type{<:SDESystem},rs::ReactionSystem;noise_scaling_param=nothing::Union{Vector{Symbol},Symbol,Nothing})
```
isaacsas, line 362: This is bikeshedding of the highest degree, so feel free to completely ignore, but maybe use `noise_scaling` since it can be a single param, or a vector of many params?

isaacsas: Or `noise_scaling_params`.

TorkelE: I first had `noise_scaling_parameters`, but the s felt weird when the likely most common case would be a single parameter. `noise_scaling` is also short, so I'm all for that.

ChrisRackauckas:
looks like the diff is messed up?

TorkelE:
Probably, but what diff you mean? There were some conflicts with later updates which I tried to merge at GitHub, that might have messed something up.

ChrisRackauckas:
The entire reactionsystem file

TorkelE:
Yeah, no idea why that happened. I think it was when I fixed the conflicts. However, I only changed the few lines marked as conflicts, so not sure why it marks the entire file as replaced. I can try to figure it out.

TorkelE:
In the end, the simple solution seemed to just make another PR, to have to avoid the merging issues: https://github.com/SciML/ModelingToolkit.jl/pull/530

PR #518, opened 2020-07-14
WIP: Unify the Term definition between MTK and SymbolicUtils

YingboMa:
Downstream PRs required if the name changes go through:

- [ ] https://github.com/SciML/ParameterizedFunctions.jl/blob/master/src/ode_def_opts.jl#L17-L19
- [ ] https://github.com/PumasAI/Pumas.jl/blob/master/src/dsl/model_macro.jl#L359-L376
- [ ] https://github.com/SciML/Catalyst.jl/blob/master/src/reaction_network.jl#L135
- [ ] https://github.com/SciML/DiffEqOperators.jl/blob/master/test/MOLtest.jl
- [ ] https://github.com/SciML/DiffEqProblemLibrary.jl/blob/master/src/ode/ode_simple_nonlinear_prob.jl#L171
- [ ] https://github.com/SciML/DiffEqTutorials.jl/blob/4e5e0c6a44efa421e2d77677097e79732a1619a0/tutorials/ode_extras/01-ModelingToolkit.jmd
- [ ] https://github.com/SciML/DataDrivenDiffEq.jl/blob/46b0ff41722f634cf76676e4ce508323080b8f30/test/basis.jl


YingboMa:
Note: `NumWrap` is wrapped and unwrapped for all user-facing APIs, so that the user can get the benefit of `NumWrap <: Number`. However, no internal computation should generate or handle `NumWrap`, since it does not add any utilities to the actual computation. Also, `NumWrap` is strictly disallowed in any subtrees.

src/ModelingToolkit.jl
```diff
@@ -50,6 +50,8 @@ SymbolicUtils.@number_methods(NumWrap,
 
 import SymbolicUtils: <ₑ, Symbolic, Term, operation, arguments
 
+Base.promote_rule(::Type{<:Number}, ::Type{<:NumWrap}) = NumWrap
+Base.promote_rule(::Type{<:Symbolic{<:Number}}, ::Type{<:NumWrap}) = NumWrap
```
shashi, line 54: Nice!!

YingboMa:
Ref: #520

test/simplify.jl
```diff
@@ -19,16 +19,16 @@ simplify(minus_op)
...
+@test toexpr(expand_derivatives(Differential(x)((x-2)^3))) == :(3 * (-2 + x)^2)
+@test toexpr(simplify(x+2+3)) == :(5 + x)
 
 d1 = Differential(x)((-2 + x)^2)
 d2 = Differential(x)(d1)
 d3 = Differential(x)(d2)
 
-@test simplified_expr(expand_derivatives(d3)) == :(0)
-@test simplified_expr(simplify(x^0)) == :(1)
+@test toexpr(expand_derivatives(d3)) == :(0)
```
ChrisRackauckas, line 30: convert(Expr,...)`?

test/direct.jl
```diff
@@ -45,19 +45,20 @@ rosenbrock(X) = sum(1:length(X)-1) do i
...
 
-@variables X[1:10]
+@variables a,b
+X = [a,b]
 
+spoly(x) = simplify(x, polynorm=true)
 rr = rosenbrock(X)
 
-reference_hes = ModelingToolkit.hessian(rosenbrock(X), X)
+reference_hes = ModelingToolkit.hessian(rr, X)
```
ChrisRackauckas, line 54: wow, user code is unchanged here great

test/derivatives.jl
```diff
@@ -90,12 +90,12 @@ t1 = ModelingToolkit.gradient(tmp, [x1, x2])
 @parameters t k
 @variables x(t)
 @derivatives D'~k
-@test convert(Variable, D(x)).name === :xˍk
```
ChrisRackauckas, line 93: why is this changed?

src/differentials.jl
```diff
@@ -174,8 +179,8 @@ julia> ModelingToolkit.derivative(myop, 2)  # wrt. y^2
 sin(x())
 ```
 """
-derivative(O::Operation, idx) = derivative(O.op, (O.args...,), Val(idx))
-derivative(O::Constant, ::Any) = Constant(0)
+derivative(O::Term, idx) = derivative(O.op, (O.args...,), Val(idx))
+derivative(O::Any, ::Any) = 0
```
ChrisRackauckas, line 183: `Number`?

ChrisRackauckas:
BTW, this is highly breaking so if we do the name change, we will need PRs prepared to all of the libraries that we maintain:

- [ ] https://github.com/SciML/ParameterizedFunctions.jl/blob/master/src/ode_def_opts.jl#L17-L19
- [ ] https://github.com/PumasAI/Pumas.jl/blob/master/src/dsl/model_macro.jl#L359-L376
- [ ] https://github.com/SciML/Catalyst.jl/blob/master/src/reaction_network.jl#L135
- [ ] https://github.com/SciML/DiffEqOperators.jl/blob/master/test/MOLtest.jl
- [ ] https://github.com/SciML/DiffEqProblemLibrary.jl/blob/master/src/ode/ode_simple_nonlinear_prob.jl#L171
- [ ] https://github.com/SciML/DiffEqTutorials.jl/blob/4e5e0c6a44efa421e2d77677097e79732a1619a0/tutorials/ode_extras/01-ModelingToolkit.jmd
- [ ] https://github.com/SciML/DataDrivenDiffEq.jl/blob/46b0ff41722f634cf76676e4ce508323080b8f30/test/basis.jl

src/build_function.jl
```diff
@@ -447,17 +447,24 @@ end
...
+	exs = [:($name[$i]) for (i, u) ∈ enumerate(vs)]
+	vs_names,exs
+end
+
+function term_to_symbol(t::Term)
+    if operation(t) isa Sym
+        s = nameof(operation(t))
+        @show s
+    else
+        error("really?")
```
ChrisRackauckas, line 460: yeah bro.

test/simplify.jl
shashi: since we don't have Constant, we'd have to pirate `convert(Expr, Any)`...

ChrisRackauckas: that would be terrifying for compile times 🤷

src/systems/jumps/jumpsystem.jl
```diff
@@ -305,12 +296,12 @@ end
...
         st = eq.lhs
-        (st.op in sts) && push!(mstates, st)
+        any(isequal(st), sts) && push!(mstates, st)
     end
 end
 
 function modified_states!(mstates, jump::MassActionJump, sts)
     for (state,stoich) in jump.net_stoch
-        (state.op in sts) && push!(mstates, state)
+        any(isequal(state), sts) && push!(mstates, state)
```
isaacsas, line 305: How does this compare performance wise to the previous check? `isequal` was very slow before when working on `Operation`s hence why we compared with the underlying `Variable`.

src/systems/jumps/jumpsystem.jl
```diff
@@ -294,9 +285,9 @@ end
 
 function get_variables!(dep, jump::MassActionJump, variables)
     sr = jump.scaled_rates
-    (sr isa Operation) && get_variables!(dep, sr, variables)
+    (sr isa Term) && get_variables!(dep, sr, variables)
     for varasop in jump.reactant_stoch
-        (varasop[1].op in variables) && push!(dep, varasop[1])
+        any(isequal(varasop[1]), variables) && push!(dep, varasop[1])
```
isaacsas, line 290: See below.

src/systems/jumps/jumpsystem.jl
```diff
@@ -305,12 +296,12 @@ end
 function modified_states!(mstates, jump::Union{ConstantRateJump,VariableRateJump}, sts)
     for eq in jump.affect!
         st = eq.lhs
-        (st.op in sts) && push!(mstates, st)
+        any(isequal(st), sts) && push!(mstates, st)
```
isaacsas, line 299: See below.

src/systems/reaction/reactionsystem.jl
```diff
@@ -93,11 +95,10 @@ end
 # calculates the net stoichiometry of a reaction as a vector of pairs (sub,substoich)
 function get_netstoich(subs, prods, sstoich, pstoich)
     # stoichiometry as a Dictionary
-    nsdict = Dict{Variable,eltype(sstoich)}(sub.op => -sstoich[i] for (i,sub) in enumerate(subs))
+    nsdict = Dict{Any, eltype(sstoich)}(sub => -sstoich[i] for (i,sub) in enumerate(subs))
```
isaacsas, line 98: Same as above.

src/systems/jumps/jumpsystem.jl
shashi: So this can actually be turned from `O(n*k)` where `n=length(net_stock), k = length(sts)` to a `O(n+k)` algorithm by just turning `sts` into a Set rather than a vector... We'd have to make a special `_in` for this since `in` may call `==`. In the new system `==` creates an Operation of type Bool which crashes comparison... This is usually a big help, e.g. https://github.com/SciML/ModelingToolkit.jl/pull/498/files#diff-2fe5177ac5b669b0517a4fea2d9d0d7aR60-R64

src/systems/jumps/jumpsystem.jl
shashi: same here. Also even doing `jacobian_sparsity(lhss(jump.effect!), sts)` would be pretty quick, `O(n+k)`.

src/systems/reaction/reactionsystem.jl
shashi: we're going to have to benchmark this one!

test/derivatives.jl
shashi: we don't have `convert(Variable,x)` anymore...

src/systems/reaction/reactionsystem.jl
```diff
@@ -311,37 +314,32 @@ end
...
     end
     #push!(rstoich, reactant_stoch)
     coef = (zeroorder || (!combinatoric_ratelaw)) ? one(eltype(substoich)) : prod(stoich -> factorial(stoich), substoich)
     (!isone(coef)) && (rate /= coef)
     #push!(rates, rate)
     net_stoch      = [Pair(var2op(p[1]),p[2]) for p in netstoich]
     #push!(nstoich, net_stoch)
-    MassActionJump(rate, reactant_stoch, net_stoch, scale_rates=false, useiszero=false)
+
+    #XXX:              vv--- this sucks
```
isaacsas, line 328: What’s this referring to?

src/systems/reaction/reactionsystem.jl
```diff
@@ -311,37 +314,32 @@ end
...
-        (rx.rate isa Operation) && get_variables!(rxvars, rx.rate)
-        haveivdep = false
-        @inbounds for i = 1:length(rxvars)
-            if rxvars[i].op.name == ivname
-                haveivdep = true
-                break
-            end
-        end
+        (rx.rate isa Term) && get_variables!(rxvars, rx.rate)
+        haveivdep = any(isequal(rs.iv), rxvars)
```
isaacsas, line 342: These types of `isequal` tests were very slow before, hence why we dropped to comparing with the underlying symbol. Is this not an issue in the new setup?

isaacsas: If I remember right, I think `any` was having issues due to the lack of concrete typing too, which is why I switched to the loop after profiling.

src/systems/jumps/jumpsystem.jl
isaacsas: That’s one change I was heading towards in https://github.com/SciML/ModelingToolkit.jl/pull/554
 but I quickly discovered that lack of concrete typing was giving issues with Dicts of Operations or Variables, and I was going to need to drop down to Symbols...

shashi: Ahh okay. I came here to say just using `in` with a `Set` does the right thing!

Maybe a nice way to improve speed across the project is, to have data structures that wrap Dict but are as fast as possible for Symbolic exprs.

src/systems/reaction/reactionsystem.jl
shashi: So what's the benchmark to beat? If I ran https://discourse.julialang.org/t/lightgraph-modeltoolkit-optimization/40696 with your PR will that be a good test of all these cases? I can make sure before I merge this that the specific benchmark does not take any major hits.

isaacsas: Yeah, that would be good. My PR got hung up per the discussion there as I was struggling to further improve performance without switching to comparisons / Dicts of the underlying Symbols.

Here is a gist of what I think is the problem size I was testing with

https://gist.github.com/isaacsas/593f9ce7f40adf05b6969ace8fe8a831

More generally, with the current MT I frequently was finding that for data structures and/or comparisons, when working with `Operation`s that I know are really just `Variable`s, it was better to use the underlying `Variable` or possibly even the `Symbol` within the `Variable`. This was both for memory allocation reasons and for performance reasons. On systems with many comparisons or many state variables this could make a huge performance difference. With the changes here what should be the preferred workflow for such comparisons / look ups?

isaacsas: That said, the PR did speed up most of the jump related methodology until the graph creation, and my previous PRs had sped up much of the `ReactionSystem` functionality.

src/ModelingToolkit.jl
```diff
@@ -38,6 +38,97 @@ Base type for a symbolic expression.
...
+end
+
+import SymbolicUtils: <ₑ, Symbolic, Term, operation, arguments
+
+Base.show(io::IO, n::Num) = show_numwrap[] ? print(io, :(Num($(value(n))))) : Base.show(io, value(n))
+
+Base.promote_rule(::Type{<:Number}, ::Type{<:Num}) = Num
+Base.promote_rule(::Type{<:Symbolic{<:Number}}, ::Type{<:Num}) = Num
+Base.getproperty(t::Term, f::Symbol) = f === :op ? operation(t) : f === :args ? arguments(t) : getfield(t, f)
+<ₑ(s::Num, x) = value(s) <ₑ value(x)
```
ChrisRackauckas, line 82: what's this?

YingboMa:
This PR needs to wait for https://github.com/JuliaRegistries/General/pull/23157

src/ModelingToolkit.jl
YingboMa: This is the ordering that SymbolicUtils.jl uses to accelerate term matching.

YingboMa:
Some things related to parameters broke. I will push the fix shortly.

src/direct.jl
```diff
@@ -1,44 +1,44 @@
...
+gradient(O::Expression, vars::AbstractVector; simplify = true)
 ```
 
 A helper function for computing the gradient of an expression with respect to
 an array of variable expressions.
 """
-function gradient(O::Expression, vars::AbstractVector{<:Expression}; simplify = true)
-    [expand_derivatives(Differential(v)(O),simplify) for v in vars]
+function gradient(O, vars::AbstractVector; simplify = true)
+    Num[expand_derivatives(Differential(v)(value(O)),simplify) for v in vars]
```
ChrisRackauckas, line 10: Why is that type specification needed?

src/function_registration.jl
```diff
@@ -1,137 +0,0 @@
...
-end
-
-for fun ∈ [:<, :>, :(==), :&, :|, :div]
-    basefun = Expr(:., Base, QuoteNode(fun))
-    sig = :($basefun(x,y))
-    @eval @register $sig
-end
-
-# special cases
-Base.:^(x::Expression,y::T) where T <: Integer = Operation(Base.:^, Expression[x, y])
```
ChrisRackauckas, line 103: where did all of this go?

src/systems/abstractsystem.jl
```diff
@@ -260,15 +272,15 @@ end
 
 struct AbstractSysToExpr
     sys::AbstractSystem
-    states::Vector{Variable}
+    states::Vector
 end
 AbstractSysToExpr(sys) = AbstractSysToExpr(sys,states(sys))
-function (f::AbstractSysToExpr)(O::Operation)
```
ChrisRackauckas, line 266: but not all things are Term? How come there's still code for Operation?

src/direct.jl
YingboMa: It's not needed but it doesn't harm anything. It speeds up the compilation.

YingboMa: Correction: I think it's needed for cases like `[1]`. We need `[Num(1)]` to dispatch it on.

src/systems/jumps/jumpsystem.jl
```diff
@@ -32,12 +32,12 @@ struct JumpSystem{U <: ArrayPartition} <: AbstractSystem
...
     """The independent variable, usually time."""
-    iv::Variable
+    iv::Any
     """The dependent variables, representing the state of the system."""
-    states::Vector{Variable}
+    states::Vector
     """The parameters of the system."""
-    ps::Vector{Variable}
-    pins::Vector{Variable}
+    ps::Vector
```
ChrisRackauckas, line 39: why was every thing in each of the systems made less specific? Reason, or just compile time?

src/function_registration.jl
YingboMa:
```julia
julia> (a + a)^3
(a + a) ^ 3
```
It just works without it.

src/utils.jl
```diff
@@ -43,95 +31,85 @@ function flatten_expr!(x)
...
-vars(exprs) = foldl(vars!, exprs; init = Set{Variable}())
-function vars!(vars, O)
-    isa(O, Operation) || return vars
-    O.op isa Variable && push!(vars, O.op)
-    for arg ∈ O.args
-        if isa(arg, Operation)
-            isa(arg.op, Variable) && push!(vars, arg.op)
-            vars!(vars, arg)
-        end
-    end
```
ChrisRackauckas, line 86: Deleting all of the utils? I guess SymbolicUtils.jl lives up to its name

src/utils.jl
```diff
@@ -146,3 +124,22 @@ macro showarr(x)
 end
 
 @deprecate substitute_expr!(expr,s) substitute(expr,s)
```
ChrisRackauckas, line 126: let's delete this.

src/systems/abstractsystem.jl
YingboMa: Which line still uses Opteration?

src/systems/jumps/jumpsystem.jl
YingboMa: There's no benefit in constraining the eltype to be `Variable` at all. It just makes some code doesn't work.

test/build_targets.jl
```diff
@@ -4,7 +4,7 @@ using ModelingToolkit, Test
 @derivatives D'~t
 eqs = [D(x) ~ a*x - x*y,
        D(y) ~ -3y + x*y]
-@test ModelingToolkit.build_function(eqs,convert.(Variable,[x,y]),convert.(Variable,[a]),t,target = ModelingToolkit.StanTarget()) ==
```
ChrisRackauckas, line 7: Nice.

test/dep_graphs.jl
```diff
@@ -17,9 +19,8 @@ rs = ReactionSystem(rxs, t, [S,I,R], [k1,k2])
 #  testing for Jumps / all dgs
 #################################
 js = convert(JumpSystem, rs)
-S  = convert(Variable,S); I = convert(Variable,I); R = convert(Variable,R)
-k1 = convert(Variable,k1); k2 = convert(Variable,k2)
-
+S  = value(S); I = value(I); R = value(R)
+k1 = value(k1); k2 = value(k2)
```
ChrisRackauckas, line 23: what's this?

test/derivatives.jl
```diff
@@ -73,16 +73,16 @@ test_equal(expand_derivatives(D(z)), -1)
...
-@test all(iszero, ModelingToolkit.gradient(ModelingToolkit.Constant(42), [t, x, y, z]))
-@test all(iszero, ModelingToolkit.hessian(ModelingToolkit.Constant(42), [t, x, y, z]))
-@test isequal(ModelingToolkit.jacobian([t, x, ModelingToolkit.Constant(42)], [t, x]),
-              Expression[ModelingToolkit.Constant(1)  ModelingToolkit.Constant(0)
-                         Differential(t)(x)           ModelingToolkit.Constant(1)
-                         ModelingToolkit.Constant(0)  ModelingToolkit.Constant(0)])
+@test iszero(ModelingToolkit.derivative(42, x))
+@test all(iszero, ModelingToolkit.gradient(42, [t, x, y, z]))
+@test all(iszero, ModelingToolkit.hessian(42, [t, x, y, z]))
+@test isequal(ModelingToolkit.jacobian([t, x, 42], [t, x]),
```
ChrisRackauckas, line 82: nice

src/utils.jl
YingboMa: yeah, SymbolicUtils.jl is very good :p

src/utils.jl
YingboMa: yeah, we haven't delete all the dead code yet. We wanted to make the test pass first.

ChrisRackauckas:
Wow, the tests look great. So many weird `convert(Variable` etc. kinds of weird interface issues are gone. Really nice.

src/utils.jl
ChrisRackauckas: Alright, follow up PR though. Since this is going to be the breaking release we'll want to do all of that kind of stuff.

test/dep_graphs.jl
YingboMa: We use the wrapper type `Num` for dispatch only. It's invalid to show up in internal functions. So when we test internal functions, we need to call `value` to unwrap it.

src/systems/abstractsystem.jl
ChrisRackauckas: There's a bunch of dead code in Operators.jl IIRC

src/function_registration.jl
ChrisRackauckas: well then.

src/direct.jl
ChrisRackauckas: I see, since `Constant` is gone.

ChrisRackauckas (approved):

YingboMa:
I will update the docs in a separate PR before making a new major release.

isaacsas:
@shashi @YingboMa nice job! This one’s a beast! I look forward to learning the new setup.

shashi:
There's a bunch of more dead code, e.g. `var2op`... We should look at coverage and get them all.

shashi:
We still have `Variable` to support the `Variable(:foo, 1)` calls... We should do something about that...

AshtonSBradley:
There was some discussion of complex support for MTK. I have in mind the simple sense for quantum mechanics applications (some fairly broad potential userbase here I think), namely a convenient representation of the (x,y) plane, rather than analytic structure. Is this still on the cards?

ChrisRackauckas:
@YingboMa I think had a working example yesterday? I think we just need to add like two definitions to the library.

PR #519, opened 2020-07-14
add option to disable rate law scaling for reactions

isaacsas:
We'll need something like this for `ReactionNetworkImporters`. 

I also dropped use of `binomial` in `jumpratelaw`, as this was not simplified down when generating the `rate` functions for `ConstantRateJump`s.

### DIFF
--- a/src/systems/reaction/reactionsystem.jl
+++ b/src/systems/reaction/reactionsystem.jl
@@ -258,7 +258,6 @@ function jumpratelaw(rx; rxvars=get_variables(rx.rate), scalerate=true)
     if !only_use_rate
         coef = one(eltype(substoich))
         for (i,stoich) in enumerate(substoich)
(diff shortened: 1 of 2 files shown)


isaacsas:
@ChrisRackauckas should be good to go when you have a chance (no rush). Thanks!

ChrisRackauckas:
Only comment is really the name `scalerate`: stoichiometric corrected rate? Combinatoric rate? I feel like it could have a more informative name, but if there isn't one in less than an hour, I'm happy to merge

isaacsas:
Yeah, I don't know a good one that is shorter. Options: `normalizerate`, `scaleby_factorials`, `use_combinatoric_rate`. `scaleratebyfactorials`?

isaacsas:
`combinatoric_ratelaw`?

ChrisRackauckas:
I like combinatoric_ratelaw

isaacsas:
OK, all changed.

PR #521, opened 2020-07-15
Fix typo + add link to lambdify in docs

briochemc:
I'm not even sure that's the `lambdify` you were referencing to but it was the first google hit for me so I thought that was it 😅


ChrisRackauckas:
That's the right one

PR #527, opened 2020-07-17
Fix hessian_sparsity documentation?

ChrisRackauckas:
Hopefully this fixes the missing docstring in https://mtk.sciml.ai/dev/highlevel/#Sparsity-Detection-1

### DIFF
--- a/src/direct.jl
+++ b/src/direct.jl
@@ -133,6 +133,7 @@ let
               end
           end] |> Rewriters.Chain |> Rewriters.Postwalk |> Rewriters.Fixpoint
 
+    global hessian_sparsity
     # we do this in a let block so that Revise works on the list of rules
     function hessian_sparsity(f, u)
         idx(i) = TermCombination(Set([Dict(i=>1)]))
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
Does that still work with revise?

src/direct.jl
```diff
@@ -124,16 +134,6 @@ let
...
-    global hessian_sparsity
-
-    """
-    ```julia
-    hessian_sparsity(ops::AbstractVector{<:Expression}, vars::AbstractVector{<:Expression})
-    ```
-
-    Return the sparsity pattern of the Hessian of an array of expressions with respect to
-    an array of variable expressions.
-    """
```
shashi, line 136: Actually you need `global hessian_sparsity` here.

src/direct.jl
```diff
@@ -124,16 +134,6 @@ let
           end] |> Rewriters.Chain |> Rewriters.Postwalk |> Rewriters.Fixpoint
 
     # we do this in a let block so that Revise works on the list of rules
-    global hessian_sparsity
```
shashi, line 127: leave this line.

PR #530, opened 2020-07-17
Noise scaling in reaction sys sd es 2

TorkelE:
After I had to make a merge with the main files for https://github.com/SciML/ModelingToolkit.jl/pull/517 everything became kinda weird and it was not obvious how to normalise it. In the end, it seemed easier to simply to create another PR.

### DIFF
--- a/src/systems/reaction/reactionsystem.jl
+++ b/src/systems/reaction/reactionsystem.jl
@@ -209,7 +209,7 @@ function assemble_diffusion(rs, noise_scaling; combinatoric_ratelaws=true)
 
     for (j,rx) in enumerate(rs.eqs)
         rlsqrt = sqrt(oderatelaw(rx; combinatoric_ratelaw=combinatoric_ratelaws))
-        (noise_scaling!==nothing) && (rlsqrt *= var2op(Variable(noise_scaling[j])))
+        (noise_scaling!==nothing) && (rlsqrt *= noise_scaling[j])
         for (spec,stoich) in rx.netstoich
             i            = species_to_idx[spec]
             signedrlsqrt = (stoich > zero(stoich)) ? rlsqrt : -rlsqrt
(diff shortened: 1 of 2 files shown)


src/systems/reaction/reactionsystem.jl
```diff
@@ -383,14 +384,15 @@ Convert a [`ReactionSystem`](@ref) to an [`SDESystem`](@ref).
 Notes:
 - `combinatoric_ratelaws=true` uses factorial scaling factors in calculating the rate
 law, i.e. for `2S -> 0` at rate `k` the ratelaw would be `k*S^2/2!`. If
-`combinatoric_ratelaws=false` then the ratelaw is `k*S^2`, i.e. the scaling factor is 
+`combinatoric_ratelaws=false` then the ratelaw is `k*S^2`, i.e. the scaling factor is
 ignored.
 """
-function Base.convert(::Type{<:SDESystem},rs::ReactionSystem, combinatoric_ratelaws=true)
+function Base.convert(::Type{<:SDESystem},rs::ReactionSystem, combinatoric_ratelaws=true; noise_scaling=nothing::Union{Vector{Symbol},Symbol,Nothing})
```
ChrisRackauckas, line 390: They should be `Operation` or `Vector{Operation}`

TorkelE: Should it really be?

Currently, when one wants to scale noise, it is enough with
```julia
rn = @reaction_network ...
...
end p1 p2 ...
prob = SDEProblem(rn,u0,span,p,scale_noise=:η)
```
If one force it to be an operation then one has to do the extra step of
```julia
rn = @reaction_network ...
...
end p1 p2 ...
@varriables η
prob = SDEProblem(rn,u0,span,p,scale_noise=η)
```
Isn't it easier to let the `Operation` be created under the hood, like it is currently? Then one can avoid having to call on MTK functionality as well?

ChrisRackauckas: for one it breaks uniformity, which will lead to maintenance and documentation issues. Every other aspect of every other system type doesn't use symbol inputs for variables, so why is this argument different? And two, it is adding variables to the system, so I don't see the issue with the user having to recognize that it is. It gives a lot of helpful features too, like:

```julia
rn = @reaction_network ...
...
end p1 p2 ...
@varriables η[1:10]
prob = SDEProblem(rn,u0,span,p,scale_noise=η)
```

whereas trying to do that with Symbols directly... not nice.

TorkelE: I see your point, still think I prefer not having to create a variable. However, I see that in the greater context of MTK that approach makes less sense, and only really is suitable when one avoids MTK entirely. I'm happy to go with this, and have uploaded the corresponding change.

The only problem left is that this does not allow you to use previously existing parameters and
```julia
rn = @reaction_network ...
...
end p1 p2 p3
@varriables p3
prob = SDEProblem(rn,u0,span,[1.,1.,1.],scale_noise=p3)
```
will fails as the system thinks the new p3 is a different one from the previous one, unintentionally creating a 4 parameter system. I will look at adding a small search and replace to cover this rather special case or to discourage the practise entirely.
(To clarify, I don't think this is an argument for whenever to use Operation/Symbol, that's more important, this is just an issue of adapting to that decision)

ChrisRackauckas: >will fails as the system thinks the new p3 is a different one from the previous one, unintentionally creating a 4 parameter system

That should be checked for in this PR. Just add a `x !in parameters` check.

TorkelE: Great, will get to it later today.

PR #539, opened 2020-07-27
making tutorial on component modeling reproducible

ramboldio:
- made usage of α/a consistent
- fixing up system of equations: 
when running it locally I get an `┌ Warning: dt <= dtmin. Aborting. There is either an error in your model specification or the true solution is unstable.`. The solution would be only one data point. I am not quite sure how to fix that. Randomly changing stuff in line 76 makes results in more data points but not the in the plot from the tutorial.


ChrisRackauckas:
Thanks for pointing this out. I opted to fix it in a slightly different way: I didn't realize that this example was slightly changed from the README. The fixes were done in https://github.com/SciML/ModelingToolkit.jl/commit/fa3914e8ec4a589c910b87b0945b6e1a48e8be62 and hopefully that should be much easier to read.

PR #542, opened 2020-08-04
Add NullParaemeters default.

TorkelE:
Should now be able to create `ReactionSystem`s without parameters, and create problems fro them. This should no longer yield an error:
```julia
using Catalyst, DiffEqBase, OrdinaryDiffEq
rn = @reaction_network begin
           exp(-(t-5)^2), 0 --> S
           1.0, S --> 0
       end 
oprob = ODEProblem(rn, [0.0], (0.0,20.0))
```
Solves https://github.com/SciML/Catalyst.jl/issues/239

### DIFF
--- a/src/systems/reaction/reactionsystem.jl
+++ b/src/systems/reaction/reactionsystem.jl
@@ -457,25 +457,25 @@ end
 
 
 # ODEProblem from AbstractReactionNetwork
-function DiffEqBase.ODEProblem(rs::ReactionSystem, u0::Union{AbstractArray, Number}, tspan, p, args...; kwargs...)
-    u0 = typeof(u0) <: Array{<:Pair} ? u0 : Pair.(rs.states,u0)
-    p = typeof(p) <: Array{<:Pair} ? p : Pair.(rs.ps,p)
+function DiffEqBase.ODEProblem(rs::ReactionSystem, u0::Union{AbstractArray, Number}, tspan, p=DiffEqBase.NullParameters(), args...; kwargs...)
+    u0 = typeof(u0) <: Union{Array{<:Pair},DiffEqBase.NullParameters} ? u0 : Pair.(rs.states,u0)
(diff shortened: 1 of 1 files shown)


src/systems/reaction/reactionsystem.jl
```diff
@@ -457,25 +457,25 @@ end
 
 
 # ODEProblem from AbstractReactionNetwork
-function DiffEqBase.ODEProblem(rs::ReactionSystem, u0::Union{AbstractArray, Number}, tspan, p, args...; kwargs...)
-    u0 = typeof(u0) <: Array{<:Pair} ? u0 : Pair.(rs.states,u0)
-    p = typeof(p) <: Array{<:Pair} ? p : Pair.(rs.ps,p)
+function DiffEqBase.ODEProblem(rs::ReactionSystem, u0::Union{AbstractArray, Number}, tspan, p=DiffEqBase.NullParameters(), args...; kwargs...)
+    u0 = typeof(u0) <: Union{Array{<:Pair},DiffEqBase.NullParameters} ? u0 : Pair.(rs.states,u0)
```
isaacsas, line 461: Why NullParameters for `u0`?

isaacsas:
Maybe add a test for this too? Even just the example I gave as each problem type?

TorkelE:
That would be sensible. Should we have it in Catalyst though? It is a bit confusing that so many of the tests are over there right now, but now mostly all the tests including creating problems and beyond are there, so might make it more consistent.

ChrisRackauckas:
The whole ReactionSystem should move over to Catalyst IMO, we just need like another week of refactoring the abstract system API

isaacsas:
Either location is fine! We just need to remember to include it.

TorkelE:
I'll add tests for it over at Catalyst later today.

isaacsas:
Sounds good!

isaacsas:
@TorkelE OK to merge?

TorkelE:
yes

PR #544, opened 2020-08-05
revert to previous ismassaction version

isaacsas:
Somehow `ismassaction` had a line added that preempted the more optimized version with an earlier version...


ChrisRackauckas:
probably just a merge issue.

PR #546, opened 2020-08-07
automatic equation reduction

ChrisRackauckas:


ChrisRackauckas:
is this ready?

test/reduction.jl
```diff
@@ -0,0 +1,97 @@
...
+       D(lorenz2.y) ~ lorenz2.x*(lorenz2.ρ-lorenz2.z)-lorenz2.x,
+       D(lorenz2.z) ~ lorenz2.x*lorenz2.y - lorenz2.β*lorenz2.z])
+
+test_equal.(observed(aliased_flattened_system), [
+       lorenz1.F ~ lorenz2.x + -1 * (lorenz2.y + lorenz2.z),
+       lorenz1.u ~ lorenz1.x + lorenz1.y + -1 * lorenz1.z,
+       lorenz2.F ~ lorenz1.x + lorenz1.y + -1 * lorenz1.z,
+       a ~ lorenz2.y + -1 * lorenz1.x,
+       lorenz2.u ~ lorenz2.x + -1 * (lorenz2.y + lorenz2.z),
+])
```
shashi, line 97: @ChrisRackauckas check this test out. It figures out all 5 outputs.

shashi:
It is ready now.

I think we should just rename

`solve_for` to `solve` and export it.

It takes a vector of Equations, and a vector of variables to solve for, and returns the expressions. Only works on linear equations now, but we can use the same function to add non linear solve functionality.

ChrisRackauckas:
using `solve` would break the CommonSolve interface and lead to a ton of ambiguities, so let's not do that.

shashi:
Good point, reverted.

ChrisRackauckas:
Well we could dispatch it specifically on abstract arrays of expressions, so that's a way to use add it to solve.

PR #554, opened 2020-08-17
WIP: speed up JumpProblem creation

isaacsas:
So far this gives about a 100x speed-up on the graph SIR model [here](https://discourse.julialang.org/t/lightgraph-modeltoolkit-optimization/40696).

### DIFF
--- a/src/systems/abstractsystem.jl
+++ b/src/systems/abstractsystem.jl
@@ -188,7 +188,15 @@ end
 namespace_operation(O::Constant,name,ivname) = O
 
 independent_variable(sys::AbstractSystem) = sys.iv
-states(sys::AbstractSystem) = unique(isempty(sys.systems) ? setdiff(sys.states, convert.(Variable,sys.pins)) : [sys.states;reduce(vcat,namespace_variables.(sys.systems))])
+
+function states(sys::AbstractSystem) 
+    if isempty(sys.systems)
+        return isempty(sys.pins) ? sys.states : unique(setdiff(sys.states, convert.(Variable,sys.pins)))
+    else
+        return unique([sys.states;reduce(vcat,namespace_variables.(sys.systems))])
+    end
+end
+
 parameters(sys::AbstractSystem) = isempty(sys.systems) ? sys.ps : [sys.ps;reduce(vcat,namespace_parameters.(sys.systems))]
 pins(sys::AbstractSystem) = isempty(sys.systems) ? sys.pins : [sys.pins;reduce(vcat,namespace_pins.(sys.systems))]
 function observed(sys::AbstractSystem)
--- a/src/systems/dependency_graphs.jl
+++ b/src/systems/dependency_graphs.jl
@@ -1,6 +1,6 @@
 """
 ```julia
-equation_dependencies(sys::AbstractSystem; variables=states(sys))
+equation_dependencies(sys::AbstractSystem; variables=Set(states(sys)))
 ```
 
 Given an `AbstractSystem` calculate for each equation the variables it depends on. 
@@ -41,7 +41,7 @@ equation_dependencies(jumpsys)
 equation_dependencies(jumpsys, variables=parameters(jumpsys))    
 ```
 """
-function equation_dependencies(sys::AbstractSystem; variables=states(sys))
+function equation_dependencies(sys::AbstractSystem; variables=Set(states(sys)))
     eqs  = equations(sys)
     deps = Set{Operation}()
     depeqs_to_vars = Vector{Vector{Variable}}(undef,length(eqs))
@@ -143,7 +143,7 @@ end
 # could be made to directly generate graph and save memory
 """
 ```julia
-asgraph(sys::AbstractSystem; variables=states(sys), 
+asgraph(sys::AbstractSystem; variables=Set(state(sys)), 
                                       variablestoids=Dict(convert(Variable, v) => i for (i,v) in enumerate(variables)))
 ```
 
@@ -164,14 +164,14 @@ Continuing the example started in [`equation_dependencies`](@ref)
 digr = asgraph(odesys)
 ```
 """
-function asgraph(sys::AbstractSystem; variables=states(sys), 
+function asgraph(sys::AbstractSystem; variables=Set(states(sys)), 
                                       variablestoids=Dict(convert(Variable, v) => i for (i,v) in enumerate(variables)))
     asgraph(equation_dependencies(sys, variables=variables), variablestoids)
 end
 
 """
 ```julia
-variable_dependencies(sys::AbstractSystem; variables=states(sys), variablestoids=nothing)
+variable_dependencies(sys::AbstractSystem; variables=Set(state(sys)), variablestoids=nothing)
 ```
 
 For each variable determine the equations that modify it and return as a [`BipartiteGraph`](@ref).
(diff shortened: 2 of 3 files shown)


isaacsas:
So constructing `JumpProblem`s from `JumpSystem`s seems to be the big time sink currently on larger systems. Benchmarking with `TimerOutput`s as follows:
```julia
    to = TimerOutput()
    @timeit to "majs" majs = MassActionJump[assemble_maj(j, statetoid, subber, invttype) for j in eqs.x[1]]
    crjs = ConstantRateJump[assemble_crj(js, j, statetoid) for j in eqs.x[2]]
    vrjs = VariableRateJump[assemble_vrj(js, j, statetoid) for j in eqs.x[3]]
    ((prob isa DiscreteProblem) && !isempty(vrjs)) && error("Use continuous problems such as an ODEProblem or a SDEProblem with VariableRateJumps")
    jset = JumpSet(Tuple(vrjs), Tuple(crjs), nothing, isempty(majs) ? nothing : majs)

    if needs_vartojumps_map(aggregator) || needs_depgraph(aggregator)
        @timeit to "set" variables = Set(sts)
        @timeit to "hdict" variablestoidsh = Dict( st.name => i for (i,st) in enumerate(sts)) 
        @timeit to "dict" variablestoids = Dict( st => i for (i,st) in enumerate(sts)) 
        @timeit to "asgraph" jdeps = asgraph(js, variables=variables, variablestoids=variablestoids)
        @timeit to "vdeps" vdeps = variable_dependencies(js, variables=variables, variablestoids=variablestoids)
        vtoj = jdeps.badjlist
        jtov = vdeps.badjlist
        @timeit to "eqeq_deps" jtoj = needs_depgraph(aggregator) ? eqeq_dependencies(jdeps, vdeps).fadjlist : nothing
    else
        vtoj = nothing; jtov = nothing; jtoj = nothing
    end
```
within `JumpProblem` gives:
```
────────────────────────────────────────────────────────────────────
                             Time                   Allocations      
                     ──────────────────────   ───────────────────────
  Tot / % measured:       17.3s / 100%            2.84GiB / 98.4%    

 Section     ncalls     time   %tot     avg     alloc   %tot      avg
 ────────────────────────────────────────────────────────────────────
 majs             1    4.37s  25.3%   4.37s    861MiB  30.1%   861MiB
 asgraph          1    3.86s  22.4%   3.86s    722MiB  25.2%   722MiB
 eqeq_deps        1    3.36s  19.5%   3.36s    524MiB  18.3%   524MiB
 vdeps            1    3.34s  19.4%   3.34s    657MiB  23.0%   657MiB
 dict             1    2.12s  12.3%   2.12s   56.5MiB  1.97%  56.5MiB
 hdict            1    121ms  0.70%   121ms   17.9MiB  0.62%  17.9MiB
 set              1   72.5ms  0.42%  72.5ms   22.8MiB  0.80%  22.8MiB
 ────────────────────────────────────────────────────────────────────
```

This is a bit frustrating now, since it appears the work is spread around somewhat uniformly... A big problem is the dependency graph generation. I think that might need to be reworked in general, or at least specialized for jumps to only build exactly what is needed for SSAs (and to only build graphs that have integer components). It also appears that it might be advantageous to make the look-up sets / Dicts use the underlying `Symbol` for a variable instead of the variable (which I wouldn't have anticipated given that `Variable` is a thin wrapper over the `Symbol`...) 

The mass action jump construction might be able to be sped up by use the vector form with integer stoichiometries within the JumpSystem from construction. This would allow aliasing of those arrays when creating the final MassActionJump for the SSAs.

ChrisRackauckas:
@shashi @YingboMa do you know what could be the cause of this?

I think we are just missing the specialization of one thing somewhere, which is why it's so uniform.

isaacsas:
A lot of the dependency graph stuff comes down to calling `get_variables!` and storing results within a `Set`. I may play around with switching the storage back to a vector since there should only be a few elements in general. (On the other hand, using a Set for the collection of many states to filter against is what gave the 100x speed up above.)

isaacsas:
From Slack on the Dict performance:
It's saying this is where all the time is spent when using Variables: 

https://github.com/JuliaLang/julia/blob/96786e22ccabfdafd073122abb1fb69cea921e17/base/dict.jl#L161

With pair creation within the iteration dominating.

When using Symbols it says the dominant cost is from 

https://github.com/JuliaLang/julia/blob/96786e22ccabfdafd073122abb1fb69cea921e17/base/dict.jl#L103

Here I'm just doing (for a JumpSystem with 300,000 variables):
```julia
using Profile, Profview
p = Profile.init()
Profile.init(p[1],.1*p[2])
symhash(sts) = Dict( st.name => i for (i,st) in enumerate(sts))
varhash(sts) = Dict( st => i for (i,st) in enumerate(sts))
sts = states(jumpsys)
@profview varhash(sts)
Profile.clear()
@profview symhash(sts)
```

isaacsas:
Here's a more minimal example that shows a smaller performance difference (only about a factor of 5). It doesn't show the same profiling results as the example above (which was from within a large graph SIR model). Given that my REPL is becoming somewhat unresponsive in both these cases I wonder if there are memory swapping issues.
```julia
using ModelingToolkit, DiffEqJump, Profile, ProfileView
@parameters t
@variables M[1:300000](t)
varhash(sts) = Dict( st => i for (i,st) in enumerate(sts))
sts = [convert(Variable,m) for m in M]
p = Profile.init()
Profile.init(p[1],.1*p[2])
@profview varhash(sts);
symhash(sts) = Dict( st.name => i for (i,st) in enumerate(sts))
Profile.clear()
@profview symhash(sts)
```

shashi:
I had made a change from using `substitute` directly to `substituter` which had sped up the same code from discourse a whole lot. Hope that's still around and working!

Am I right in understanding that the main problem is in constructing the dict?

isaacsas:
The timing reports above indicates it is one of several problems that are contributing at about the same order of magnitude, so it would be good to understand why constructing a Dict of variables is so much slower than a Dict of Symbols (is it just that the former isn't concretely typed when the Variables come from `states(sys)`?). 

I don't think this is a substitution issue -- your fix there definitely helped! Profiling this example is problematic as I find my REPL frequently becomes near unresponsive after one or two rounds, and I'm not sure I trust the profiling results that I get back... With that said, I think part of the issue is related to the performance of `Variable`s within Dicts as much of the graph code relies on `Set`s of `Variable`s. I'm planning to play around with switching to using Symbols to see if that helps, and to convert the `Set`s that will normally only contain a few elements to vectors.

shashi:
I see a smaller difference:

```
julia> @btime varhash($(Symbol.(convert.(Variable, M))));
  27.129 ms (44 allocations: 18.75 MiB)

julia> @btime varhash(map(x->x.name, $(convert.(Variable, M)))); # includes taking the name out.
  27.612 ms (46 allocations: 21.04 MiB)

julia> @btime varhash($(convert.(Variable, M)));
  51.723 ms (984256 allocations: 33.77 MiB)

```
I would bet that `hash` on Variable is type unstable.

isaacsas:
As I mentioned on Slack, it did seem on profiling that the slower `Variable`-based version was spending the most time in `Pair` creation, so you think the problem is the generator and not the Dict?

shashi:
> I would bet that hash on Variable is type unstable.

it's not. But even this is weird:

```
julia> Base.hash(v::Variable{T}, u::UInt) where {T} = hash(v.name, u)

julia> @btime varhash($(convert.(Variable, M)));
  79.393 ms (2479220 allocations: 56.58 MiB)
```
it should be doing the same exact work as making a dict of Symbols in terms of hashing. but it looks like the slowdown happens when storing Variables in the vectors internal to the `Dict`... I am not sure why!

shashi:
> spending the most time in Pair creation

Oh that's another possibility! let me see!

isaacsas:
Yeah, I tried adding that same hash definition and it made no performance different.

shashi:
```
julia> foo(sts) = [st => i for (i,st) in enumerate(sts)]

julia> @btime foo(M);
  944.872 μs (2 allocations: 6.87 MiB)
```
It seems ok.

PR #556, opened 2020-08-17
Change parameters in README

devmotion:
Maybe you did this on purpose (then just close this PR), but it confused me quite a bit that the second order example uses the standard values for the parameters of the Lorenz equations - but with the values of `σ` and `ρ` being swapped.

With the default set of parameters I obtain the following plot:
lorenz_mtk


ChrisRackauckas:
Oh oops... but also... the other plot looks better. Hmm...

ChrisRackauckas:
I gotta say... I like the look 😆

PR #562, opened 2020-08-26
Add ControlSystem for Nonlinear Optimal Control

ChrisRackauckas:
This pull request adds a `ControlSystem` construct for specifying nonlinear optimal control problems. The format is as follows:

```julia
using ModelingToolkit

@variables t x(t) v(t) u(t)
@derivatives D'~t

loss = (4-x)^2 + 2v^2 + u^2
eqs = [
    D(x) ~ v
    D(v) ~ u^3
]

sys = ControlSystem(loss,eqs,t,[x,v],[u],[])
```

Over time it will get full standardization with the rest of ModelingToolkit's abstract system interface, which means that things like pins and observables should get supported. Through component-based modeling should already work.

This is an abstract specification, so it is symbolic in nature and does not do the computation itself. Instead, all computation on a ControlSystem occurs through the transformation into another system type. In order to demonstrate this, I created the first transformation function, `runge_kutta_discretize`, which takes in a ControlSystem and outputs an OptimizationSystem that is defined by the collocation of some Runge-Kutta method (defaulting to a 5th order RadauIIA). That means that all of the machinery for OptimizationSystem can then be used to transform this into a OptimizationProblem to be numerically handled, and sparse symbolic Hessians should already work, etc.

As a demonstration of this:

```julia
dt = 0.1
tspan = (0.0,1.0)
runge_kutta_discretize(sys,dt,tspan)
```

creates an OptimizationSystem, which one can then transform to a numerical problem to solve. Note that for this transformation to work, I did the long-overdue thing of expanding OptimizationSystem to have constraints. We will need to update OptimizationProblem to have an interface for this, and then setup GalacticOptim to appropriately handle constraints, and then this will be a "direct to numerics" thing, but for now it's purely symbolic.

### DIFF
--- a/docs/src/systems/ControlSystem.md
+++ b/docs/src/systems/ControlSystem.md
@@ -1,0 +1,20 @@
+# ControlSystem
+
+## System Constructors
+
+```@docs
+ControlSystem
+```
+
+## Composition and Accessor Functions
+
+- `sys.eqs` or `equations(sys)`: The equations that define the system.
+- `sys.states` or `states(sys)`: The set of states in the system.
+- `sys.parameters` or `parameters(sys)`: The parameters of the system.
+- `sys.controls` or `controls(sys)`: The control variables of the system
+
+## Transformations
+
+```@docs
+runge_kutta_discretize
+```
--- a/src/ModelingToolkit.jl
+++ b/src/ModelingToolkit.jl
@@ -108,6 +108,8 @@ include("systems/nonlinear/nonlinearsystem.jl")
 
 include("systems/optimization/optimizationsystem.jl")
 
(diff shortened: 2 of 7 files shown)


freemin7:
What other approaches will there be to work on except `runge_kutta_discretize`?

ChrisRackauckas:
`universal_diffeq_control` which would setup a DiffEqFlux training problem, `variational_solve` which transforms to a variational PDE, etc.

freemin7:
Are pseudo spectral approaches considered or do they need much MTK work?
How about collocation techniques based on other classes of ODE solvers?

ChrisRackauckas:
Other collocations can be done as well, yes

Issue #532, opened 2020-07-19
The alternative to `ifelse`


ChrisRackauckas:
`ifelse` is a bit problematic for symbolic libraries because it's in Core and not in Base and dispatch cannot be added to  it. So, while we want to represent branching, we need to do so without using `ifelse`. The opportunities are something like:

- `ModelingToolkit.ifelse`
- `_ifelse` (exported)
- A different name

We need to just pick one and start using it, and whatever is chosen is what the eventual tracer should transform things to.

ChrisRackauckas:
https://github.com/JuliaLang/julia/issues/32844 hopefully this changes in Base.

Issue #531, opened 2020-07-18
Support for piecewise functions


thowell:
Is it possible to use this package with piecewise functions (e.g., [Huber loss](https://en.wikipedia.org/wiki/Huber_loss))?

```JULIA
@variables a
function huber(a)
    δ = 1.0
    if abs(a) <= δ
        return 0.5*a^2
    else
        return δ*(abs(a) - 0.5*δ)
    end
end
da = huber(a)
```

This errors on the last line:
> TypeError: non-boolean (Operation) used in boolean context

ChrisRackauckas:
For that it's probably best to `@register` it. You cannot directly trace a function with branches right now, though we're working on that.

ChrisRackauckas:
`@register` and `IfElse.ifelse` are the solutions.

PR #568, opened 2020-09-02
Handle conditionals and abs

ChrisRackauckas:
It's time we finally handle https://github.com/SciML/ModelingToolkit.jl/issues/531, https://github.com/SciML/ModelingToolkit.jl/issues/532, and https://discourse.julialang.org/t/differentiation-method-for-element-wise-abs-function-applied-on-operation-types-in-modelingtoolkit-jl/45867/4 . This PR creates a `ModelingToolkit.ifelse` with derivative fixes for `abs` to make it so standard conditional code can work. While tracing cannot correctly make these operations, this would at least allow directly written symbolic code to handle conditions, so we're at least as good (or bad) as TensorFlow.


ChrisRackauckas:
@celrod should we make a IfElse.jl package so that way we use the same one as LoopVectorization.jl? @c42f should we inject this as Base.ifelse?

ChrisRackauckas:
```julia
using ModelingToolkit
function func!(du,u)
    du = abs.(u)
end
@variables du[1:2] u[1:2]
func!(du,u)
sjac= ModelingToolkit.sparsejacobian(vec(du),vec(u));

function func!(du,u)
    for j = 1:2
        du[j] = abs(u[j])
    end
 end
@variables du[1:2] u[1:2]
func!(du,u)
sjac= ModelingToolkit.sparsejacobian(vec(du),vec(u));
```

works with this PR.

PR #575, opened 2020-09-10
start PDESystem documentation

ChrisRackauckas:


anandijain:
`DG` acronym in Vision section stands for Discrete Geometry?

PR #599, opened 2020-10-01
Add GalacticOptim-compatible OptimizationProblem generation

ChrisRackauckas:
Requires hoisting OptimizationProblem/OptimizationFunction back to DiffEqBase (which was intended anyways), and then defines a new AD version here that defines the OptimizationProblem from the symbolic code.

Requires https://github.com/SciML/DiffEqBase.jl/pull/590 and https://github.com/SciML/GalacticOptim.jl/pull/52

### DIFF
--- a/src/systems/optimization/optimizationsystem.jl
+++ b/src/systems/optimization/optimizationsystem.jl
@@ -90,10 +90,6 @@ struct AutoModelingToolkit <: DiffEqBase.AbstractADType end
 DiffEqBase.OptimizationProblem(sys::OptimizationSystem,args...;kwargs...) =
     DiffEqBase.OptimizationProblem{true}(sys::OptimizationSystem,args...;kwargs...)
 
-OptimizationProblemExpr(sys::OptimizationSystem,args...;kwargs...) =
-    OptimizationProblemExpr{true}(sys::OptimizationSystem,args...;kwargs...)
-
-
 """
 ```julia
 function DiffEqBase.OptimizationProblem{iip}(sys::OptimizationSystem,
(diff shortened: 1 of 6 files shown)


src/systems/optimization/optimizationsystem.jl
```diff
@@ -98,30 +105,53 @@ function DiffEqBase.OptimizationProblem{iip}(sys::OptimizationSystem,
 Generates an OptimizationProblem from an OptimizationSystem and allows for automatically
 symbolically calculating numerical enhancements.
 """
-function DiffEqBase.OptimizationProblem{iip}(sys::OptimizationSystem,
+function DiffEqBase.OptimizationProblem{iip}(sys::OptimizationSystem, u0,
```
Vaibhavdixit02, line 108: I am wondering if it might be better to follow the same pattern as in GalacticOptim and keep `OptimizationFunction` and problem separate?

src/systems/optimization/optimizationsystem.jl
```diff
@@ -98,30 +105,53 @@ function DiffEqBase.OptimizationProblem{iip}(sys::OptimizationSystem,
 Generates an OptimizationProblem from an OptimizationSystem and allows for automatically
 symbolically calculating numerical enhancements.
 """
-function DiffEqBase.OptimizationProblem{iip}(sys::OptimizationSystem,
+function DiffEqBase.OptimizationProblem{iip}(sys::OptimizationSystem, u0,
                                           parammap=DiffEqBase.NullParameters();
-                                          u0=nothing, lb=nothing, ub=nothing,
-                                          hes = false, sparse = false,
+                                          lb=nothing, ub=nothing,
+                                          grad = false,
```
Vaibhavdixit02, line 111:
```suggestion
                                          grad = nothing,

```

src/systems/optimization/optimizationsystem.jl
ChrisRackauckas: they are separate. Building the problem type requires building the function though. Building problems is a MTK interface thing.

src/systems/optimization/optimizationsystem.jl
ChrisRackauckas: that would not conform to the standard interface.

Vaibhavdixit02:
LGTM once tests pass

PR #607, opened 2020-10-13
A little bit more variable conversion

ChrisRackauckas:
Someone needs to work out the general case, but this is starting to show how to implement it.

```julia
using ModelingToolkit
@variables t x(t)
@derivatives D'~t

ex = D(D(x))
convert(Variable,ex) # xˍtt
```


shashi:
I think we have `tosymbol` now.

PR #613, opened 2020-10-18
exprs_occur_in

shashi:


ChrisRackauckas:
Are there places for this to be used internally?

shashi:
Yeah @isaacsas code has a lot of uses for this.

ChrisRackauckas:
Yeah but we should really change the internal code that needs it.

isaacsas:
What’s the recommended, preferably allocation free, way to check if a variable is within a list of other variables? Similarly to check if a given variable is within an operation, or which of a collection of variables is within an operation? These are among the places that cause performance issues with reaction and jump systems currently as large systems can require many of such checks. Comparing against sets has at times been slow due to slow equality comparisions and lack of concrete typing. I had gotten reaction system performance pretty good, but I’m not sure where we are at now. Jump system performance was being worked on in that PR that is open, which I’ll have to come back to once the conversion to the new types is done.

PR #617, opened 2020-10-23
make states_to_sym fast in checking membership

shashi:


shashi:
@ChrisRackauckas we should just use this closure instead of `ODEExpr` and other `*Expr` types.

ChrisRackauckas:
>@ChrisRackauckas we should just use this closure instead of ODEExpr and other *Expr types.

That sounds good to me. I've been wondering how to get rid of that.

PR #618, opened 2020-10-23
More serious Symbol and makesym

YingboMa:
```julia
x(t) becomes x⦗t⦘

@derivative D~t
D(x)(t) becomes x_t⦗t⦘
(with the weird _)
```

### DIFF
--- a/src/systems/diffeqs/abstractodesystem.jl
+++ b/src/systems/diffeqs/abstractodesystem.jl
@@ -64,13 +64,30 @@ function generate_jacobian(sys::AbstractODESystem, dvs = states(sys), ps = param
                           conv = ODEToExpr(sys), kwargs...)
 end
 
-function makesym(t::Term{T}) where {T}
-    t.op isa Sym && return makesym(t.op)
-    t.op isa Differential && return Sym{T}(Symbol(nameof(makesym(t.args[1])), :ˍ, nameof(makesym(t.op.x))))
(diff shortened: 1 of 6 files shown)


ChrisRackauckas:
Time varying parameters still work okay?

BTW, since we'll be doing a breaking release, I'd like to really deprecate that old derivative syntax for a new one.

ChrisRackauckas (approved):

src/systems/diffeqs/abstractodesystem.jl
```diff
@@ -64,18 +68,49 @@ function generate_jacobian(sys::AbstractODESystem, dvs = states(sys), ps = param
...
 
-function makesym(t::Term{T}) where {T}
-    t.op isa Sym && return makesym(t.op)
-    t.op isa Differential && return Sym{T}(Symbol(nameof(makesym(t.args[1])), :ˍ, nameof(makesym(t.op.x))))
+Base.Symbol(x::Union{Num,Symbolic}) = tosymbol(x)
+tosymbol(x; kwargs...) = x
+tosymbol(x::Sym; kwargs...) = nameof(x)
+tosymbol(t::Num; kwargs...) = tosymbol(value(t); kwargs...)
+
+"""
```
ChrisRackauckas, line 76: seems like it's the wrong file for this function.

shashi: Yes it does.

PR #632, opened 2020-11-01
WIP: speed up JumpProblem conversion 2

shashi:
https://gist.github.com/isaacsas/593f9ce7f40adf05b6969ace8fe8a831

with

```julia
const nedgess = 30
const nvertss = 1000
```
gives

```julia
  0.000425 seconds (14.00 k allocations: 641.219 KiB)
Completed: ReactionSystem
  0.033972 seconds (519.34 k allocations: 16.500 MiB)
Completed: convert(JumpSystem)
  0.025654 seconds (585.68 k allocations: 15.495 MiB)
Completed: DiscreteProblem
  2.788977 seconds (2.35 M allocations: 2.143 GiB, 5.16% gc time)
Completed: JumpProblem
  0.000594 seconds (30 allocations: 353.297 KiB)
```

This is probably really bad. I'm thinking of better hash-table operations for the purposes here, hopefully we can turn something from `O(n^2)` to `O(n)`

### DIFF
--- a/src/systems/jumps/jumpsystem.jl
+++ b/src/systems/jumps/jumpsystem.jl
@@ -284,7 +284,7 @@ function get_variables!(dep, jump::Union{ConstantRateJump,VariableRateJump}, var
     dep
 end
 
-function get_variables!(dep, jump::MassActionJump, variables::Set)
+function get_variables!(dep, jump::MassActionJump, variables::AbstractSet)
     sr = value(jump.scaled_rates)
     (sr isa Symbolic) && get_variables!(dep, sr, variables)
     for varasop in jump.reactant_stoch
--- a/src/systems/jumps/jumpsystem.jl
+++ b/src/systems/jumps/jumpsystem.jl
@@ -288,7 +288,7 @@ function get_variables!(dep, jump::MassActionJump, variables::AbstractSet)
     sr = value(jump.scaled_rates)
     (sr isa Symbolic) && get_variables!(dep, sr, variables)
     for varasop in jump.reactant_stoch
-        isequal(varasop[1]) in variables && push!(dep, varasop[1])
+        varasop[1] in variables && push!(dep, varasop[1])
     end
     dep
 end
--- a/src/systems/jumps/jumpsystem.jl
+++ b/src/systems/jumps/jumpsystem.jl
@@ -147,8 +147,8 @@ function numericnstoich(mtrs::Vector{Pair{V,W}}, statetoid) where {V,W}
 end
 
 # assemble a numeric MassActionJump from a MT MassActionJump representing one rx.
-function assemble_maj(maj::MassActionJump, statetoid, subber, invttype)
-    rval = subber(maj.scaled_rates)
+function assemble_maj(maj::MassActionJump, statetoid, parammap, invttype)
+    rval = substitute(maj.scaled_rates, parammap)
     rs   = numericrstoich(maj.reactant_stoch, statetoid)
     ns   = numericnstoich(maj.net_stoch, statetoid)
     maj  = MassActionJump(convert(invttype, value(rval)), rs, ns, scale_rates = false)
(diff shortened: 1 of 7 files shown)


shashi:
This is pretty interesting, most of the time is spent in substitute (the call to `subber` in `assemble_maj`), and not the dictionary construction. I guess this should be even faster than what we had before though.

ChrisRackauckas:
What was it before?

isaacsas:
It was pretty bad before, though in the other open PR I had gotten it down a fair amount for everything but the final JumpProblem construction, which was performing badly due to the dependency graph construction among other things.

ChrisRackauckas:
3 seconds still doesn't seem great.

shashi:
Got another 10x

it was doing wasteful dict construction.

```
  0.000401 seconds (14.00 k allocations: 641.219 KiB)
Completed: ReactionSystem
  0.030159 seconds (519.34 k allocations: 16.500 MiB)
Completed: convert(JumpSystem)
  0.024397 seconds (585.68 k allocations: 15.495 MiB)
Completed: DiscreteProblem
  0.311764 seconds (2.18 M allocations: 57.016 MiB)
Completed: JumpProblem
  0.000451 seconds (30 allocations: 353.297 KiB)
Completed: solve
```

shashi:
I think we should only leave `substitue(expr, dict::Dict)` method since everything essentially calls that. This way it's much easier to reason about substitute's performance.

shashi:
Okay another 3x

```
  0.093674 seconds (2.18 M allocations: 57.018 MiB)
```

shashi:
Now with
```julia
const nedgess = 3000
const nvertss = 100000
```
```julia
  0.058344 seconds (1.50 M allocations: 64.072 MiB)
Completed: ReactionSystem
  5.699431 seconds (44.17 M allocations: 1.322 GiB, 44.66% gc time)
Completed: convert(JumpSystem)
  4.851998 seconds (52.66 M allocations: 1.283 GiB, 18.10% gc time)
Completed: DiscreteProblem
 17.310221 seconds (211.11 M allocations: 5.235 GiB, 12.65% gc time)
Completed: JumpProblem
  0.055617 seconds (30 allocations: 34.334 MiB)
Completed: solve
retcode: Default
Interpolation: Piecewise constant interpolation
t: 6-element Array{Float64,1}:
  0.0
  0.1
  0.2
  0.3
  0.4
 20.0
u: 6-element Array{Array{Int64,1},1}:
 [0, 1, 1, 1, 1, 1, 1, 1, 1, 1  …  0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
 [0, 1, 1, 1, 1, 1, 1, 1, 1, 1  …  0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
 [0, 1, 1, 1, 1, 1, 1, 1, 1, 1  …  0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
 [0, 1, 1, 1, 1, 1, 1, 1, 1, 1  …  0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
 [0, 1, 1, 1, 1, 1, 1, 1, 1, 1  …  0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
 [0, 1, 1, 1, 1, 1, 1, 1, 1, 1  …  0, 0, 0, 0, 0, 0, 0, 0, 0, 0]

```

shashi:
```
function states(sys::AbstractSystem)
    unique(isempty(sys.systems) ?
           setdiff(sys.states, value.(sys.pins)) :
           [sys.states;reduce(vcat,namespace_variables.(sys.systems))])
end
```

This is pretty expensive, especially `setdiff`. We should cache this.

Similarly `get_variables!` does `unique` which is expensive too.

isaacsas:
Could also just return `states` if `pins` and `systems` are both empty.

shashi:
> Could also just return states if pins and systems are both empty. I think that’s what I had put in in the other PR.

That's kind of a kludge but ok!

Do you know what's causing the test to fail now?

src/systems/jumps/jumpsystem.jl
```diff
@@ -280,30 +279,30 @@ end
...
     dep
 end
 
-function get_variables!(dep, jump::MassActionJump, variables)
+function get_variables!(dep, jump::MassActionJump, variables::AbstractSet)
     sr = value(jump.scaled_rates)
     (sr isa Symbolic) && get_variables!(dep, sr, variables)
     for varasop in jump.reactant_stoch
-        any(isequal(varasop[1]), variables) && push!(dep, varasop[1])
+        isequal(varasop[1]) in variables && push!(dep, varasop[1])
```
isaacsas, line 291: Do you really want the `isequal here?

shashi: Oh damn my bad!!!

isaacsas:
@shashi Before we merge, can copy the latest timings over here so we have a reference for future comparisons?

ChrisRackauckas:
Was there something done in place of this?

isaacsas:
No, this never got finished and merged unfortunately. If no one else takes a pass I’ll come back to it once I have some time (probably when classes end in a month). There was still more to do.

Issue #636, opened 2020-11-02
Variable names incorrect in plot legends on 4.0+


isaacsas:
On Catalyst master
```julia
rs = @reaction_network begin
  c1, X --> 2X
  c2, X --> 0
  c3, 0 --> X
end c1 c2 c3
p = (1.0,2.0,50.) # [c1,c2,c3]
tspan = (0.,4.)
u0 = [5.]         # [X]

# solve ODEs
oprob = ODEProblem(rs, u0, tspan, p)
osol  = solve(oprob, Tsit5())
plot(osol)

# solve SDEs
sprob = SDEProblem(rs, u0, tspan, p)
ssol  = solve(sprob, EM(), dt=.01)
plot(ssol)

# solve JumpProblem
u0 = [5]
dprob = DiscreteProblem(rs, u0, tspan, p)
jprob = JumpProblem(rs, dprob, Direct())
jsol = solve(jprob, SSAStepper())
plot(jsol)
```
gives
![osol](https://user-images.githubusercontent.com/9385167/97916225-b18e8b00-1d20-11eb-9782-d5c65bd48fcd.png)
![ssol](https://user-images.githubusercontent.com/9385167/97916230-b3584e80-1d20-11eb-823d-0ec5a5041723.png)
![jsol](https://user-images.githubusercontent.com/9385167/97916235-b5221200-1d20-11eb-877f-0db38a844296.png)

ChrisRackauckas:
@shashi @YingboMa should be easy to fix with a find replace heuristic in the DiffEq plot recipe

isaacsas:
@YingboMa Legends are now incorrect again.

ODE
![ode](https://user-images.githubusercontent.com/9385167/100883381-469ab600-347e-11eb-9ed2-257c223578aa.png)

SDE
![sde](https://user-images.githubusercontent.com/9385167/100883400-4bf80080-347e-11eb-8466-8ff4f49ba0bd.png)

Jump
![jump](https://user-images.githubusercontent.com/9385167/100883416-4ef2f100-347e-11eb-807e-072b1de829d2.png)

isaacsas:
Maybe we need a test for the generated string / Symbol?

YingboMa:
This is not from MTK. DiffEqBase add the extra `(t)`. I will make a PR to fix it.

isaacsas:
Great, thanks!

PR #646, opened 2020-11-09
Fix names in plot legends

isaacsas:
Fixes https://github.com/SciML/ModelingToolkit.jl/issues/636

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -159,7 +159,6 @@ tovar(s::Sym) = s
 Base.Symbol(x::Union{Num,Symbolic}) = tosymbol(x)
 tosymbol(x; kwargs...) = x
 tosymbol(x::Sym; kwargs...) = nameof(x)
-tosymbol(x::Term; kwargs...) = nameof(operation(x))
 tosymbol(t::Num; kwargs...) = tosymbol(value(t); kwargs...)
 
 """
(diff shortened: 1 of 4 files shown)


ChrisRackauckas (approved):

isaacsas:
@shashi so `nameof.(operation.(states(sys)))` works if the states are terms like `A(t)` but doesn't work if they are syms like `A`? Is there an existing way to handle both?

src/utils.jl
```diff
@@ -159,6 +159,7 @@ tovar(s::Sym) = s
 Base.Symbol(x::Union{Num,Symbolic}) = tosymbol(x)
 tosymbol(x; kwargs...) = x
 tosymbol(x::Sym; kwargs...) = nameof(x)
+tosymbol(x::Term; kwargs...) = nameof(operation(x))
```
isaacsas, line 162: @shashi @YingboMa is this reasonable to do, or will it potentially break things elsewhere that rely on the pass through?

shashi:
> Is there an existing way to handle both?

`opname`

isaacsas:
I don't see a merged `opname` function in either ModelingToolkit or SymbolicUtils. Is that something you have in a PR? If you want to give me it I can add it here, or if you merge it I'll rebase and switch to it.

isaacsas:
OK, the old `tosymbol` seems to work for terms if escaping is disabled so I've switched to that. Assuming tests pass this should be good to merge.

YingboMa:
No, this not the right fix. It will totally break when plotting states with derivatives.

PR #648, opened 2020-11-09
fix symbol replacement in build_function

shashi:
This is my fix for https://discourse.julialang.org/t/coulomb-and-static-friction-with-modelingtoolkit-jl-and-differentialequations-jl/49802/2


ChrisRackauckas:
Looks like a real test failure

PR #649, opened 2020-11-12
A nicer error message

contradict:
When extra variables are present or when variables are missing.


ChrisRackauckas:
Thanks! Great addition.

PR #651, opened 2020-11-16
Fix sym_lu

YingboMa:


src/solve.jl
```diff
@@ -28,7 +28,7 @@ function sym_lu(A)
 
         for j = k+1:m
             L[j,k] = U[j, k] / U[k, k]
-            U[j,k:m] .= U[j,k:m] .- L[j,k] .* U[k,k:m]
+            U[j,k:m] .= U[j,k:m] .- Ref(L[j,k]) .* U[k,k:m]
```
ChrisRackauckas, line 31: why Ref?

YingboMa: So it behaves like a scalar in broadcasting. There's no guarantee that a symbolic object is a scalar.

ChrisRackauckas: tuple it?

ChrisRackauckas, line 31:
```suggestion
            U[j,k:m] .= U[j,k:m] .- (L[j,k],) .* U[k,k:m]
```

PR #658, opened 2020-11-19
Macro-free docs extra parens removed

anandijain:


docs/src/tutorials/symbolic_functions.md
```diff
@@ -382,9 +382,9 @@ the user's code. For these cases, ModelingToolkit.jl allows for fully
 macro-free usage. For example:
 
 ```julia
-x = Sym{Float64}(:x)()
-y = Sym{Float64}(:y)()
-x+y^2.0
+x = Sym{Float64}(:x)
+y = Sym{Float64}(:y)
+x+y^2.0 # isa Num
```
shashi, line 387: It's not a Num actually

shashi: it would be a Num only if one of x or y was a Num

anandijain: I see, but given the prior two lines it makes more sense than nothing(?). I just tried `typeof(x+y^2.0)`.

I'm happy to remove it though

shashi: Yeah given the previous two lines we should probably have

```
x = Num(Sym{Float64}(:x))
```
etc.

And then `# isa Num` would make sense...

shashi: Let's do this, it's ugly but faithful

anandijain:
Thanks, maybe this is not best practice, but re: your slack comment, I also changed `Expr(::Num)` because it errors, and `toexpr` doesn't

PR #660, opened 2020-11-22
Switch to GitHub action

YingboMa:


ChrisRackauckas:
Turn off nightlies?

ChrisRackauckas:
This is probably the one borderline case in DiffEq. Nightlies usually shouldn't matter, but here we do hook into some compiler things in nefarious ways.

PR #663, opened 2020-11-27
Wrap registered fn output in Num if any input is Num

shashi:


ChrisRackauckas:
is there a test for why? Is this linked to an issue?

PR #681, opened 2020-12-03
Allow evaluation module to be specified when building an ODEFunction

### LINKED ISSUE #678: [Regression] RuntimeGeneratedFunctions tagged under ModelingToolkit instead of precompiled user module
I am writing a package that builds some ODEFunctions at precompile time dependent on some ModelingToolkit models.
Supporting this is what ultimately drove my previous PR #477.

It seems that since ModelingToolkit's v4.0 (I upgraded from v3.20.1), the above functionality is once again broken.

Using the first test from https://github.com/SciML/ModelingToolkit.jl/blob/master/test/function_registration.jl as an example, the regression I'm seeing is as follows:

```julia
julia> # NOTE: this is an illustrative example, the keys shown may not match exactly.
julia> # I have a package MyModule (as per the test case above).
julia> using MyModule  
[precompiling MyModule...]

julia> # As per the test, I expect an answer of [30.0]
julia> MyModule.fun([0.5], [5.0], 0.)  
ERROR: KeyError: key (0xd6493547, 0xaf3999e1, 0x66674bbf, 0xb716e034, 0xa6bb5c0f) not found
Stacktrace:
 [1] getindex(::Dict{Any,Any}, ::NTuple{5,UInt32}) at .\dict.jl:467
 [2] (::RuntimeGeneratedFunctions.var"#5#6"{DataType,NTuple{5,UInt32}})() at C:\Users\DPad\.julia\packages\RuntimeGeneratedFunctions\uvbuv\src\RuntimeGeneratedFunctions.jl:133
 [3] lock(::RuntimeGeneratedFunctions.var"#5#6"{DataType,NTuple{5,UInt32}}, ::Base.Threads.SpinLock) at .\lock.jl:161
 [4] _lookup_body(::Type{T} where T, ::NTuple{5,UInt32}) at C:\Users\DPad\.julia\packages\RuntimeGeneratedFunctions\uvbuv\src\RuntimeGeneratedFunctions.jl:131
 [5] generated_callfunc_body(::NTuple{4,Symbol}, ::Type{T} where T, ::NTuple{5,UInt32}, ::NTuple{4,DataType}) at C:\Users\DPad\.julia\packages\RuntimeGeneratedFunctions\uvbuv\src\RuntimeGeneratedFunctions.jl:76
 [6] #s12#1 at C:\Users\DPad\.julia\packages\RuntimeGeneratedFunctions\uvbuv\src\RuntimeGeneratedFunctions.jl:163 [inlined]     
 [7] #s12#1(::Any, ::Any, ::Any, ::Any, ::Any) at .\none:0
 [8] (::Core.GeneratedFunctionStub)(::Any, ::Vararg{Any,N} where N) at .\boot.jl:527
 [9] RuntimeGeneratedFunction at C:\Users\DPad\.julia\packages\RuntimeGeneratedFunctions\uvbuv\src\RuntimeGeneratedFunctions.jl:68 [inlined]
 [10] f at C:\Users\DPad\.julia\packages\ModelingToolkit\HTjKG\src\systems\diffeqs\abstractodesystem.jl:133 [inlined]
 [11] ODEFunction at C:\Users\DPad\.julia\packages\DiffEqBase\b38LC\src\diffeqfunction.jl:248 [inlined]
...

julia> # It seems my function is tagged as being in ModelingToolkit...
julia> typeof(MyModule.fun.f.f_oop).parameters[2:3] 
svec(ModelingToolkit.var"#_RGF_ModTag", (0xd6493547, 0xaf3999e1, 0x66674bbf, 0xb716e034, 0xa6bb5c0f))

julia>   # The ModelingToolkit's function cache is empty.
julia> using ModelingToolkit
julia> using RuntimeGeneratedFunctions
julia> getfield(ModelingToolkit, RuntimeGeneratedFunctions._cachename)
Dict{Any,Any}()

julia> # If I define the module at runtime, it works fine.
julia> module MyModule   
               ...
               fun = ODEFunction(sys)
       end
julia> MyModule.fun([0.5], [5.0], 0.)
1-element Array{Float64,1}:
 30.0
julia> getfield(ModelingToolkit, RuntimeGeneratedFunctions._cachename)
Dict{Any,Any} with 2 entries:
  (0x7091df0e, 0xaddd82b1, 0xaa32db83, 0x… => WeakRef(quote…
  (0x61321fab, 0x744b53c8, 0xe99e5a49, 0x… => WeakRef(quote…
```

Note that the regression does not seem to apply when a module is defined at runtime, as I assume is the case for the actual test cases above.

I would expect my function `MyModule.fun` to actually get tagged under `MyModule` instead of `ModelingToolkit` (and that I also should initialise my own cache using `RuntimeGeneratedFunctions.init(MyModule)` inside my module definition, at least if I understood correctly).

dpad:
This adds the ability to manually specify which module's cache to use when an RGF is built as part of an ODEFunction. Basically, together with `eval_expression` you can now specify `eval_module=@__MODULE__` (the default, which simply uses ModelingToolkit's RGF cache).

I've added an example test case, which consists of building an ODEFunction as part of a precompiled user module. 
In this case, "f_bad" throws a `KeyError` because it does not use `eval_module`, so it's tagged in the ModelingToolkit RGF cache which is not kept after precompilation of the user module is done.
On the other hand, "f_good" works because it supplies the user module to `eval_module`, allowing ModelingToolkit to build the RGF within the user module's cache.

NOTE: there's a few other uses of @RuntimeGeneratedFunctions that probably could do with a similar fix to this, or some better solution. I am not sure how best to tackle this other than the current way, so I would appreciate any comments, but I'll leave that out of scope for this PR as it at least fixes #678 for me.


ChrisRackauckas:
I think this should work. I'll just want @c42f's opinion on the safety.

c42f:
Perhaps there's two separate things desired here?
1. Ability to specify which module's precompile cache the RGF lives in
2. Ability to specify which module is used to look up global symbols used in that RGF (including generic functions)

In principle these might be separate modules, I suppose, so https://github.com/SciML/RuntimeGeneratedFunctions.jl/pull/20 might only be half the story?

ChrisRackauckas:
Oh I just read this. Yes, good point.

dpad:
Updated this to work with https://github.com/SciML/RuntimeGeneratedFunctions.jl/pull/20. Note this still only adds the support to ODE types.

ChrisRackauckas:
Can this get updated for https://github.com/SciML/RuntimeGeneratedFunctions.jl/pull/22 and handle all of the system types?

PR #682, opened 2020-12-05
leave D(foo(t)) untouched if foo is not a known function

shashi:
cc @YingboMa

### DIFF
--- a/src/differentials.jl
+++ b/src/differentials.jl
@@ -101,9 +101,11 @@ function expand_derivatives(O::Term, simplify=true; occurances=nothing)
             x = if _iszero(t2)
                 t2
             elseif _isone(t2)
-                derivative(o, i)
+                d = derivative(o, i)
+                d isa NoDeriv ? D(o) : d
             else
                 t1 = derivative(o, i)
+                t1 = t1 isa NoDeriv ? D(o) : t1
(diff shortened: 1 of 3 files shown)


ChrisRackauckas (approved):

YingboMa:
`derivative` or `derivative_idx` probably should not be a user API, since it doesn't apply the chain rule. But the name makes it super confusing, as there are `gradient` and `jacobian`, and they are meant to be used by the users.

YingboMa:
Can we define `ModelingToolkit.derivative(f, x) = expand_derivatives(Differential(x)(f))`?

ChrisRackauckas:
>Can we define ModelingToolkit.derivative(f, x) = expand_derivatives(Differential(x)(f))?

Yes, we should.

shashi:
> Can we define ModelingToolkit.derivative(f, x) = expand_derivatives(Differential(x)(f))?

Of course we can, but in that case I'd like to rename the `derivative` function which is meant to be extended to define differentiation on new functions.

deprecating that should be possible?

YingboMa:
It seem hard to deprecate. `derivative` was a public API before and was in the documentation. It's just that the definition doesn't make sense as it doesn't use chain rules.

PR #690, opened 2020-12-18
Fix nested diff

### LINKED ISSUE #688: expand_derivatives issues
I face up with two problems, using`expand_derivatives`

```julia
@parameters x
@variables u(..)
@derivatives Dx'~x
julia> expand_derivatives(Dx(u(1)))
0
```     
but it should probably work out `Differential(u(1))`
e.g. if u(x) = x+x^2,
then Dx(u(1)) = 1+2x(where x=1) = 3 but not 0 .
The same, if `u` is `Flux.chain `.

Second, with mixed derivative.
```julia
@parameters x y
@variables u(..)
@derivatives Dy'~y
@derivatives Dx'~x
julia> expand_derivatives(Dx(Dy(u(x,y))) )

Julia has exited.
Press Enter to start a new session.
```

 ModelingToolkit v4.3.0

YingboMa:
Fix #688

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "ModelingToolkit"
 uuid = "961ee093-0014-501f-94e3-6117800e7a78"
 authors = ["Chris Rackauckas <accounts@chrisrackauckas.com>"]
-version = "4.3.0"
+version = "4.3.1"
 
 [deps]
 ArrayInterface = "4fba245c-0d91-5ea0-9b3e-6abc04ee57a9"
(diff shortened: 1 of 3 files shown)


ChrisRackauckas (approved):

shashi (approved):

YingboMa:
The CI failure seems unrelated, so I will merge it. The test passes locally.

ChrisRackauckas:
It would be good to figure out what that is because a failing master is going to block any other work.

PR #691, opened 2020-12-18
WIP: use dict instead of dispatch for derivative

shashi:
it's faster:

```
julia> @btime ModelingToolkit.expand_derivatives(D(atan(sin(x)*cos(x))), false)
  12.658 μs (163 allocations: 7.27 KiB)
((1 + ((sin(x) * cos(x)) ^ 2)) ^ -1) * ((cos(x) * cos(x)) + (sin(x) * (-sin(x))))

julia> @btime ModelingToolkit.expand_derivatives(D(atan(sin(x)*cos(x))), false)
  21.009 μs (175 allocations: 7.61 KiB)
((1 + ((sin(x) * cos(x)) ^ 2)) ^ -1) * ((cos(x) * cos(x)) + (sin(x) * (-sin(x))))

```

### DIFF
--- a/src/differentials.jl
+++ b/src/differentials.jl
@@ -181,19 +181,18 @@ sin(x())
 ```
 """
 derivative_idx(O::Any, ::Any) = 0
+
+# Indicate that no derivative is defined.
+struct NoDeriv
+end
 function derivative_idx(O::Term, idx)
-    d = derivatives[O.op]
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
So this will require users to change the Dict and is breaking? Can we hide the details with a macro?

YingboMa:
I'll close this PR, since this doesn't actually speed up the derivative computation and is breaking

PR #692, opened 2020-12-21
namespace_expr should not change type

hexaeder:
If one changes the namespace of an `Term` symbolic, the type parameter get lost. This leads to strange behaviour, for example one can't performe substitutions on namespaced equations as expected:
```julia
using ModelingToolkit
using ModelingToolkit: getname, renamespace, rename
@parameters t p(t) pnew(t)
@variables x(t)
ode = ODESystem([x ~ p], name=:ode)
eq = namespace_equations(ode)[1]
substitute(eq.rhs, ode.p=>pnew)
# ode₊p(t) without patch
# pnew(t) with patch
```

### DIFF
--- a/src/systems/abstractsystem.jl
+++ b/src/systems/abstractsystem.jl
@@ -188,11 +188,11 @@ function namespace_expr(O::Sym,name,ivname)
     O.name == ivname ? O : rename(O,renamespace(name,O.name))
 end
 
-function namespace_expr(O::Term,name,ivname)
+function namespace_expr(O::Term{T},name,ivname) where {T}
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
Add a test?

hexaeder:
I've added a simple type check for `Sym` as well as `Term` in a file that seemed suitable. But I'm not sure whether this is the right test in the right place 😅

ChrisRackauckas (approved):

ChrisRackauckas:
@YingboMa the test error needs to be fixed since it's blocking CI.

Issue #686, opened 2020-12-05
Alias elimination


YingboMa:
Aliasing cases:
1. `u = number` => replace `u` with `number`
2. `u1 = u2` => replace `u1` with `u2`
3. `u = x` => when `u` is a differential variable and `x` is an algebraic variable, then replace `x` with `u`

shashi:
`u1 = u2 => replace u1 with u2`

here do you allow `u2` to be any expression? (I'm guessing not., it just has to be a variable?)

YingboMa:
Yes, that's correct.

YingboMa:
Ah, actually, we allow u2 to be any expr but `u1` must be a simple expr, like `alpha * x`, where `x` is a symbol.

PR #695, opened 2020-12-21
Simpler alias elimination

shashi:
as described in #686

### DIFF
--- a/src/systems/diffeqs/odesystem.jl
+++ b/src/systems/diffeqs/odesystem.jl
@@ -119,26 +119,32 @@ function ODESystem(eqs, iv=nothing; kwargs...)
     end
     iv === nothing && throw(ArgumentError("Please pass in independent variables."))
     for eq in eqs
-        for var in vars(eq.rhs for eq ∈ eqs)
-            if isparameter(var) || isparameter(var.op)
-                isequal(var, iv) || push!(ps, var)
-            else
-                push!(allstates, var)
-            end
-        end
-        if !(eq.lhs isa Symbolic)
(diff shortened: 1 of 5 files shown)


src/systems/diffeqs/odesystem.jl
```diff
@@ -149,3 +149,5 @@ Base.:(==)(sys1::ODESystem, sys2::ODESystem) =
 function rename(sys::ODESystem,name)
     ODESystem(sys.eqs, sys.iv, sys.states, sys.ps, sys.pins, sys.observed, sys.tgrad, sys.jac, sys.Wfact, sys.Wfact_t, name, sys.systems)
 end
+
+isdiffeq(eq) = eq.lhs isa Term && operation(eq.lhs) isa Differential
```
ChrisRackauckas, line 153: this should go in AbstractODESystem

src/systems/reduction.jl
```diff
@@ -27,59 +29,99 @@ function substitute_aliases(diffeqs, dict)
...
+            return prod(nums), syms[1]
+        end
     else
-        0 ~ eq.lhs - eq.rhs
+        return nothing
     end
 end
 
-function alias_elimination(sys::ODESystem)
+function alias_elimination(sys)
```
ChrisRackauckas, line 52: since the output is an ODESystem, this is currently limited to ODESystem

YingboMa:
This is still not quite right. I will fix it tomorrow.

YingboMa:
I am happy with this PR at its current state. I'll merge once tests pass.

PR #696, opened 2020-12-22
Rewrite latexify support.

### LINKED ISSUE #621: Latexify fails for nested derivatives
The `Latexify` recipe fails for nested derivative operators. It seems like this case was just missed in the current implementation / tests. 

The branch to fix this should be aware of #502. 

```julia
using ModelingToolkit, Latexify

@parameters t, x

@variables u(..)

@derivatives Dt'~t
@derivatives Dx'~x

eqs = [Dt(u(t, x)) ~ Dx(Dx(u(t, x)))]

latexify(eqs)
```

```
in Latexify.jl: 
You are trying to create latex-maths from a `String` that cannot be parsed as
an expression. 

`latexify` will, by default, try to parse any string inputs into expressions
and this parsing has just failed.

If you are passing strings that you want returned verbatim as part of your input,
try making them `LaTeXString`s first. 

If you are trying to make a table with plain text, try passing the keyword
argument `latex=false`. You should also ensure that you have chosen an output
environment that is capable of displaying not-maths objects. Try for example
`env=:table` for a latex table or `env=:mdtable` for a markdown table.


Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] latexraw(::String; kwargs::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}) at /home/dan/.julia/packages/Latexify/7DHiI/src/latexraw.jl:114
 [3] latexraw(::String) at /home/dan/.julia/packages/Latexify/7DHiI/src/latexraw.jl:110
 [4] (::ModelingToolkit.var"#414#430")(::Expr) at /home/dan/.julia/packages/ModelingToolkit/BHRXD/src/latexify_recipes.jl:14
 [5] walk(::Expr, ::Function, ::ModelingToolkit.var"#414#430") at /home/dan/.julia/packages/MacroTools/gME9C/src/utils.jl:112
 [6] postwalk at /home/dan/.julia/packages/MacroTools/gME9C/src/utils.jl:122 [inlined]
 [7] #413 at ./none:0 [inlined]
 [8] iterate at ./generator.jl:47 [inlined]
 [9] collect(::Base.Generator{Array{Expr,1},ModelingToolkit.var"#413#429"}) at ./array.jl:686
 [10] macro expansion at /home/dan/.julia/packages/ModelingToolkit/BHRXD/src/latexify_recipes.jl:14 [inlined]
 [11] apply_recipe(::Array{Equation,1}; kwargs::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}) at /home/dan/.julia/packages/Latexify/7DHiI/src/recipes.jl:164
 [12] apply_recipe at /home/dan/.julia/packages/Latexify/7DHiI/src/recipes.jl:158 [inlined]
 [13] latexify(::Array{Equation,1}; kwargs::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}) at /home/dan/.julia/packages/Latexify/7DHiI/src/latexify_function.jl:4
 [14] latexify(::Array{Equation,1}) at /home/dan/.julia/packages/Latexify/7DHiI/src/latexify_function.jl:4
 [15] top-level scope at In[6]:12
 [16] include_string(::Function, ::Module, ::String, ::String) at ./loading.jl:1091
```

korsbo:
Makes everything nice.

Rewrite the latexify recipes pretty much completely. 

- Fixes @isaacsas's `\mathrm` gripes. It's now all consistent.
- Fixes higher-order derivatives. Closes #621.

Requires _latest_ latexify - I'm in the process of testing and releasing so this PR has to wait a bit.

Defaults are nicer so it's already "useful" but I'm also adding some config options that need some documentation. 

Should we also consider future developers who will read this and include a phone number to some good therapist?


test/latexify.jl
```diff
@@ -27,51 +27,134 @@ eqs = [D(x) ~ σ*(y-x)*D(x-y)/D(z),
...
 
+@test latexify(sys; cdot=true) == latexify(equations(sys); cdot=true)
+
 @parameters t
 @variables x(t)
 @derivatives D'~t
 eqs = [D(x) ~ (1+cos(t))/(1+2*x)]
+eqs[1].rhs.args[1].args[2].f isa Function
+
+latexify(eqs) |> render
```
korsbo, line 112: Oops. That has to be removed.

korsbo:
Waiting for https://github.com/JuliaRegistries/General/pull/26826

YingboMa:
What's the status of this PR? Is it good to merge?

korsbo:
The PR was pretty much fine before MKT 5. Now it seems to need some love again - but this time it might make sense to place some of the latexify support directly in SymbolicUtils. However, I think I just reached the breaking point where I'm just going to massively refactor Latexify.jl to make it a lot more extensible. I'll come back to this one when I have settled on a new API for extending latexify.

PR #702, opened 2020-12-29
Simplify type handling in non-concrete varmap

ChrisRackauckas:
It's hard to come up with a useful example, but:

```julia
using OrdinaryDiffEq, Catalyst, DiffEqFlux, ModelingToolkit

NN(S, I, R, p) = FastChain(FastDense(3,10,tanh),FastDense(10,1))([S, I, R], p)[1]
@register NN(S, I, R, p)

rn = @reaction_network begin
    β, S + I --> 2I
    γ, I --> R
    NN(S, I, R, p3n), I --> Q
    δ, Q --> R
end β γ δ p3n

_p3n = Float64.(initial_params(FastChain(FastDense(3,10,tanh),FastDense(10,1))))
@parameters β γ δ p3n
p     = [β=>1.0,γ=>1.0,δ=>1.0,p3n=>_p3n]
tspan = (0.0,250.0)
u0    = [999.0,0.0,1.0,0.0]
op    = ODEProblem(rn, u0, tspan, p)
sol   = solve(op,Tsit5())
```

is such a thing. We can try to handle this better in the future but at least for now it gives `Vector{Pair{Num,Any}}` and keeps `Vector{Any}`.

### DIFF
--- a/src/systems/reaction/reactionsystem.jl
+++ b/src/systems/reaction/reactionsystem.jl
@@ -487,24 +487,18 @@ end
 
 # ODEProblem from AbstractReactionNetwork
 function DiffEqBase.ODEProblem(rs::ReactionSystem, u0::Union{AbstractArray, Number}, tspan, p=DiffEqBase.NullParameters(), args...; kwargs...)
-    u0 = typeof(u0) <: Array{<:Pair} ? u0 : Pair.(rs.states,u0)
-    p = typeof(p) <: Union{Array{<:Pair},DiffEqBase.NullParameters} ? p : Pair.(rs.ps,p)
     return ODEProblem(convert(ODESystem,rs),u0,tspan,p, args...; kwargs...)
 end
 
 # SDEProblem from AbstractReactionNetwork
 function DiffEqBase.SDEProblem(rs::ReactionSystem, u0::Union{AbstractArray, Number}, tspan, p=DiffEqBase.NullParameters(), args...; noise_scaling=nothing, kwargs...)
     sde_sys = convert(SDESystem,rs,noise_scaling=noise_scaling)
-    u0 = typeof(u0) <: Array{<:Pair} ? u0 : Pair.(rs.states,u0)
-    p = typeof(p) <: Union{Array{<:Pair},DiffEqBase.NullParameters} ? p : Pair.(sde_sys.ps,p)
     p_matrix = zeros(length(rs.states), length(rs.eqs))
     return SDEProblem(sde_sys,u0,tspan,p,args...; noise_rate_prototype=p_matrix,kwargs...)
 end
 
 # DiscreteProblem from AbstractReactionNetwork
 function DiffEqBase.DiscreteProblem(rs::ReactionSystem, u0::Union{AbstractArray, Number}, tspan::Tuple, p=DiffEqBase.NullParameters(), args...; kwargs...)
-    u0 = typeof(u0) <: Array{<:Pair} ? u0 : Pair.(rs.states,u0)
-    p = typeof(p) <: Union{Array{<:Pair},DiffEqBase.NullParameters} ? p : Pair.(rs.ps,p)
     return DiscreteProblem(convert(JumpSystem,rs), u0,tspan,p, args...; kwargs...)
 end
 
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
@isaacsas 🎉

isaacsas:
Nice, I was just looking at what was going wrong in this case! (I'd like us to have something like this work, and it would be awesome to have for the paper.)

ChrisRackauckas:
I think to make it work better, we need some way to represent the `p` here as a vector so we can adjoint over it, but that's a larger MTK issue. Now I'm interested though... this might be a good use of "vacation time"

isaacsas:
Hah, vacation time is exactly what I'm using right now to catch up on issues...

ChrisRackauckas:
The true solution to "adjoints require vectors" might be to just abstract it to VectorOfArray, ArrayPartition, ComponentArray, etc. so that adjoints work on these extended types, and allow them to be the input types in MTK. `[β=>1.0,γ=>1.0,δ=>1.0,p3n=>_p3n]` represented as a ComponentArray is then just a flat vector of Float64s, which would then make it be efficient. So then if we allow that as our state vector in the adjoint expression, we can do all of these crazy arrays of arrays in a very sane "just a vector" way.

isaacsas:
That makes sense, and ComponentArrays are easy enough to recommend to general users.

src/systems/reaction/reactionsystem.jl
```diff
@@ -487,24 +487,18 @@ end
 
 # ODEProblem from AbstractReactionNetwork
 function DiffEqBase.ODEProblem(rs::ReactionSystem, u0::Union{AbstractArray, Number}, tspan, p=DiffEqBase.NullParameters(), args...; kwargs...)
-    u0 = typeof(u0) <: Array{<:Pair} ? u0 : Pair.(rs.states,u0)
-    p = typeof(p) <: Union{Array{<:Pair},DiffEqBase.NullParameters} ? p : Pair.(rs.ps,p)
```
isaacsas, line 491: @ChrisRackauckas Was it intentional to disallow passing straight vectors instead of maps for the initial condition and parameters? Before `u0` and `p` could just be numeric vectors but this doesn't seem to work anymore?

ChrisRackauckas: `varmap_to_vars` has the right dispatches to handle this, so it doesn't need to be handled at this level.

isaacsas: We'll need to update the Catalyst docs if this is no longer allowed since we had previously used things like
```julia
repressilator = @reaction_network begin
    hillr(P₃,α,K,n), ∅ --> m₁
    hillr(P₁,α,K,n), ∅ --> m₂
    hillr(P₂,α,K,n), ∅ --> m₃
    (δ,γ), m₁ ↔ ∅
    (δ,γ), m₂ ↔ ∅
    (δ,γ), m₃ ↔ ∅
    β, m₁ --> m₁ + P₁
    β, m₂ --> m₂ + P₂
    β, m₃ --> m₃ + P₃
    μ, P₁ --> ∅
    μ, P₂ --> ∅
    μ, P₃ --> ∅
end α K n δ γ β μ;

# parameters [α,K,n,δ,γ,β,μ]
p = (.5, 40, 2, log(2)/120, 5e-3, 20*log(2)/120, log(2)/60)

# initial condition [m₁,m₂,m₃,P₁,P₂,P₃]
u₀ = [0.,0.,0.,20.,0.,0.]

# time interval to solve on
tspan = (0., 10000.)

# create the ODEProblem we want to solve
oprob = ODEProblem(repressilator, u₀, tspan, p)
```

ChrisRackauckas: We should have a test on the MTK side then, because that shouldn't error.

ChrisRackauckas: The error for this case is unrelated and has to do with Tuple handling. I'll have a fix in a minute.

ChrisRackauckas: https://github.com/SciML/ModelingToolkit.jl/pull/704 is the way to do it without requiring a special case handling for ReactionSystem.

isaacsas: Awesome, thanks!

isaacsas: I hadn't noticed that was supported throughout MT systems now. Very nice!

ChrisRackauckas: Yup, just cleaning up all of the little details. In a few months it should be very straightforward and things you expect to work should work. It's getting very close. 2019 and 2020 MTK was definitely a developer's library but it should be ready to be using facing quite soon.

PR #706, opened 2020-12-29
Register set operations

dpsanders:


src/extra_functions.jl
```diff
@@ -11,3 +11,12 @@ ModelingToolkit.derivative(::typeof(abs), args::NTuple{1,Any}, ::Val{1}) = IfEls
...
 ModelingToolkit.derivative(::typeof(IfElse.ifelse), args::NTuple{3,Any}, ::Val{2}) = IfElse.ifelse(args[1],1,0)
 ModelingToolkit.derivative(::typeof(IfElse.ifelse), args::NTuple{3,Any}, ::Val{3}) = IfElse.ifelse(args[1],0,1)
+
+@register ∈(x::Num, y::AbstractArray)
+
+@register ∪(x, y)
+@register ∩(x, y)
+@register ∨(x, y)
+@register ∧(x, y)
+@register ≠(x, y)
```
YingboMa, line 21: This is not right. Don't we already have `x != y`?

src/extra_functions.jl
```diff
@@ -11,3 +11,11 @@ ModelingToolkit.derivative(::typeof(abs), args::NTuple{1,Any}, ::Val{1}) = IfEls
 ModelingToolkit.derivative(::typeof(IfElse.ifelse), args::NTuple{3,Any}, ::Val{1}) = 0
 ModelingToolkit.derivative(::typeof(IfElse.ifelse), args::NTuple{3,Any}, ::Val{2}) = IfElse.ifelse(args[1],1,0)
 ModelingToolkit.derivative(::typeof(IfElse.ifelse), args::NTuple{3,Any}, ::Val{3}) = IfElse.ifelse(args[1],0,1)
+
+@register ∈(x::Num, y::AbstractArray)
+
+@register ∪(x, y)
```
YingboMa, line 17: I am not sure if this is correct. We have `Num <: Real`, so `∪(x, y)` doesn't really make sense.

dpsanders: It makes sense for intervals...

dpsanders: Although it's not clear that it makes sense for intervals to be `<: Real`.

ChrisRackauckas: I'll take it. It won't be useful to trace through, that's for sure.

PR #709, opened 2020-12-30
fix DiscreteProblem varmap usage

isaacsas:
This is causing a test failure for Catalyst and needed to be updated now that `varmap_to_vars` handles a variety of input types.


isaacsas:
This should be good to go. I just incremented the MT version as it would be good to get a release with this fix for Catalyst to use.

ChrisRackauckas:
Ahh yes, I should've checked for this sorry. But yes, now all of the special casing should be gone 🎉

isaacsas:
No worries, thanks!

PR #710, opened 2020-12-30
Fix conversion of reactionsys to nonlinsys

isaacsas:


ChrisRackauckas:
Looks like a real failure in the dependency graphs.

isaacsas:
Yeah, have to look at it more closely later to see what’s going on.

src/systems/reaction/reactionsystem.jl
```diff
@@ -188,25 +188,35 @@ function oderatelaw(rx; combinatoric_ratelaw=true)
     rl
 end
 
-function assemble_drift(rs; combinatoric_ratelaws=true)
-    D   = Differential(rs.iv)
-    eqs = [D(x) ~ 0 for x in rs.states]
+function assemble_oderhs(rs; combinatoric_ratelaws=true)
     species_to_idx = Dict((x => i for (i,x) in enumerate(rs.states)))
+    rhsvec         = Any[0 for i in eachindex(rs.states)]
```
isaacsas, line 193: Is this a reasonable initialization for a vector that is going to store the rhs of ODEs, but constructed in an incremental fashion as below? (I guess I'm still worrying about performance issues with using `Any`, but perhaps everything is already not concretely typed so it shouldn't matter here...)

ChrisRackauckas: Equation?

isaacsas: I'm using that function to generate the rhs of the ODEs, and then wrapping into an equation in the calling function (where the user can pass a flag whether they want the SS or time-dependent eq).

I guess part of my problem is I don't know whether I should be keeping things wrapped in a `Num` or not internally here. The `assemble_diffusion` code that @YingboMa and @shashi made seems to use `Num(0)`, which keeps everything as a `Num` in the generated diffusion expressions, while the old code here was unwrapped. Which is preferred for the internal representation?

isaacsas: The problem with the dependency graphs is they assume the expressions are unwrapped currently.

isaacsas:
Looks like this passes test now. @shashi and I can tweak for performance as we work on BCR. 

I do think we should unify whether the drift and diffusion expressions are stored as `Num`s or as `Term`s at some point though.

src/systems/reaction/reactionsystem.jl
YingboMa: Internally, the expression should have no `Num`. Otherwise many other internal operations could fail.

isaacsas: Ok, thanks. I’ll update `assemble_diffusion’s tomorrow then.

PR #717, opened 2021-01-07
add the liouville transformation

ChrisRackauckas:
@yingboma this would be a nice case for a transformation to be able to set an initial condition, another indicator that we need some form of default values

### DIFF
--- a/docs/src/systems/ODESystem.md
+++ b/docs/src/systems/ODESystem.md
@@ -17,6 +17,7 @@ ODESystem
 
 ```@docs
 ode_order_lowering
+liouville_transform
 ```
 
 ## Applicable Calculation and Generation Functions
--- a/src/ModelingToolkit.jl
+++ b/src/ModelingToolkit.jl
@@ -207,7 +207,7 @@ export SteadyStateProblem, SteadyStateProblemExpr
 export JumpProblem, DiscreteProblem
 export NonlinearSystem, OptimizationSystem
 export ControlSystem
-export ode_order_lowering
+export ode_order_lowering, liouville_transform
 export runge_kutta_discretize
 export PDESystem
 export Reaction, ReactionSystem, ismassaction, oderatelaw, jumpratelaw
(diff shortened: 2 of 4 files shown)


ChrisRackauckas:
@agerlach

agerlach:
I’m not at my computer to look closer but I don’t think the implementation is correct. It should be dpdt = -p * tr(jac).

Also, why do you state the IC should be zero?

ChrisRackauckas:
Oh yes, I found a good source:

https://abhishekhalder.bitbucket.io/F16ACC2013Final.pdf

Indeed I was missing a term, and that IC would come from the characteristic.

ChrisRackauckas:
@YingboMa I think I'll go with this for now, though I don't quite like exposing a hardcoded symbol. I can't think of another good way to do it though.

agerlach:
> I don't quite like exposing a hardcoded symbol

Why not just have the user create and pass the symbol

```julia
function liouville_transform(sys, var)
    t = independent_variable(sys)
    D = ModelingToolkit.Differential(t)
    neweq = D(var) ~ var*-tr(calculate_jacobian(sys))
    neweqs = [equations(sys);neweq]
    vars = [states(sys);var]
    ODESystem(neweqs,t,vars,parameters(sys))
end

@variables q
liouville_transform(sys, q)
```

or wrap it in a type?

```julia
struct Liouville
    var
end

function transform(sys, alg::Liouville)
    var = alg.var
    t = independent_variable(sys)
    D = ModelingToolkit.Differential(t)
    neweq = D(var) ~ var*-tr(calculate_jacobian(sys))
    neweqs = [equations(sys);neweq]
    vars = [states(sys);var]
    ODESystem(neweqs,t,vars,parameters(sys))
end

@variables q
transform(sys,q)
```

ChrisRackauckas:
Having the user pass a symbol might be a better option than having one hardcoded. @YingboMa what do you think?

PR #719, opened 2021-01-07
Make sure equations returns Vector{Equation}

### LINKED ISSUE #716: Empty Equation array leads to incorrect type for connections function. Causes flatten to fail.
If one of the component ODESystems is purely made up of observations, the joint ODESystem has the wrong type for equations and parameters. This makes flatten fail:

```julia
## Setup
using ModelingToolkit
const MTK = ModelingToolkit

@parameters t
@derivatives D'~t

## Define a system

@variables x(t), u(t), y(t)
@parameters a, b, c, d
ol = ODESystem([D(x) ~ a * x + b * u, y ~ c * x], t, name=:ol)

## Define another system that is purely observing

@variables u_c(t), y_c(t)
@parameters k_P
pc = ODESystem(Equation[], t, pins=[y_c], observed = [u_c ~ k_P * y_c], name=:pc)

## Connect them

connections = [
    ol.u ~ pc.u_c,
    y_c ~ ol.y
]

connected = MTK.ODESystem(connections, t, systems=[ol, pc])

##

# equations(connected) has eltype Any:
@show eltype(equations(connected))
@show eltype(parameters(connected)) # So does parameters
@show eltype(parameters(pc)) # So does parameters of pc

## hence flatten doesn't work:
MTK.flatten(connected) # MethodError

```

YingboMa:
Fix #716

### DIFF
--- a/src/systems/abstractsystem.jl
+++ b/src/systems/abstractsystem.jl
@@ -176,7 +176,11 @@ function namespace_pins(sys::AbstractSystem)
     [renamespace(sys.name,x) for x in pins(sys)]
 end
 
-namespace_equations(sys::AbstractSystem) = namespace_equation.(equations(sys),sys.name,sys.iv.name)
+function namespace_equations(sys::AbstractSystem)
+    eqs = equations(sys)
+    isempty(eqs) && return Equation[]
+    map(eq->namespace_equation(eq,sys.name,sys.iv.name), eqs)
+end
 
 function namespace_equation(eq::Equation,name,ivname)
     _lhs = namespace_expr(eq.lhs,name,ivname)
@@ -236,11 +240,10 @@ function equations(sys::ModelingToolkit.AbstractSystem)
     if isempty(sys.systems)
         return sys.eqs
     else
-        eqs = [sys.eqs;
+        eqs = Equation[sys.eqs;
                reduce(vcat,
                       namespace_equations.(sys.systems);
                       init=Equation[])]
-
         return eqs
     end
 end
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "ModelingToolkit"
 uuid = "961ee093-0014-501f-94e3-6117800e7a78"
 authors = ["Chris Rackauckas <accounts@chrisrackauckas.com>"]
-version = "4.5.0"
+version = "4.5.1"
 
 [deps]
 ArrayInterface = "4fba245c-0d91-5ea0-9b3e-6abc04ee57a9"
--- a/src/systems/reduction.jl
+++ b/src/systems/reduction.jl
@@ -1,4 +1,4 @@
-export alias_elimination
+export alias_elimination, flatten
 
 function flatten(sys::ODESystem)
     if isempty(sys.systems)
--- a/test/reduction.jl
+++ b/test/reduction.jl
@@ -109,7 +109,7 @@ let
         x ~ y
     ];
     sys = ODESystem(eqs, t, [x], []);
-    asys = alias_elimination(ModelingToolkit.flatten(sys))
+    asys = alias_elimination(flatten(sys))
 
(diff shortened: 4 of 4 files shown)


ChrisRackauckas (approved):

src/systems/abstractsystem.jl
```diff
@@ -236,11 +240,10 @@ function equations(sys::ModelingToolkit.AbstractSystem)
     if isempty(sys.systems)
         return sys.eqs
     else
-        eqs = [sys.eqs;
+        eqs = Equation[sys.eqs;
```
isaacsas, line 243: I think this might break for both jump and reaction systems, neither of which store `Equation`s as their internal `eqs`. @YingboMa did you check if this works for those system types?

isaacsas: I guess we probably don't have any tests on systems storing systems for those, since I'm not sure anyone has used that functionality.

YingboMa: Do jump and reaction systems use different types?

isaacsas: Yeah, one stores `Reactions` and the other `AbstractJumps` as the `eqs`. But my guess is that subsystems won’t currently work for either, and didn’t before, so maybe this is worth worrying about if/when we try to make connections work for them.

YingboMa: Do you mind to open a PR and add tests for that? I'd be happy to fix it if it fails.

isaacsas: Sure! I probably need to do a fair amount of work if we want sub-systems to work for both since they were designed before sub-systems were really finalized. (Though for `ReactionSystem`s having sub-systems and connections seems natural for developing modular signaling libraries so we should definitely plan to support it.)

ChrisRackauckas: yeah. Just so you're aware, what's going on in this next month is Yingbo is writing a full Modelica-style library for something, so there's a few features of AbstractSystem we noticed are missing, namely things like default initial conditions and parameters for systems. So those will get added in the next month to get the first demo of a "full" MTK library, we can get ReactionSystem to be a fully compliant example of how we think AbstractSystems should look, and then I think it's good to say everything is pretty finalized in that domain and we should offload it to be part of Catalyst itself. But yeah... subsystems are new and almost there, but this churn on the system interface should be starting to converge to a final form.

isaacsas: We should maybe discuss how we'd want reaction subsystems to work. I guess it can be similar to odes, where you define reactions that couple two systems? Should species be shared though, or assumed distinct? I can imagine sometimes wanting a reaction system to be a compartment, so species shouldn't be shared, but other times we might want modularity for building a pathway from many components.

For jumps I guess connections would just be jumps that involve states from two different systems in one of the rate or affect?

ChrisRackauckas: In AbstractSystems we always have that species are assumed distinct. In ODEs though, what we then do is `lorenz1.x ~ lorenz2.x`, which lets us say "these two are actually the same, and when generating the code we eliminate the aliasing to get efficient code. But this makes it easy to do separation of concerns: I make a model of a lung with outflux, I make a model of the throat with influx, and then I make a model of models where I say those two are equal, and the symbolic system understands that and generates everything it needs. That's the "acausal" way of doing system modeling. I think what we need to somehow support in ReactionSystem is some non-reaction equations which allow for defining explicit equations, like `a ~ 2b` so that `a` could then be used in other reactions. If we can then mix reactions and equations, then a ReactionSystem with two subsystems can say `system1.right ~ system2.left` and an appropriate overload of `alias_elimination(sys)` would take care of it all. But indeed it's a little bit different since reaction systems don't generally write equations in eqs, so it's something to think about.

isaacsas: Ahh, ok. I haven't tried the ODE connections out yet. I'll have to take a look at what is done there.

Issue #684, opened 2020-12-05
What to do with `derivative`


YingboMa:
Should we define `ModelingToolkit.derivative(f, x) = expand_derivatives(Differential(x)(f))`? It's technically breaking, but I doubt the current behavior of not using chain rules is desirable.

ChrisRackauckas:
The current `derivative` is "symbolic rules", not the derivative function. We should definitely change this because it's a confusing trap for most users.

PR #720, opened 2021-01-07
WIP: upgrade to SymbolicUtils w/ fast terms

YingboMa:
This PR also deprecates `.op` and `.args` on `Term`s.

Most of the functionalities are upgraded except Hessian sparsity. @shashi is looking into that.


ChrisRackauckas:
Is this breaking? @YingboMa take a look at Pumas and make sure it stays fine.

isaacsas:
We'll need some Catalyst updates too.

YingboMa:
Yeah, it's a little breaking, unfortunately. I will test those when this PR and the SymbolicUtils PR is finalized.

ChrisRackauckas:
Yeah so this probably qualifies for a breaking release.

src/utils.jl
```diff
@@ -309,10 +309,10 @@ xˍtt(t)
 ```
 """
 function diff2term(O)
-    isa(O, Term) || return O
+    (isa(O, Symbolic) && !isa(O, Sym)) || return O
```
YingboMa, line 312:
```suggestion
    istree(O)) || return O
```

ChrisRackauckas:
Yeah no worries. I am fine with having it a bit common in this stage in the repo, especially for a 100x performance gain. It just needs to be done right and take care of our downstream buddies

YingboMa:
If we are going to do a breaking release anyway. Let's also sneak in https://github.com/SciML/ModelingToolkit.jl/issues/684 :P

src/systems/jumps/jumpsystem.jl
```diff
@@ -75,7 +75,7 @@ function generate_rate_function(js, rate)
 end
 
 function generate_affect_function(js, affect, outputidxs)
-    build_function(affect, states(js),
+    bf = build_function(map(x->x isa Equation ? x.rhs : x , affect), states(js),
```
shashi, line 78: @isaacsas this should be correct, am I right?

isaacsas: Looks good. Thanks!

YingboMa:
All tests pass locally for me now.

shashi:
@isaacsas @ChrisRackauckas (ping anyone else using this package downstream)

This is a breaking change! Let me summarize what this means to ModelingToolkit users:

- `Term` is not the only expression type now, there is also `Add`, `Mul`, and `Pow`.
- If you were dispatching on `Term`, now use `istree(x)` to check if `x` is of any of these expression types. All of them support `operation` and `arguments` accessors, so:
- `x.op` and `x.args` are deprecated in favor of `operation(x)` and `arguments(x)`.
Use these sed commands to make make this change across all files:
  - `sed -i 's/\(\b[a-zA-Z0-9_]*\b\)\.args/arguments\(\1\)/g' **/*.jl`
  - `sed -i 's/\(\b[a-zA-Z0-9_]*\b\)\.op/operation\(\1\)/g' **/**/*.jl`
 Note that `expr.args` where `expr` is an `Expr` will also be rewritten by this change, so be careful of what you commit (use `git add -p` to select the right changes, commit, and then run `git reset --hard` -- or however you usually do this.).

Here is what changed (and a migration guide) in SymbolicUtils.jl https://github.com/JuliaSymbolics/SymbolicUtils.jl/pull/154#issuecomment-756897615 if you need to get into the finer details.

ChrisRackauckas:
Thanks, makes sense. Let's make sure own downstream libraries all get PRs too.

YingboMa:
I’d like to do some more breaking changes before we merge it, like changing
the BipartiteGraph struct and the derivative.

ChrisRackauckas:
Sounds good.

YingboMa:
It's the annoying string LaTeX test again... Can we just delete it to test for nonempty string?

ChrisRackauckas:
what's going on there?

YingboMa:
Oh, this is actually due to the exp change. I will fix it.

YingboMa:
Fingers crossed. CI should pass in this commit.

Issue #727, opened 2021-01-09
` MethodError: no method matching lower_mapnames(::Array{Any,1}) ` when trying to pass distributions as parameters


claudio20497:
Hello,

If I try to `modelingtoolkitize` a DifferentialEquations.jl SDE that takes a parameters array `p` that contains both `Float64`s and distributions, I get an error.

So I reproduced the behaviour in the following MWE:

```julia

using DifferentialEquations, Distributions

# Model parameters

β = 0.01# infection rate
λ_R = 0.05 # inverse of transition time from  infected to recovered
λ_D = 0.83 # inverse of transition time from  infected to dead
σ_β = 0.01 
σ_R = 0.01 
σ_D = 0.01 

𝒫 = [β, λ_R, λ_D, Normal(), Normal(), Normal()]


# regional contact matrix and regional population

## regional contact matrix
const C = [3.45536   0.485314  0.506389  0.123002 ; 0.597721  2.11738   0.911374  0.323385 ; 0.906231  1.35041   1.60756   0.67411 ; 0.237902  0.432631  0.726488  0.979258] # 4x4 contact matrix

## regional population stratified by age
N= [723208 , 874150, 1330993, 1411928] # array of 4 elements, each of which representing the absolute amount of population in the corresponding age class.


# Initial conditions 
i₀ = 0.075 # fraction of initial infected people in every age class
I₀ = repeat([i₀],4)
S₀ = N.-I₀
R₀ = [0.0 for n in 1:length(N)]
D₀ = [0.0 for n in 1:length(N)]
D_tot₀ = [0.0]
ℬ = vcat([S₀, I₀, R₀, D₀, D_tot₀]...) 

# Time 
final_time = 20
𝒯 = (1.0,final_time); 

# initialize this parameter (death probability stratified by age, taken from literature)
#const δ₁, δ₂, δ₃, δ₄ = [0.003/100, 0.004/100, (0.015+0.030+0.064+0.213+0.718)/(5*100), (2.384+8.466+12.497+1.117)/(4*100)]
const δ = vcat(repeat([0.003/100],1),repeat([0.004/100],1),repeat([(0.015+0.030+0.064+0.213+0.718)/(5*100)],1),repeat([(2.384+8.466+12.497+1.117)/(4*100)],4-1-1-1))





function SIRD!(du,u,p,t)  
    # extract parameters
    β, λ_R, λ_D, _,_,_= p


    # State variables
[...]

PR #732, opened 2021-01-13
don't require numbers for varmap lowering

ChrisRackauckas:
The only specialization really needed is on Pair.

Fixes https://github.com/SciML/ModelingToolkit.jl/issues/727


ChrisRackauckas:
test failures from master @YingboMa ?

PR #734, opened 2021-01-13
fix "if parameterizedfunction just return" return

ChrisRackauckas:
I don't think we need a test because it's such a weird feature... found in MATLAB tests


src/build_function.jl
```diff
@@ -478,8 +478,11 @@ get_varnumber(varop, vars::Vector) =  findfirst(x->isequal(x,varop),vars)
 
 function numbered_expr(O::Symbolic,args...;varordering = args[1],offset = 0,
                        lhsname=gensym("du"),rhsnames=[gensym("MTK") for i in 1:length(args)])
-    O = value(O)
-  if O isa Sym || isa(operation(O), Sym)
+
+  O = value(O)
+  if ModelingToolkit.value(O) isa Sym && ModelingToolkit.value(O).name == args[3].name
```
YingboMa, line 483: Should use `nameof(x)`

YingboMa:
What exactly does this PR fix? `args` may not have length 3.

ChrisRackauckas:
It fixes MATLABDiffEq.jl. Its build_function is just... old. It needs more work, but I was hoping to at least keep the benchmarks running.

PR #736, opened 2021-01-14
Offset fix for ctarget matrix equations

### LINKED ISSUE #735: Matrix-output functions for CTarget build_function calls
# Overview

[Discourse topic](https://discourse.julialang.org/t/minimum-working-example-for-modelingtoolkits-ctarget/53303)

I'm trying to produce a C function for a Jacobian. As a minimum-working example, I tried to leave the `lhs` variables out of the state vector argument to `build_function`, but this produces an error. I can retrieve run-able C code if I add the `lhs` variables to the state vector, but then the generated C code has indices which are offset.

I don't see any big reason why vectors __should__ work, and matrices __should not__. If we know ahead of time how C indexes 2D arrays (row major, right?) then __we should be able to treat the output as a vector with single indices__. I'm not sure what the cause of the `+(::Nothing, ::Int64)` error shown below is.

# More Detail

#### Setup
First, some setup code:
```Julia
julia> import Pkg
julia> Pkg.activate(tempdir())
julia> Pkg.add("ModelingToolkit")
julia> using ModelingToolkit
```

#### Desired Usage (errors)
Here, I try to leave the `lhs` variables out of the state vector argument. This produces an error.

```Julia
julia> @variables x[1:2,1:2] y[1:4]
julia> eqs = x[:] .~ y

julia> build_function(eqs, y, target=ModelingToolkit.CTarget())
ERROR: MethodError: no method matching +(::Nothing, ::Int64)
Closest candidates are:
  +(::Any, ::Any, ::Any, ::Any...) at operators.jl:538
  +(::ChainRulesCore.One, ::Any) at /home/joe/.julia/packages/ChainRulesCore/cpHLu/src/differential_arithmetic.jl:94
  +(::ChainRulesCore.Zero, ::Any) at /home/joe/.julia/packages/ChainRulesCore/cpHLu/src/differential_arithmetic.jl:63
```

#### Working Usage (offset indices)
If I __add__ the `lhs` variables to the state vector argument, I do get C-code, but the indices are offset due to additional (dead) states in the state vector. 

```Julia
julia> @variables x[1:2,1:2] y[1:4]
julia> eqs = x[:] .~ y

julia> build_function(eqs, [y..., x...]; target=ModelingToolkit.CTarget())
"void diffeqf(double* du, double* RHS1) {\n  du[4] = RHS1[0];\n  du[5] = RHS1[2];\n  du[6] = RHS1[1];\n  du[7] = RHS1[3];\n}\n"
```

cadojo:
## Overview

* Closes #735 
* Adds one helper function, `get_symnumber` (just a copy of existing code, moved to a function)
* Adds some offset math for handling matrix equations in `_build_function(::CTarget)`.

## More Detail

Before this PR (`master`), calling the following code would result in an error.

```Julia
julia> @variables x[1:2,1:2] y[1:4]
julia> eqs = x[:] .~ y

julia> build_function(eqs, y, target=ModelingToolkit.CTarget())
ERROR: MethodError: no method matching +(::Nothing, ::Int64)
Closest candidates are:
  +(::Any, ::Any, ::Any, ::Any...) at operators.jl:538
  +(::ChainRulesCore.One, ::Any) at /home/joe/.julia/packages/ChainRulesCore/cpHLu/src/differential_arithmetic.jl:94
  +(::ChainRulesCore.Zero, ::Any) at /home/joe/.julia/packages/ChainRulesCore/cpHLu/src/differential_arithmetic.jl:63
```

Upon further inspection, this issue was caused on `master:build_function.jl:500`. 

```Julia
# build_function.jl:500
i = findfirst(x->isequal(tosymbol(x isa Sym ? x : operation(x), escape=false), tosymbol(var, escape=false)),varordering)
```

Because the none of the output variables `x` appear in the state vector provided (here the state vector provided is the second argument to `build_function`, `y`), the `findfirst` call returns `nothing`.

## Changes
* I used the above `findfirst` code in one new place, so I moved it to a new function called `get_symnumber`, and replaced that line (line 500) with a call to `get_symnumber`
* In `numbered_expr(::Equation...)`, I changed the `lhs` and `rhs` offsets to the following logic:
  * If the output variable is not found in the provided state vector, `lhs` offset is equal to `1 + the index of the output vector element`, and `rhs` offset is equal to the index of each state vector element
  * If the output variable _is_ found in the provided state vector, then the previous logic was used
* In `_build_function(::CTarget...)`, I added the following logic at the top:
  * If no output is in the provided state vector, use new logic (`offset = index of output - 2`)
  * If all outputs are in the provided state vector, use previous logic (`offset = -1`)

## Conclusion
I'm definintely open to other implementations, but this was the least-optrusive solution I could think of. With this PR, the output of the following code block is as expected.

```Julia
julia> @variables x[1:2,1:2] y[1:4]
julia> eqs = x[:] .~ y

julia> build_function(eqs, y, target=ModelingToolkit.CTarget())
"void diffeqf(double* du, double* RHS1) {\n  du[0] = RHS1[0];\n  du[1] = RHS1[1];\n  du[2] = RHS1[2];\n  du[3] = RHS1[3];\n}\n"
```

### DIFF
--- a/src/build_function.jl
+++ b/src/build_function.jl
@@ -536,7 +536,7 @@ function _build_function(target::CTarget, eqs::Array{<:Equation}, args...;
     count(lhs_in_rhs) ∈ (0, length(lhs_in_rhs)) || throw(ArgumentError("If one output is included in the state vector, all must be."))
     offsets = [val ? -1 : i-2 for (i, val) ∈ enumerate(lhs_in_rhs)]
     differential_equation = string(join([numbered_expr(eq,args...,lhsname=lhsname,
-                                  rhsnames=rhsnames,offset=-1) for
+                                  rhsnames=rhsnames,offset=offsets[i]) for
                                   (i, eq) ∈ enumerate(eqs)],";\n  "),";")
 
     argstrs = join(vcat("double* $(lhsname)",[typeof(args[i])<:Array ? "double* $(rhsnames[i])" : "double $(rhsnames[i])" for i in 1:length(args)]),", ")
--- a/src/build_function.jl
+++ b/src/build_function.jl
@@ -475,6 +475,7 @@ function vars_to_pairs(name,vs, symsdict)
 end
 
 get_varnumber(varop, vars::Vector) =  findfirst(x->isequal(x,varop),vars)
+get_symnumber(varsym, vars)  = findfirst(x->isequal(tosymbol(x isa Sym ? x : operation(x), escape=false), tosymbol(varsym, escape=false)),vars)
 
 function numbered_expr(O::Symbolic,args...;varordering = args[1],offset = 0,
                        lhsname=gensym("du"),rhsnames=[gensym("MTK") for i in 1:length(args)])
@@ -497,8 +498,8 @@ function numbered_expr(de::ModelingToolkit.Equation,args...;varordering = args[1
 
     varordering = value.(args[1])
     var = var_from_nested_derivative(de.lhs)[1]
-    i = findfirst(x->isequal(tosymbol(x isa Sym ? x : operation(x), escape=false), tosymbol(var, escape=false)),varordering)
-    :($lhsname[$(i+offset)] = $(numbered_expr(de.rhs,args...;offset=offset,
+    i = get_symnumber(var, varordering)
+    :($lhsname[$((i === nothing ? 1 : i)+offset)] = $(numbered_expr(de.rhs,args...;offset=(i === nothing ? -1 : offset),
 											  varordering = varordering,
 											  lhsname = lhsname,
 											  rhsnames = rhsnames)))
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
Can this get a test?

cadojo:
I added the test case shown in #735 and this PR. For some reason, checks are failing for unrelated tests. Tests for `derivates` are now failing for 049ac92 and 8795d27, which is odd because 8795d27 tests succeeded just a few hours ago.

ChrisRackauckas:
@YingboMa ?

YingboMa (commented):
This logic is too hard to understand just from the code. I don't understand what
```julia
@variables x[1:2,1:2] y[1:4]
eqs = x[:] .~ y
build_function(eqs, y, target=ModelingToolkit.CTarget())
```
is supposed to do.

ChrisRackauckas:
The test failure on master got fixed though? We should update the branch when there's a passing master.

ChrisRackauckas:
Yeah I think the issue is that the CTarget, MATLABTarget, and StanTarget are old and need some work. `build_function`'s first argument should be the term or list of terms, not an equation. `build_function(y, y, target=ModelingToolkit.CTarget())` is what it should probably be changing to. I'd be okay with accepting the patch, but I would expect this function to get overhauled.

YingboMa:
I wouldn't want to merge this. It basically impossible to understand how `+1` or some random integer is supposed to be the offset to differentiate different cases just from the code.

Also,
```julia
build_function(eqs, y, ...)
```
is just wrong. We shouldn't support it.

ChrisRackauckas:
>is just wrong. We shouldn't support it.

It is just wrong, these days. It's the MTK function signature from 2 years ago. It still lives on in the non-Julia targets because it hasn't been updated. And those haven't been deleted because they are still there because they are used in a bunch of tests and benchmarks. We need to completely overhaul them.

cadojo:
Permission to close this PR in favor of the draft PR #746?

ChrisRackauckas:
I was thinking the same thing. Yeah doing the full update of CTarget is great and I'm happy you're interested in taking it on. Thanks!

PR #741, opened 2021-01-18
don't simplify by default in expand_derivatives

shashi:


YingboMa:
The diff looks messed up. I opened #742

PR #746, opened 2021-01-19
Build Function Updates

### LINKED ISSUE #745: CTarget, MATLABTarget, StanTarget Rewrite
### Overview

Non-Julia `build_function` targets currently use outdated function signatures, which were first introduced with `ModelingToolkit` around two years ago (as described in #736).

A rewrite for these `_build_function` implementations would be useful. Improvements could include...

* Supporting the `_build_function(::JuliaTarget)` function signature, e.g. `_build_function(::CTarget, ex, args)`
* Support for matrix and vector equations (described in #735 and PR #736)

cadojo:
### Work in Progress!

Closes #745 

This PR will include new `_build_function` implementations alongside the existing implementations. Existing implementations will produce new deprecation warnings, and the new implementations will match the function signatures used by `_build_function(::JuliaTarget, ...)`.

### Completed

- [ ]  `MATLABTarget` Rewrite

- [x] `CTarget` Rewrite 

- [ ] `StanTarget` Rewrite

### DIFF
--- a/src/build_function.jl
+++ b/src/build_function.jl
@@ -529,19 +529,21 @@ control the compilation:
 function _build_function(target::CTarget, eqs::Array{<:Equation}, args...;
                          conv = toexpr, expression = Val{true},
                          fname = :diffeqf,
-						 lhsname=:du,rhsnames=[Symbol("RHS$i") for i in 1:length(args)],
-						 libpath=tempname(),compiler=:gcc)
+                         lhsname=:du,rhsnames=[Symbol("RHS$i") for i in 1:length(args)],
+                         libpath=tempname(), compiler=:gcc)
 
-    differential_equation = string(join([numbered_expr(eq,args...,lhsname=lhsname,
-                                  rhsnames=rhsnames,offset=-1) for
-                                  (i, eq) ∈ enumerate(eqs)],";\n  "),";")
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
@shashi what needs to be done to make this compatible with what you've been doing?

cadojo:
Current status: 

Matrix expressions are working for `CTarget`. I've added a `columnmajor` as a keyword argument to `_build_function(::Ctarget...)`, which defaults to true because Eigen [defaults](https://eigen.tuxfamily.org/dox/group__TopicStorageOrders.html) to column-major order. I have private tests in another repo which I'm using this functionality in, and they are passing.

I'm planning on doing the same re-write for MATLAB targets, and Stan targets, though I have no experience with Stan.

ChrisRackauckas:
Stan is very similar to C except for the function headers.

ChrisRackauckas:
`build_function` is now part of Symbolics.jl because it was considered part of a standard CAS, like `latexify` for SymPy. Thus this PR should move to https://github.com/JuliaSymbolics/Symbolics.jl. See https://juliasymbolics.org/roadmap/ for more details. It doesn't look like I can automatically transfer it either, so you'll need to paste the code over.

PR #748, opened 2021-01-20
Introduce default_ps and default_u0

### LINKED ISSUE #739: Initial conditions and default parameters handling
We need
```julia
    initial_conditions::Any # [x => 1.0, y => 2.0, z => 2y]
    # observed: [y ~ 2x]
    # [x => 1.0, y => 2*(2x)]
    default_parameters::Any
```
fields in various systems, so that a user can write
```julia
eqs(params) = ...
model = (name, iv, (structural_params, params)) -> ODESystem(eqs(structural_params), pins=[F, H, K...], default_params = ...)
```
to instantiate a new model. Default parameters are going to be merged with `params` using `merge(default, params)` so that `params` is preferred.

YingboMa:
Closes #739

### DIFF
--- a/src/systems/abstractsystem.jl
+++ b/src/systems/abstractsystem.jl
@@ -297,3 +297,6 @@ function (f::AbstractSysToExpr)(O)
     end
     return build_expr(:call, Any[operation(O); f.(arguments(O))])
 end
+
+get_default_p(sys) = sys.default_p
+get_default_u0(sys) = sys.default_u0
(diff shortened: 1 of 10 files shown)


shashi:
lgtm!

YingboMa:
@ChrisRackauckas do we want to use `default_ps` or `default_p`? In ODEProblem, we have `prob.p`. It feels a little awkward to call it `ps` in here.

ChrisRackauckas:
`default_p` makes more sense to me.

This seems to omit SDESystem, NonlinearSystem, ControlSystem, NonlinearSystem, etc.

PR #751, opened 2021-01-21
Convert observed equations as assignments when lowering

YingboMa:


ChrisRackauckas (approved):

ChrisRackauckas (approved):

YingboMa:
I will merge this once tests pass, and I will add more features in this area.

PR #752, opened 2021-01-21
Update symbolic_functions.md to make derivative example work

GlenHertz:
Without `D` defined the next code snippet will fail.


YingboMa:
Thanks!

PR #754, opened 2021-01-25
Complete NonlinearSystem

ChrisRackauckas:


YingboMa (commented):
Looks good to me. Could you add a test?

PR #765, opened 2021-02-02
Pretty print and some fixes

YingboMa:


src/register_function.jl
```diff
@@ -48,7 +48,7 @@ macro register(expr, Ts = [Num, Symbolic, Real])
     push!(
           ex.args,
           quote
-              if $!($hasmethod($promote_symtype, $Tuple{$typeof($f), $Vararg}))
+              if $define_promotion
                   (::$typeof($promote_symtype))(::$typeof($f), args...) = $ret_type
```
YingboMa, line 52: When I use the `hasmethod` check, the `promote_symtype` method doesn't seem to be defined only when I run `@register` in a file. Any ideas? @shashi

src/ModelingToolkit.jl
```diff
@@ -152,13 +152,13 @@ end
...
 tosymbolic(a::Sym) = tovar(a)
 tosymbolic(a) = a
-@num_method Base.isless isless(tosymbolic(a), tosymbolic(b)) (Real,)
-@num_method Base.:(<) (tosymbolic(a) < tosymbolic(b)) (Real,)
-@num_method Base.:(<=) (tosymbolic(a) <= tosymbolic(b)) (Real,)
-@num_method Base.:(>) (tosymbolic(a) > tosymbolic(b)) (Real,)
-@num_method Base.:(>=) (tosymbolic(a) >= tosymbolic(b)) (Real,)
-@num_method Base.isequal isequal(tosymbolic(a), tosymbolic(b)) (AbstractFloat, Number, Symbolic)
-@num_method Base.:(==) tosymbolic(a) == tosymbolic(b) (AbstractFloat,Number)
+@num_method Base.isless  Num(isless(tosymbolic(a), tosymbolic(b))) (Real,)
```
YingboMa, line 155: We need these `Num`s to make sure the invariance `foo(::Num) -> ::Num` is satisfied.

shashi: Okay, what we had was I think what `Constant` did.

src/register_function.jl
shashi: I'm not sure

src/ModelingToolkit.jl
YingboMa: I changed this to `val isa Bool ? val : Num(val)`, so that `Num(1) == 1` is true to be more user-friendly

YingboMa:
Test failure is unrelated. All tests pass locally.

PR #768, opened 2021-02-04
Add @add_kwonly to PDESystem

KirillZubov:


src/systems/pde/pdesystem.jl
```diff
@@ -1,7 +1,30 @@
-struct PDESystem <: ModelingToolkit.AbstractSystem
+struct PDESystem
```
ChrisRackauckas, line 1: why?

src/systems/pde/pdesystem.jl
```diff
@@ -1,7 +1,30 @@
...
+
+struct PDESystem{iip}
+  eq
+  bcs
+  domain
+  indvars
+  depvars
+
+  @add_kwonly function PDESystem{iip}(eq, bcs, domain, indvars, depvars) where iip
+      new{iip}(eq, bcs, domain, indvars, depvars)
```
ChrisRackauckas, line 22: that doesn't make sense: systems don't have in-place properties since they aren't functions?

src/systems/pde/pdesystem.jl
KirillZubov: https://github.com/SciML/ModelingToolkit.jl/issues/770

ChrisRackauckas: we should rename the field `eqs`.

PR #769, opened 2021-02-05
Work around to specify cache module when building RuntimeGeneratedFunctions

dpad:
The RuntimeGeneratedFunctions (RGF) macro `@RuntimeGeneratedFunction` allows the caller to specify the "context" module, but sets the "cache" module (i.e. which RGF cache will contain this function) to `@__MODULE__`, which ends up being ModelingToolkit. 

However, if a user is indirectly calling ModelingToolkit's `build_function` at precompile time, ModelingToolkit's cache eventually disappears, and the resulting RGF cache lookup does not work. 

Therefore, the user needs to set up an RGF cache in their own module, and ensure `build_function` is called with `expression=false` and `expression_module=UserModule`. The work-around in this PR then sets *both* the RGF's "cache" *and* "context" modules to be the `UserModule`.

### DIFF
--- a/src/build_function.jl
+++ b/src/build_function.jl
@@ -101,7 +101,10 @@ function _build_and_inject_function(mod::Module, ex)
     elseif ex.head == :(->)
         return _build_and_inject_function(mod, Expr(:function, ex.args...))
     end
-    @RuntimeGeneratedFunction(mod, ex)
+    # XXX: Workaround to specify the module as both the cache module AND context module.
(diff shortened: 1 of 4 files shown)


ChrisRackauckas:
It would be good to get @c4tf's opinion here.

c42f:
This seems valid to me (getting the caching right is tricky).

We just need some support in RuntimeGeneratedFunctions to make it officially supported. See https://github.com/SciML/RuntimeGeneratedFunctions.jl/pull/26

PR #777, opened 2021-02-09
Add support for symbolic parameters

pepijndevos:
With this change you can express parameters as an expression of another parameter
```
par = [
    σ => 1,
    ρ => 0.1+σ,
    β => ρ*1.1
]
```
In combination with `flatten` this allows to express parameter dependencies across different levels, where you can define a slew of lower level parameter in terms of a single top-level definition for example.

To make this usage more convenient I have added assignment to parameters to set the default. Currently there does not appear to be an API to generate symbol names in the flattened system other than by specifying the full path with `\_+` or using the hierarchical model to get it, which might be fine.

I'm currently not checking parameters for cycles.

I'm not 100% sure if my usage of `namespace_expr` is correct because of the unused `iv` parameter.

I could not find a way to properly test the `NonlinearProblem` parameters and solution. So I've resorted to manually calling `varmap_to_vars` and just running the solver to see that nothing breaks.

### DIFF
--- a/src/systems/abstractsystem.jl
+++ b/src/systems/abstractsystem.jl
@@ -184,7 +184,7 @@ end
 function Base.setproperty!(sys::AbstractSystem, prop::Symbol, val)
     param = Sym{Parameter{Real}}(prop)
     if param in parameters(sys)
-        sys.default_p[param] = val
+        sys.default_p[param] = value(val)
     else
         setfield!(sys, prop, val)
     end
@@ -212,7 +212,7 @@ end
 
 function namespace_default_p(sys)
     d_p = default_p(sys)
-    Dict(parameters(sys, k) => namespace_expr(d_p[k], nameof(sys), []) for k in keys(d_p))
+    Dict(parameters(sys, k) => namespace_expr(d_p[k], nameof(sys), independent_variable(sys)) for k in keys(d_p))
 end
 
 function namespace_equations(sys::AbstractSystem)
--- a/src/systems/abstractsystem.jl
+++ b/src/systems/abstractsystem.jl
@@ -182,9 +182,10 @@ function Base.getproperty(sys::AbstractSystem, name::Symbol)
 end
 
 function Base.setproperty!(sys::AbstractSystem, prop::Symbol, val)
-    param = Sym{Parameter{Real}}(prop)
-    if param in parameters(sys)
-        sys.default_p[param] = value(val)
+    if (pa = Sym{Parameter{Real}}(prop); pa in parameters(sys))
+        sys.default_p[pa] = value(val)
+    elseif (st = Sym{Real}(prop); st in states(sys))
+        sys.default_u0[st] = value(val)
     else
         setfield!(sys, prop, val)
     end
--- a/src/systems/abstractsystem.jl
+++ b/src/systems/abstractsystem.jl
@@ -181,6 +181,15 @@ function Base.getproperty(sys::AbstractSystem, name::Symbol)
     throw(error("Variable $name does not exist"))
 end
 
+function Base.setproperty!(sys::AbstractSystem, prop::Symbol, val)
+    param = Sym{Parameter{Real}}(prop)
+    if param in parameters(sys)
+        sys.default_p[param] = val
+    else
+        setfield!(sys, prop, val)
+    end
+end
+
 function renamespace(namespace, x)
     if x isa Num
         renamespace(namespace, value(x))
@@ -203,7 +212,7 @@ end
 
 function namespace_default_p(sys)
     d_p = default_p(sys)
-    Dict(parameters(sys, k) => d_p[k] for k in keys(d_p))
+    Dict(parameters(sys, k) => namespace_expr(d_p[k], nameof(sys), []) for k in keys(d_p))
 end
 
 function namespace_equations(sys::AbstractSystem)
--- a/src/systems/abstractsystem.jl
+++ b/src/systems/abstractsystem.jl
@@ -184,7 +184,8 @@ end
 function Base.setproperty!(sys::AbstractSystem, prop::Symbol, val)
     if (pa = Sym{Parameter{Real}}(prop); pa in parameters(sys))
         sys.default_p[pa] = value(val)
-    elseif (st = Sym{Real}(prop); st in states(sys))
+    # comparing a Sym returns a symbolic expression
+    elseif (st = Sym{Real}(prop); any(s->s.name==st.name, states(sys)))
         sys.default_u0[st] = value(val)
     else
         setfield!(sys, prop, val)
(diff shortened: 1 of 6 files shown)


src/systems/abstractsystem.jl
```diff
@@ -181,6 +181,15 @@ function Base.getproperty(sys::AbstractSystem, name::Symbol)
     throw(error("Variable $name does not exist"))
 end
 
+function Base.setproperty!(sys::AbstractSystem, prop::Symbol, val)
```
YingboMa, line 184: Don't make it `Symbol` and assume it must be a `Parameter`. We can use another function like `setdefault(sys, sym, val)` to do this.

src/systems/abstractsystem.jl
```diff
@@ -181,6 +181,15 @@ function Base.getproperty(sys::AbstractSystem, name::Symbol)
     throw(error("Variable $name does not exist"))
 end
 
+function Base.setproperty!(sys::AbstractSystem, prop::Symbol, val)
+    param = Sym{Parameter{Real}}(prop)
+    if param in parameters(sys)
+        sys.default_p[param] = val
```
YingboMa, line 187: This should be
```julia
sys.default_p[value(param)] = value(val)
```
to unwrap potential `Num`.

src/systems/abstractsystem.jl
```diff
@@ -203,7 +212,7 @@ end
 
 function namespace_default_p(sys)
     d_p = default_p(sys)
-    Dict(parameters(sys, k) => d_p[k] for k in keys(d_p))
+    Dict(parameters(sys, k) => namespace_expr(d_p[k], nameof(sys), []) for k in keys(d_p))
```
YingboMa, line 215: I am not sure what this is trying to do. Let's take a step back and don't handle multiple level yet.

src/systems/abstractsystem.jl
pepijndevos: Isn't that what `getparameter` is already doing? If I write `sys.par` that will convert the `par` symbol and return the parameter, if it is present.

This function is so that you can write `flat.b = ns.σ*0.5` as I do in the test case. Else you have to write `flat.default_p[flat.b] = ns.σ*0.5` and IIRC it acts weird because of `Num` wrapping.

src/systems/abstractsystem.jl
pepijndevos: It works perfectly, the only issue is that `namespace_expr` has a special case where it does not namespace the independent variable. There should be no independent variable in the parameters, so I pass a dummy value.

Imagine you have a parameter dict like this inside a nested system
```
par = [
    σ => 1,
    ρ => 0.1+σ,
    β => ρ*1.1
]
```
If you call `default_p` on this, it will return
```
par = [
    sys₊σ => 1,
    sys₊ρ => 0.1+σ,
    sys₊β => ρ*1.1
]
```
Which is wrong because now  `σ` doesn't resolve anymore. The correct dict is
```
par = [
    sys₊σ => 1,
    sys₊ρ => 0.1+sys₊σ,
    sys₊β => sys₊ρ*1.1
]
```
Any this is what `namespace_expr` accomplishes. You always have to namespace the keys together with the values to maintain correctness.

src/systems/abstractsystem.jl
pepijndevos: Could you explain this? `param` was just constructed a few lines earlier. For `val` it might make sense, I just don't fully understand `Num` wrapping. The reason for it, and when something needs to be a Num or not.

pepijndevos:
I ran into a problem with this PR, which is that if you don't pass any explicit parameters, it uses a `NullParameters`, and does not call `varmap_to_vars` at all. This then leaks symbolic expressions from the defaults into the solver.

I worked around this in my code by passing `Pair[]` as the parameters which makes it call `varmap_to_vars`, but the problem is that if there are no parameters at all, it will fail at `Base.promote_typeof` on `Any`.

src/systems/abstractsystem.jl
YingboMa: That's not what `getproperty` is doing. It doesn't assume the name refers to a parameter. Could you add a check for states and set the `u0`, too?

test/symbolic_parameters.jl
```diff
@@ -0,0 +1,35 @@
...
+@variables a
+@parameters b
+top = NonlinearSystem([0 ~ -a + ns.x+1], [a], [b], systems=[ns], name=:top)
+flat = flatten(top)
+flat.b = ns.σ*0.5
+
+flatres = ModelingToolkit.varmap_to_vars(Dict(), parameters(flat), defaults=ModelingToolkit.default_p(flat))
+@test flatres == [0.5, 1, 0.1+1, (0.1+1)*1.1]
+
+prob = NonlinearProblem(flat, ones(4), Pair[])
```
YingboMa, line 34: Could you add another test for the
```julia
prob = NonlinearProblem(flat, ones(4))
```
case?

pepijndevos:
Thanks for the review.

I've updated the PR to properly handle u0 defaults as well, and added this to the testcase.
And also added a basic test of the NullParameters.

YingboMa (approved):
LGTM!

test/symbolic_parameters.jl
```diff
@@ -0,0 +1,43 @@
...
+
+prob = NonlinearProblem(ns, [u=>1.0], Pair[])
+@show sol = solve(prob,NewtonRaphson())
+
+@variables a
+@parameters b
+top = NonlinearSystem([0 ~ -a + ns.x+b], [a], [b], systems=[ns], name=:top)
+top.b = ns.σ*0.5
+
+res = ModelingToolkit.varmap_to_vars(Dict(), parameters(top), defaults=ModelingToolkit.default_p(top))
```
ChrisRackauckas, line 35: can we get a test on the defaults with the component notation?

pepijndevos: What do you mean?

I was a bit puzzled by testing parameters with `NonlinearProblem` because you can not easily access the final parameter list.
I guess I could add a test on the computer problem u0, which you can access it appears.
Is that what you mean?

ChrisRackauckas: I mean the high level overriding the default with like `ns.σ => b`

pepijndevos: Good point! It's actually running into a problem where `in` compares the states and... returns a symbolic expression rather than a boolean.
```
julia> s = Sym{ModelingToolkit.Parameter{Real}}(:foo)
foo

julia> s == s
true

julia> s = Sym{Real}(:foo)
foo

julia> s == s
foo == foo
```

pepijndevos: Is there like a `in` that uses `===`? That one seems to work (was just a wild guess based on JS experience)
Other alternative is to compare `s.name` with a custom HOF.

pepijndevos:
Okay I got the states figured out, and added more tests.

One thing I ran into a few times is when Julia infers the type of a datastructure, and later complains when you try to add a different type. I managed to solve this a few times by adding the right type annotations, and other cases I worked around it by purposefully making it infer to Any.

pbouffard:
Friendly appeal to some updates to docs explaining this cool new feature 😄 .

PR #789, opened 2021-02-12
small improvements to JumpProblem dispatches

ChrisRackauckas:


src/systems/jumps/jumpsystem.jl
```diff
@@ -262,7 +262,7 @@ function DiffEqJump.JumpProblem(js::JumpSystem, prob, aggregator; kwargs...)
 
     statetoid = Dict(value(state) => i for (i,state) in enumerate(states(js)))
     eqs       = equations(js)
-    invttype  = typeof(1 / prob.tspan[2])
+    invttype  = prob.tspan[1] === nothing ? Float64 : typeof(1 / prob.tspan[2])
```
YingboMa, line 265: Do you mean `prob.tspan === nothing`?

ChrisRackauckas: No, DiscreteProblem is weird. Since `f` is optional, it has to type `tspan` as `Tuple`, so the null value is `(nothing,nothing)`.

PR #794, opened 2021-02-16
turn coveralls back on

isaacsas:
We still have the badge in the readme but it hasn't actually run since November it seems.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,3 +30,7 @@ jobs:
       - uses: codecov/codecov-action@v1
         with:
           file: lcov.info
+      - uses: coverallsapp/github-action@master
+        with:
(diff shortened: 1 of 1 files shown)


isaacsas:
The integration test failure doesn't seem to have anything to do with this PR.

ChrisRackauckas:
Indeed, the integration test failure is known (found yesterday, and so I added the test and mentioned it to @shashi )

isaacsas:
Sounds good. Please do let me know if you see Catalyst failures occurring regularly going forward. I'm hoping that switching to StableRNGs with a fixed seed will have eliminated all the random failures we were getting...

ChrisRackauckas:
The plan with the integration tests is to not allow anything that breaks Catalyst or NeuralPDE. Once updated, DiffEqOperators will be in that list as well.

PR #795, opened 2021-02-16
support complex numbers in equalitions

pepijndevos:


ChrisRackauckas:
If this is fine (calling in other reviews), it will need a test.

YingboMa (changes requested):
this will cause stack overflow when one writes `0 ~ 0+0im`

pepijndevos:
I added some tests and cannot reproduce a stack overflow.

shashi:
Can you say a bit about why this is needed?

pepijndevos:
To be able to use complex numbers in equations. For spice frequency analysis you use complex voltages to represent phase and magnitude. Some initial works was done in #759 to allow mathematical operations on complex numbers, but to actually use them in a model, they also need to be defined for `~` else you get ambiguity errors.

pepijndevos:
Alright so after some discussion with @YingboMa and @ChrisRackauckas I learned that it is not as simple and clear cut as I hoped to support complex numbers, so they asked me to go into a bit more detail to start a discussion about the best way to solve the problem.

One of the reservations of adding support for complex numbers to equations is what do you do when the user writes something like `abs(Complex(a, b))~Complex(x, y)` which doesn't really make sense. So maybe there is an argument to be made for only supporting `Complex~Complex` and `Real~Real`. Or I guess you could just assume that `Complex~Real` implies `Complex(n, m)~Complex(x, 0)`. I think there was another issue with `abs` but I forgot what that was about.

Another thing to think about is that in fact an equation of complex numbers is two equations, for the real and imaginary part. So maybe it should not be supported at all and the user should write out the individual equations. On the other hand this is more verbose than supporting complex numbers directly, and quite nasty when trying to build a general framework for spice modeling.

Here is a minimal example that demonstrates the intended usage. It requires this PR, as well as #777, which I have merged into https://github.com/pepijndevos/ModelingToolkit.jl/tree/both for convenient testing. It constructs an RC filter with complex impedance, to obtain a transfer function in the frequency domain (using `s=jω` in the Laplace transform). By sweeping the ω parameter a bode plot can be obtained, which is used frequently in electrical engineering.

With this PR applied, it currently crashes on an ambiguity error on trying to do `Num(n::Complex{Num})`, and I'm not sure what the  solution is.

```
ERROR: MethodError: Num(::Complex{Int64}) is ambiguous. Candidates:
  (::Type{T})(z::Complex) where T<:Real in Base at complex.jl:37
  Num(val) in ModelingToolkit at /home/pepijn/.julia/packages/ModelingToolkit/lQobm/src/ModelingToolkit.jl:51
Possible fix, define
  Num(::Complex)
```

```
using ModelingToolkit
using NonlinearSolve

function ac_voltage_component(name::Symbol, voltage=missing)
    @parameters ω ac
    @variables i1 i2 v1 v2
    eqns = [
        0 ~ -ac + (v1 - v2)  # Ohms law
        0 ~ i1 + i2              # device KCL: currents leaving device sum to zero within device
    ]
    par = ismissing(voltage) ? [] : [ac => voltage]
    sys = NonlinearSystem(eqns, [i1, i2, v1, v2], [ω, ac], name=name, default_p=par)
end

function ac_capacitor_component(name::Symbol, cap=missing)
    @parameters ω c
    @variables i1 i2 v1 v2
    eqns = [
        0 ~ -1/(im*ω*c) * i2 + (v1 - v2)  # Ohms law
        0 ~ i1 + i2              # device KCL: currents leaving device sum to zero within device
    ]
    par = ismissing(cap) ? [] : [c => cap]
    sys = NonlinearSystem(eqns, [i1, i2, v1, v2], [ω, c], name=name, default_p=par)
end

function ac_resistor_component(name::Symbol, res=missing)
    @parameters ω r
    @variables i1 i2 v1 v2
    eqns = [
        0 ~ -r * i2 + (v1 - v2)  # Ohms law
        0 ~ i1 + i2              # device KCL: currents leaving device sum to zero within device
    ]
    par = ismissing(res) ? [] : [r => res]
    sys = NonlinearSystem(eqns, [i1, i2, v1, v2], [ω, r], name=name, default_p=par)
end

cur((sys, num)::Tuple{ModelingToolkit.AbstractSystem, Number}) = getproperty(sys, Symbol(:i, num))
cur(num::Number) = iszero(num) ? 0 : Sym{Real}(Symbol(:i, num))
vol((sys, num)::Tuple{ModelingToolkit.AbstractSystem, Number}) = getproperty(sys, Symbol(:v, num))
vol(num::Number) = iszero(num) ? 0 : Sym{Real}(Symbol(:v, num))

function connect(ports)
    currents = map(cur , ports)
    voltages = map(vol , ports)
    v0 = voltages[1]
    voltages = voltages[2:end]
    eqs = map(v -> 0 ~ v - v0, voltages)
    if !(0 in ports) # skip groud KCL
        push!(eqs, 0 ~ sum(currents))
    end
    eqs
end

r1 = ac_resistor_component(:r1, 1e3)
c1 = ac_capacitor_component(:ci, 1e-6)
v1 = ac_voltage_component(:v1, 1)

eqs = [
    connect([(v1, 1), (r1, 1)]);
    connect([(r1, 2), (c1, 1)]);
    connect([(v1, 2), (c1, 2), 0]);
]

function freq_defaults(sys)
    @parameters ω
    for p in parameters(sys)
        if endswith(String(nameof(p)), "ω")
            sys.default_p[ModelingToolkit.value(p)] = ModelingToolkit.value(ω)
        end
    end
end

@parameters ω
sys = NonlinearSystem(eqs, [], [ω], systems=[r1, c1, v1])
freq_defaults(sys)
u0 = [s => Complex(0.0) for s in states(sys)]
prob = NonlinearProblem(sys, u0, [ω => 1e3])
@show sol = solve(prob,NewtonRaphson())
```

YingboMa:
This code has been moved to Symbolics.

PR #798, opened 2021-02-16
Bring back multithreading

shashi:
Uses https://github.com/JuliaSymbolics/SymbolicUtils.jl/pull/204/files

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -67,7 +67,6 @@ Unitful = "1.1"
 julia = "1.2"
 
 [extras]
-Dagger = "d58978e5-989f-55fb-8d15-ea34adc7bf54"
 ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
 GalacticOptim = "a75be94c-b780-496d-a8a9-0878b188d577"
 NonlinearSolve = "8913a72c-1f9b-4ce2-8d82-65094dcecaec"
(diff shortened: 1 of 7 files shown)


shashi:
Right now it creates 2 * nthreads() taks -- just to oversubscribe a little, but maybe this should be configurable as an argument to build_function.

shashi:
It probably makes more sense to use `MultithreadedForm(N)` where `N=2*nthreads()` if not specified.

ChrisRackauckas:
I think we should default to `N=nthreads()` since the overhead is high and so clumping to reduce overhead is a good idea?

ChrisRackauckas:
Looks good. Though:

https://github.com/SciML/ModelingToolkit.jl/pull/798/files#diff-ab049598c645cd39917d2003f9c80ffe47dc0a68e688465029db6239a30c30f4R163-R166

Remove those? Or make Distributed work?

shashi:
@ChrisRackauckas that link does not work for me... Do you mean DistributedForm and DaggerForm? I would be in favor of removing those because I can't imagine it having low enough overhead to be useful apart from being a proof of concept.

ChrisRackauckas:
Yeah that's what it was. We can remove it, that's fine. I think we need to redo it really: I think it can be useful if it's pre-split. So remove it for now: it needs more program analysis to be useful.

ChrisRackauckas (approved):

PR #801, opened 2021-02-19
use unflatten_long_ops in build_function again

shashi:


src/build_function.jl
```diff
@@ -52,14 +52,19 @@ function build_function(args...;target = JuliaTarget(),kwargs...)
...
-                 +(+((~~x)[1:N]...) + (+)((~~x)[N+1:end]...)) : nothing)
-    rule2 = @rule((*)((~~x)) => length(~~x) > N ?
-                 *(*((~~x)[1:N]...) * (*)((~~x)[N+1:end]...)) : nothing)
+function unflatten_args(f, args, N=4)
+    length(args) < N && return Term{Number}(f, args)
+    unflatten_args(f, [Term{Number}(f, group)
+                                       for group in Iterators.partition(args, N)], N)
+end
 
+function unflatten_long_ops(op, N=4)
```
YingboMa, line 61: Does this function return nothing, sometimes?

https://github.com/SciML/ModelingToolkit.jl/pull/801/checks?check_run_id=1933640784#step:6:260

src/build_function.jl
```diff
@@ -53,18 +53,18 @@ function build_function(args...;target = JuliaTarget(),kwargs...)
...
 
 function unflatten_long_ops(op, N=4)
     op = value(op)
-    !istree(op) && return
+    !istree(op) && return Num(op)
     rule1 = @rule((+)(~~x) => length(~~x) > N ? unflatten_args(+, ~~x, 4) : nothing)
     rule2 = @rule((*)(~~x) => length(~~x) > N ? unflatten_args(*, ~~x, 4) : nothing)
 
-    Rewriters.Postwalk(Rewriters.Chain([rule1, rule2]))(op)
+    Num(Rewriters.Postwalk(Rewriters.Chain([rule1, rule2]))(op))
```
shashi, line 67: I don't understand why this fixed it...

YingboMa: lol. It fixed the sparse array case. We need `Num` to make sure that `zero` is defined.

shashi: Ahh ok

PR #802, opened 2021-02-22
add single eq ODESystem constructor

anandijain:
should be harmless additional constructor. there doesn't seem to be any reason not to have this.

### DIFF
--- a/src/systems/diffeqs/odesystem.jl
+++ b/src/systems/diffeqs/odesystem.jl
@@ -208,3 +208,5 @@ function flatten(sys::ODESystem)
                         )
     end
 end
+
+ODESystem(eq::Equation; kwargs...) = ODESystem([eq]; kwargs...)
--- a/src/systems/diffeqs/odesystem.jl
+++ b/src/systems/diffeqs/odesystem.jl
@@ -209,4 +209,4 @@ function flatten(sys::ODESystem)
     end
 end
 
-ODESystem(eq::Equation; kwargs...) = ODESystem([eq]; kwargs...)
+ODESystem(eq::Equation, args...; kwargs...) = ODESystem([eq], args...; kwargs...)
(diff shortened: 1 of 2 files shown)


src/systems/diffeqs/odesystem.jl
```diff
@@ -208,3 +208,5 @@ function flatten(sys::ODESystem)
                         )
     end
 end
+
+ODESystem(eq::Equation; kwargs...) = ODESystem([eq]; kwargs...)
```
YingboMa, line 212: It should be
```suggestion
ODESystem(eq::Equation, args...; kwargs...) = ODESystem([eq], args...; kwargs...)
```

YingboMa:
Also, could you add a test?

anandijain:
thanks @YingboMa ! I don't have a great intuition for what is a good test. happy to revise this

YingboMa:
Add a single equation system like `u' = alpha*u` and test for `equations(sys)`.

anandijain:
hows this

test/odesystem.jl
```diff
@@ -267,7 +267,10 @@ eqs = D(x) .~ nums
 
 systems = ODESystem.(eqs)
 sys = ODESystem(eqs[1])
+eq = D(x) ~ r*x
```
YingboMa, line 270: This should be moved to the end to make definition and testing local.

test/odesystem.jl
```diff
@@ -256,7 +256,7 @@ sys = ODESystem(eqs, t)
 @test isempty(parameters(sys))
 
 # one equation ODESystem test
-@parameters t
+@parameters t r
```
YingboMa, line 259: so does this

anandijain:
thanks. i think this might be a better test. if this works then so will the broadcast, and this is more readable

PR #804, opened 2021-02-22
Define inplace and out of place methods in ODEFunctionExpr

mohamed82008:
In this PR, I attempt to define an inplace and out of place methods for the ODE function in `ODEFunctionExpr`. I did this on GitHub for some reason so I didn't run local tests yet. Will add a test after I open the PR.


mohamed82008:
This is ready for review.

ChrisRackauckas:
Can this get a test in https://github.com/SciML/ModelingToolkit.jl/blob/master/test/odesystem.jl to check at least that `f` works with both calls after this is evaled?

mohamed82008:
Tests added.

ChrisRackauckas (approved):

ChrisRackauckas:
Looks like a real test failure.

mohamed82008:
>Looks like a real test failure.

Ya the loop masked some variables used outside.

PR #810, opened 2021-02-24
add isautonomous

ChrisRackauckas:
Checks for whether the ODESystem is autonomous

@shashi is there a better way to do this?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "ModelingToolkit"
 uuid = "961ee093-0014-501f-94e3-6117800e7a78"
 authors = ["Chris Rackauckas <accounts@chrisrackauckas.com>"]
-version = "5.6.4"
+version = "5.7.0"
 
 [deps]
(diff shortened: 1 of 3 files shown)


YingboMa (approved):
LGTM

shashi:
You can use `jacobian_sparsity` for speed.

YingboMa:
`jacobian_sparsity` will think that `x(t)` is time dependent.

ChrisRackauckas:
Yeah. But the method of jacobian_sparsity could be used w.r.t. `t` and that would speed it up.

ChrisRackauckas:
Opened an issue in https://github.com/SciML/ModelingToolkit.jl/issues/816 . The issue is that sparsity w.r.t. `t` still has `D(x(t))` as non-zero, so we'd then have to zero all of those, and then use the sparsity. So it's not trivial enough to do right now, but it's something we could think about for the future.

PR #820, opened 2021-02-25
Add default parameter argument to the PDESystem

ashutosh-b-b:
Needed for SciML/NeuralPDE.jl#256

### DIFF
--- a/src/systems/pde/pdesystem.jl
+++ b/src/systems/pde/pdesystem.jl
@@ -5,8 +5,9 @@ struct PDESystem <: ModelingToolkit.AbstractSystem
   indvars
   depvars
   ps
-  @add_kwonly function PDESystem(eqs, bcs, domain, indvars, depvars, ps = SciMLBase.NullParameters())
-      new(eqs, bcs, domain, indvars, depvars, ps)
+  default_p
+  @add_kwonly function PDESystem(eqs, bcs, domain, indvars, depvars, ps = SciMLBase.NullParameters(), default_p = nothing)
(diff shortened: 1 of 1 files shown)


ChrisRackauckas (approved):

ChrisRackauckas:
@AlCap23 do you know what the downstream issue is? It's not related to the PR so it's kind of odd.

AlCap23:
I think its due to the recent updates with the getters / setters and maybe some changes within symbolic utils. I will have a look now and see whats causing this, since the PRs in DataDriven seems to be fine. But this PR has nothing to do with it :+1:

AlCap23:
Quick update: This is related to 

```julia
@variables x y
a = Num[x;x;y;x;y] 
unique(a) # Works
unique!(a) # Errors with
```

```julia
unique!(a)
ERROR: TypeError: non-boolean (Num) used in boolean context
Stacktrace:
 [1] issorted(::Array{Num,1}, ::Base.Order.ForwardOrdering) at ./sort.jl:62
 [2] #issorted#1 at ./sort.jl:91 [inlined]
 [3] issorted at ./sort.jl:91 [inlined]
 [4] unique!(::Array{Num,1}) at ./set.jl:341
 [5] top-level scope at none:1
```

ChrisRackauckas:
Interesting. Open an issue. I haven't seen that before. @shashi might know

AlCap23:
I found the real error. Sorry. It was hidden in `create_linear_independent_eqs`, which returned `Vector{Any}` instead of `Num` .

PR #837, opened 2021-02-28
Handle dependencies of observed variables properly

### LINKED ISSUE #834: Calculating observed variables which are functions of algebraic variables fails
The new acausal modeling tutorial is the MWE for this.

YingboMa:
Fix #834

### DIFF
--- a/docs/src/tutorials/acausal_components.md
+++ b/docs/src/tutorials/acausal_components.md
@@ -406,10 +406,15 @@ observed(sys)
  capacitor₊n₊v(t) ~ 0.0
  source₊n₊v(t) ~ 0.0
  ground₊g₊i(t) ~ 0.0
- ⋮
+ source₊n₊i(t) ~ capacitor₊p₊i(t)
+ source₊p₊i(t) ~ -capacitor₊p₊i(t)
+ capacitor₊n₊i(t) ~ -capacitor₊p₊i(t)
+ resistor₊n₊i(t) ~ -capacitor₊p₊i(t)
  ground₊g₊v(t) ~ 0.0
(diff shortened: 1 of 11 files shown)


src/systems/nonlinear/nonlinearsystem.jl
```diff
@@ -60,7 +61,7 @@ function NonlinearSystem(eqs, states, ps;
     default_p isa Dict || (default_p = Dict(default_p))
     default_u0 = Dict(value(k) => value(default_u0[k]) for k in keys(default_u0))
     default_p = Dict(value(k) => value(default_p[k]) for k in keys(default_p))
-    NonlinearSystem(eqs, value.(states), value.(ps), observed, name, systems, default_u0, default_p, nothing)
+    NonlinearSystem(eqs, value.(states), value.(ps), observed, name, systems, default_u0, default_p, nothing, [])
```
ChrisRackauckas, line 64: so we need this in all systems?

YingboMa: Only for systems that support `structural_simplify`.

docs/src/tutorials/acausal_components.md
```diff
@@ -12,7 +12,7 @@ equalities before solving. Let's see this in action.
 ## Copy-Paste Example
 
 ```julia
-using ModelingToolkit, Plots
+using ModelingToolkit, Plots, OrdinaryDiffEq
```
ChrisRackauckas, line 15:
```suggestion
using ModelingToolkit, Plots, DifferentialEquations
```

PR #840, opened 2021-03-01
Allow u0 and p defaults to refer to each other

pepijndevos:
Currently symbolic defaults can be expressions of other defaults of the same kind, but @ChrisRackauckas asked me to make a PR that also allows initial conditions to be expressions of parameters.

The easiest way I saw to do this is to merge the default dictionaries before calling `varmap_to_var`. I don't think this will have any unintended consequences unless you have a parameter and a state that share the same name, which would be silly. But I'm open to alternative approaches of course.

### DIFF
--- a/src/systems/diffeqs/abstractodesystem.jl
+++ b/src/systems/diffeqs/abstractodesystem.jl
@@ -273,20 +273,20 @@ function process_DEProblem(constructor, sys::AbstractODESystem,u0map,parammap;
                            kwargs...)
     dvs = states(sys)
     ps = parameters(sys)
+    defaults = merge(default_p(sys), default_u0(sys))
 
     if u0map !== nothing
         u0map′ = lower_mapnames(u0map,get_iv(sys))
-        u0 = varmap_to_vars(u0map′,dvs; defaults=default_u0(sys))
+        u0 = varmap_to_vars(u0map′,dvs; defaults=defaults)
     else
         u0 = nothing
     end
 
-    defp = default_p(sys)
     if !(parammap isa DiffEqBase.NullParameters)
         parammap′ = lower_mapnames(parammap)
-        p = varmap_to_vars(parammap′,ps; defaults=defp)
-    elseif !isempty(defp)
-        p = varmap_to_vars(Dict(),ps; defaults=defp)
+        p = varmap_to_vars(parammap′,ps; defaults=defaults)
+    elseif !isempty(defaults)
+        p = varmap_to_vars(Dict(),ps; defaults=defaults)
     else
         p = ps
     end
--- a/src/systems/jumps/jumpsystem.jl
+++ b/src/systems/jumps/jumpsystem.jl
@@ -204,8 +204,9 @@ dprob = DiscreteProblem(js, u₀map, tspan, parammap)
 """
 function DiffEqBase.DiscreteProblem(sys::JumpSystem, u0map, tspan::Union{Tuple,Nothing},
                                     parammap=DiffEqBase.NullParameters(); kwargs...)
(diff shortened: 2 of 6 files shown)


src/systems/nonlinear/nonlinearsystem.jl
```diff
@@ -211,14 +211,14 @@ function process_NonlinearProblem(constructor, sys::NonlinearSystem,u0map,paramm
...
+    defaults = merge(default_p(sys), default_u0(sys))
+    u0 = varmap_to_vars(u0map′,dvs; defaults=defaults)
 
     if !(parammap isa DiffEqBase.NullParameters)
         parammap′ = lower_mapnames(parammap)
-        p = varmap_to_vars(parammap′,ps; defaults=defp)
-    elseif !isempty(defp)
-        p = varmap_to_vars(Dict(),ps; defaults=defp)
+        p = varmap_to_vars(parammap′,ps; defaults=defaults)
+    elseif !isempty(default_p(sys))
```
YingboMa, line 220: This is very minor, but ideally `default_p(sys)` should be called only once as it is potentially expensive. We need to be careful about this.

src/systems/jumps/jumpsystem.jl
```diff
@@ -232,8 +233,9 @@ dprob = DiscreteProblem(js, u₀map, tspan, parammap)
 """
 function DiscreteProblemExpr(sys::JumpSystem, u0map, tspan::Union{Tuple,Nothing},
                                     parammap=DiffEqBase.NullParameters(); kwargs...)
-    u0 = varmap_to_vars(u0map, states(sys); defaults=default_u0(sys))
-    p  = varmap_to_vars(parammap, parameters(sys); defaults=default_p(sys))
+    defaults = merge(default_p(sys), default_u0(sys))
+    u0 = varmap_to_vars(u0map, states(sys); defaults=defaults)
```
YingboMa, line 237: So does `states`.

src/systems/jumps/jumpsystem.jl
```diff
@@ -204,8 +204,9 @@ dprob = DiscreteProblem(js, u₀map, tspan, parammap)
 """
 function DiffEqBase.DiscreteProblem(sys::JumpSystem, u0map, tspan::Union{Tuple,Nothing},
                                     parammap=DiffEqBase.NullParameters(); kwargs...)
-    u0 = varmap_to_vars(u0map, states(sys); defaults=default_u0(sys))
-    p  = varmap_to_vars(parammap, parameters(sys); defaults=default_p(sys))
+    defaults = merge(default_p(sys), default_u0(sys))
+    u0 = varmap_to_vars(u0map, states(sys); defaults=defaults)
```
YingboMa, line 208: Also here.

src/systems/diffeqs/abstractodesystem.jl
```diff
@@ -273,20 +273,20 @@ function process_DEProblem(constructor, sys::AbstractODESystem,u0map,parammap;
...
     end
 
-    defp = default_p(sys)
     if !(parammap isa DiffEqBase.NullParameters)
         parammap′ = lower_mapnames(parammap)
-        p = varmap_to_vars(parammap′,ps; defaults=defp)
-    elseif !isempty(defp)
-        p = varmap_to_vars(Dict(),ps; defaults=defp)
+        p = varmap_to_vars(parammap′,ps; defaults=defaults)
+    elseif !isempty(defaults)
```
YingboMa, line 288: Also here.

YingboMa (approved):
Just a few minor comments. I am fine with merge this PR as-is.

ChrisRackauckas:
We should generally walk through the library with this in mind.

PR #849, opened 2021-03-02
Actually report correct error for a mass matrix that's not semi-explicit

pepijndevos:
This code was doing an equality on a symbolic expression, which of course results in a symbolic result rather than a boolean.

Now I can actually go and fix the offending equation ;)


pepijndevos:
It's actually kinda weird to me that MTK doesn't seem to support equations with a non-zero algebraic lhs. This makes it so you need to write `a ~ b` as `0 ~ -a + b` all the time, which seems like a fairly mechanical transformation MTK could just do by itself.

YingboMa:
Surely that's a mechanical transformation. That's why there's the `structural_simplify` function.

YingboMa:
`_iszero(x)` is used to check if `x` is a concrete number and is zero. `x === 0` would be false when `x` is `0.0`.

PR #868, opened 2021-03-10
Add hasproperty implementation

pepijndevos:
Currently there is a disconnect between what `hasproperty` says and what `getproperty` returns.

```julia
@variables a b c t
sys = ODESystem(Equation[], t, [a, b, c], [])
hasproperty(sys, :a) # false
sys.a # var"##ODESystem#458₊a"
```

This PR is a simple adaptation of the `getproperty` behavior to match `hasproperty`.


YingboMa (changes requested):
You should overload `Base.propertynames` instead of `hasproperty`. BTW, the signature will be `propertynames(x, private::Bool)` when `private=true` then the output should be `fieldnames(typeof(x))`.

YingboMa:
Closed by #882

PR #882, opened 2021-03-14
Overload `propertynames` and add `@nonamespace`

YingboMa:


ChrisRackauckas:
Where is `@nonamespace` used?

YingboMa:
It's not used, but could be helpful if the user define models hierarchically, and want to share a variable across two parallel level models.

ChrisRackauckas:
Then if it's user-facing, it should get a user-facing docstring. Or at least, a sentence above that which describes what it does and be put into the docs. Right now it looks like an internal.

YingboMa:
Yeah, I haven't decided how this should be used yet, so I haven't documented it yet. I think it will be helpful later when we implement inheritance, so I added it preemptively.

ChrisRackauckas:
alright.

ChrisRackauckas (approved):

PR #884, opened 2021-03-14
add preprocessing step: move all non-differentials from lhs to rhs

### LINKED ISSUE #860: Automatically shuffle derivative terms to the left hand side of an ODESystem?
I am trying to pass from a Lagrangian to something solvable by a DiffEq package. 
* Is this possible?
* Is my approach the right one?

Here is an MWE:
```julia
# harmonic oscillator
using Symbolics
using ModelingToolkit
using OrdinaryDiffEq

# define lagrangian
@variables t m k q q̇
L = k*q^2/2 - m*q̇^2/2

# build euler lagrange equations
Dt = Differential(t)
res = derivative(L, q) - Dt(derivative(L, q̇))
@parameters t
@variables u(t)
res = substitute(res, q => u)
res = substitute(res, q̇ => Dt(u))
display(res)

sys = ODESystem([Dt(Dt((x))) ~ -(k/m)*x]) # works if I shuffle the Dt to lhs by hand
sys = ODESystem([res ~ 0]) # fails
```
```
𝑘∗𝑢(𝑡)−(𝐷𝑖𝑓𝑓𝑒𝑟𝑒𝑛𝑡𝑖𝑎𝑙(𝑡)(−𝑚∗𝐷𝑖𝑓𝑓𝑒𝑟𝑒𝑛𝑡𝑖𝑎𝑙(𝑡)(𝑢(𝑡))))
MethodError: isequal(::Sym{ModelingToolkit.Parameter{Real}}, ::Missing) is ambiguous. Candidates:
  isequal(::Any, ::Missing) in Base at missing.jl:82
  isequal(s::SymbolicUtils.Symbolic, x) in SymbolicUtils at /home/jan/.julia/packages/SymbolicUtils/oQKim/src/types.jl:59
Possible fix, define
  isequal(::SymbolicUtils.Symbolic, ::Missing)

Stacktrace:
 [1] collect_var!(states::OrderedCollections.OrderedSet{Any}, parameters::OrderedCollections.OrderedSet{Any}, var::Sym{ModelingToolkit.Parameter{Real}}, iv::Missing)
   @ ModelingToolkit ~/.julia/packages/ModelingToolkit/Df3sc/src/systems/diffeqs/odesystem.jl:192
 [2] collect_vars!(states::OrderedCollections.OrderedSet{Any}, parameters::OrderedCollections.OrderedSet{Any}, expr::SymbolicUtils.Add{Real, Int64, Dict{Any, Number}}, iv::Missing)
   @ ModelingToolkit ~/.julia/packages/ModelingToolkit/Df3sc/src/systems/diffeqs/odesystem.jl:185
 [3] ODESystem(eqs::Vector{Equation}, iv::Nothing; kwargs::Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
   @ ModelingToolkit ~/.julia/packages/ModelingToolkit/Df3sc/src/systems/diffeqs/odesystem.jl:160
 [4] ODESystem
   @ ~/.julia/packages/ModelingToolkit/Df3sc/src/systems/diffeqs/odesystem.jl:142 [inlined]
 [5] top-level scope
   @ In[41]:24
 [6] eval
   @ ./boot.jl:360 [inlined]
 [7] include_string(mapexpr::typeof(REPL.softscope), mod::Module, code::String, filename::String)
   @ Base ./loading.jl:1090
```

iliailmer:
This is a pr to fix #860

### DIFF
--- a/src/systems/diffeqs/odesystem.jl
+++ b/src/systems/diffeqs/odesystem.jl
@@ -142,6 +142,12 @@ function ODESystem(eqs, iv=nothing; kwargs...)
     diffvars = OrderedSet()
     allstates = OrderedSet()
     ps = OrderedSet()
+    
+    # preprocess eqs
+    args_lhs = map(x->arguments(x.lhs), eqs)
+    derivatives = map(y -> filter(x -> (operation(x) isa Differential), y), args_lhs) 
+    eqs = [ derivatives[i][1] ~ derivatives[i][1] - eqs[i].lhs + eqs[i].rhs for i in 1:length(eqs)]
+    
     # reorder equations such that it is in the form of `diffeq, algeeq`
     diffeq = Equation[]
     algeeq = Equation[]


YingboMa (commented):
We shouldn't do the conversion during the construction of `ODESystem(eqs,...)`. Also, this conversion is very ad-hoc and incorrect in general. We can relax the ODESystem definition and add a pass that solves for the derivative variables like `D(x)`. However, doing this in general needs tearing and state selection. If you are inclined, you can read [this paper](https://2017.international.conference.modelica.org/proceedings/html/submissions/ecp17132565_OtterElmqvist.pdf) to get a general feel about structural transformation passes. Section 4.7 is about state selections.

ChrisRackauckas:
I think I misjudged how hard this is. I think a similar problem that is a better first issue would be generating `DAEProblem`s from the ODESystem, similar to `ODEProblem` except in the implicit form.

iliailmer:
I supposed my submission was too soon done, since the methods I put in return an empty array on some inputs that I have not considered which I think is why the CI failed

PR #899, opened 2021-03-16
Add check to avoid accidental singulatity error after structural_simplify

pepijndevos:
Currently the following will give a singularity error
```julia
u0 = zeros(length(states(sys)))
prob = NonlinearProblem(sys, u0)
sol = solve(prob, NewtonRaphson())
```
This is because structural simplify removed redundant states, and now u0 is too big, resulting in zeros in the jacobian which cause a singularity error. This PR stops you from making that mistake.

I'm not sure if other types of systems also need this check because I don't really know what they do. (jumpsystem etc)

### DIFF
--- a/src/systems/diffeqs/abstractodesystem.jl
+++ b/src/systems/diffeqs/abstractodesystem.jl
@@ -304,7 +304,7 @@ function process_DEProblem(constructor, sys::AbstractODESystem,u0map,parammap;
         p = ps
     end
 
-    @assert length(dvs) == length(u0) "states and initial conditions are of different lengths"
+    length(dvs) == length(u0) || throw(ArgumentError("States ($(length(dvs))) and initial conditions ($(length(u0))) are of different lengths."))
 
     f = constructor(sys,dvs,ps,u0;tgrad=tgrad,jac=jac,checkbounds=checkbounds,
                     linenumbers=linenumbers,parallel=parallel,simplify=simplify,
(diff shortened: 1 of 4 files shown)


src/systems/diffeqs/abstractodesystem.jl
```diff
@@ -304,6 +304,8 @@ function process_DEProblem(constructor, sys::AbstractODESystem,u0map,parammap;
         p = ps
     end
 
+    @assert length(dvs) == length(u0) "states and initial conditions are of different lengths"
```
YingboMa, line 307: If you check the doc string of `@assert`, you'll see that
```julia
  │ Warning
  │
  │  An assert might be disabled at various optimization levels. Assert should therefore only be used as
  │  a debugging tool and not used for authentication verification (e.g., verifying passwords), nor
  │  should side effects needed for the function to work correctly be used inside of asserts.
```
It's always recommended to use actually explicit throws when checking user errors.

YingboMa: Something like
```julia
length(dvs) == length(u0) || throw(ArgumentError("States $(length(dvs)) and initial conditions $(length(u0)) are of different lengths."))
```
would be good.

Also, could you use `@test_throws` to test this?

pepijndevos: Yea I saw the doc string, but figured this check would mainly be useful for debugging. It seems unlikely that a deployed system would suddenly get a wrong u0. But sure, I'll make it an explicit throw and add some tests.

pepijndevos:
While running the tests I got `LoadError: ArgumentError: Package OrdinaryDiffEq not found in current path` so I'm not sure how that's done on CI, but when adding that to my local system it worked for me.

PR #900, opened 2021-03-16
Implement variable scope

pepijndevos:
In hierarchical systems, it is possible that a subsystem needs to depend on a parameter or state that is from a parent system or even a global. Currently the only way to express this is to manually alias local paramters and states from the outside.

I've been told this feature is analogous to outer variables in Modelica.

I have implemented it by defining a `SymScope` type that is stored in the symbol metadata. The default scope is `LocalScope`, but it is possible to make variables of wider scopes. `GlobalScope(a)` causes `a` to never be namespaced at all, while `ParentScope(a)` makes `a` belong to one level up in the hierarchy. `ParentScope` can be nested to refer to variables higher up the hierarchy. I have added some tests that demonstrate the behavior.

The problem I'm facing is that most calls to `renamespace` don't pass the `Num` or the `Sym` but the _name_ of the `Sym`. This means that the metadata is not available to correctly namespace outer variables, hence I'm marking this as a draft.

In most places this probably comes down to replacing `renamespace(namespace,nameof(x))` with `nameof(renamespace(namespace,x))`, which is a bit involved operation, so I thought it'd be good to post the result so far.

This PR also depends on https://github.com/JuliaSymbolics/SymbolicUtils.jl/pull/240 because it uses `getmetadata(sys, type, default)` which did not work for me.

### DIFF
--- a/src/systems/abstractsystem.jl
+++ b/src/systems/abstractsystem.jl
@@ -268,11 +268,32 @@ function Base.setproperty!(sys::AbstractSystem, prop::Symbol, val)
     end
 end
 
+abstract type SymScope end
+
+struct LocalScope <: SymScope end
+LocalScope(sym::Union{Num, Sym}) = setmetadata(sym, SymScope, LocalScope())
+
+struct ParentScope <: SymScope
+    parent::SymScope
+end
+ParentScope(sym::Union{Num, Sym}) = setmetadata(sym, SymScope, ParentScope(getmetadata(value(sym), SymScope, LocalScope())))
+
+struct GlobalScope <: SymScope end
+GlobalScope(sym::Union{Num, Sym}) = setmetadata(sym, SymScope, GlobalScope())
+
 function renamespace(namespace, x)
     if x isa Num
         renamespace(namespace, value(x))
     elseif x isa Symbolic
-        rename(x, renamespace(namespace, getname(x)))
+        let scope = getmetadata(x, SymScope, LocalScope())
+            if scope isa LocalScope
+                rename(x, renamespace(namespace, getname(x)))
+            elseif scope isa ParentScope
+                setmetadata(x, SymScope, scope.parent)
+            else # GlobalScope
+                x
+            end
+        end
     else
         Symbol(namespace,:₊,x)
     end
--- a/src/systems/abstractsystem.jl
+++ b/src/systems/abstractsystem.jl
@@ -226,14 +226,14 @@ function Base.getproperty(sys::AbstractSystem, name::Symbol; namespace=true)
     i = findfirst(x->getname(x) == name, sts)
 
     if i !== nothing
-        return namespace ? rename(sts[i],renamespace(sysname,name)) : sts[i]
+        return namespace ? renamespace(sysname,sts[i]) : sts[i]
     end
 
     if has_ps(sys)
         ps = get_ps(sys)
         i = findfirst(x->getname(x) == name,ps)
         if i !== nothing
-            return namespace ? rename(ps[i],renamespace(sysname,name)) : ps[i]
+            return namespace ? renamespace(sysname,ps[i]) : ps[i]
         end
     end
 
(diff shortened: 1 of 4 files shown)


pepijndevos:
It seems to me that `rename(O,renamespace(name,nameof(O)))` is the same as `renamespace(name,O))` because `renamespace` on a `Sym` recurs with `Sym{symtype(x)}(renamespace(namespace,nameof(x)))`, and as far as I can tell `rename` is not an in-place update.

I hope this change does not have any side-effects I'm missing. I left the case where it renames a system, which clearly makes sense. And I'm not 100% how observed work, so that one could certainly be wrong.

Other than that the only thing missing from my side is https://github.com/JuliaSymbolics/SymbolicUtils.jl/pull/240 but I'm sure there will be other things to polish before this gets merged.

src/systems/abstractsystem.jl
```diff
@@ -265,13 +265,34 @@ function Base.setproperty!(sys::AbstractSystem, prop::Symbol, val)
...
 
+abstract type SymScope end
+
+struct LocalScope <: SymScope end
+LocalScope(sym::Union{Num, Sym}) = setmetadata(sym, SymScope, LocalScope())
+
+struct ParentScope <: SymScope
+    parent::SymScope
+end
+ParentScope(sym::Union{Num, Sym}) = setmetadata(sym, SymScope, ParentScope(getmetadata(value(sym), SymScope, LocalScope())))
```
YingboMa, line 276: Does this actually work? I thought we typed the keys to be `DataType` not instances.

pepijndevos: They key is a data type, the value is an instance of that data type.

YingboMa: Ah ha, that's right. I missed that.

pepijndevos:
https://github.com/JuliaSymbolics/SymbolicUtils.jl/pull/240 has been merged, so this pr should work now.

YingboMa:
Looks there's a real CI failure.

pepijndevos:
It's broken alright. What I can't figure out is why it ever worked and why my change broke it haha

At this point `u` basically never had a default or initial condition assigned, so something funky happened.
https://github.com/SciML/ModelingToolkit.jl/blob/f1a070be319ce49ecd80c9530e196eb186139b5f/test/symbolic_parameters.jl#L25

But the a bit down the same thing doesn't work. Really strange. Well, this will need more debugging later.

pepijndevos:
Ok I located the problem. It turned out that parameters were silently converted to variables. I think my new implementation fixes that, and should also keep working when #943 lands.

YingboMa (approved):
Let's merge this after #943

pepijndevos:
Well **** now #943 got merged this has conflicts all over the place. I tried resolving them which took a lot of cursing because I had to do it like 5 times before I managed to coerce git and vscode into not committing merge conflicts, and I can't even run the tests because `ERROR: empty intersection between Symbolics@0.1.11 and project compatibility 0.1.14-0.1` no matter what I type in the pkg shell.

@YingboMa it seems you made a ton of changes regarding the preservation of metadata, which is the reason I removed a bunch of renames, but then you *added* a bunch of renames, so I have no idea what's going on anymore. For example, I think
```julia
renamed_op = rename(O,renamespace(name, getname(O)))
```
Will just pass a Symbol to `renamespace` which doesn't have any metadata to tell if it's an outer variable or not.
In my branch I did the following.
```julia
renamed_op = renamespace(name,operation(O))
Term{_symparam(O)}(renamed_op,renamed)
```
I _think_ it needs to be like so, but I can't run the tests so... just going to blindly commit it.
```julia
renamed_op = rename(O,getname(renamespace(name, O)))
```
Actually while I was typing this it seems like there is a CI failure in simplification. Not sure how that could be related to my changes. Oh god this PR is a nightmare.

pepijndevos:
Well, on my local system the simplify test succeeds, but then it fails on some latex test that compares a huge string which looks kinda fragile. No idea what's going on.

Modified a test to namespace a term to see if that works correctly, which it seems to do.

YingboMa:
Yeah, the test failure is unrelated. Could you just update the tests? Some internal handling is changed.

pepijndevos:
What do you want me to update about the tests?

YingboMa:
You can rebase this PR, and tests should pass in master now.

pepijndevos:
Seems like that did the trick!

ChrisRackauckas:
Follow up with docs

pepijndevos:
:partying_face: cool so what about docs? What would be a good place to write about this? (and maybe also the symbolic parameters stuff from earlier)

PR #904, opened 2021-03-16
Point to ColPrac and forums within the docs

pbouffard:
It might be good to make the ColPrac easier to find in the README as well (I missed it, wasn't looking for a badge).


YingboMa:
Thanks!

Issue #918, opened 2021-03-19
Visual model representation and interaction


asprionj:
History cf. #907, for some ideas see [this comment therein](https://github.com/SciML/ModelingToolkit.jl/issues/907#issuecomment-802651432).

It would be desirable to have some sort of visual representation of a model, ideally interactive. Interactivity in a unidirectional sense (from model to visualization) that "browsing" through the model is possible, as well as copying to clipboard paths to state variables, parameters, etc.

The goal is to get the important benefits of a visual representation (such as in Simulink), namely understandig the structure of a complex model, without the hassle of full-fledged (i.e. bidirectional) interaction.

anandijain:
@asprionj https://github.com/SciML/ModelingToolkit.jl/pull/919#issue-596754630 
does this get you part of the way there?

it's not interactive, but should at least give you a way of cleanly viewing states and subsystems

asprionj:
Sure, it's a beginning :wink:
BTW, this issue is not primarily "about me"; it's just an idea that came up which would probably be appreciated by many working with large / complex systems (that they got from someone else and need to understand it).

anandijain:
What's the next step you'd like to see?

asprionj:
This is not a concrete goal with a predefined path. So I think next steps could be:

- Discussing how to get as much benefit with the least amount of work
- Brainstorming ideas for possible solutions / approaches
- Collect actual wishes / requirements / potential features

So I'll try to start a feature list here. One could (and I will, if I find the time) just go through some existing tools, e.g. the [OpenModelica Connection Editor](https://openmodelica.org/doc/OpenModelicaUsersGuide/latest/omedit.html), and sort all features of those tools into the categories below...

However, I'd suggest to stick to a unidirectional information flow (from the model to the visualization) for now. That means also to exclude the notion of a "visual modelling tool"; i.e. having a model-component library and being able to fully define new components in the tool, building up models from components, visually add connections, etc. This is a tremendous amount of work, and IMO it does not add much benefit - all of this can be (at least as) efficiently done in code. Do you agree?

So here's the lists for the features and possible approaches for implementation. I'll add whatever comes up from further discussions.

---

# Features

**Required (base)**
- Visualization of the entire model hierarchy (tree view)
- Visualization of the connections within one system (i.e. hierarchy level)

**Very helpful / important**
- List of state variables and parameters per (sub-)system (like opening a dialog box or something)
- Default values indicated for those variables/parameters for which they're set
[...]

Issue #907, opened 2021-03-17
Display of systems not copyable


baggepinnen:
I find that the printing of types from MTK sometimes adds friction because they are printed in a way that does not allow copy-paste of code. The example below illustrates the problem. Is there any particular reason to print the `₊` instead of a `.`?
```julia
julia> fb
Model fb with 13 equations
States (13):
  u(t)
  loopgain₊y(t)
  loopgain₊u(t)
  y(t)
  loopgain₊C₊u(t)
  loopgain₊C₊y(t)
  loopgain₊nonlinear_P₊u(t)
  loopgain₊nonlinear_P₊y(t)
  loopgain₊C₊x₁(t) [defaults to 0.0]
  loopgain₊C₊x₂(t) [defaults to 0.0]
  loopgain₊nonlinear_P₊P₊u(t)
  loopgain₊nonlinear_P₊P₊y(t)
⋮
Parameters (0):

julia> loopgain₊C₊u
ERROR: UndefVarError: loopgain₊C₊u not defined

julia> loopgain.C.u
loopgain₊C₊u(t)
```

YingboMa:
There's a reason to print `₊`.

What exactly do you want to achieve? Do you want to just copy-paste or to serialize the model.

baggepinnen:
It's related to my post here
https://discourse.julialang.org/t/accessing-variables-in-a-modelingtoolkit-system/57249/8?u=baggepinnen
basically, when I try to figure out how to supply initial conditions for a sub system deeply nested in a structure like
```julia
julia> states(fb)
13-element Vector{Term{Real, Nothing}}:
 u(t)
 loopgain₊y(t)
 loopgain₊u(t)
 y(t)
 loopgain₊C₊u(t)
 loopgain₊C₊y(t)
 loopgain₊nonlinear_P₊u(t)
 loopgain₊nonlinear_P₊y(t)
 loopgain₊C₊x₁(t)
 loopgain₊C₊x₂(t)
 loopgain₊nonlinear_P₊P₊u(t)
 loopgain₊nonlinear_P₊P₊y(t)
 loopgain₊nonlinear_P₊P₊x₁(t)
```
it would be convenient if I could copy the line ` loopgain₊nonlinear_P₊P₊x₁` to my code. Even better would be if I could just supply the initial condition for `P.x₁` without knowing the exact structure of the bigger system `P` is a part of.

YingboMa:
[...]

PR #919, opened 2021-03-19
[wip] pretty print system

anandijain:
addresses https://github.com/SciML/ModelingToolkit.jl/issues/907#issuecomment-802651432 and #918 

provides 
```julia
julia> print_tree(rc_model)
rc_model
├─ resistor
│  ├─ p
│  └─ n
├─ capacitor
│  ├─ p
│  └─ n
├─ source
│  ├─ p
│  └─ n
└─ ground
   └─ g
```


test/print_tree.jl
```diff
@@ -0,0 +1,26 @@
+
+using ModelingToolkit, AbstractTrees, Test
+
+include("components.jl")
+
+open("rc_tree.txt", "w") do io
```
YingboMa, line 6: You can test it like this
```julia
julia> io = IOBuffer()
IOBuffer(data=UInt8[...], readable=true, writable=true, seekable=true, append=false, size=0, maxsize=Inf, ptr=1, mark=-1)

julia> print(io, "hey")

julia> String(take!(io))
"hey"
```

anandijain: thanks i was looking for a good way to do that

PR #920, opened 2021-03-19
Composing `ReactionSystem`s 

isaacsas:
This is a start towards composing `ReactionSystem`s:
1. Composing via derived converted `ODESystem`s works and has a test now (this needs more testing though before advertising). 
2. I added some overloads for `ReactionSystem` and `Reaction` namespacing functions to get around printing issues, and to allow `ReactionSystem`s to have other system types as sub-systems. Please let me know if these are a bad idea and should be removed.
3. One complication that arises with `ReactionSystem`s is that they can generate ODEs where the rhs is zero for connection species, which seems to crash structural simplification during tearing. I added an explicit check to drop generated ODEs where the rhs is zero, but perhaps it would be better to allow users to specify which variables are connection variables in the system, and then filter them out when generating ODE models?
4. I fixed up usage of `states` and `equations` with the appropriate setters in a number of places.

Some things we still need for `ReactionSystem`s: 
1. defaults
2. Right now we can't compose `ReactionSystem`s with `ODESystem`s to store algebraic relations without also changing the `ODESystem` system handling to allow `ODESystem`s to store `ReactionSystem` subsystems. i.e. it seemed like I would have to make the root system be an `ODESystem` for this to work. It seemed undesirable that the root of a collection of `ReactionSystem`s with couplings would need to be an `ODESystem`, so I wasn't quite sure how to proceed.
 
I'm thinking that maybe we should just add an explicit `alg_eqs` field to `ReactionSystem`s that they can use to store coupling equations? Any thoughts? I guess one question then is whether such a field should be merged into calls to `equations` or not. One advantage is that all `ReactionSystem` sub-systems could then be assumed to be `ReactionSystem`s.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@
 *.jl.*.cov
 *.jl.mem
 Manifest.toml
+.vscode
+.vscode/*
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,6 +1,0 @@
-{
-    "cSpell.ignoreWords": [
-        "get",
-        "systems"
-    ]
-}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,6 @@
+{
+    "cSpell.ignoreWords": [
+        "get",
+        "systems"
+    ]
+}
--- a/src/systems/reaction/reactionsystem.jl
+++ b/src/systems/reaction/reactionsystem.jl
@@ -187,7 +187,7 @@ function oderatelaw(rx; combinatoric_ratelaw=true)
 end
 
 function assemble_oderhs(rs; combinatoric_ratelaws=true)
-    sts = states(rs)
+    sts = get_states(rs)
     species_to_idx = Dict((x => i for (i,x) in enumerate(sts)))
     rhsvec         = Any[0 for i in eachindex(sts)]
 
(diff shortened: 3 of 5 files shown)


YingboMa:
> I'm thinking that maybe we should just add an explicit alg_eqs field to ReactionSystems that they can use to store coupling equations? Any thoughts? I guess one question then is whether such a field should be merged into calls to equations or not. One advantage is that all ReactionSystem sub-systems could then be assumed to be ReactionSystems.

That's a good question. We are still in the process of figuring that all. Since `ReactionSystem`s are intrinsically ODEs and all the algebraic equations are from acausal modeling, I can see the appeal of making that a separate field. However, all the connections are going to be reduced away by the tearing algorithm given that there aren't any cycles in the connections. This makes them not special in some way. I think it's cleaner to just keep connection equations in the `equation` field.

YingboMa (approved):
LGTM

isaacsas:
`ReactionSystems` also encode SDEs and jump systems, and hopefully we’ll get some functionality for analyzing the network structure too, so it would be good to separate `Reactions` from constraint equalities. I can use an `ArrayPartition` for the `eqs` as we use for jumps though so it can still be iterated through, then we can keep the Catalyst api `reactions` function to allow access to the underlying `Reaction` vector.

PR #927, opened 2021-03-22
default_u0 / default_p with substituted expressions

### LINKED ISSUE #910: StackOverflow with defaults containing expressions with variables from outer scope
```
using ModelingToolkit
using DifferentialEquations

function Model(; name=gensym(:model))
    @parameters t
    @named soma = MySoma()
    eqs = [
        soma.nc.v ~ soma.v,
        soma.INa ~ soma.nc.I,
    ]
    return ODESystem(eqs,t,name=name,systems=[soma])
end

function MySoma(; name=gensym(:Soma))
    @parameters t
    D = Differential(t)

    @parameters C_m
    @variables v(t), INa(t), I_inj(t)

    @named nc = MyNavChannel()

    eqs = [
        D(v) ~ (-(INa) + I_inj) / C_m,

        I_inj ~ 15.0,

    ]
    defaults = [
        v => -65.1,
        INa => 0.0,
        I_inj => 0.0,
        C_m => 1.0,
    ]

    return ODESystem(eqs,t,[v,INa,I_inj],[C_m],defaults=defaults,name=name,systems=[nc])
end


function MyNavChannel(;name=gensym(:NavC))
    @parameters t
    D = Differential(t)

    @parameters g, vr, E
    @variables v(t), a(t), i(t), I(t)

    dV = v - vr

    α_a = (2.5 - 0.1*dV)/(exp(2.5 - 0.1*dV)- 1)
    β_a = 4/(exp(dV/18))
    a_∞ = α_a / (α_a + β_a)

    α_i = 0.07/(exp(0.05*dV))
    β_i = 1/(exp(3 - 0.1*dV)+ 1)
    i_∞ = α_i / (α_i + β_i)

    I_eq = g * a^3 * i * (v - E)

    eqs = [
        D(a) ~ α_a*(1.0-a) - β_a*a,
        D(i) ~ α_i*(1.0-i) - β_i*i,
        I ~ I_eq, # after up: ~ 0,
    ]
    defaults = [
        #v => -65.1,
        #a => 0.05293248525724958,
        i => 0.5961207535084603,
        I => 0.0,
        g => 120.0,
        vr => -65.0,
        E => 50.0,
        a => a_∞,
    ]
    return ODESystem(eqs,t,defaults=defaults,name=name)
end

# container
@named model = Model()
soma = getproperty(model,:soma,namespace=false)

defs = ModelingToolkit.get_defaults(sys)
u0 = filter(e->!ModelingToolkit.isparameter(e.first),defs)
p = filter(e->ModelingToolkit.isparameter(e.first),defs)

# u0 = ModelingToolkit.get_defaults(sys)
# p = ModelingToolkit.get_default_p(sys)
prob = ODEProblem(sys,collect(u0),(0.0,160.0),collect(p),jac=true)
sol = solve(prob,Rosenbrock23())

# StackOverflowError:
# in recursive_unitless_bottom_eltype at RecursiveArrayTools/69f7F/src/utils.jl:86
...
```
This fails, beacuse of the default `a => a_∞,` inside MyNavChannel(). `a_∞` depends on `v` coming from outside.
It works, if I define a default for `v` inside the channel.

Can this be done without defining a default for `v` inside the channel?

lungd:
Fix #910

AFAIK, you already have some plans for the get_default_(u0|p) functions. These functions will be available again after it is possible to create variables/parameters with `@variables VAR = 1.0`?

Maybe I've overlooked the description for how to get u0 and p I need for the DEProblem initialization but I added a simple function to split the dict coming from `get_defaults(sys)`

### DIFF
--- a/src/systems/abstractsystem.jl
+++ b/src/systems/abstractsystem.jl
@@ -543,3 +543,22 @@ Base.showerror(io::IO, e::InvalidSystemException) = print(io, "InvalidSystemExce
 AbstractTrees.children(sys::ModelingToolkit.AbstractSystem) = ModelingToolkit.get_systems(sys)
 AbstractTrees.printnode(io::IO, sys::ModelingToolkit.AbstractSystem) = print(io, nameof(sys))
 AbstractTrees.nodetype(::ModelingToolkit.AbstractSystem) = ModelingToolkit.AbstractSystem
+
+function get_default_u0(sys::AbstractSystem)
+    defs = get_defaults(sys)
+    u0 = filter(e->!isparameter(e.first),defs)
+    obs = Dict([e.lhs=>e.rhs for e in get_observed(sys)])
+    u0s = Dict([k=>substitute(v, obs) for (k,v) in u0])
+    u0s = Dict([k=>substitute(v, defs) for (k,v) in u0s])
+    return u0s
+end
+
+function get_default_p(sys::AbstractSystem)
+    defs = get_defaults(sys)
+    p = filter(e->isparameter(e.first),defs)
+    obs = Dict([e.lhs=>e.rhs for e in get_observed(sys)])
+    ps = Dict([k=>substitute(v, obs) for (k,v) in p])
+    ps = Dict([k=>substitute(v, defs) for (k,v) in ps])
+    return ps
+end
+
--- a/test/simplify.jl
+++ b/test/simplify.jl
@@ -1,7 +1,7 @@
 using ModelingToolkit
 using ModelingToolkit: value, get_defaults, get_default_u0, get_default_p
 using Test
-using DifferentialEquations
+using OrdinaryDiffEq
 
 @parameters t
 @variables x(t) y(t) z(t)
@@ -69,13 +69,13 @@ _p = filter(e->ModelingToolkit.isparameter(e.first),defs)
 prob = ODEProblem(sys,collect(_u0),(0.0,160.0),collect(_p),jac=true)
 
 # SHOULD NOT HAPPEN; TODO: remove this test
-@test_throws StackOverflowError solve(prob)
+@test_throws StackOverflowError solve(prob,Tsit5())
 
 u0 = get_default_u0(sys)
 p = get_default_p(sys)
 @test isequal(u0, Dict([a=>0.1, sub.y=>0.05]))
 @test isequal(p, Dict())
 prob = ODEProblem(sys,collect(u0),(0.0,160.0),collect(p),jac=true)
-sol = solve(prob)
+sol = solve(prob,Tsit5())
 @test isequal(sol[a], sol[sub.x])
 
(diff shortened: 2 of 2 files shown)


lungd:
I just realized that it is possible to create a DEProblem without explicitly defining u0 and p. They will get extracted from the DESystem automatically. Probably, that's why get_default_u0 is not needed anymore.

The problem still exists. But where to put this code? Separate helper function, inside the DEProblem constructor?

```
function fix_defaults(sys::AbstractSystem)
    defs = get_defaults(sys)
    obs = Dict([e.lhs=>e.rhs for e in get_observed(sys)])
    defs = Dict([k=>substitute(v, obs) for (k,v) in defs])
    defs = Dict([k=>substitute(v, defs) for (k,v) in defs])
    return defs
end
```

ChrisRackauckas:
What's the use case?

lungd:
```
@parameters t
D = Differential(t)
@variables x(t),y(t)
@variables a(t)
@parameters p
@named sub = ODESystem([D(y) ~ p*x],t,[x,y],[p],defaults=Dict(y=>x/2, p=>2.0))
@named model = ODESystem([D(a) ~ 1, sub.x ~ a],t,[a],[],defaults=Dict(a=>0.1),systems=[sub])

defs = get_defaults(sys)
# sub₊y(t) => 0.5sub₊x(t)
# a(t) => 0.100
# sub₊p => 2.00

get_observed(sys)
# sub₊x(t) ~ a(t)
```
The creation of the ODEProblem fails (StackOverflowError) because sys.defaults contain a pair `sub₊y(t) => 0.5sub₊x(t)`
This can be solved by substituting `sub.x` with `a` (based on sys.observed) first and then `a` with the default of `a` (based on sys.defaults)
1) `sub₊y(t) => 0.5sub₊x(t)` -> `sub₊y(t) => 0.5a(t)`
2) `sub₊y(t) => 0.5a(t)` -> `sub₊y(t) => 0.05`

I think this would improve #777?

Please have a look at #910 to see another example.

ChrisRackauckas:
https://github.com/SciML/ModelingToolkit.jl/issues/910 should just use `prob = ODEProblem(sys,Pair[],(0.0,160.0),jac=true)`?

ChrisRackauckas:
More precisely, no where in the docs do we recommend users to deconstruct the defaults themselves like:

```julia
u0 = filter(e->!ModelingToolkit.isparameter(e.first),defs)
p = filter(e->ModelingToolkit.isparameter(e.first),defs)

# u0 = ModelingToolkit.get_defaults(sys)
# p = ModelingToolkit.get_default_p(sys)
prob = ODEProblem(sys,collect(u0),(0.0,160.0),collect(p),jac=true)
```

and I don't think we want to have users do that when it's already handled automatically. So, I don't get who's supposed to use it.

lungd:
> #910 should just use `prob = ODEProblem(sys,Pair[],(0.0,160.0),jac=true)`?

Sorry, my bad. The error pops up when I try to solve the problem not during the construction with ODEProb()
I still get the same error with your code.

I thought I had to manually split the defaults into u0 and p because I didn't know it is possible to just pass sys to ODEProb() while writing the MWE, so this part of my code is obsolete already.

PR #942, opened 2021-03-29
Support bursty reactions

kaandocal:
As pointed out at [MomentClosure.jl](https://github.com/augustinas1/MomentClosure.jl) and [Catalyst.jl](https://github.com/SciML/Catalyst.jl/issues/308), MT does not currently support reactions with stochastic affects (e.g. bursty reactions in which the number of reactants is geometrically distributed). 

I have started working on a simple implementation of such reactions which can take distributions (from Distributions.jl) as stoichiometric coefficients. A demo is below:

```
@parameters t k[1:2] q
@variables A(t)
rxs = [Reaction(k[1], nothing, [A], nothing, [Geometric(q)]),            # 0 -> A
       Reaction(k[2], [A], nothing)]            # B -> 0]
rs = ReactionSystem(rxs,t,[A],[k;b])
```

Conversion to ODE and SDE systems is performed as pointed out in my reply at Catalyst.jl. For Jump systems every time the reaction occurs the stoichiometric coefficients are sampled randomly.

Limitations:
- Requires Statistics.jl for the `mean `and `var` functions
- I overloaded Distributions for Num types; this probably belongs elsewhere
- For Jump systems no RNGs are supported at the moment
- You have to register each distribution with Symbolics, my way of doing this is a bit hacky (related to the point above)
- As of now there is no support for the type T in Reaction (I'm not sure what it does) - I assumed Int for now
- Bursty reactions cannot be MA (limitation of DiffEqJump.jl)
- So far limited testing


src/systems/reaction/reactionsystem.jl
```diff
@@ -333,12 +422,26 @@ function ismassaction(rx, rs; rxvars = get_variables(rx.rate),
...
         (rxvars[i] in stateset) && return false
     end
     return true
 end
 
+function get_affect(rx::Reaction)
+    affect = Vector{Equation}()
+    for (spec,stoich) in rx.netstoich
+        if stoich isa VarStoich
+#            throw("not yet implemented")
```
ChrisRackauckas, line 436: ?

src/systems/reaction/reactionsystem.jl
```diff
@@ -42,40 +42,105 @@ Notes:
...
+sample_negbinomial(r,p) = rand(NegativeBinomial(r,p))
+sample_poisson(λ) = rand(Poisson(λ))
+
+@register sample_bernoulli(p)
+@register sample_betabinomial(n,α,β)
+@register sample_binomial(n,p)
+@register sample_geom(p)
+@register sample_discunif(a,b)
+@register sample_negbinomial(r,p)
+@register sample_poisson(λ)
```
ChrisRackauckas, line 84: We want to just register all of the distributions and `rand` in Symbolics.jl.

ChrisRackauckas: You might be able to loop through the subtypes of Distributions.Sampleable to do this quickly

src/systems/reaction/reactionsystem.jl
kaandocal: Whoopsie, I should have removed that line

isaacsas:
Very nice! On a quick glance this seems allocation heavy via the `get_burst` methods (but perhaps I'm misreading since I only looked quickly). Is there a way to set this up that we can avoid much in the way of additional allocations during system construction and especially conversion (which are already a potential issue on large systems)?

src/systems/reaction/reactionsystem.jl
```diff
@@ -42,40 +42,105 @@ Notes:
...
+        substoich = Vector{Int}()
     end
     if isnothing(prods)
         prods = Vector{Term}()
-        !isnothing(prodstoich) && error("If products are nothing, product stiocihometries have to be so too.")
-        prodstoich = typeof(substoich)()
+        !isnothing(prodstoich) && error("If products are nothing, product stoichiometries have to be so too.")
+        prodstoich = Vector{Union{Int,VarStoich}}()
+    else
+        prodstoich = convert_prodstoich(prodstoich)
```
isaacsas, line 143: If we could set this up to avoid copying over all the product stoichiometry vectors it would be good. We could make the vector type in the array be `Any`, avoiding the copy, but I’m not sure if that would have even worse performance implications.

src/systems/reaction/reactionsystem.jl
```diff
@@ -42,40 +42,105 @@ Notes:
...
+
+function Base.:(*)(coeff::Int, vs::VarStoich) 
+    coeff == 1 && VarStoich(vs.offs, vs.dist)
+    throw("Cannot multiply distributions")
+end 
+
+Statistics.mean(vs::VarStoich) = vs.offs + mean(vs.dist)
+Statistics.var(vs::VarStoich) = var(vs.dist)
+
+sample_bernoulli(p) = rand(Bernoulli(p))
```
isaacsas, line 70: Does this mean you create a new distribution instance anytime a sample is generated in an `affect` call? What about using a functor that stores the distribution instead?

isaacsas:
@ChrisRackauckas i guess one question is whether conversions from distribution to `rand` call is something that should be done here, or lower-level in MT by the system types we convert to (or even `build_function`?).

ChrisRackauckas:
Maybe. In the DiscreteSystem I want to do that. But you can't in ODEs and SDEs without making it an RODE.

src/systems/reaction/reactionsystem.jl
kaandocal: Yes. The overhead should be minimal, but the best way to bypass this would involve storing some distributions/a RNG as auxiliary data somehow...

src/systems/reaction/reactionsystem.jl
kaandocal: > Very nice! On a quick glance this seems allocation heavy via the `get_burst` methods (but perhaps I'm misreading since I only looked quickly). Is there a way to set this up that we can avoid much in the way of additional allocations during system construction and especially conversion (which are already a potential issue on large systems)?

I will try to fix get_bursts. One option in general would be to reinstate the template type T in Reaction without requiring it to be a subtype of Number. That way the code for pre-existing ReactionSystems would effectively not change.

isaacsas: I think dropping the requirement that `T` be a number should be fine.

kaandocal: I've changed it to a Vector{Any} for now and got rid of the superfluous copying. `get_bursts` now works with iterators; as I am not very familiar with the internals of Julia I don't know what the best way to avoid allocations is here save for writing out the loop...

src/systems/reaction/reactionsystem.jl
```diff
@@ -42,41 +42,107 @@ Notes:
...
+struct Reaction{S}
     """The rate function (excluding mass action terms)."""
     rate
     """Reaction substrates."""
     substrates::Vector
     """Reaction products."""
     products::Vector
     """The stoichiometric coefficients of the reactants."""
-    substoich::Vector{T}
+    substoich::Vector{Int}
```
isaacsas, line 104: I think we don't want to restrict to `Int`. For example, a number of users that only care about ODE models have asked us to support non-integer coefficients in the past (and that is something we should get working again eventually -- I thought it worked through `ReactionSystem`s at one point).

src/systems/reaction/reactionsystem.jl
```diff
@@ -42,41 +42,107 @@ Notes:
...
     """The stoichiometric coefficients of the products."""
-    prodstoich::Vector{T}
+    prodstoich::Vector
     """The net stoichiometric coefficients of all species changed by the reaction."""
-    netstoich::Vector{Pair{S,T}}
+    netstoich::Vector{Pair{S, Any}}
```
isaacsas, line 108: I think this should actually be ok to use a union type on like you had before since we construct it generally. We can force users that want to pass it in to make it concrete since it is optional. Sorry if my comment caused confusion!

So maybe we have `substoich` be type `T`, and `netstoich` be a union of `T` and `VarStoich`. I guess with `prodstoich` we should just leave it unspecified for now. This will require some Catalyst updates I think or API functions may break.

src/systems/reaction/reactionsystem.jl
```diff
@@ -247,19 +321,37 @@ end
 
 function assemble_diffusion(rs, noise_scaling; combinatoric_ratelaws=true)
     sts  = get_states(rs)
-    eqs  = Matrix{Any}(undef, length(sts), length(get_eqs(rs)))
+    n_bursts = count(f -> f !== nothing, get_bursts(rs))
```
isaacsas, line 324: Why can't you count on `isbursty` iterating over the reactions?

src/systems/reaction/reactionsystem.jl
```diff
@@ -176,6 +242,12 @@ function equations(sys::ModelingToolkit.ReactionSystem)
     end
 end
 
+isbursty(reac::Reaction) = any(stoich -> stoich isa VarStoich, reac.prodstoich)
+
+get_bursty_eqs(sys::ReactionSystem) = filter(reac -> isbursty(reac), get_eqs(sys))
```
isaacsas, line 247: `Iterators.filter`?

src/systems/reaction/reactionsystem.jl
isaacsas: Ahh, nevermind. It isn't the same as the number of reactions with bursts...

src/systems/reaction/reactionsystem.jl
isaacsas: Probably also should be `equations(sys)` as that will include subsystems.

isaacsas:
I wonder if this might be cleaner if we added a new reaction type, `BurstyReaction`, leaving the old `Reaction` type as is? This makes it easy to know a reaction needs special processing. We could even store them with an `ArrayPartition` as we do for different jump types in `JumpSystem`s. Maybe it wouldn't make a difference, but perhaps it would better for performance when the typical use case will not have bursty reactions?

src/systems/reaction/reactionsystem.jl
isaacsas: Things like `map`, `flatten`, `filter` allocate I believe, while the `Iterators.` versions are lazy. That said, in Julia it is fine to write loops too (that's what is great about the language, loops are fast).

src/systems/reaction/reactionsystem.jl
kaandocal: Any chance one could make `equations(sys)` return an iterator to save on allocations?

src/systems/reaction/reactionsystem.jl
kaandocal: Thanks for the tip! I mostly wanted to put these into functions for encapsulation, if the `Iterators.` versions are too slow we can write out the loops where necessary.

src/systems/reaction/reactionsystem.jl
kaandocal: I can think of two approaches atm:
- `substoich = T; netstoich = Union{T,VarStoich{T}}; prodstoich = netstoich`
- `substoich = T; netstoich = P; prodstoich = netstoich` where `P` defaults to either `T` or `VarStoich` (this would have to be specified in the constructors)

isaacsas:
Closing as this should really be moved to Catalyst now, and it looks like bursty reactions now work with Catalyst's recent support for symbolic stoichiometry (https://github.com/SciML/Catalyst.jl/pull/493). This PR had a cleaner notation in basically allowing random variables, but we can update Catalyst to that in the future or wait till Symbolics/ModelingToolkit begin supporting them in a more official way.

Issue #881, opened 2021-03-13
Use metadata to store "parameter" flag


shashi:
instead of `Symbolic{Parameter{Real}}`.

This will allow us to restrict number methods by default to `Symbolic{<:Number}`

YingboMa:
This could be breaking. We might want to have an overload that makes `Symbolic{Parameter{Real}}(:x)` build `Symbolic{Real}(:x, metadata=Parameter)`

ChrisRackauckas:
I think we should break it though. It's internally breaking, but not necessarily very breaking downstream if we overload it correctly.

PR #943, opened 2021-03-29
Use metadata to denote parameters

shashi:
fixes https://github.com/SciML/ModelingToolkit.jl/issues/881


shashi:
requires https://github.com/JuliaSymbolics/Symbolics.jl/pull/163

src/parameters.jl
```diff
@@ -1,24 +1,35 @@
-import SymbolicUtils: symtype, term
-struct Parameter{T} end
+import SymbolicUtils: symtype, term, hasmetadata
+struct MTKParameterCtx end
 
+isparameter(x::Num) = isparameter(value(x))
+isparameter(x::Symbolic) = hasmetadata(x, MTKParameterCtx) && getmetadata(x, MTKParameterCtx)
```
YingboMa, line 5: Can we do
```suggestion
isparameter(x::Symbolic) = getmetadata(x, MTKParameterCtx, false)
```
?

YingboMa (commented):
Could you update the relavent docs as well?

src/systems/abstractsystem.jl
```diff
@@ -274,7 +274,7 @@ function renamespace(namespace, x)
     elseif istree(x)
         renamespace(namespace, operation(x))(arguments(x)...)
     elseif x isa Sym
-        Sym{symtype(x)}(renamespace(namespace,nameof(x)))
+        @set! x.name = renamespace(namespace,nameof(x))
```
shashi, line 277: The previous behavior was to actually not propagate the metadata... This is the cause of the current test failure I believe.

shashi:
@YingboMa legit ci fail

PR #944, opened 2021-03-30
Add Dockerfile for Codespaces

giordano:


ChrisRackauckas:
Is there any way to do this in https://github.com/SciML/.github and make it add the right package?

giordano:
That's a great question, but apparently the organisation-wide health files are only for the `.github` directory, while Codespaces settings to the `.devcontainer` directory.  There is an open issue about this: https://github.community/t/codespaces-by-organization-not-repository/137072.  They may fix it in the future

PR #948, opened 2021-03-31
forward kwargs to `ODESystem`, for things like `defaults`

anandijain:


isaacsas:
Looks good. We can merge assuming tests pass. Thanks!

PR #951, opened 2021-04-04
change latexify test to make sure -1*x is rendered as -x

catethos:
Originally `Symbolics.jl` renders the negative of a variable as `-1*x`, I have made a PR(https://github.com/JuliaSymbolics/Symbolics.jl/pull/179) so that it can be rendered as `-x`. So I would like to change the tests here to reflect the change.


adamslc:
Tests pass locally with this patch, and the one for `Symbolics.jl`.

PR #953, opened 2021-04-05
Update usage of DestructuredArgs to use inbounds keyword

matthieubulte:
Hi there, this PR implements the changes from JuliaSymbolics/SymbolicUtils.jl#265.

This looks very innocent but was actually a little bit tricky so here are a couple of comments which might help the reviewer:
1. I added the `checkbounds` keyword to functions that use `DestructuredArgs` and have a caller that already used the `checkbounds` keyword. This was done to make sure that the use of `checkounds` was properly propagated. However, this will change the code generated since currently (before changes) the codegen behaves as if `checkbounds=true` while with the propagation of the existing `checkbounds`, the new codegen will have `checkbounds=false`. This simply means that we are adding `@inbounds` to the generated code by default now in functions (and children of funcs) that already had a `checkbounds` keyword.
2. If the call to `DestructuredArgs` was in a function directly exposed to the user, I added the `checkbounds` with default `checkbounds=true` to make sure that the generated code stays the same with the current defaults. The user then is free to manually set `checkbounds=false` for better perfs.
3. If the call to `DestructuredArgs` was far from the user-level functions, and that these user-level functions didn't have a `checkbounds` argument, I didn't change the code. This means that `DestructuredArgs` will take the default value of **not** generating the `@inbounds`, which is slower, but it's the current behavior so it's safer. The reason for that is that I would otherwise have to add many many `checkbounds` in many different places. I could do that if you want, but I thought it might be quite cumbersome and verbose to add this everywhere.

EDIT: I don't know if/how the dependency to `SymbolicUtils.jl` has to be updated, so it's missing from the PR. **Merging this will break master**.


src/systems/diffeqs/abstractodesystem.jl
```diff
@@ -158,6 +158,7 @@ function DiffEqBase.ODEFunction{iip}(sys::AbstractODESystem, dvs = states(sys),
                                      eval_expression = true,
                                      sparse = false, simplify=false,
                                      eval_module = @__MODULE__,
+                                     checkbounds=false,
```
YingboMa, line 161: This is not quite right. If you add `checkbounds` here, then it won't be forwarded to `generate_function` via `kwargs`.

YingboMa (approved):

YingboMa (changes requested):
Oops, I clicked the wrong button

src/systems/diffeqs/abstractodesystem.jl
matthieubulte: My bad, I missed this one. Just added a commit to properly forward `checkbounds`.

YingboMa (approved):

matthieubulte:
Just added an additional commit to have the proper dependency on `SymbolicUtils` v0.10.1

YingboMa:
Please rebase so the tests can pass.

matthieubulte:
Hi, I just rebased on master but there is still is a failing test. I was a little bit confused so I cloned `SciML/ModelingToolkit.jl/master` in a fresh new directory to run the tests, and it seems that the same test fails with the same error.

YingboMa:
I fixed the master branch. Could you rebase again? Thanks!

YingboMa:
The upstream fix is merged. I'll restart the CI in the morning.

ChrisRackauckas:
Awesome!

PR #954, opened 2021-04-06
Update Downstream.yml

anandijain:
add CellMLToolkit


anandijain:
and just the same for Symbolics?

ChrisRackauckas:
Yes.

PR #959, opened 2021-04-09
update create_array signature

shashi:
along with the Symutils version.

let's run CI after a Symutils tag.


YingboMa (approved):

YingboMa:
@KirillZubov could you take a look at the NeuralPDE test failure?

KirillZubov:
> @KirillZubov could you take a look at the NeuralPDE test failure?

ok. I try to figure it out.
After the update, `StochasticTraining, QuadratureTraining, QuasiRandomTraining` gives out "LoadError: Mutating arrays is not supported" when the gradient is taken in reverse mode.

KirillZubov:
I got it. I got it, the problem is in the new release of Adapt.
```julia
using Zygote ,Adapt
f(x,p) = sum(sin.(adapt(typeof(p),x).* p))
x = Float32.([1.,2.])
function testf(p)
    f(x,p)
end
p = [1.5,2.0]
dp1 = Zygote.gradient(testf,p)
ERROR: Mutating arrays is not supported
```
I will try to fix it today

ChrisRackauckas:
Make that `f(x,p) = sum(sin.(adapt(DiffEqBase.parameterless_type(p),x).* p))`

KirillZubov:
will this work with `CuArray`?

ChrisRackauckas:
yes

PR #964, opened 2021-04-12
Fix faulty alias elimination

YingboMa:


src/systems/systemstructure.jl
```diff
@@ -227,7 +227,7 @@ function find_linear_equations(sys)
             c = expand_derivatives(Differential(var)(term), false)
             # test if `var` is linear in `eq`.
             if !(c isa Symbolic) && c isa Number
-                if isinteger(c) && !iszero(c)
+                if (c == 1 || c == -1) && !iszero(c)
```
ChrisRackauckas, line 230: why doesn't this work on other integers?

YingboMa: The algorithm can probably be extended for any integers, but it currently assumes that all coefficients are -1 or 1.

tshort:
Wow, that was some fast bug fixing! :rocket:

PR #966, opened 2021-04-12
Changed constructor to use sparsity_pattern() for sparse matrices

SamuelzXu:
Attempt at closing #871 with help from @YingboMa


ChrisRackauckas (approved):

SamuelzXu:
@ChrisRackauckas how do I approach the SafeTestSets bug?

ChrisRackauckas:
What Safetestsets bug? I don't see one.

The error seems to be that you need to change the matrix from a matrix of booleans to a matrix of numbers.

src/systems/diffeqs/abstractodesystem.jl
```diff
@@ -220,7 +220,7 @@ function DiffEqBase.ODEFunction{iip}(sys::AbstractODESystem, dvs = states(sys),
                      jac = _jac === nothing ? nothing : _jac,
                      tgrad = _tgrad === nothing ? nothing : _tgrad,
                      mass_matrix = _M,
-                     jac_prototype = sparse ? similar(get_jac(sys)[],Float64) : nothing,
+                     jac_prototype = sparse ? jacobian_sparsity(sys) : nothing, # similar(sys.jac[],Float64)
```
ChrisRackauckas, line 223:
```suggestion
                     jac_prototype = sparse ? convert.((eltype(u0),),jacobian_sparsity(sys)) : nothing, # similar(sys.jac[],Float64)
```

ChrisRackauckas: Actually hmm, how do you do that without a `u0` type? I guess you just guess Float64 for now?

YingboMa (changes requested):
This is incorrect. We want use `sparsity_pattern` only when `jac=false` and `sparse=true`.

src/systems/diffeqs/abstractodesystem.jl
```diff
@@ -26,7 +26,7 @@ function calculate_jacobian(sys::AbstractODESystem;
     dvs = states(sys)
 
     if sparse
-        jac = sparsejacobian(rhs, dvs, simplify=simplify)
+        jac = sparsity_pattern(sys)
```
ChrisRackauckas, line 29: This should probably be reverted.

ChrisRackauckas:
I think you can achieve that by just using `jacobian_sparsity(sys)` in the ODEFunction construction? The issue from before was that `get_jac(sys)` would be building the full version just to create the Jacobian, which ended up full.

YingboMa:
You should also promote the `jac_prototype` to have the same `eltype` with `u0` in `ODEProblem` when you have `u0`.

SamuelzXu:
@YingboMa Hopefully this does the trick

PR #968, opened 2021-04-13
Return jacobian_sparsity() when sparse && !jac

### LINKED ISSUE #871: Sparsity pattern without calculating the Jacobian?
@shashi could we easily get support for:

```julia
using OrdinaryDiffEq, ModelingToolkit

const N = 32
const xyd_brusselator = range(0,stop=1,length=N)
brusselator_f(x, y, t) = (((x-0.3)^2 + (y-0.6)^2) <= 0.1^2) * (t >= 1.1) * 5.
limit(a, N) = a == N+1 ? 1 : a == 0 ? N : a
function brusselator_2d_loop(du, u, p, t)
  A, B, alpha, dx = p
  alpha = alpha/dx^2
  @inbounds for I in CartesianIndices((N, N))
    i, j = Tuple(I)
    x, y = xyd_brusselator[I[1]], xyd_brusselator[I[2]]
    ip1, im1, jp1, jm1 = limit(i+1, N), limit(i-1, N), limit(j+1, N), limit(j-1, N)
    du[i,j,1] = alpha*(u[im1,j,1] + u[ip1,j,1] + u[i,jp1,1] + u[i,jm1,1] - 4u[i,j,1]) +
                B + u[i,j,1]^2*u[i,j,2] - (A + 1)*u[i,j,1] + brusselator_f(x, y, t)
    du[i,j,2] = alpha*(u[im1,j,2] + u[ip1,j,2] + u[i,jp1,2] + u[i,jm1,2] - 4u[i,j,2]) +
                A*u[i,j,1] - u[i,j,1]^2*u[i,j,2]
    end
end
p = (3.4, 1., 10., step(xyd_brusselator))

function init_brusselator_2d(xyd)
  N = length(xyd)
  u = zeros(N, N, 2)
  for I in CartesianIndices((N, N))
    x = xyd[I[1]]
    y = xyd[I[2]]
    u[I,1] = 22*(y*(1-y))^(3/2)
    u[I,2] = 27*(x*(1-x))^(3/2)
  end
  u
end
u0 = init_brusselator_2d(xyd_brusselator)
prob = ODEProblem(brusselator_2d_loop,u0,(0.,11.5),p)

sparseprob = ODEProblem(sys,Pair[],(0.,11.5),sparse=true)

using BenchmarkTools
@btime solve(prob,TRBDF2(),save_everystep=false) # 51.714 s (7317 allocations: 70.12 MiB)
@btime solve(sparseprob,TRBDF2(),save_everystep=false)
```

i.e. allow `sparse = true` and then if that's the case, then generate the sparsity pattern even if we don't do the Jacobian. Basically, can we skip `sparsejacobian` and go straight to the pattern?

akashkgarg:
Return jacobian_sparsity() when sparse && !jac. Resolves #871 

At the time of writing this, I realized there's another open PR that addresses this ticket: 
https://github.com/SciML/ModelingToolkit.jl/pull/966


ChrisRackauckas:
I think this is a better solution because it doesn't double compute the sparsity pattern when it doesn't need to. But this doesn't do the last little bit: it needs to upconvert from the `SparseMatrixCSC{Bool}` to the `eltype(u0)` during the construction of the ODEProblem.

akashkgarg:
Ah okay, so we want to add something like `convert.(eltype(u0), jacobian_sparsity(sys))`? I had assumed that if a user wanted the sparsity "pattern" they would expect it in Bool format, but that probably doesn't compose well.

ChrisRackauckas:
The differential equation solver just makes a copy of the `jac_prototype`, so it will error when it tries to fill it with floats. This is why it's called "prototype": it's a prototype of what it's going to copy, not just the pattern itself. Though in theory the differential equation solver could do the conversion in its setup phase: that would be another way to fix it.

YingboMa:
Yeah, so something like `similar(f.jac_prototype, uEltype)`.

akashkgarg:
Okay, fixed the typing issue in latest push.

YingboMa:
No, this's not right. It will error when a user just call ODEFunction.

akashkgarg:
@YingboMa Could you expand on your comment? I tried running tests in ModelingToolkit and all of them pass except for some in latexify.jl, but I think those are unrelated to this change? It's likely I'm missing something or we need to add another test that captures the failure case you're thinking of?

src/systems/diffeqs/abstractodesystem.jl
```diff
@@ -215,12 +215,13 @@ function DiffEqBase.ODEFunction{iip}(sys::AbstractODESystem, dvs = states(sys),
         end
     end
 
+    uElType = eltype(u0)
```
ChrisRackauckas, line 218: I think this isn't safe for the `u0 == nothing` case.

ChrisRackauckas:
>No, this's not right. It will error when a user just call ODEFunction.

I don't think it will?

ChrisRackauckas:
This is missing a test.

YingboMa:
> I don't think it will?

It will fail when u0 isn't passed into ODEFunction.

ChrisRackauckas:
Yes, that's what I meant by "I think this isn't safe for the u0 == nothing case.". Is the solution here to add a convert in each of the solvers? Use `SetField.jl` in ODEProblem? Etc.

YingboMa:
Handling this in the solvers' level would be great. @akashkgarg do you want to file a PR to OrdinaryDiffEq, StochasticDiffEq, and Sundials?

akashkgarg:
Okay, I updated the code to explicitly check for the case when `u0` is `nothing` and also added some tests. I can take a stab at modifying the solvers as well.

YingboMa:
It might be easier to overload `get_concrete_problem` in https://github.com/SciML/DiffEqBase.jl/blob/7824bd8cec9d98ff6122acef99fe80d7dee8dddb/src/solve.jl#L138-L192

Note that not all `jac_prototype`s have `similar` defined, so you need to check `isa AbstractArray` before calling `similar(jac_prototype, eltype(u0))` and such, e.g.
```julia
julia> using DiffEqOperators, LinearAlgebra

julia> A = 0.01*rand(3, 3)
3×3 Matrix{Float64}:
 0.00423074   0.00891582  0.00839771
 0.00382897   0.00291319  0.00918495
 0.000318021  0.00235121  0.0080333

julia> rn = (du, u, p, t) -> begin
           mul!(du, A, u)
       end
#1 (generic function with 1 method)

julia> u0 = rand(3)
3-element Vector{Float64}:
 0.6075074514491132
 0.8949122350449714
 0.8993517833428093

julia> A = JacVecOperator{Float64}(rn, u0; autodiff=false);

julia> similar(A, Int)
ERROR: MethodError: no method matching similar(::JacVecOperator{Float64, var"#1#2", Vector{Float64}, Vector{Float64}, Vector{Float64}, Any, Any, Bool}, ::Type{Int64})
Closest candidates are:
  similar(::Union{Adjoint{T, var"#s832"}, Transpose{T, var"#s832"}} where {T, var"#s832"<:(AbstractVector{T} where T)}, ::Type{T}) where T at /Users/scheme/.bin/julia-release/usr/share/julia/stdlib/v1.6/LinearAlgebra/src/adjtrans.jl:230
  similar(::Union{Adjoint{T, S}, Transpose{T, S}} where {T, S}, ::Type{T}) where T at /Users/scheme/.bin/julia-release/usr/share/julia/stdlib/v1.6/LinearAlgebra/src/adjtrans.jl:234
  similar(::Union{Adjoint{T, S}, Transpose{T, S}} where {T, S}, ::Type{T}, ::Tuple{Vararg{Int64, N}}) where {T, N} at /Users/scheme/.bin/julia-release/usr/share/julia/stdlib/v1.6/LinearAlgebra/src/adjtrans.jl:235
  ...
Stacktrace:
 [1] top-level scope
   @ REPL[6]:1

julia> A isa AbstractArray
false
```

ChrisRackauckas:
Yes, adding that to `get_concrete_problem` and using SetField.jl could be the right way to do this.

akashkgarg:
I took a stab at this here: SciML/DiffEqBase.jl#657 . Not sure if it catches all the cases, but as far as I can tell, promote_f gets called before any solve that happens related to ODEFunction. Let me know if something like this is what you had in mind or if there might be a better solution.

ChrisRackauckas:
There seems to still be a test failure here. Could you take a look at it?

https://github.com/SciML/ModelingToolkit.jl/pull/968/checks?check_run_id=2362637867#step:6:370

akashkgarg:
Weird. These tests pass locally for me. Probably some weird tolerance checking with floating point.

ChrisRackauckas:
Increasing the tolerance a bit is probably fine.

test/jacobiansparsity.jl
```diff
@@ -0,0 +1,73 @@
...
+end
+
+u0 = init_brusselator_2d(xyd_brusselator)
+prob_ode_brusselator_2d = ODEProblem(brusselator_2d_loop,
+                                        u0,(0.,11.5),p)
+sys = modelingtoolkitize(prob_ode_brusselator_2d)
+
+# test sparse jacobian pattern only.
+prob = ODEProblem(sys, u0, (0, 11.5), sparse=true, jac=false)
+@test isapprox(similar(Symbolics.jacobian_sparsity(map(x->x.rhs, equations(sys)), states(sys)),Float64), prob.f.jac_prototype; atol=1.e-4)
```
YingboMa, line 43: What does this do? Are you comparing uninitialize memory with another piece of uninitialized memory?

YingboMa (changes requested):
The tests are wrong.

test/jacobiansparsity.jl
ChrisRackauckas: Yes, for reference `similar` is uninitialized memory, so its values are essentially random, which is why this check fails. I think you should check `zero(A)` instead of `similar(A)`, or use the `(I,J,V) = findnz(A)` and compare the `I` and `J`.

YingboMa: Since it's `jac_prototype`, we should only check the sparsity pattern, but not the values.

akashkgarg: Hmm, I did not realize that `similar` returns an uninitialized array. Shouldn't we be using `convert` in this case? My understanding is that you get the sparsity pattern using`jacobian_sparsity(sys)`, which is an array of `bool`, but we want to make sure the type matches that of `u0`. Seems like `convert` is what we need over `similar`?

ChrisRackauckas: `convert.(Float64,x)` would guarantee it to be 1's and 0's, and so would `float.`, though for a test directly checking `I` and `J` might be safer.

akashkgarg:
It seems to me that for `jac_prototype` all we care about is the sparsity pattern? If so, is there ever a reason to return `similar(get_jac(sys)[],uElType))`? Can we simply just always return `similar(jacobian_sparsity(sys),uElType)` when `sparse = true`?

YingboMa:
We should not use `get_jac(sys)` but just call `J = calculate_jacobian(sys)`. And to get the sparsity pattern, we just do `similar(J, uElType)`. Since if we have the Jacobian already, we might as well just generate the prototype from the symbolic Jacobian.

ChrisRackauckas:
Awesome, that one had a bit more too it than I had anticipated.

PR #973, opened 2021-04-16
Add connectors

YingboMa:


ChrisRackauckas:
Needs a downstream PR to NeuralPDE

Issue #526, opened 2020-07-17
PDESystem boundary specification


ChrisRackauckas:
@thomasgibson @sandreza @simonbyrne let's discuss how the PDESystem specification needs to change in order to accommodate the types of meshes that are used for Clima. 

@KirillZubov @emmanuellujan should stay informed here, since we will need to make sure the PINN and finite difference methods update for any change.

The current PDE system is like:

```julia
@parameters x y θ
@variables u(..)
@derivatives Dxx''~x
@derivatives Dyy''~y

# 2D PDE
eq  = Dxx(u(x,y,θ)) + Dyy(u(x,y,θ)) ~ -sin(pi*x)*sin(pi*y)

# Boundary conditions
bcs = [u(0,y) ~ 0.f0, u(1,y) ~ -sin(pi*1)*sin(pi*y),
       u(x,0) ~ 0.f0, u(x,1) ~ -sin(pi*x)*sin(pi*1)]
# Space and time domains
domains = [x ∈ IntervalDomain(0.0,1.0),
                   y ∈ IntervalDomain(0.0,1.0)]

pde_system = PDESystem(eq,bcs,domains,[x,y],[u])
```

One of the difficulties that I saw was that, in this form, it's "easy" to know how to identify the boundaries and write a condition to them. However, a breakdown occurs on harder domains. An even simpler example than unstructured meshes is `CircleDomain`, we we can no long write things like `u(1,x) = ... for all x`, since we now have to say things like `u(x,y) = ... for all x^2 + y^2 = 1` which is a fundamentally implicit specification, whereas it can be made explicit only by directly knowing a representation of the boundary.

So it looks like we need to be able to query for boundaries of domains and specify the BCs based on them. For example, maybe something like:

```julia
domains = [x ∈ IntervalDomain(0.0,1.0),
                   y ∈ IntervalDomain(0.0,1.0)]

bcs = [boundary(domains[1])[1] => 0.f0, 
           boundary(domains[1])[2] => -sin(pi*1)*sin(pi*y),
           boundary(domains[2])[1] => 0.f0,
           boundary(domains[2])[2] => -sin(pi*x)*sin(pi*1)]
```

That looks a bit awful though, but gets the point across. But the real question is:

1. Does this solve the issues with BCs?
2. Is there a nicer way to get the same information?

ChrisRackauckas:
Probably good to add @kburns here as well,

kburns:
[...]

PR #976, opened 2021-04-21
domain deprecations

tinosulzer:
Start implementing the DomainSets.jl interface as discussed in #526 

Questions:
- What to do about the following warnings:
```
WARNING: using DomainSets.islinear in module ModelingToolkit conflicts with an existing identifier.
WARNING: using DomainSets.jacobian in module ModelingToolkit conflicts with an existing identifier.
WARNING: using DomainSets.gradient in module ModelingToolkit conflicts with an existing identifier.
```
- What to do with the existing `IntervalDomain`, `ProductDomain`, `CircleDomain`? Looks like only `IntervalDomain` is used (but it is used a lot).
- Specifying polar/spherical coordinates in `DomainSets.jl`?
- How to access the equivalent of `lower` and `upper` in domains defined by `DomainSets.jl`?
- Should users have to import DomainSets to access the domains or should they be exported by MTK?


tinosulzer:
Should the package export (at least) `Interval`? Or should we always require a `using DomainSets` import?

tinosulzer:
This is ready for review, just waiting to know how to fix the failing test (export `Interval` or import `DomainsSets`)

ChrisRackauckas:
https://github.com/JuliaApproximation/DomainSets.jl/blob/master/src/DomainSets.jl#L88

DomainSets.jl's exports are very unclean and isn't up to standard. We should not `@reexport` it, and inside of our packages we should only do `import DomainSets` as otherwise having a bunch of things named `matrix`, `vector`, `∂`, etc. is going to be a mess down the road (I think this is worth making an issue about).

Also, I think the domain support needs to be in Symbolics.jl to be used with integrals?

tinosulzer:
I'll open a PR in `Symbolics.jl`. What are your thoughts on automatically constructing an `Interval` domain from a tuple to avoid having to explicitly import `DomainSets` in each script just for the domain definition?
```julia
Base.:∈(variable::SymbolicUtils.Sym,domain::NTuple{2,Real}) = VarDomainPairing(variable,Interval(domain...))
```

ChrisRackauckas:
>What are your thoughts on automatically constructing an Interval domain from a tuple to avoid having to explicitly import DomainSets in each script just for the domain definition?

I like that, it's a good shorthand for the common case.

Project.toml
```diff
@@ -14,6 +14,7 @@ DiffRules = "b552c78f-8df3-52c6-915a-8e097449b14b"
 Distributed = "8ba89e20-285c-5b6f-9357-94700520ee1b"
 Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
 DocStringExtensions = "ffbed154-4ef7-542d-bbb7-c09d3a79fcae"
+DomainSets = "5b8099bc-c8ec-5219-889f-1d9e522a28bf"
```
YingboMa, line 17: Add compat, too.

tinosulzer: should this be done manually or with compathelper?

ChrisRackauckas:
The patch for that has been released.

tinosulzer:
Ok perfect. What's the fix for the integration test in Neural PDE? The error is because `domain.lower` has been replaced by `infimum(domain)`

ChrisRackauckas:
That's rather unfortunate. Can we do a type piracy getproperty overload to handle the deprecation? Double check that domains don't already have a getproperty and make it just on `Interval` and we should be okay. It's definitely not the nicest thing but it'll be a much easier upgrade path.

src/domains.jl
```diff
@@ -1,4 +1,15 @@
...
 
 @deprecate IntervalDomain(a,b) Interval(a,b)
-@deprecate CircleDomain() Ball()
\ No newline at end of file
+@deprecate CircleDomain() Ball()
+
+# type piracy on Interval for downstream compatibility to be reverted once upgrade is complete
+function Base.getproperty(domain::Interval, sym::Symbol)
+    if sym === :lower
+        return infimum(domain) # or domain.left also defined by IntervalSets.jl
```
ChrisRackauckas, line 9: throw a warning?

tinosulzer: since `domain.left` and `domain.right` are already defined do we want to recommend using that or infimum/supremum? At this level I don't think we care about open vs closed sets?

ChrisRackauckas: we should recommend `infimum(domain)` etc. as the upgrade path.

ChrisRackauckas (approved):

PR #978, opened 2021-04-22
Round-trip system printing

YingboMa:


src/ModelingToolkit.jl
```diff
@@ -154,6 +154,7 @@ export ode_order_lowering, liouville_transform
 export runge_kutta_discretize
 export PDESystem
 export Reaction, ReactionSystem, ismassaction, oderatelaw, jumpratelaw
+export system2expr
```
ChrisRackauckas, line 157: ehh, too cute of a name? system_to_expr? `toexpr(::System)`?

src/systems/abstractsystem.jl
```diff
@@ -411,6 +411,98 @@ function (f::AbstractSysToExpr)(O)
...
+    var2name = Dict{Any,Symbol}()
+    for v in Iterators.flatten((sts, ps))
+        var2name[v] = getname(v)
+    end
+
+    eqs_name = push_eqs!(stmt, equations(sys), var2name)
+    defs_name = push_defaults!(stmt, defaults(sys), var2name)
+
+    if sys isa ODESystem
+        push!(stmt, :($ODESystem($eqs_name, $ivname, $stsname, $psname; defaults=$defs_name)))
```
ChrisRackauckas, line 498: seems like this machinery needs to dispatch per system?

src/ModelingToolkit.jl
YingboMa: Yeah, overloading `toexpr` is better.

src/systems/abstractsystem.jl
YingboMa: Yeah, I plan to do that when I clean up the code.

ChrisRackauckas:
Needs a test, but it does look good to go with a name change. And yeah, it's going to need to be refactored to each system type, etc., but that can come later.

YingboMa:
There are tests for ODESystem and NonlinearSystem.

ChrisRackauckas (approved):

ChrisRackauckas:
oh missed that.

PR #982, opened 2021-04-23
Add system promotion (toodesystem)

### LINKED ISSUE #817: System promotion
When we combine a nonlinear system and an ODE system, we need to change the independent variable of the nonlinear system before concatenating.

YingboMa:
Close #817

This PR adds the `toodesystem` function which converts a NonlinearSystem to an ODESystem or converts an ODESystem to a new ODESystem with a different independent variable.


src/systems/abstractsystem.jl
```diff
@@ -737,3 +737,40 @@ end
...
 end
+
+###
+### system promotion
+###
+
+# We have a stand-alone function to convert a `NonlinearSystem` or `ODESystem`
+# to an `ODESystem` to connect systems, and we later can reply on
+# `structural_simplify` to convert `ODESystem`s to `NonlinearSystem`s.
+function toodesystem(sys, t; name=nameof(sys))
```
ChrisRackauckas, line 748: we should name this more generally, `promote(ODESystem,sys, args...)` or something? Since we will want to promote ODESystem to SDESystem and everything too.

YingboMa: No, we cannot. `promote` is a generic julia function. Maybe `convert_system`?

ChrisRackauckas: `convert_system` works.

PR #984, opened 2021-04-23
start of using Systems with io ops

anandijain:
could use a little help though.

reading back into system isn't working

### DIFF
--- a/src/ModelingToolkit.jl
+++ b/src/ModelingToolkit.jl
@@ -53,6 +53,7 @@ import Symbolics: rename, get_variables!, _solve, hessian_sparsity,
 import DiffEqBase: @add_kwonly
 
 import LightGraphs: SimpleDiGraph, add_edge!
+import Base.write, Base.read
 
 using Requires
 
@@ -181,5 +182,6 @@ export build_function
 export modelingtoolkitize
 export @variables, @parameters
 export @named, @nonamespace
+export read, write
 
 end # module
--- a/src/ModelingToolkit.jl
+++ b/src/ModelingToolkit.jl
@@ -53,7 +53,6 @@ import Symbolics: rename, get_variables!, _solve, hessian_sparsity,
 import DiffEqBase: @add_kwonly
 
 import LightGraphs: SimpleDiGraph, add_edge!
-import Base.write, Base.read
 
 using Requires
 
--- a/src/ModelingToolkit.jl
+++ b/src/ModelingToolkit.jl
@@ -182,6 +182,5 @@ export build_function
 export modelingtoolkitize
 export @variables, @parameters
 export @named, @nonamespace
-export read, write
 
 end # module
--- a/src/systems/abstractsystem.jl
+++ b/src/systems/abstractsystem.jl
@@ -739,5 +739,5 @@ function connect(syss...)
 end
 
 # idk where to put them 
-Base.write(io::IO, sys::T) where {T <: ModelingToolkit.AbstractODESystem} = write(io, string(toexpr(sys)))
-Base.read(io::IO, sys::T) where {T <: ModelingToolkit.AbstractODESystem} = eval(Meta.parse(read(io, String)))
+Base.write(io::IO, sys::T) where {T <: AbstractODESystem} = write(io, string(toexpr(sys)))
+Base.read(io::IO, sys::T) where {T <: AbstractODESystem} = eval(Meta.parse(read(io, String)))
--- a/src/systems/abstractsystem.jl
+++ b/src/systems/abstractsystem.jl
@@ -739,5 +739,5 @@ function connect(syss...)
 end
 
 # idk where to put them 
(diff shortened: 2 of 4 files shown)


anandijain:
FYI I did see the serialization testset, but I figure that we will also want easy IO usage too.

anandijain:
for a little context the usage I'm aiming for is `write("rc.jl", rc_model)` and `read("rc.jl", ODESystem)`

src/ModelingToolkit.jl
```diff
@@ -53,6 +53,7 @@ import Symbolics: rename, get_variables!, _solve, hessian_sparsity,
 import DiffEqBase: @add_kwonly
 
 import LightGraphs: SimpleDiGraph, add_edge!
+import Base.write, Base.read
```
YingboMa, line 56: We shouldn't import it when we just overload it in one place.

src/ModelingToolkit.jl
```diff
@@ -181,5 +182,6 @@ export build_function
 export modelingtoolkitize
 export @variables, @parameters
 export @named, @nonamespace
+export read, write
```
YingboMa, line 185: The are already in Base, so we shouldn't export it.

src/systems/abstractsystem.jl
```diff
@@ -737,3 +737,7 @@ end
 function connect(syss...)
     connect(promote_connect_type(map(get_connection_type, syss)...), syss...)
 end
+
+# idk where to put them 
+write(io::IO, sys::T) where {T <: ModelingToolkit.AbstractODESystem} = write(io, string(toexpr(sys)))
```
YingboMa, line 742: Just write `Base.write`.

src/ModelingToolkit.jl
anandijain: What is the proper way to do this? REPL told me to import, so I did

src/ModelingToolkit.jl
anandijain: Thanks

test/systemio.jl
```diff
@@ -0,0 +1,16 @@
...
+
+# I can't get buffer to work with read
+# io = IOBuffer() 
+
+open("rc.jl", "w") do io 
+    write(io, rc_model)
+end
+
+open("rc.jl", "r") do io 
+    ser = read(io, ODESystem)
```
YingboMa, line 13: Does this even work? You are declaring `ser` in an anonymous function.

anandijain: no I mentioned it doesn't. How should I do this?

src/ModelingToolkit.jl
YingboMa: Just don't do this. Write `Base.write`.

src/systems/abstractsystem.jl
anandijain: kk

anandijain:
also what file should this go in? abstractsystem.jl okay?

src/systems/abstractsystem.jl
```diff
@@ -737,3 +737,7 @@ end
 function connect(syss...)
     connect(promote_connect_type(map(get_connection_type, syss)...), syss...)
 end
+
+# idk where to put them 
+Base.write(io::IO, sys::T) where {T <: ModelingToolkit.AbstractODESystem} = write(io, string(toexpr(sys)))
```
YingboMa, line 742: You don't need to write `ModelingToolkit.AbstractODESystem` when this is already in `ModelingToolkit`...

anandijain: haha kk

YingboMa: Also, `sys:: AbstractSystem` is enough.

test/systemio.jl
anandijain: and by this, I don't mean just moving ser outside haha I mean getting `write("rc.jl", rc_model)` and `read("rc.jl", ODESystem)` to work

src/systems/abstractsystem.jl
```diff
@@ -737,3 +737,7 @@ end
 function connect(syss...)
     connect(promote_connect_type(map(get_connection_type, syss)...), syss...)
 end
+
+# idk where to put them 
+Base.write(io::IO, sys::T) where {T <: AbstractODESystem} = write(io, string(toexpr(sys)))
```
YingboMa, line 742: `sys::AbstractSystem`

YingboMa: Also, do you need `string` to `write`?

anandijain: So `toexpr` works for all Systems? Nice!

YingboMa: No, it doesn't. But it just doesn't make sense to constraint the type in here.

anandijain:
yeah what is the right way to do `read` cuz that looks wrong

test/rc.jl
```diff
@@ -0,0 +1,8 @@
+begin
```
anandijain, line 1: ill remove this in the actual PR

PR #989, opened 2021-04-26
Fixes for symbolic arrays change

shashi:
Running MTK off of https://github.com/JuliaSymbolics/Symbolics.jl/pull/197


src/systems/diffeqs/odesystem.jl
```diff
@@ -86,8 +86,8 @@ function ODESystem(
                    connection_type=nothing,
                   )
     iv′ = value(iv)
-    dvs′ = value.(dvs)
-    ps′ = value.(ps)
+    dvs′ = value.(collect(dvs))
+    ps′ = value.(collect(ps))
```
YingboMa, line 90: Does `collect` work recursively?

shashi: no

YingboMa: Then how would `collect(dvs)` work for `[A[1:3] x]`?

shashi: ideally `[A[1:3]; x]` should be a `vcat` node but we don't have that yet. I would imagine Julia generics will take care of making that into an array and collect will be a no-op.

YingboMa: Collect on an array is a copy though. Maybe call `collect` only when `dvs` is a symbolic array?

src/parameters.jl
```diff
@@ -21,6 +31,16 @@ Maps the variable to a state.
 tovar(s::Symbolic) = setmetadata(s, MTKParameterCtx, false)
 tovar(s::Num) = Num(tovar(value(s)))
 
+function recurse_and_apply(f, x)
```
YingboMa, line 34: Looks like this function is never called.

src/parameters.jl
```diff
@@ -10,7 +10,17 @@ isparameter(x) = false
...
 Maps the variable to a paramter.
 """
-toparam(s::Symbolic) = setmetadata(s, MTKParameterCtx, true)
+function toparam(s)
+    if s isa Symbolics.Arr
+        Symbolics.wrap(toparam(Symbolics.unwrap(s)))
+    elseif s isa AbstractArray
+        map(toparam, s)
+    elseif symtype(s) <: AbstractArray
+        Symbolics.recurse_and_apply(toparam, s)
```
YingboMa, line 19: Is it moved to Symbolics?

shashi: no we copied it here.

shashi: but forgot to use the copy lol

YingboMa: Maybe change it?

YingboMa:
DataDrivenDiffEq needs a separate update. It tries to use implementation details to handle array variables.

PR #1017, opened 2021-05-20
support symbolic parameters at problem level

### LINKED ISSUE #991: Simulation fails when using parameters in the initial value map
Hi, 

Version:
[961ee093] ModelingToolkit v5.16.0

I have the following Lotka Volterra model:
```
#= Lotka-Volterra, John Tinnerholm=#
using ModelingToolkit    
using DiffEqBase    
using DifferentialEquations    
function LotkaVolterraSimulate(tspan = (0.0, 100.0))                
  pars = @parameters(begin             
                     alpha
                     x0
                     gamma
                     y0
                     delta
                     beta            
                     t
                     end)        
  vars = @variables(begin             
                    y(t)
                    x(t)                    
                    end)        
  der = Differential(t)        
  eqs = [der(y) ~ (delta * x - gamma) * y, der(x) ~ (alpha - beta * y) * x]        
  nonLinearSystem =
    ODESystem(eqs, t, vars, pars, name = :($(Symbol("LotkaVolterra"))))        
  pars = [
    alpha => 0.1,
    x0 => 10.0,
    gamma => 0.4,
    y0 => 10.0,
    delta => 0.02,
    beta => 0.02,
    t => tspan[1],
  ]        
  initialValues = [
    y => 10.0,
    x => 10.0
  ]        
  firstOrderSystem = ModelingToolkit.ode_order_lowering(nonLinearSystem)        
  reducedSystem = dae_index_lowering(firstOrderSystem)        
  problem = ODEProblem(reducedSystem, initialValues, tspan, pars)
  return problem
end
```

This works and simulates just fine! 
image

However, if I use the parameters for the initial values that is 
```
  initialValues = [
    y => x0,
    x => y0
  ]        
```
and running it with:

`sol = solve(LotkaVolterraSimulate(), Tsit5())
`
I get the following error: 

```
ERROR: LoadError: StackOverflowError:
Stacktrace:
 [1] recursive_unitless_bottom_eltype(a::Type{Any}) (repeats 79984 times)
   @ RecursiveArrayTools ~\.julia\packages\RecursiveArrayTools\K3lTR\src\utils.jl:86
in expression starting at C:<Path>\LotkaVolterra.jl:40
```

Is this intended or a bug? 

Cheers,
John

pepijndevos:
fix #991

### DIFF
--- a/src/systems/diffeqs/abstractodesystem.jl
+++ b/src/systems/diffeqs/abstractodesystem.jl
@@ -401,17 +401,19 @@ function process_DEProblem(constructor, sys::AbstractODESystem,u0map,parammap;
         u0defs = merge(parammap, defs)
     elseif eltype(parammap) <: Pair
         u0defs = merge(Dict(parammap), defs)
-    elseif parammap isa SciMLBase.NullParameters
-        u0defs = defs
-    else
+    elseif eltype(parammap) <: Number
         u0defs = merge(Dict(zip(ps, parammap)), defs)
+    else
+        u0defs = defs
     end
     if u0map isa Dict
         pdefs = merge(u0map, defs)
     elseif eltype(u0map) <: Pair
         pdefs = merge(Dict(u0map), defs)
-    else
+    elseif eltype(u0map) <: Number
         pdefs = merge(Dict(zip(dvs, u0map)), defs)
+    else
+        pdefs = defs
     end
 
     u0 = varmap_to_vars(u0map,dvs; defaults=u0defs)
--- a/src/systems/diffeqs/abstractodesystem.jl
+++ b/src/systems/diffeqs/abstractodesystem.jl
@@ -401,6 +401,8 @@ function process_DEProblem(constructor, sys::AbstractODESystem,u0map,parammap;
         u0defs = merge(parammap, defs)
     elseif eltype(parammap) <: Pair
         u0defs = merge(Dict(parammap), defs)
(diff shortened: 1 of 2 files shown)


pepijndevos:
Yea uh that doesn't really work for initial conditions passed as an array.

pepijndevos:
I mean you could just turn them into a map, but I'm not convinced this is the right approach.

ChrisRackauckas:
oh the issue is that the initial conditions are an array not a vector?

pepijndevos:
Well no just that my quick fix only works when `parammap` amd `u0map` are a dict or list of pairs, rather than a flat array.

pepijndevos:
If you set initial conditions to `[10]` rather than `[x0 => 10]` you can't really pass that as a dict to defaults.
Of course you can do something like `Dict(zip(states(sys), parammap)` but seems like a bit ugly, so we should first answer the question if this is the correct approach.

ChrisRackauckas:
isa Dict || eltype isa Pair?

pepijndevos:
```julia
eltype(parammap)
Pair{Num, Float64}
eltype(parammap) isa Pair
false
```

ChrisRackauckas:
`eltype(parammap) <: Pair`

pepijndevos:
huh, I guess I never really thought about the exact difference between `isa` and `<:`. Well anyway I think it should work now.

PR #1021, opened 2021-05-24
add kwarg to disable checking number of equations and states agree

isaacsas:
CC @YingboMa

### DIFF
--- a/src/systems/abstractsystem.jl
+++ b/src/systems/abstractsystem.jl
@@ -670,13 +670,13 @@ function check_eqs_u0(eqs, dvs, u0; check_length=true, kwargs...)
     if u0 !== nothing
         if check_length
             if !(length(eqs) == length(dvs) == length(u0))
-                throw(ArgumentError("Equations ($(length(eqs))), states ($(length(dvs))), and initial conditions ($(length(u0))) are of different lengths. \nTo allow a different number of equations than states use kwarg check_length=false."))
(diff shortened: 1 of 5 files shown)


src/systems/abstractsystem.jl
```diff
@@ -665,15 +665,17 @@ AbstractTrees.children(sys::ModelingToolkit.AbstractSystem) = ModelingToolkit.ge
 AbstractTrees.printnode(io::IO, sys::ModelingToolkit.AbstractSystem) = print(io, nameof(sys))
 AbstractTrees.nodetype(::ModelingToolkit.AbstractSystem) = ModelingToolkit.AbstractSystem
 
-function check_eqs_u0(eqs, dvs, u0)
+function check_eqs_u0(eqs, dvs, u0; check_length=true, kwargs...)
```
YingboMa, line 668: Cannot we just do `check_length && return` in the first line of this function?

isaacsas: Seemed like it still makes sense to check the length of `u0` is the length of the `states`, or are there cases where those could disagree?

YingboMa: Ah, cool.

isaacsas:
@YingboMa ok to make a new release? (Or can you?) I'll then update ReactionNetworkImporters.

YingboMa:
Thanks! I just registered a new release.

isaacsas:
Great, thanks!

PR #1022, opened 2021-05-24
Have ODEs from reactions include zero ODEs by default

isaacsas:
@YingboMa unfortunately `calculate_massmatrix` was also not happy with non-square systems, so I've just dropped a bunch of kwargs into `ReactionSystem`s and changed the default behavior to generate ODEs with zero rhs. These are needed as including zero ODEs breaks composing of reaction-based ODE systems based on how `structural_simplify` works (since one now gets equations like `dA/dt~0` and `0 ~ A + B` for `A` a connection variable, so I had to add a kwarg to disable the generation of zero ODEs for such use cases). This is a pretty ugly solution overall, but I'm not sure what else can be done without changing around the mass matrix calculation to avoid mass-matrix generation for a non-square system where it is not needed.

I've tested now that with these changes ReactionNetworkImporters tests seem to pass locally.

### DIFF
--- a/src/systems/reaction/reactionsystem.jl
+++ b/src/systems/reaction/reactionsystem.jl
@@ -185,7 +185,7 @@ end
 """
     oderatelaw(rx; combinatoric_ratelaw=true)
 
-Given a [`Reaction`](@ref), return the reaction rate law [`Operation`](@ref) used in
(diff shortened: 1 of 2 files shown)


YingboMa (approved):
This is a bit unfortunate, but it this makes downstream work, then let's do it.

YingboMa:
New version is registered.

PR #1028, opened 2021-06-05
Put observed in a wrapper type to allow dispatching

ChrisRackauckas:


src/systems/diffeqs/abstractodesystem.jl
```diff
@@ -139,6 +139,18 @@ for F in [:ODEFunction, :DAEFunction]
     end
 end
 
+struct GeneratedObserved{S}
+    sys::S
+    dict::Dict
```
YingboMa, line 144: add a checkbounds::Bool field here.

PR #1036, opened 2021-06-09
Compute Jacobian sparsity when u0 is not provided

YingboMa:


ChrisRackauckas (approved):
oh, haha

YingboMa:
The Jacobian caching code is also very wrong.

src/systems/diffeqs/abstractodesystem.jl
```diff
@@ -31,7 +34,7 @@ function calculate_jacobian(sys::AbstractODESystem;
         jac = jacobian(rhs, dvs, simplify=simplify)
     end
 
-    get_jac(sys)[] = jac  # cache Jacobian
+    get_jac(sys)[] = jac, (sparse, simplify) # cache Jacobian
```
ChrisRackauckas, line 37: do the same for nonlinear systems?

YingboMa: Nonlinear systems don't cache stuff yet.

ChrisRackauckas: you're two days late https://github.com/SciML/ModelingToolkit.jl/pull/1030 😅

YingboMa: That code never uses the cache, i.e. it writes into the cache but never reads it.

ChrisRackauckas: it uses it because the constructors in abstractodesystem are used in problem construction. It actually threw an error before because it was trying to fill a cache that didn't exist.

PR #1040, opened 2021-06-11
[WIP] Local Identifiability Algorithm via Power Series

iliailmer:
This is a work in progress on the algorithm for local identifiability of ODE parameters.

I am mainly working on building a power series solution first as it is crucial for the algorithm to work.

Once this is done, I will add the actual identifiability determination step.


ChrisRackauckas:
Is there a reason to not just copy the power series solution from https://github.com/siravan/SymPertExamples.jl ? @shahriariravanian

Project.toml
```diff
@@ -24,6 +24,7 @@ LightGraphs = "093fc24a-ae57-5d10-9952-331d41423f4d"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 MacroTools = "1914dd2f-81c6-5fcd-8719-6d5c9610ff09"
 NaNMath = "77ba4419-2d1f-58cd-9bb1-8ffee604a2e3"
+Nemo = "2edaba10-b0f1-5616-af89-8c11ac63239a"
```
YingboMa, line 27: Do we need Nemo or only a subset of it? Also remember to add compat.

Project.toml
```diff
@@ -70,7 +71,7 @@ Setfield = "0.7"
 SpecialFunctions = "0.7, 0.8, 0.9, 0.10, 1.0"
 StaticArrays = "0.10, 0.11, 0.12, 1.0"
 SymbolicUtils = "0.11.0"
-Symbolics = "0.1.21"
+Symbolics = "0.1.21, 1.0"
```
YingboMa, line 74: No, don't add this in this PR.

iliailmer: missed it when experimenting with compatibility, this is removed now.

Project.toml
iliailmer: I will, yes. So far, I am using all of it, will reduce to a subset later.

Issue #1041, opened 2021-06-11
`ControlSystem` without defined loss function


cadojo:
I typically use many different loss functions for the same control system / model. Defining the loss function once for each `ControlSystem` is a bit limiting, because I can't define one common `System` (model) for a bunch of analysis cases. 

For reference, I've extended `ModelingToolkit` to [`PolynomialGTM`](https://github.com/cadojo/PolynomialGTM). Right now,  I'm representing the model aircraft dynamics as an `ODESystem`. It would be great to instead represent this model as a `ControlSystem` because I could then separate states and controls. But I have no common loss function I'd want to use as a default. I'd prefer not to make a factory function, since lots of controls analysis is not optimal control analysis.

Would changing `ControlSystem` to `OptimalControlSystem`, and then introducing a new `struct ControlSystem <: AbstractControlSystem` be appropriate here? The new `ControlSystem` type would have all of the same fields, except for the loss function. The `generate_jacobian` and `calculate_jacobian` functions would essentially produce the symbolic "A" matrix (the homogeneous dynamics), and we could add functions to generate the full symbolic state space model pretty easily.

This is a breaking change, obviously feel free to throw this out. I do think that non-optimal control problem modeling would be a valuable addition to `ModelingToolkit.jl`, and I'm definitely interested in helping out with that.

ChrisRackauckas:
You're on the right track. I was going to talk to @YingboMa about modifying every every system to allow a control variable. I don't think control needs to be separate from `ODESystem`, but instead should be a part of it. And then we need to think about how to do optimal control as just a loss function on top of that, instead of duplicating system code.

ChrisRackauckas:
ControlSystem was removed. OptimziationSystem as it grows should just cover this.

PR #1059, opened 2021-06-18
Control parameter specification for `AbstractODESystem`, `DiscreteSystem`

cadojo:
Adds control parameter specification, related to #1041. Also adds a control Jacobian cached reference and an associated function. If the `calculate_jacobian` function generates a state space model `A` matrix, then the new `calculate_control_jacobian` function calculates the `B` matrix. 

@ChrisRackauckas @YingboMa what do you think about this design? __Thoughts before I proceed? Should I proceed?__ The rules are...

1. Control parameters are specified with a `kwarg` to `ODESystem`, and are empty by default
2. Control parameters __must__ be some subset of the `parameters` of a system (I think this makes sense, since we don't want the control parameters to have their own dynamics)
3. Users can call `calculate_control_jacobian` to calculate the Jacobian of the system's dynamics _with respect to the control parameters_, __not__ with respect to the state variables as is the case in `calculate_jacobian`

It would be nice to add a `generate_statespace` function to produce expressions for calculating the tuple `A,B`. That way, users can easily linearize an `ODESystem` about any operating point.

### DIFF
--- a/src/ModelingToolkit.jl
+++ b/src/ModelingToolkit.jl
@@ -161,11 +161,13 @@ export Differential, expand_derivatives, @derivatives
 export Equation, ConstrainedEquation
 export Term, Sym
 export SymScope, LocalScope, ParentScope, GlobalScope
-export independent_variable, states, parameters, equations, controls, observed, structure
+export independent_variable, states, parameters, equations, controls, observed, structure, defaults
+export ssmodel, linearize
 export structural_simplify
 export DiscreteSystem, DiscreteProblem
 
 export calculate_jacobian, generate_jacobian, generate_function
+export calculate_control_jacobian
 export calculate_tgrad, generate_tgrad
 export calculate_gradient, generate_gradient
 export calculate_factorized_W, generate_factorized_W
--- a/src/ModelingToolkit.jl
+++ b/src/ModelingToolkit.jl
@@ -161,7 +161,7 @@ export Differential, expand_derivatives, @derivatives
 export Equation, ConstrainedEquation
 export Term, Sym
 export SymScope, LocalScope, ParentScope, GlobalScope
-export independent_variable, states, parameters, equations, controls, observed, structure, defaults
+export independent_variable, states, parameters, equations, controls, observed, structure
 export structural_simplify
 export DiscreteSystem, DiscreteProblem
 
--- a/src/ModelingToolkit.jl
+++ b/src/ModelingToolkit.jl
@@ -167,7 +167,7 @@ export structural_simplify
 export DiscreteSystem, DiscreteProblem
 
 export calculate_jacobian, generate_jacobian, generate_function
-export calculate_control_jacobian
+export calculate_control_jacobian, generate_control_jacobian
 export calculate_tgrad, generate_tgrad
 export calculate_gradient, generate_gradient
 export calculate_factorized_W, generate_factorized_W
--- a/src/ModelingToolkit.jl
+++ b/src/ModelingToolkit.jl
@@ -162,7 +162,6 @@ export Equation, ConstrainedEquation
 export Term, Sym
 export SymScope, LocalScope, ParentScope, GlobalScope
 export independent_variable, states, parameters, equations, controls, observed, structure, defaults
-export ssmodel, linearize
 export structural_simplify
 export DiscreteSystem, DiscreteProblem
 
--- a/src/systems/abstractsystem.jl
+++ b/src/systems/abstractsystem.jl
@@ -36,6 +36,18 @@ function calculate_jacobian end
 
 """
 ```julia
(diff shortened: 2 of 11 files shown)


src/systems/abstractsystem.jl
```diff
@@ -346,11 +360,17 @@ function states(sys::AbstractSystem)
...
 end
+
 function parameters(sys::AbstractSystem)
     ps = get_ps(sys)
     systems = get_systems(sys)
     isempty(systems) ? ps : [ps;reduce(vcat,namespace_parameters.(systems))]
 end
+
+function controls(sys::AbstractSystem)
+    get_ctrl(sys)
```
ChrisRackauckas, line 371: this should return the flattened form.

cadojo: I changed the `get_ctrl(sys)` line to `get_ctrl(flatten(sys))`, is this what you had in mind?

cadojo: Nevermind, I'm copying the format for `states(...)`, etc.

ChrisRackauckas:
Analogues of this same functionality should be added to SDESystem and DiscreteSystem.

cadojo:
Added a simple spring-mass-damper test for the control Jacobian, which should pass. Is there anything else y'all would recommend for this PR? I'm thinking a future PR could add the capability of replacing the control parameters with functions in `ODEFunction` so you can simulate control laws, but I won't be able to help with that until August or so.

ChrisRackauckas:
I think this looks good, though note CI is failing. @YingboMa what do you think of the design?

test/odesystem.jl
```diff
@@ -370,13 +370,18 @@ let
...
     eqs = [δ(x) ~ ẋ, δ(ẋ) ~ f - k*x - d*ẋ]
-    sys = ODESystem(eqs, t, [x, ẋ], [f, d, k])
+    sys = ODESystem(eqs, t, [x, ẋ], [f, d, k]; controls = [f])
 
     calculate_control_jacobian(sys)
 
     @test isequal(
-        sys.ctrl_jac,
-        Num[0, 1]
+        ModelingToolkit.get_ctrl_jac(sys)[][1],
```
YingboMa, line 378: Why not test the return of the function but test the internal state?

cadojo: My understanding was `get_ctrl_jac` directly accesses the internal state, even if the internal state hasn't been set yet (e.g. `calculate_control_jacobian` hasn't yet been called). 

When I left `sys.ctrl_jac`, it left warnings about deprecation. Something like: `sys.name has been deprecated, use get_... instead`.

I can change it back to `sys.ctrl_jac` if you prefer. My reasoning was `get_ctrl_jac` would do the same thing, and wouldn't clog the test log output with a deprecation warning.

YingboMa: That accesses the internal state. How about just
```julia
jac = calculate_control_jacobian(sys)
```
and test that.

cadojo:
I don't think the `DataDrivenDiffEq` integration test failure is related to this PR. Could someone confirm?

ChrisRackauckas:
Yes, unrelated. That's the Symbolic arrays stuff.

cadojo:
I've added the control input to the `DiscreteSystem` test, hopefully that meets the code coverage requirements.

src/ModelingToolkit.jl
```diff
@@ -161,11 +161,12 @@ export Differential, expand_derivatives, @derivatives
 export Equation, ConstrainedEquation
 export Term, Sym
 export SymScope, LocalScope, ParentScope, GlobalScope
-export independent_variable, states, parameters, equations, controls, observed, structure
+export independent_variable, states, parameters, equations, controls, observed, structure, defaults
```
ChrisRackauckas, line 164: I don't think `defaults` should be exported.

ChrisRackauckas:
What happens with the controls when the ODEProblem is built? Looks like nothing?

ChrisRackauckas:
I guess let's move one step at a time. The `ODEProblem` step needs a solution to https://github.com/SciML/ModelingToolkit.jl/issues/1088 .

ChrisRackauckas:
Now if we do this, we should put a deprecation warning on ControlSystem and remove it.

src/ModelingToolkit.jl
cadojo: Fixed with [47c0039](https://github.com/SciML/ModelingToolkit.jl/pull/1059/commits/47c00394bde77b3034693c1a48326cc70ffbefc3).

cadojo:
> Now if we do this, we should put a deprecation warning on ControlSystem and remove it.

Should this be done in this PR?

ChrisRackauckas:
Do it in the next.

PR #1060, opened 2021-06-19
Bound Documenter

YingboMa:


ChrisRackauckas:
We need 0.27.

YingboMa:
But we cannot use it if it doesn't build.

ChrisRackauckas:
It's almost better that way then 😆 . That's kind of a sad place to be. @pfitzseb could you help us?

Issue #141, opened 2019-05-29
Differential Equation Function Transformations


ChrisRackauckas:
Basic things like http://www.math.mcgill.ca/jakobson/courses/ma261/lecture5.pdf

This can really help enforce conditions like non-negativity since if your space doesn't allow it, then the numerical solver can guarantee it won't happen if it solves for `exp(x)` instead of `x`.

stevengogogo:
> it won't happen if it solves for exp(x) instead of x

Do you have **a sample equation** that has the problem you mentioned and can be benefited from this method?

ChrisRackauckas:
https://www.radford.edu/~thompson/vodef90web/problems/demosnodislin/Demos_MOL/DemoPolymer/nonnegative.pdf

The one on the bottom of page 4 should be good.

PR #1073, opened 2021-06-23
[WIP] Differential Equation Function Transformations #141

SteffenPL:
For https://github.com/SciML/ModelingToolkit.jl/issues/141
I'm not quite done yet, but I wanted to ask for feedback to check that the direction is alright.

For example, the change of variables `z = exp(x)` looks currently like this:
```
eqs = [D(x) ~ α*x]
u0 = [x => 1.0]
sys = ODESystem(eqs; defaults=u0)

@variables z(t)
forward_subs  = [exp(x) => z]
backward_subs = [x => log(z)]
new_sys = changeofvariables(sys, forward_subs, backward_subs)
@test equations(new_sys)[1] == (D(z) ~ α*z*log(z))
```
Since there is not symbolic way (AFAIK) to get the inverse of general nonlinear function, one has to
provide both: the forward and backward substitution (`forward_subs` and `backward_subs`).

In the test file, I also added a Riccati ODE and a 3D linear transformation as examples.

Details:
- If `sys.defaults` exists, then those will be transformed as well.
- The substituted variables become observables.
- Transformation can depend on time. (For transformation if initial data, one needs to provide `t0`)
- It only works for first-order ODEs

Questions:
- Is the function interface ok? I picked this form, since it looks mathematically correct. I use expressions instead of `Function`s because everything else is also symbolic.

----

Things I will add for final PR/known bugs:
- [ ] support cases where `length(forward_subs) < length(states(sys))`, i.e. keep old state variables.
- [ ] add proper doc string once the interface is fixed.

Optional ideas:
- [ ] compute forward/backward substitution automatically if possible (but I don't know where to get inveres from). One could use inverse function theorem, but then the transformed ODEs are ugly/too complex.
- [ ] allow to pass known invariants, to cover cases like Riccati equations better.

### DIFF
--- a/src/ModelingToolkit.jl
+++ b/src/ModelingToolkit.jl
@@ -153,7 +153,7 @@ export JumpProblem, DiscreteProblem
 export NonlinearSystem, OptimizationSystem
 export ControlSystem
 export alias_elimination, flatten, connect, @connector
-export ode_order_lowering, liouville_transform
+export ode_order_lowering, liouville_transform, changeofvariables
 export runge_kutta_discretize
 export PDESystem
 export Reaction, ReactionSystem, ismassaction, oderatelaw, jumpratelaw
--- a/src/systems/diffeqs/basic_transformations.jl
+++ b/src/systems/diffeqs/basic_transformations.jl
@@ -54,3 +54,85 @@ function liouville_transform(sys)
       vars = [states(sys);trJ]
       ODESystem(neweqs,t,vars,parameters(sys))
 end
+
+
+
+
+
+
+
+
+"""
+$(TYPEDSIGNATURES)
+
+Generates the set of ODEs after change of variables.
+
+
+Example:
+
+```julia
+using ModelingToolkit, OrdinaryDiffEq, Test
+
+@parameters t α
+@variables x(t)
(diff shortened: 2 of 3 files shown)


ChrisRackauckas:
>Since there is not symbolic way (AFAIK) to get the inverse of general nonlinear functions, one has to
provide both: the forward and backward substitution (forward_subs and backward_subs).

I think that's fine for now.

EricTulowetzke:
@ChrisRackauckas and @SteffenPL are we wanting to get this PR finished? Do either of have issues with me picking this up? I'm looking for a good first issue to get my feet wet.

ChrisRackauckas:
It's gone stale. Feel free to take this on yourself either using this as a starting point or from scratch.

SteffenPL:
Hi, sorry for my inactivity. It wasn't really my focus for a while. I can work on it in the next week. Initially a issue in symbolics was halting this issue. Maybe we can directly communicate, my name email is plunder.steffen.2a@kyoto-u.ac.jp.

SteffenPL:
Maybe to add more details.  The main reason why I didn't ask for merging was that SDRs and DAEs were not supported at the time.

PR #1092, opened 2021-07-02
Variable scope: don't lose metadata by passing around bare symbols

pepijndevos:
I ran into *another* instance where the code takes a symbolic value, namespaces it, takes the name (discarding metadata) and uses that to rename the original symbol (with the old metadata).

This was not caught in the tests because the tests use `renamespace` directly and mostly used `Sym` rather than `Term` which trigger this code path. I've added a testcase for that now.

I seem to recall there was some test failure in a completely different test, so while I can't see why, I'm semi-expecting a CI failure. Let's see...

### DIFF
--- a/src/systems/abstractsystem.jl
+++ b/src/systems/abstractsystem.jl
@@ -345,7 +345,7 @@ function namespace_expr(O,name,iv) where {T}
     if istree(O)
         renamed = map(a->namespace_expr(a,name,iv), arguments(O))
         if operation(O) isa Sym
-            rename(O,getname(renamespace(name, O)))
+            renamespace(name, O)
         else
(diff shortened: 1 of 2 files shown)


pepijndevos:
Yeah that's the CI error I was expecting. Except I can't remember where it happened before and what I did to solve it. @YingboMa you remember by any chance?

I guess I need to figure out how to reproduce this testcase locally... bleg

pepijndevos:
This failure does not seem to be cause by my changes. I checked out  DataDrivenDiffEq.jl locally and it gives this test failure without my changes.

ChrisRackauckas:
That's the symbolic array issues. Don't worry about DataDriven's failure.

PR #1097, opened 2021-07-04
Fix second example in README.md

mayl:
I'm not very familiar with either `Julia` or `ModelingToolkit`, but copy pasting the example as it currently stands leads to the following error:

```
ArgumentError: Variable a is not a function of independent variable t.
```

Making the above change seems to resolve it.


ChrisRackauckas (approved):

ChrisRackauckas:
Thanks, yes was a typo.

PR #1116, opened 2021-07-13
add fix for extending a Nonlinear system with no independent variables

### LINKED ISSUE #1115: Extending a nonlinear system with no iv
Attempting to extend a nonlinear system with no iv leads to an error as the convert_system call made in uses the iv which has a value of nothing:

```
function extend(sys::AbstractSystem, basesys::AbstractSystem; name::Symbol=nameof(sys))
    T = SciMLBase.parameterless_type(basesys)
    iv = independent_variable(basesys)
    # iv is nothing here
    sys = convert_system(T, sys, iv)
    ... 
```
Adding a check for iv being nothing and calling convert_system without iv fixes the problem:
```
function extend(sys::AbstractSystem, basesys::AbstractSystem; name::Symbol=nameof(sys))
    T = SciMLBase.parameterless_type(basesys)
    iv = independent_variable(basesys)
    if iv === nothing
        sys = convert_system(T, sys)
    else
        sys = convert_system(T, sys, iv)
    end
```

stonepreston:
fixes #1115

### DIFF
--- a/src/systems/abstractsystem.jl
+++ b/src/systems/abstractsystem.jl
@@ -826,7 +826,12 @@ by default.
 function extend(sys::AbstractSystem, basesys::AbstractSystem; name::Symbol=nameof(sys))
     T = SciMLBase.parameterless_type(basesys)
     iv = independent_variable(basesys)
-    sys = convert_system(T, sys, iv)
+    if iv === nothing
+        sys = convert_system(T, sys)
+    else
+        sys = convert_system(T, sys, iv)
+    end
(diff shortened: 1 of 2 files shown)


YingboMa:
Nice! Thanks a lot!

lamorton:
`independent_variable` is being deprecated. The preferred way to handle this situation would be to use the new `independent_variables`.

YingboMa:
It's not deprecated before the PR is merged.

Issue #1088, opened 2021-07-01
Parameter representation for controls


ChrisRackauckas:
The underlying problem is that we need to represent the generated `f(du,u,p,t)` in a way where `p` can hold the necessary parameters while also being a valid object for the optimal control. Without that, https://github.com/SciML/ModelingToolkit.jl/pull/1059 will not be able to generate an ODEProblem. But what kind of ODEProblem should it be generating? This issue is to specify the interface for the forward simulation and describe why that needs to be the case.

## Proposed code

In a modified version of https://diffeqflux.sciml.ai/dev/examples/optimal_control/ where `dx[1]` also has a parameter, i.e.

```julia
function dxdt_(dx,x,p,t)
    x1, x2 = x
    dx[1] = p.p[1] * x[2]
    dx[2] = p.ctrl[1](t)^3
end
```

i.e. the parameters are held in some kind of struct:

```julia
ParameterController <: AbstractParameterData
  p::P
  ctrl::C
end
```

In this form, we need that each `ctrl` is a `ctrl(t)` which internally holds its parameterization. Then we would just build it like:

```julia
ODEProblem(sys,[x1 => ...], [p1 => ...],ctrls = [ctrl1 => DataInterpolations.ConstantInterpolation(data,data_t)])
```

Sounds good? Okay now the hard part.

## Parameter Representation for Optimal Control

Now we want to optimize `p`. That can mean multiple things: optimize `p.p`, optimize `p.ctrl`, or optimize `p` in full. How do we make this happen? Well we can have `p` itself act as a vector. If `trainctrl == false`, then `p[i] = p.p[i]`. If `trainp == false` then `p[i] == ...?`. So let's talk about that part.

We need to make the parameter interface recursive. `ctrl` should be something that follows the interface that it acts like a parameter vector, so that `ctrl[i]` is something that acts like a parameter vector, so that `ctrl[i][j]` is something that acts like a parameter vector. If that's the case, we just map `i` to continue over each next parameter vector.

[...]

PR #1119, opened 2021-07-13
Use Difference operator for discrete system

sharanry:
https://github.com/SciML/ModelingToolkit.jl/issues/1088#issuecomment-876649631

TODO:

- [x] Add utilities which enable handling Difference operators.
- [x] Add difference function generators
- [x] Add tests

### DIFF
--- a/src/systems/discrete_system/discrete_system.jl
+++ b/src/systems/discrete_system/discrete_system.jl
@@ -164,3 +164,33 @@ function collect_difference_variables(sys::DiscreteSystem)
     end
     return difference_vars
 end
+
+@noinline function throw_invalid_difference(difvar, eq)
+    msg = "The difference variable must be isolated to the left-hand " *
+    "side of the equation like `$difvar ~ ...`.\n Got $eq."
+    throw(InvalidSystemException(msg))
+end
+
+function check_difference_variables(eq, expr=eq.rhs)
+    istree(expr) || return nothing
+    if operation(expr) isa Difference
+        throw_invalid_difference(expr, eq)
+    end
+    foreach(Base.Fix1(check_difference_variables, eq), arguments(expr))
+end
+
+function generate_function(
+        sys::DiscreteSystem, dvs = states(sys), ps = parameters(sys);
+        kwargs...
+    )
+    eqs = equations(sys)
+    foreach(check_difference_variables, eqs)
+    # substitute x(t) by just x
+    rhss = [isdifference(eq.lhs) ? arguments(eq.lhs)[1] + eq.rhs : eq.rhs for eq in eqs]
+
+    u = map(x->time_varying_as_func(value(x), sys), dvs)
+    p = map(x->time_varying_as_func(value(x), sys), ps)
+    t = get_iv(sys)
+    
+    build_function(rhss, u, p, t; kwargs...)
+end
--- a/src/systems/discrete_system/discrete_system.jl
+++ b/src/systems/discrete_system/discrete_system.jl
@@ -186,7 +186,8 @@ function generate_function(
     eqs = equations(sys)
     foreach(check_difference_variables, eqs)
     # substitute x(t) by just x
-    rhss = [isdifference(eq.lhs) ? arguments(eq.lhs)[1] + eq.rhs : eq.rhs for eq in eqs]
+    # rhss = [isdifference(eq.lhs) ? arguments(eq.lhs)[1] + eq.rhs : eq.rhs for eq in eqs]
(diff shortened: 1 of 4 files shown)


sharanry:
@ChrisRackauckas @YingboMa The simple SIR test now works fine Difference operator. Is there any other more complex scenarios I can look at and support or is this sufficient for this initial PR?

sharanry:
Also, I think currently FunctionMap defaults `dt=1` [here](https://github.com/SciML/OrdinaryDiffEq.jl/blob/ba2a3a841614182a75c37e9e48b2c9ee82c60a0a/src/solve.jl#L26). I guess we will have to change that?

sharanry:
One other thing we are assuming right now is that we have a common difference operator on all states with equal timesteps. Do we intend to support multi-timestep processes?

ChrisRackauckas:
>Also, I think currently FunctionMap defaults to dt=1 here. I guess we will have to change that?

Yes.

src/systems/discrete_system/discrete_system.jl
```diff
@@ -115,8 +115,82 @@ function DiffEqBase.DiscreteProblem(sys::DiscreteSystem,u0map,tspan,
     u = dvs
     p = varmap_to_vars(parammap,ps)
 
-    f_gen = build_function(rhss, dvs, ps, t; expression=Val{eval_expression}, expression_module=eval_module)
-    f_oop,f_iip = (@RuntimeGeneratedFunction(eval_module, ex) for ex in f_gen)
+    f_gen = generate_function(sys; expression=Val{eval_expression}, expression_module=eval_module)
+    f_oop, _ = (@RuntimeGeneratedFunction(eval_module, ex) for ex in f_gen)
     f(u,p,t) = f_oop(u,p,t)
```
YingboMa, line 120: Shouldn't there also be an in-place version as well?

src/systems/discrete_system/discrete_system.jl
```diff
@@ -115,8 +115,82 @@ function DiffEqBase.DiscreteProblem(sys::DiscreteSystem,u0map,tspan,
...
     DiscreteProblem(f,u0,tspan,p;kwargs...)
 end
+
+isdifference(expr) = istree(expr) && operation(expr) isa Difference
+isdifferenceeq(eq) = isdifference(eq.lhs)
+
+difference_vars(x::Sym) = Set([x])
```
YingboMa, line 127: Do you really need to copy-paste the `vars` code? What did you change?

src/systems/discrete_system/discrete_system.jl
```diff
@@ -115,8 +115,82 @@ function DiffEqBase.DiscreteProblem(sys::DiscreteSystem,u0map,tspan,
...
+
+    symtype(operation(O)) <: FnType && push!(difference_vars, O)
+    for arg in arguments(O)
+        difference_vars!(difference_vars, arg)
+    end
+
+    return difference_vars
+end
+
+function collect_difference_variables(sys::DiscreteSystem)
```
YingboMa, line 153: Let's try not duplicate too much code. This code could be easily reused by making it a higher order function and pass in `isdifference`.

YingboMa (changes requested):
The code copy-pasting seems easily avoidable.

src/systems/discrete_system/discrete_system.jl
sharanry: Yes. It does get generated. But not sure if we have any use for it in discrete systems as even solvers like `FunctionMap` uses the not-in-place function (not sure what it is called 😅 ).

sharanry:
Yup. I will try and refactor the code.

sharanry:
@YingboMa @ChrisRackauckas Does the ODESystem assume some kind of ordering of the equations in the differential system? It looks like the `generate_function` for ODESystem does not take into consideration what the LHS of each equation is. I also noticed [this TODO comment](https://github.com/SciML/ModelingToolkit.jl/blob/master/src/systems/diffeqs/abstractodesystem.jl#L112).

YingboMa (approved):
LGTM

YingboMa:
Oh, remember to also generalize the `vars!` function.

sharanry:
> Discrete system doesn't require it. Will add it in the difference operator in ODE system PR on which I have started a discussion [here](https://github.com/SciML/ModelingToolkit.jl/issues/1122)?

YingboMa:
Yes, handling `vars!` will be required for that issue, but let's do that in a future PR.

Issue #1122, opened 2021-07-15
Allowing discrete Difference equations in ODESystem 


sharanry:
The [proposed plan](https://github.com/SciML/ModelingToolkit.jl/issues/1088#issuecomment-876649631) is to support difference equations containing `DIfference` operator in `ODESystem`s using `PeriodicCallback`s.

This could involve the following workflow:

Difining difference and differential equations together:
```julia
using ModelingToolkit
@parameters t σ ρ β
@variables x(t) y(t) z(t)
δ = Differential(t)
D = Difference(t; dt=0.01)

eqs = [
       δ(x) ~ σ*(y-x),
       δ(y) ~ x*(ρ-z)-y,
       δ(z) ~ x*y - β*z,
       D(y) ~ x
       ]
```

Manually mentioning which are the difference variables while defining the ODESystem using an optional kwarg:
```julia
sys = ODESystem(eqs,t,[x,y,z],[σ,ρ,β]; differencevars=[y])
```

Producing the ODEProblem and the callback
```julia
prob = ODEProblem(sys,u0,tspan,p,jac=true)
difference_cb = generate_difference_cb(sys)
```

Mention the callback while solving the problem:
```julia
solve(prob, Rodas5(); cb=difference_cb)
```

@ChrisRackauckas @YingboMa 
Does this kind of workflow seem reasonable to begin with? 
OR do we want more automation that might involve API changes? 
OR did you have something else in mind?

ChrisRackauckas:
No, just put the callback in the problem. The user shouldn't have to deal with it. And automatically detect the difference equations and move them from eqs.

PR #1126, opened 2021-07-16
Enable usage of difference operator in ODESystem

sharanry:
Issue: https://github.com/SciML/ModelingToolkit.jl/issues/1122
Initial discussion: https://github.com/SciML/ModelingToolkit.jl/issues/1088#issuecomment-876649631

TODO:
- [x] Add tests

@ChrisRackauckas @YingboMa


sharanry:
Looking into the failed tests

sharanry:
The [following tests](https://github.com/SciML/ModelingToolkit.jl/blob/d3f46aac7ac4adae32fde1dfc6cf385eb5a84a32/test/odesystem.jl#L233) seem to be failing consistently due to what seems to be numerical errors since I have added support for the `Difference` operator to ODESystem. What’s weird is that it is failing for tests that do not even contain any difference equations.  So I am taking a closer look -- but if you have any thoughts on what might be wrong, do let me know.

sharanry:
@YingboMa @ChrisRackauckas This is ready for a review. I think the failing tests are unrelated.

src/systems/diffeqs/abstractodesystem.jl
```diff
@@ -90,7 +90,7 @@ function generate_function(
     #obsvars = map(eq->eq.lhs, observed(sys))
     #fulldvs = [dvs; obsvars]
 
-    eqs = equations(sys)
+    eqs = [eq for eq in equations(sys) if !isdifferenceeq(eq)]
```
ChrisRackauckas, line 93: @YingboMa do we want to lower by moving these to a separate array? Will structural simplify ignore them as well?

src/systems/diffeqs/abstractodesystem.jl
```diff
@@ -109,6 +109,36 @@ function generate_function(
...
+    p = map(x->time_varying_as_func(value(x), sys), ps)
+    t = get_iv(sys)
+
+    f_oop, f_iip = build_function(rhss, u, p, t; kwargs...)
+
+    f = @RuntimeGeneratedFunction(@__MODULE__, f_oop)
+
+    function cb_affect!(int) 
+        int.u += f(int.u, int.p, int.t) 
+    end
```
ChrisRackauckas, line 134: this is allocating a whole array as output?

src/systems/diffeqs/abstractodesystem.jl
```diff
@@ -109,6 +109,36 @@ function generate_function(
...
+
+    dts = [ operation(eq.lhs).dt for eq in eqs if isdifferenceeq(eq)] 
+    all(dts .== dts[1]) || error("All difference variables should have same time steps.")
```
ChrisRackauckas, line 137: fine for a first pass, we'll need to improve this, and that will require scalarizing the effect as well.

ChrisRackauckas (approved):
I'm good with this as a first pass. The comments can be done as a future pass, splitting the callback `affect!`s and possibly making structural simplify more robust.

src/systems/diffeqs/abstractodesystem.jl
sharanry: Yes. I aim to avoid this. But it is still a bit tricky with the current inplace function generator offered by Symbolics.

src/systems/diffeqs/abstractodesystem.jl
YingboMa: Like, adding another field to ODESystems? Yeah, that would work.

src/systems/diffeqs/abstractodesystem.jl
```diff
@@ -109,6 +109,46 @@ function generate_function(
     end
 end
 
+@inline function allequal(x)
```
YingboMa, line 112: No need to expand everything. You can just use an generator.
```julia
julia> alldts(dts) = all(dts[1] == dt for dt in dts)
alldts (generic function with 1 method)

julia> dts = collect(0:0.2:10);

julia> alldts(dts);

julia> @time alldts(dts)
  0.000005 seconds
false
```

YingboMa (approved):
I am fine with merging this PR in its current state. I agree with Chris that discrete states should be processed differently. All the filters on equations seems very unsustainable.

YingboMa:
NNPDE failure unrelated?

PR #1127, opened 2021-07-16
preserve context vars in `toexpr`

anandijain:
- fix a misspelling in Contextual variable metadata types, reserve "default" metadata key.
- add utility accessor/setters for reserved metadata
- preserve context var metadata in `toexpr`


anandijain:
blocked by https://github.com/PainterQubits/Unitful.jl/issues/466

PR #1129, opened 2021-07-19
adding defaults to ReactionSystem constructors

anandijain:
I'm not sure if I need to do any scaling of defaults during conversion

### DIFF
--- a/src/systems/reaction/reactionsystem.jl
+++ b/src/systems/reaction/reactionsystem.jl
@@ -181,20 +181,6 @@ function ReactionSystem(iv; kwargs...)
     ReactionSystem(Reaction[], iv, [], []; kwargs...)
 end
 
-# function equations(sys::ModelingToolkit.ReactionSystem)
-#     eqs = get_eqs(sys)
-#     systems = get_systems(sys)
-#     if isempty(systems)
-#         return eqs
-#     else
-#         eqs = [eqs;
-#                reduce(vcat,
-#                       namespace_equations.(get_systems(sys));
-#                       init=[])]
-#         return eqs
-#     end
-# end
-
 """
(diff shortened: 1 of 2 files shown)


isaacsas (approved):

isaacsas:
Looks ok to me. I guess we merge here and then retest the Catalyst PR once there is a new release. @ChrisRackauckas that ok?

isaacsas:
@anandijain can you update the direct problem constructors too, which also call convert? I missed those.

ChrisRackauckas (approved):
looks good to me.

isaacsas:
I’ll merge once those last updates are in and tests pass.

ChrisRackauckas:
wait, isn't `defaults` usually a keyword argument in system constructors? It should match the format of the others.

anandijain:
i need to merge u0 and default_u0, similarly for `p` 
where u0 and p that are passed as args are preferred

isaacsas:
Just mimic ODESystem.

anandijain:
would appreciate a sanity check, but I think it works as it is now, see the tests I added. it seems like defaults get handled just fine as is

ChrisRackauckas (approved):

isaacsas:
Duh, yeah it just gets passed through as a kwarg.

ChrisRackauckas:
Why is Catalyst failing?

anandijain:
it uses the direct rxnsys constructors so its missing that last Dict() arg. see that pr to see why

anandijain:
we are about to walk home, would u mind patch versioning and rerunning the catalyst pr test

ChrisRackauckas:
already started.

PR #1132, opened 2021-07-21
Allow indicating variables as input/output

sharanry:
This PR aims to prevent certain variables/states from being eliminated by declaring them as input/output at the time of their instantiation. 

For example, in the following system, `z` variable will be prevented from being eliminated structural_simplify.
```julia
@parameters t σ ρ β
@variables x(t) y(t) z(t) [output=true] a(t) u(t) F(t)
D = Differential(t)

eqs = [
       D(x) ~ σ*(y-x)
       D(y) ~ x*(ρ-z)-y + β
       0 ~ z - x + y
       0 ~ a + z
       u ~ z + a
      ]

lorenz1 = ODESystem(eqs,t,name=:lorenz1)
lorenz1_reduced = structural_simplify(lorenz1)


@assert any(isequal.(z, states(lorenz1_reduced)))
```

### DIFF
--- a/src/variables.jl
+++ b/src/variables.jl
@@ -2,10 +2,14 @@ struct VariableUnit end
 struct VariableConnectType end
 struct VariabelNoiseType end
 struct VariabelDescriptionType end
+struct VariabelInput end
+struct VariabelOutput end
 Symbolics.option_to_metadata_type(::Val{:unit}) = VariableUnit
 Symbolics.option_to_metadata_type(::Val{:connect}) = VariableConnectType
 Symbolics.option_to_metadata_type(::Val{:noise}) = VariabelNoiseType
 Symbolics.option_to_metadata_type(::Val{:description}) = VariabelDescriptionType
+Symbolics.option_to_metadata_type(::Val{:input}) = VariabelInput
+Symbolics.option_to_metadata_type(::Val{:output}) = VariabelOutput
 
 """
 $(SIGNATURES)
--- a/src/variables.jl
+++ b/src/variables.jl
@@ -1,15 +1,15 @@
 struct VariableUnit end
 struct VariableConnectType end
-struct VariabelNoiseType end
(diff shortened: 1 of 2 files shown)


anandijain:
variable is misspelled in those new metadata types

sharanry:
> variable is misspelled in those new metadata types

Ah thanks! 😅

sharanry:
Reposted from slack--
@YingboMa  I got the chance to take a closer look at this. I wanted to confirm something — do we have to modify [each of the simplification/reduction steps called by structural simplify ](https://github.com/SciML/ModelingToolkit.jl/blob/892dffe90aad92a00bd1d44050ead95c753ab2e6/src/systems/abstractsystem.jl#L765-L775)(namely alias_elimination and tearing) to check if a state is an input/output before deciding to eliminate it?

sharanry:
@YingboMa 
In the above example, if one declared state `a(t)` as input/output, we might have to retain the equations or some simplified version of it that define what state `a` equates to.  Is there already some mechanism to accomplish this?

YingboMa:
https://github.com/SciML/ModelingToolkit.jl/blob/892dffe90aad92a00bd1d44050ead95c753ab2e6/src/systems/systemstructure.jl#L118

You can skip adding the variables when a state is an input/output.

sharanry:
> https://github.com/SciML/ModelingToolkit.jl/blob/892dffe90aad92a00bd1d44050ead95c753ab2e6/src/systems/systemstructure.jl#L118
> 
> You can skip adding the variables when a state is an input/output.

Yes but this doesn't include the necessary equation along with the state. It errors out like `ERROR: LoadError: InvalidSystemException: The system is missing an equation for z(t).`

I am trying to figure out a way to also add the necessary equations.

sharanry:
@YingboMa I have removed the unrelated tests. Please feel free to merge if and when CI passes.

Issue #405, opened 2020-05-28
Linear Noise Approximation (LNA) model from ReactionNetworks


ChrisRackauckas:
Saw this in: https://tanhevg.github.io/GpABC.jl/latest/overview-lna/

It would be nice to automatically allow for generating this set of ODEs from the ReactionNetwork.

isaacsas:
Tau-leaping too.

isaacsas:
What's the master equation as a PDESystem here refer to? Do you mean a PDE for the corresponding generating function, or for spatial particle systems?

ChrisRackauckas:
For the probability distribution

isaacsas:
Ahh, ok so you are thinking of it as a differential-difference equation. Could we use some of the infinite dimensional LA tooling for representing the adjoint of the generator in the master equation? (I know next to nothing about whether it would be worthwhile, but that would presumably allow use to "exactly" represent the equation prior to trying to solve it numerically.)

ChrisRackauckas:
We can represent it via an InfiniteArray of equations, and then find ways to deal with it. Closed systems would be finite of course.

isaacsas:
Yup, it would just be nice to have a representation that is "exact", so that different methods can truncate / approximate as they see fit.

vnminin:
@ChrisRackauckas, UCI-MCSB PhD student @yushang-lai is trying to implement LNA inference in Turing. We have run into a problem with the covariance matrix ODE: https://discourse.julialang.org/c/domain/probprog/48

Do you have any advice for us? Thanks!

ChrisRackauckas:
@vnminin oh, someone just put up a whole library of moment closures: https://github.com/augustinas1/MomentClosure.jl

vnminin:
Thanks, @ChrisRackauckas. Maybe helpful, but hard to tell, because there is almost no documentation.

Our main problem is with Turing not liking our way of enforcing positive definiteness of the covariance  matrix -- ODE returns a nearly positive definite matrix

ChrisRackauckas:
https://augustinas1.github.io/MomentClosure.jl/dev/

ChrisRackauckas:
For undocumented libraries, look at their tests: https://github.com/augustinas1/MomentClosure.jl/blob/main/test/moment_convert_tests.jl

vnminin:
[...]

PR #1138, opened 2021-07-22
lowering ReactionSystem to CME ODEs

anandijain:
Should I add a kwarg to `convert(ODESystem, rs)` to specify which method is chosen? How should `ODESystem_cme` be integrated?

Note that the BFS method fails to terminate on ReactionSystems like Catalyst's [birth death](https://catalyst.sciml.ai/stable/tutorials/basic_examples/#Example:-Birth-Death-Process). I am wondering if I should try to detect cycles where you end up with more than you started (or am I just doing this all wrong).

This was copied over quite hastily but I'd like some comments/review, particularly someone to verify that I have generated the right equations (I think I have but haven't investigated enough).

addresses one of the todos in #405

### DIFF
--- a/src/systems/reaction/reactionsystem.jl
+++ b/src/systems/reaction/reactionsystem.jl
@@ -559,20 +559,33 @@ function modified_states!(mstates, rx::Reaction, sts::Set)
 end
 
 "Performs a breadth-first graph traversal to enumerate possible states"
-function BFS(G, v)
+function traverse_reactionsystem(G, v; truncation=nothing)
+    num_molecules = sum(v)
+    num_species = length(u0)
     state_space = []
-    Q = []
-    push!(Q, v)
+    Q = Queue{typeof(v)}()
+    enqueue!(Q, v)
     push!(state_space, v)
     nsm = netstoichmat(G)
     edges = eachrow(nsm)
     while !isempty(Q)
-        w = popfirst!(Q)
+        w = dequeue!(Q)
         for e in edges
             x = w .+ e
-            if x ∉ state_space && all(x .>= 0)
-                push!(state_space, x) 
-                push!(Q, x)
+            if x ∉ state_space && all(y >= 0 for y in x)
+              if truncation === nothing 
+                if sum(x) > num_molecules 
+                  error("need to provide truncation for infinite reaction system")
+                else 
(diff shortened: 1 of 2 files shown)


isaacsas:
Very cool. I'll try to look it over this weekend and leave a review. For general systems the CME will be infinite-dimensional, so I guess we either need a way to symbolically represent infinite systems of differential-difference equations, or to make the user choose a truncation method.

On a quick glance I see you are using some functions from Catalyst like `netstoichmat`. It's probably time to make the move of `ReactionSystem` to Catalyst, so that this functionality all lives in the same place. I can try to migrate it over next week (unless it would be better to wait till after Juliacon so as not to immediately break the SBML-related presentations).

src/systems/reaction/reactionsystem.jl
```diff
@@ -557,3 +557,63 @@ function modified_states!(mstates, rx::Reaction, sts::Set)
         (species in sts) && push!(mstates, species)
     end
 end
+
+"Performs a breadth-first graph traversal to enumerate possible states"
+function BFS(G, v)
+    state_space = []
```
YingboMa, line 563: Should be a queue to make is linear time.

src/systems/reaction/reactionsystem.jl
```diff
@@ -557,3 +557,63 @@ function modified_states!(mstates, rx::Reaction, sts::Set)
...
+    Q = []
+    push!(Q, v)
+    push!(state_space, v)
+    nsm = netstoichmat(G)
+    edges = eachrow(nsm)
+    while !isempty(Q)
+        w = popfirst!(Q)
+        for e in edges
+            x = w .+ e
+            if x ∉ state_space && all(x .>= 0)
```
YingboMa, line 573: This is allocating.

src/systems/reaction/reactionsystem.jl
```diff
@@ -557,3 +557,63 @@ function modified_states!(mstates, rx::Reaction, sts::Set)
         (species in sts) && push!(mstates, species)
     end
 end
+
+"Performs a breadth-first graph traversal to enumerate possible states"
+function BFS(G, v)
```
YingboMa, line 562: Call something more descriptive ;-p

PR #1143, opened 2021-07-24
switch MassActionJumps to use parameter vectors

isaacsas:
Extends symbolic JumpSystems to create MassActionJumps that remember the mapping from parameters to rate constants. This should enable updating MassActionJump parameters via callbacks. 

Before merging requires
- [x] Merging https://github.com/SciML/DiffEqJump.jl/pull/193
- [x] A new DiffEqJump release 
- [x] Updating the DiffEqJump version in the Project.toml of this PR


isaacsas:
@TorkelE this should update ModelingToolkit, and hence Catalyst too, to use the new MassActionJump form that stores parameters (so updating in callbacks works for Catalyst models too).

ChrisRackauckas (approved):

PR #1150, opened 2021-07-28
WIP: add algebraic equations to ReactionSystems

isaacsas:
@paulflang @anandijain is this what you wanted / does it cover your use cases?

There are a few limitations in this version. Right now all variables must be declared as states of the ReactionSystem (i.e. species), even variables that appear only within the algebraic equations, which in turn causes ODEs/SDEs to be generated for them. This can be disabled in `convert` with the `include_zero_odes=false` kwarg, see the test example, (but note the ODEs will still appear if they are non-zero even if you want them dropped). 

An alternative would be to also pass ReactionSystems a list of `algebraic_states`, which will not be considered species or have ODEs/SDEs generated. I'm thinking this could be a better approach, but I'd be interested to hear your thoughts.

### DIFF
--- a/src/systems/reaction/reactionsystem.jl
+++ b/src/systems/reaction/reactionsystem.jl
@@ -154,14 +154,16 @@ struct ReactionSystem <: AbstractSystem
     parameters are not supplied in `ODEProblem`.
     """
     defaults::Dict
+    """Algebraic equations that impose constraints on species."""
+    algebraic_eqs::Vector{Equation}
 
-    function ReactionSystem(eqs, iv, states, ps, observed, name, systems, defaults)
+    function ReactionSystem(eqs, iv, states, ps, observed, name, systems, defaults, algebraic_eqs)
         iv′ = value(iv)
         states′ = value.(states)
         ps′ = value.(ps)
         check_variables(states′, iv′)
         check_parameters(ps′, iv′)
-        new(collect(eqs), iv′, states′, ps′, observed, name, systems, defaults)
+        new(collect(eqs), iv′, states′, ps′, observed, name, systems, defaults, algebraic_eqs)
     end
 end
 
@@ -171,10 +173,11 @@ function ReactionSystem(eqs, iv, species, params;
                         name = gensym(:ReactionSystem),
                         default_u0=Dict(),
                         default_p=Dict(),
-                        defaults=_merge(Dict(default_u0), Dict(default_p)))
+                        defaults=_merge(Dict(default_u0), Dict(default_p)),
+                        algebraic_eqs=nothing)
 
     #isempty(species) && error("ReactionSystems require at least one species.")
-    ReactionSystem(eqs, iv, species, params, observed, name, systems, defaults)
+    ReactionSystem(eqs, iv, species, params, observed, name, systems, defaults, algebraic_eqs)
 end
 
 function ReactionSystem(iv; kwargs...)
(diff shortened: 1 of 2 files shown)


paulflang:
I am not using Rules in my own models, but judging from the SBML spec this looks good to me.

> There are a few limitations in this version. Right now all variables must be declared as states of the ReactionSystem (i.e. species)

This should be fine, as I see in the test code that not only species, but also parameters can appear in the algebraic equations. But are parameters allowed to change over time (e.g. `gamma ~ A + delta`, where A is a time-variable state)? Cause in SBML it seems they can.

isaacsas:
Since we would want those to end up as observables that can be evaluated, it sounds like it might be better to explicitly list the algebraic states too. Are you able to determine and list the algebraic variables from the SBML file?

I wonder if it might be better to just store a NonlinearSystem attached to the ReactionSystem, in a dedicated field or as a sub-system.

paulflang:
In this case `gamma` may be more than an observable: it could appear as a parameter in any other equation.

What do you mean with _"Since we would want those to end up as observables that can be evaluated, it sounds like it might be better to explicitly list the algebraic states too"_?

> Are you able to determine and list the algebraic variables from the SBML file?

SBML does not make a distinction whether species or parameters occur only in AlgebraicRules or Reactions or both as long as the model is not overdetermined. So I guess the answer to your question is probaly: I can easily list all variables but I think it takes some effort to tell in which type of elements (Reactions, AssignmentRules, etc.) they appear.

ChrisRackauckas:
>I wonder if it might be better to just store a NonlinearSystem attached to the ReactionSystem, in a dedicated field or as a sub-system.

That sounds like a good way to handle it, and then you can only create a JumpProblem if there's no non-JumpProblem subsystems.

isaacsas:
@paulflang an observable is essentially a variable (state in the ModelingToolkit terminology) that is completely determined from the ones you actually solve for, so can be eliminated to reduce the system size. Storing such variables as observables allows one to say stuff like `sol[gamma]` for a solution object and get the right values out (the algebraic formula will be evaluated based on the variables that were actually solved for). It'd be like a species that one can eliminate from the ODEs due to a conservation law, but one still might want to be able to evaluate its value.

isaacsas:
I believe anything that is not a fixed valued should be an observable/state and not a parameter. So gamma would not be appropriate to specify as a parameter in your example. That's why I was thinking it makes sense to make it a state of an attached NonlinearSystem.

paulflang:
> I believe anything that is not a fixed valued should be an observable/state and not a parameter.

Agreed. Idk why SBML allows time varying parameters.

Just to be sure: would combining the following ReactionSystem with NonlinearSystem work?

```julia
@paramters t k1
@variables A(t) B(t) E(t)

r1 = Reaction(k1, [A, E], [B, E], [1,1], [1,1])
rs = ReactionSystem([r1], t, [A, B, E], [k1])

NonlinearSystem([A ~ E^2], [A, E], [])
```
Cause `E` is not just an observable here.

isaacsas:
Yes, that should work as long as in `convert` you explicitly use the `include_zero_odes=false` keyword to drop zero ODEs. Then `E` will have no generated ODE in the `ReactionSystem`. Otherwise the number of equations would be unbalanced in the combined ODE system.

isaacsas:
Handling in https://github.com/SciML/Catalyst.jl/pull/404

PR #1151, opened 2021-07-29
Linearizing delayed equation systems

sharanry:
This PR aims to automatically linearize delayed equations in discrete systems.


src/systems/discrete_system/discrete_system.jl
```diff
@@ -125,6 +127,43 @@ function DiffEqBase.DiscreteProblem(sys::DiscreteSystem,u0map,tspan,
...
+            end
+            var_sym = Symbol((~t).f.name, Symbol("("), join(Symbolics.tosymbol.(Symbolics.arguments(~t)), Symbol(",")), Symbol(")"))
+            Symbolics.variable(var_sym)
+        else
+            ~t
+        end
+    end
+    rw = SymbolicUtils.Postwalk(r)
+    [
+        Symbolics.value(eq.lhs) ~ Symbolics.value(rw(eq.rhs))
```
sharanry, line 146: @shashi @YingboMa Any idea what might lead to the `rhs` to remain as a `Num` as opposed to a composition of `Mul`s and `Add`s even after calling `value`?

sharanry: Not using this anymore

Project.toml
```diff
@@ -73,7 +73,7 @@ Setfield = "0.7"
 SpecialFunctions = "0.7, 0.8, 0.9, 0.10, 1.0"
 StaticArrays = "0.10, 0.11, 0.12, 1.0"
 SymbolicUtils = "0.12, 0.13"
-Symbolics = "1.4.1"
+Symbolics = "2"
```
YingboMa, line 76: Don't use Symbolics v2

src/systems/discrete_system/discrete_system.jl
```diff
@@ -125,6 +126,55 @@ function DiffEqBase.DiscreteProblem(sys::DiscreteSystem,u0map,tspan,
     DiscreteProblem(f,u0,tspan,p;kwargs...)
 end
 
+function linearize_eqs(sys, eqs=sys.eqs; return_max_delay=false)
+    unique_states = unique(operation.(sys.states))
+    max_delay = Dict(v=>0.0 for v in unique_states)
+
+    r = @rule ~t::(t -> t isa Symbolics.Term && any(isequal((t).f), Symbolics.operation.(sys.states)) && is_delay_var(sys.iv, t)) => begin
```
YingboMa, line 133: Aren't they already imported in MTK? You can just write `operation`, also, do you really mean to check `Term` but not `istree`?

src/systems/discrete_system/discrete_system.jl
```diff
@@ -125,6 +126,55 @@ function DiffEqBase.DiscreteProblem(sys::DiscreteSystem,u0map,tspan,
...
+        delay = get_delay_val(sys.iv, first(Symbolics.arguments(~t)))
```
YingboMa, line 134: `Symbolics.` is not necessary.

src/utils.jl
```diff
@@ -111,10 +111,25 @@ function check_parameters(ps, iv)
...
+        args = arguments(var)
+    catch
+        return false
+    end
+    length(args) > 1 && return false
+    isequal(first(args), iv) && return false
+    delay = iv - first(args)
+    delay isa Integer || 
+    delay isa AbstractFloat ||
+    (delay isa Num && isreal(delay.val))
```
YingboMa, line 126: Never use `.val`

Project.toml
sharanry: I previously needed `Symbolics.variable` method which is part of Symbolics v2. But now I have managed without it. Will revert back.

src/systems/discrete_system/discrete_system.jl
sharanry: Yup. `istree` is what I was looking for!

test/discretesystem.jl
```diff
@@ -55,4 +55,37 @@ p = [0.05,10.0,0.25,0.1];
...
+@test_throws ErrorException z(t)
+
+# Equations
+eqs = [
+    D1(x(t)) ~ 0.4x(t) + 0.3x(t-1.5) + 0.1x(t-3),
+    D2(y(t)) ~ 0.3y(t) + 0.7y(t-2) + 0.1z,
+]
+
+# System
+sys = DiscreteSystem(eqs,t,[x(t),x(t-1.5),x(t-3),y(t),y(t-2),z],[])
```
sharanry, line 78: @YingboMa Currently we need to mention all delay variables manually. Is this fine or do we want to collect them automatically somehow?

sharanry:
I don't think the Catalyst build fail is related.

sharanry:
@YingboMa I have used states function. Let me know if any other changes need to be made before merge. Thanks!

PR #1153, opened 2021-07-30
move connection_type to ReactionSystems from Reactions

isaacsas:
I noticed in https://github.com/SciML/ModelingToolkit.jl/pull/1150 that this was not setup on the system but the reactions, which was causing me some issues.


isaacsas:
Catalyst will require an update once this is merged for the updates to `Reaction` and `ReactionSystem`.

ChrisRackauckas (approved):

Issue #1158, opened 2021-08-02
`modelingtoolkitize` does not match form for `Tuple` and `NamedTuple` parameters


superckl:
`modelingtoolkitize` fails on problems expecting parameters in a form other than a `Vector` since it always reduces the passed parameters to a `Vector`. The case of a `Tuple` is somewhat trivial since a `Vector` will also suffice, but the case of a `NamedTuple` is more important for when calculations must refer to parameters by name (e.g., smaller, interchangeable parts of a larger calculation). Is it possible to have `modelingtoolkitize` match the form of the passed in parameters? Examples with `ModelingToolkit.jl` v5.26.0:

`NamedTuple`:
```julia
using ModelingToolkit
using OrdinaryDiffEq

function ode_prob(du, u, p::NamedTuple, t)
    du[1] = u[1]+p.α*u[2]
    du[2] = u[2]+p.β*u[1]
end

params = (α = 1, β = 1)
prob = ODEProblem(ode_prob, [1 1], (0, 1), params)
modelingtoolkitize(prob)
```

error:
```julia
ERROR: LoadError: MethodError: no method matching ode_prob(::Matrix{Num}, ::Matrix{Num}, ::Vector{Num}, ::Num)
Closest candidates are:
  ode_prob(::Any, ::Any, ::NamedTuple, ::Any) at g:\My Drive\Research\Code\julia solver\minimal.jl:4
Stacktrace:
 [1] (::ODEFunction{true, typeof(ode_prob), LinearAlgebra.UniformScaling{Bool}, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, typeof(SciMLBase.DEFAULT_OBSERVED), Nothing})(::Matrix{Num}, ::Vararg{Any, N} where N)
   @ SciMLBase C:\Users\super\.julia\packages\SciMLBase\cU5k7\src\scimlfunctions.jl:334
[...]

PR #1162, opened 2021-08-04
fix modelingtoolkitize for tuple and namedtuple parameters

ChrisRackauckas:
Fixes https://github.com/SciML/ModelingToolkit.jl/issues/1158

### DIFF
--- a/src/systems/diffeqs/modelingtoolkitize.jl
+++ b/src/systems/diffeqs/modelingtoolkitize.jl
@@ -8,12 +8,7 @@ function modelingtoolkitize(prob::DiffEqBase.ODEProblem; kwargs...)
                             return prob.f.sys
     @parameters t
 
-    if prob.p isa Tuple || prob.p isa NamedTuple
-        p = [x for x in prob.p]
-    else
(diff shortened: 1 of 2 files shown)


src/systems/diffeqs/modelingtoolkitize.jl
```diff
@@ -91,6 +95,14 @@ function define_params(p::Union{SLArray,LArray})
     [Num(toparam(Sym{Real}(nameof(Variable(x))))) for x in LabelledArrays.symnames(typeof(p))]
 end
 
+function define_params(p::Tuple)
+    tuple((Num(toparam(Sym{Real}(nameof(Variable(:α, i))))) for i in eachindex(p))...)
+end
+
+function define_params(p::NamedTuple)
+    @show NamedTuple(x=>Num(toparam(Sym{Real}(nameof(Variable(x))))) for x in keys(p))
```
YingboMa, line 103: :-p

ChrisRackauckas: yeah...

Issue #1037, opened 2021-06-09
Move expressions to RHS during lowering


akashkgarg:
I noticed this when creating a simple NonlinearProblem (see https://github.com/SciML/DiffEqOperators.jl/pull/400#issuecomment-858036932): 
```julia
    eqs = [a ~ 1,
           a + c - 2*b ~ 0,
           b + d - 2*c ~ 0,
           d ~ 1]
```

It turns out that this produces an invalid generated function. The equations need to be setup like this: 
```julia
    eqs = [0 ~ a - 1,
           0 ~ a + c - 2*b,
           0 ~ b + d - 2*c,
           0 ~ d - 1]
```

Automatically moving everything over to the RHS should happen during lowering as suggested by Chris here: https://github.com/SciML/DiffEqOperators.jl/pull/400#issuecomment-858039286

asprionj:
Also noticed during development of the intro example that
```julia
@named fol_separate = ODESystem([ τ*D(x) ~ 1-x ])
```
does not work; throws:
```
ERROR: MethodError: isequal(::Sym{Real, Base.ImmutableDict{DataType, Any}}, ::Missing) is ambiguous. [...]
```

So I changed it to `D(x) ~ (1-x)/τ`, but isn't one advantage of using a symbolic framework not needing to manually do such rearrangements?

ChrisRackauckas:
Yeah, we just haven't done this one. The first one is absurdly simple and just an oversight. The latter one has an interaction with tearing that needs to be a bit careful.

YingboMa:
I don't why the first case is an oversight. It's not clear to me on why that's a problem. What do you need to do?

YingboMa:
I see. We can move everything over to the RHS during construction time for NonlinearSystems.

PR #1163, opened 2021-08-04
automatically put nonlinear systems in canonical form

ChrisRackauckas:
Fixes https://github.com/SciML/ModelingToolkit.jl/issues/1037

### DIFF
--- a/src/systems/nonlinear/nonlinearsystem.jl
+++ b/src/systems/nonlinear/nonlinearsystem.jl
@@ -66,7 +66,7 @@ function NonlinearSystem(eqs, states, ps;
                          connection_type=nothing,
                          )
     # Move things over, but do not touch array expressions
-    eqs = eqs isa Vector{Equation} ? [0 ~ x.rhs - x.lhs : x for x in eqs] : eqs
+    eqs = eqs isa Vector{Equation} ? [0 ~ x.rhs - x.lhs for x in eqs] : eqs
 
     if !(isempty(default_u0) && isempty(default_p))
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
Unrelated CI error @YingboMa ?

src/systems/nonlinear/nonlinearsystem.jl
```diff
@@ -65,7 +65,9 @@ function NonlinearSystem(eqs, states, ps;
                          systems=NonlinearSystem[],
                          connection_type=nothing,
                          )
-    eqs = collect(eqs)
```
YingboMa, line 68: This should be kept.

src/systems/nonlinear/nonlinearsystem.jl
```diff
@@ -65,7 +65,9 @@ function NonlinearSystem(eqs, states, ps;
...
+    # Move things over, but do not touch array expressions
+    eqs = eqs isa Vector{Equation} ? [0 ~ x.rhs - x.lhs for x in eqs] : eqs
```
YingboMa, line 69: The branch looks unnecessary.

ChrisRackauckas: it's not. `D.(x) .~ ...` symbolic array turns into symbolic broadcast expression. I'm just going to ignore that case for now by skipping it 😅

YingboMa: That's why there is `collect`.

YingboMa:
Maybe squash the commits next time?

ChrisRackauckas:
yeah

PR #1164, opened 2021-08-04
Automatically move differentials to the left side

ChrisRackauckas:
Fixes https://github.com/SciML/ModelingToolkit.jl/issues/860

### DIFF
--- a/src/systems/diffeqs/odesystem.jl
+++ b/src/systems/diffeqs/odesystem.jl
@@ -95,15 +95,15 @@ end
 function move_diffs(eq::Equation,r)
     # Do not modify `D(x) ~ ...`, already correct
     # Ignore `δ(x) ~ ...` for now
-    if !(eq.lhs isa Term && ModelingToolkit.operation(eq.lhs) isa Differential) &&
-       !(eq.lhs isa Term && ModelingToolkit.operation(eq.lhs) isa Difference)
+    if !(eq.lhs isa Term && operation(eq.lhs) isa Differential) &&
+       !(eq.lhs isa Term && operation(eq.lhs) isa Difference)
        _eq = eq.rhs-eq.lhs
        rhs = r(_eq)
        if isnothing(rhs)
            eq
        else
            lhs = _eq - rhs
-           if !(lhs isa Number) && ModelingToolkit.operation(lhs) isa Differential
+           if !(lhs isa Number) && operation(lhs) isa Differential
                lhs ~ -rhs
            else
                -lhs ~ rhs
--- a/src/systems/diffeqs/odesystem.jl
+++ b/src/systems/diffeqs/odesystem.jl
@@ -103,7 +103,6 @@ function move_diffs(eq::Equation,r)
            eq
        else
            lhs = _eq - rhs
-           @show lhs
            if !(lhs isa Number) && ModelingToolkit.operation(lhs) isa Differential
                lhs ~ -rhs
            else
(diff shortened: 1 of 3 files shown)


src/systems/diffeqs/odesystem.jl
```diff
@@ -92,6 +92,31 @@ struct ODESystem <: AbstractODESystem
...
 end
 
+function move_diffs(eq::Equation,r)
+    # Do not modify `D(x) ~ ...`, already correct
+    # Ignore `δ(x) ~ ...` for now
+    if !(eq.lhs isa Term && ModelingToolkit.operation(eq.lhs) isa Differential) &&
+       !(eq.lhs isa Term && ModelingToolkit.operation(eq.lhs) isa Difference)
+       _eq = eq.rhs-eq.lhs
+       rhs = r(_eq)
+       if isnothing(rhs)
```
YingboMa, line 102:
```suggestion
       if rhs === nothing
```
`isnothing` probably should never be used, and just removed from Base.

src/systems/diffeqs/odesystem.jl
```diff
@@ -92,6 +92,31 @@ struct ODESystem <: AbstractODESystem
     end
 end
 
+function move_diffs(eq::Equation,r)
+    # Do not modify `D(x) ~ ...`, already correct
+    # Ignore `δ(x) ~ ...` for now
+    if !(eq.lhs isa Term && ModelingToolkit.operation(eq.lhs) isa Differential) &&
```
YingboMa, line 98: This code already lives in MTK.

test/odesystem.jl
```diff
@@ -439,10 +439,39 @@ function periodic_difference_affect!(int)
...
+D = Differential(t)
+eqs = [0 ~ x + y - D(x)
+       x ~ y]
+sys = ODESystem(eqs)
+eqs2 = equations(sys)
+eqs3 = [D(x) ~ x + y,
+        0 ~ y-x]
+@test isequal(eqs2,eqs3)
+
+eqs = [0 ~ x + y + D(x)
```
YingboMa, line 462: What happens if there's `D(x)/x ~ x`?

ChrisRackauckas: `D(x)/x ~ x` will be untouched like before. `D(x)*x ~ x` will give you `-D(x)*x ~ -x`, which is a little odd but works fine for `DAEProblem`.

YingboMa: Could you test these cases as well? Also `sin(D(x) * x) ~ x` etc.

ChrisRackauckas:
@YingboMa one last test now throws the wrong error. It errored before, but it shouldn't throw this error.

```julia
julia> equations(sys)
3-element Vector{Equation}:
 0 ~ x(t) + sin(u1(t) + u2(t))
 Differential(t)(x(t)) ~ x(t) + y1(t)
 0 ~ sin(y2(t)) - cos(x(t))
fullvars = Any[Differential(t)(x(t)), x(t), u1(t), u2(t), y2(t)]
```

How come `y2(t)` was changed from a parameter to a state in structural_simplify? That seems like it should never be able to happen?

YingboMa:
Maybe rule drops the metadata.

YingboMa:
What's the MWE?

ChrisRackauckas:
```julia
# Example 5 from Pantelides' original paper
@parameters t
params = collect(@parameters y1(t) y2(t))
sts = collect(@variables x(t) u1(t) u2(t))
D = Differential(t)
eqs = [
       0 ~ x + sin(u1 + u2)
       D(x) ~ x + y1
       cos(x) ~ sin(y2)
      ]
@named sys = ODESystem(eqs, t, sts, params)
@test_throws ModelingToolkit.InvalidSystemException structural_simplify(sys)
```

Issue #1179, opened 2021-08-09
PDESystem fails


KirillZubov:
ModelingToolkit v6.0.0
```julia
using ModelingToolkit
@parameters x
@variables t u(..)
Dxx = Differential(x)^2
Dtt = Differential(t)^2
Dt = Differential(t)
#2D PDE
C=1
eq  = Dtt(u(t,x)) ~ C^2*Dxx(u(t,x))
# Initial and boundary conditions
bcs = [u(t,0) ~ 0.,# for all t > 0
       u(t,1) ~ 0.,# for all t > 0
       u(0,x) ~ x*(1. - x), #for all 0 < x < 1
       Dt(u(0,x)) ~ 0. ] #for all  0 < x < 1]
# Space and time domains
domains = [t ∈ (0.0,1.0),
           x ∈ (0.0,1.0)]
julia> pde_system = PDESystem(eq,bcs,domains,[t,x],[u])
PDESystem
Equations: Differential(t)(Differential(t)(u(t, x))) ~ Differential(x)(Differential(x)(u(t, x)))
Boundary Conditions: Equation[u(t, 0) ~ 0.0, u(t, 1) ~ 0.0, u(0, x) ~ x*(1.0 - x), Differential(t)(u(0, x)) ~ 0.0]
Domain: Symbolics.VarDomainPairing[Symbolics.VarDomainPairing(t, 0.0..1.0), Symbolics.VarDomainPairing(x, 0.0..1.0)]
Error showing value of type PDESystem:
Error showing value of type PDESystem:
ERROR: type PDESystem has no field depvars
Stacktrace:
  [1] get_depvars(sys::PDESystem)
    @ ModelingToolkit ~/.julia/packages/ModelingToolkit/fEwAm/src/systems/abstractsystem.jl:194
  [2] show(io::IOContext{Base.TTY}, #unused#::MIME{Symbol("text/plain")}, sys::PDESystem)
    @ Main ./none:6
  [3] (::REPL.var"#38#39"{REPL.REPLDisplay{REPL.LineEditREPL}, MIME{Symbol("text/plain")}, Base.RefValue{Any}})(io::Any)
    @ REPL /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.6/REPL/src/REPL.jl:220
  [4] with_repl_linfo(f::Any, repl::REPL.LineEditREPL)
[...]

PR #1181, opened 2021-08-09
fix PDESystem REPL

KirillZubov:
https://github.com/SciML/ModelingToolkit.jl/issues/1179


src/systems/abstractsystem.jl
```diff
@@ -185,6 +185,8 @@ for prop in [
              :domain
              :depvars
              :indvars
+             :ivs
+             :dvs
```
YingboMa, line 189: Instead of having `ivs` and `dvs`, is it possible to rename the fields and always use `get_devars` and such?

YingboMa: Oh, `depvars` and `indvars` are just never used. We can remove them and keep using `ivs` and `dvs`.

PR #1187, opened 2021-08-10
mv ReactionSystems to Catalyst and fix dependencies for JumpSystems

isaacsas:
Hopefully I'll have the complementary Catalyst PR up today and tested at least locally.

If tests pass here we should make a new release once the new Catalyst PR is passing local tests, and then we can quickly merge that and release at Catalyst.

CC: @ChrisRackauckas @paulflang @anandijain @kaandocal @augustinas1 @bosonbaas, this will move `ReactionSystem` functionality to all live in Catalyst so it is no longer split across here and Catalyst. Just FYI. In particular,
```julia
Reaction
ReactionSystem
ismassaction
oderatelaw
jumpratelaw
```
would now all live in Catalyst.


ChrisRackauckas (approved):
Since Catalyst hasn't updated, might as well just put this in a minor.

isaacsas:
Maybe we should do a major so the AlgebraicPetri folks catch the change this time? Up to you.

bosonbaas:
Thanks, I appreciate the heads up! I'll make sure the AlgebraicPetri interop tooling is updated to work with this, no need to make it a major version for our sake.

isaacsas:
@bosonbaas I'll tag you (and everyone else) when we get a working Catalyst release with these changes out.

isaacsas:
This has now propagated through and will be the default on MT >= 6.0 and Catalyst >= 8.0.

ChrisRackauckas:
Awesome! Was a longtime coming. I think we're almost finally "there". I'd still say everything is "beta" for the next 3-6 months 😅

isaacsas:
Yeah, I'm really looking forward to the array stuff and geometry specification. What @YingboMa has done so far for the former is great. I'm looking forward to starting on the Catalyst spatial interface once the work on them settles down.

For now we can plan on keeping an eye on system changes here and porting them to Catalyst when MT makes major releases (and before updating Catalyst's compatibility).

Issue #1200, opened 2021-08-16
Documentation of the function `ODESystem`


RainerHeintzmann:
Fantastic package!
I was trying to test the example that you provided in the great [video](https://www.youtube.com/watch?v=HEVOgSLBzWA) but found that the current version of `ODESystem()` requires the `name` argument to be provided.
However, this is not seen in its documentation and the example given in the documentation does also fail due to the missing `name` argument. It's also not really clear from the documentation whether `name` expects a string or a symbol.
 Happy to create a pull request, if wanted, but you are probably way quicker to fix this.
Thanks again for the great work!

RainerHeintzmann:
... this also concerns the examples given on the landing page, like the combination of two Lorenz systems...

ChrisRackauckas:
>just noted that, if you want to force the user to provide a name, as it seems to be the case for NonlinearSystem, then you may want to also introduce this requirement in PDESystem, which currently does not require a name.

That is a good point.

ChrisRackauckas:
>It's also not really clear from the documentation whether name expects a string or a symbol.

Thanks , it expects a symbol.

PR #1201, opened 2021-08-16
add required name for PDESystem

ChrisRackauckas:
Fixes https://github.com/SciML/ModelingToolkit.jl/issues/1200


ChrisRackauckas:
@KirillZubov this will require NeuralPDE.jl examples to all to `@named sys = PDESystem(...)` in the examples.

ChrisRackauckas:
We should've done this at v6, and it was intended for v6, so I'm just going to rip off the band-aid.

PR #1205, opened 2021-08-16
Optimization and support DiscreteUpdate operator

YingboMa:


YingboMa:
StructuralIdentifiability failure seems unrelated.

PR #1209, opened 2021-08-17
Adding inputs and outputs functions

dannys4:
This adds functions to access inputs and outputs of a system

### DIFF
--- a/src/systems/abstractsystem.jl
+++ b/src/systems/abstractsystem.jl
@@ -392,6 +392,14 @@ function states(sys::AbstractSystem)
            [sts; reduce(vcat,namespace_variables.(systems))])
 end
 
+function inputs(sys::AbstractSystem)
+    [st for st in states(sys) if isinput(st)]
+end
+
+function outputs(sys::AbstractSystem)
+    [st for st in states(sys) if isoutput(st)]
+end
+
 function parameters(sys::AbstractSystem)
     ps = get_ps(sys)
     systems = get_systems(sys)


src/systems/abstractsystem.jl
```diff
@@ -392,6 +392,14 @@ function states(sys::AbstractSystem)
            [sts; reduce(vcat,namespace_variables.(systems))])
 end
 
+function inputs(sys::AbstractSystem)
+    [st for st in states(sys) if isinput(st)]
+end
+
+function outputs(sys::AbstractSystem)
+    [st for st in states(sys) if isoutput(st)]
```
YingboMa, line 400: This may not be true. Do you want it to be true?

dannys4: In what case?

ChrisRackauckas: yeah they could be eliminated to observed?

dannys4: ah, so should I `cat` any states in `observed` that are also decorated as outputs?

YingboMa: But then, you cannot calculate the Jacobian.

dannys4: Could you elaborate on what you mean? What jacobian can I not calculate?

YingboMa: Suppose we have the reduced output as an observed variable `x ~ a + b` with the equation `D(y) ~ a + sin(a) + b`. What's the derivative of the right hand wrt `x ~ a + b`? We cannot differentiate this thing.

dannys4: I disagree with you on the idea that the `a+b` needs to be "from" `x`, but I see your point nonetheless. Does that have anything to do with it being an `output` though? Your argument seems to be valid regardless of that. Right now, don't we just expect people to, for lack of better words, "not do that"?

YingboMa: Given `x` is an output, what's the output Jacobian of
```julia
D(y) ~ a + b + x
0 ~ a + b - x
```
and what's the output Jacobian of
```
```julia
D(y) ~ 2a + 2b
```
wrt `a+b`?

dannys4: Say `x` isn't an output though. Isn't this still a problem? I'm failing to see where it being labelled as `output=true` comes in

YingboMa: When linearizing control problems, we need to compute the Jacobian wrt the inputs or the outputs.

PR #1227, opened 2021-08-21
Add spring-mass system tutorial

AayushSabharwal:
Add a simple spring-mass system as the second tutorial. It produces some nice plots too!

I noticed the plots aren't stored here, so I didn't upload them or add them to the markdown. Here they are:

`plot(sol)`
plotsol

`plot(sol, vars = (mass.pos[1], mass.pos[2]))`
plotmass

`plot(sol, vars = (mass.pos[1], mass.v[1]))`
plotposv

The lower two plots look much better if the problem is solved over a longer interval (such as `(0., 15.)`). However, then the first plot becomes crowded.

I've tried to copy the style of the acausal example. In fact, some parts are the same.

### DIFF
--- a/docs/src/tutorials/spring_mass.md
+++ b/docs/src/tutorials/spring_mass.md
@@ -57,7 +57,7 @@ plot(sol)
 
 ## Explanation
 ### Building the components
-For each component we use a Julia function that returns an `ODESystem`. At the top, we define the fundamental properties of a `Mass`: it has a mass `m`, a position `pos` and a velocity `vel`. We also define that the velocity is the rate of change of position with respect to time.
+For each component we use a Julia function that returns an `ODESystem`. At the top, we define the fundamental properties of a `Mass`: it has a mass `m`, a position `pos` and a velocity `v`. We also define that the velocity is the rate of change of position with respect to time.
 
 ```julia
 function Mass(; name, m = 1.0, xy = [0., 0.], u = [0., 0.])
@@ -68,7 +68,7 @@ function Mass(; name, m = 1.0, xy = [0., 0.], u = [0., 0.])
 end
 ```
 
-Note that this is an incompletely specified `ODESystem`. It cannot be simulated on its own since the equations for `pos[1:2](t)` are unknown. Notice the addition of a `name` keyword. This allows us to generate different masses with different names. A `Mass` can now be constructed as:
+Note that this is an incompletely specified `ODESystem`. It cannot be simulated on its own since the equations for the velocity `v[1:2](t)` are unknown. Notice the addition of a `name` keyword. This allows us to generate different masses with different names. A `Mass` can now be constructed as:
 
 ```julia
 Mass(name = :mass1)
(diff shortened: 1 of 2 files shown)


docs/src/tutorials/spring_mass.md
```diff
@@ -0,0 +1,229 @@
...
+```julia
+function Mass(; name, m = 1.0, xy = [0., 0.], u = [0., 0.])
+    ps = @parameters m=m
+    sts = @variables pos[1:2](t)=xy v[1:2](t)=u
+    eqs = collect(D.(pos) .~ v)
+    ODESystem(eqs, t, [pos..., v...], ps; name)
+end
+```
+
+Note that this is an incompletely specified `ODESystem`. It cannot be simulated on its own since the equations for `pos[1:2](t)` are unknown. Notice the addition of a `name` keyword. This allows us to generate different masses with different names. A `Mass` can now be constructed as:
```
ChrisRackauckas, line 71:
```suggestion
Note that this is an incompletely specified `ODESystem`. It cannot be simulated on its own since the equations for the velocity `v(t)` are unknown. Notice the addition of a `name` keyword. This allows us to generate different masses with different names. A `Mass` can now be constructed as:
```

docs/src/tutorials/spring_mass.md
```diff
@@ -0,0 +1,229 @@
...
+sys = structural_simplify(model)
+
+prob = ODEProblem(sys, [], (0., 3.))
+sol = solve(prob, Rosenbrock23())
+plot(sol)
+```
+
+## Explanation
+### Building the components
+For each component we use a Julia function that returns an `ODESystem`. At the top, we define the fundamental properties of a `Mass`: it has a mass `m`, a position `pos` and a velocity `vel`. We also define that the velocity is the rate of change of position with respect to time.
```
ChrisRackauckas, line 60:
```suggestion
For each component we use a Julia function that returns an `ODESystem`. At the top, we define the fundamental properties of a `Mass`: it has a mass `m`, a position `pos` and a velocity `v`. We also define that the velocity is the rate of change of position with respect to time.
```

docs/src/tutorials/spring_mass.md
```diff
@@ -0,0 +1,229 @@
...
+
+```julia
+function Spring(; name, k = 1e4, l = 1.)
+    ps = @parameters k=k l=l
+    @variables x(t), dir[1:2](t)
+    ODESystem(Equation[], t, [x, dir...], ps; name)
+end
+```
+
+We now define functions that help construct the equations for a mass-spring system. First, the `connect_spring` function connects a `spring` between two positions `a` and `b`. Note that `a` and `b` can be the `pos` of a `Mass`, or just a fixed position such as `[0., 0.]`.
```
ChrisRackauckas, line 93: In that sense, the length of the spring would be given by ... while ...

The next equations just need a bit more explanation in case someone doesn't catch it's just length and direction. Just a quick reinforcement.

ChrisRackauckas (approved):
Just that one reinforcement and I think it's good

AayushSabharwal:
What about the plot(s)?

ChrisRackauckas:
Just use the link to them from your PR's comment: 

https://user-images.githubusercontent.com/23384717/130322185-52ff1523-4ad8-4b24-94d3-3aa2c4a87082.png
https://user-images.githubusercontent.com/23384717/130322197-cff35eb7-0739-471d-a3d9-af83d87f1cc7.png
https://user-images.githubusercontent.com/23384717/130322202-efe967e6-7a44-43b0-aeec-8d8d54e993e5.png

ChrisRackauckas:
Use `name`

AayushSabharwal:
I'll add the first one at the top, and modify the last couple of lines to plot the second one then

ChrisRackauckas:
Thanks!

AayushSabharwal:
Happy to help!

I'm not sure why it says I've started a review. Might just be on my screen.

ChrisRackauckas:
https://github.com/AayushSabharwal/MTK-Exploration/

plot1
plot2
plot3

Putting the other pretty plots here for posterity.

Issue #1232, opened 2021-08-25
StructuralIdentifiability tutorial?


ChrisRackauckas:
@iliailmer it would be good to get a tutorial here showcasing the StructuralIdentifiability.jl integration.

iliailmer:
Yep, working on it. There is more thing to fix in StructuralIdentifiability.jl.

PR #1243, opened 2021-08-29
Identifiability tutorial

iliailmer:
Addressing #1232


docs/src/tutorials/parameter_identifiability.md
```diff
@@ -0,0 +1,165 @@
...
+    D(x3) ~ γ * x2 - δ * x3,
+    D(x4) ~ σ * x4 * (γ * x2 - δ * x3)/x3
+]
+
+observed = [
+    y~x1
+]
+
+# no inputs
+inputs = []
```
ChrisRackauckas, line 117: that should get handled via the metadata instead.

docs/src/tutorials/parameter_identifiability.md
```diff
@@ -0,0 +1,165 @@
...
+# indicate inputs
+inputs = [u1, u2]
+
+# check only 2 parameters
+to_check = [b, c]
+
+ode = ODESystem(eqs, t, [x1, x2, x3, x4], [b, c, α, β, γ, δ, σ], observed=observed, name=:GoodwinOsc)
+
+@time global_id = assess_identifiability(ode, inputs, to_check, 0.9)
+```
```
ChrisRackauckas, line 155: show the result

docs/src/tutorials/parameter_identifiability.md
```diff
@@ -0,0 +1,165 @@
...
+inputs = []
+
+# check all parameters
+to_check = []
+
+ode = ODESystem(eqs, t, [x1, x2, x3, x4], [b, c, α, β, γ, δ, σ], observed=observed, name=:GoodwinOsc)
+
+@time global_id = assess_identifiability(ode, inputs, to_check, 0.99)
+```
+
```
ChrisRackauckas, line 126: show the result

docs/src/tutorials/parameter_identifiability.md
```diff
@@ -0,0 +1,165 @@
...
+to_check = []
+
+# query local identifiability
+# we pass the ode-system
+local_id_all = assess_local_identifiability(de, inputs, to_check, 0.99)
+
+# let's try to check specific parameters and their combinations
+to_check = [k5, k7, k10/k9, k5+k6]
+local_id_some = assess_local_identifiability(de, inputs, to_check, 0.99)
+
```
ChrisRackauckas, line 74: show the result

docs/src/tutorials/parameter_identifiability.md
iliailmer: You mean via `ODESystem`? Does it accept input functions?

ChrisRackauckas: https://github.com/SciML/ModelingToolkit.jl/pull/1208/files#diff-6bdc4e7000e8945a6acaed152fc6e77aa61f661ea4e66795d9e432d178ac5b0bR287

iliailmer:
@ChrisRackauckas 👋 how does this look?

ChrisRackauckas:
That looks really good. But is there a way we can make `assess_local_identifiability(de, data_series, to_check, 0.99)`? The issue is that in some cases like pharmaceutical models the model definition might not be the place where the data is known, i.e. knowing that `y1 ~ x1` with `y1` output may be delayed. I am wondering a bit about the assumptions in there: is it assuming that we have a time series for each output? The outputs of a model can be different from the observed series which I guess could be another issue.

iliailmer:
The assumption is that at any point in time we know `y1(t) ~ g(x, p, u)` which is a function of states, parameters and inputs. The output is assumed to be without noise for determining structural identifiability. Would the `data_series` argument serve as a specified output value?

ChrisRackauckas:
>The assumption is that at any point in time we know y1(t) ~ g(x, p, u) which is a function of states, parameters and inputs. The output is assumed to be without noise for determining structural identifiability. Would the data_series argument serve as a specified output value?

Yes. That's not generally assumed of an output variable in a modeling environment (Simulink, Modelica, ModelingToolkit), so people will mark things as output variables for different reasons which have more to do with the causal connection structures. That would mean that in order for people to run identifiability analysis they would need to rewrite the model which isn't the best. So we should split the idea of output values and observables, and so here `data_series` is corresponding to the observables for which at any point in time the value is known. That would allow people to take any existing model, state which time series they have, and learn parameter identifiability without changing any model code.

iliailmer:
So instead of designating `y1 [output=true]` it is better to have a separate array containing required observed functions like `data_series = [y1~x1+x2]` and call `assess_identifiability(de, data_series)`. Is my understanding here correct?

In the terminology of identifiability analysis `y`-functions are commonly called outputs, and observability is a model property which is why I got a bit confused.

ChrisRackauckas:
>So instead of designating y1 [output=true] it is better to have a separate array containing required observed functions like data_series = [y1~x1+x2] and call assess_identifiability(de, data_series). Is my understanding here correct?

Yeah that seems necessary. 

>In the terminology of identifiability analysis y-functions are commonly called outputs, and observability is a model property which is why I got a bit confused.

I think it's just a domain thing. People use words differently. Sometimes it collides 😅

docs/src/tutorials/parameter_identifiability.md
```diff
@@ -140,23 +142,23 @@ Let us consider the same system but with two inputs and we will try to find out
...
 
 eqs = [
     D(x1) ~ -b * x1 + 1/(c + x4),
     D(x2) ~ a * x1 - beta * x2 - u1,
     D(x3) ~ g * x2 - delta * x3 + u2,
-    D(x4) ~ sigma * x4 * (g * x2 - delta * x3)/x3,
-    y~x1
+    D(x4) ~ sigma * x4 * (g * x2 - delta * x3)/x3
 ]
+measured_quantities = [y~x1]
```
ChrisRackauckas, line 154: We should probably showcase in one of them something more complex, like `measured_quantities = [y~x1+x3]`

ChrisRackauckas:
Beautiful. Thanks for seeing this through @iliailmer.

Trying to run it locally, does Singular.jl not work on Windows?

```julia
julia> using Singular
[ Info: Precompiling Singular [bcd08a7b-43d2-5ff7-b6d4-c458787f915c]
ERROR: LoadError: UndefVarError: libsingular not defined
Stacktrace:
 [1] getproperty(x::Module, f::Symbol)
   @ Base .\Base.jl:35
 [2] top-level scope
   @ C:\Users\accou\.julia\packages\Singular\nGiup\src\Singular.jl:66
 [3] include
   @ .\Base.jl:418 [inlined]
 [4] include_package_for_output(pkg::Base.PkgId, input::String, depot_path::Vector{String}, dl_load_path::Vector{String}, load_path::Vector{String}, concrete_deps::Vector{Pair{Base.PkgId, UInt64}}, source::Nothing)
   @ Base .\loading.jl:1318
 [5] top-level scope
   @ none:1
 [6] eval
   @ .\boot.jl:373 [inlined]
 [7] eval(x::Expr)
   @ Base.MainInclude .\client.jl:453
 [8] top-level scope
   @ none:1
in expression starting at C:\Users\accou\.julia\packages\Singular\nGiup\src\Singular.jl:1
ERROR: Failed to precompile Singular [bcd08a7b-43d2-5ff7-b6d4-c458787f915c] to C:\Users\accou\.julia\compiled\v1.7\Singular\jl_72EC.tmp.
Stacktrace:
 [1] error(s::String)
   @ Base .\error.jl:33
 [2] compilecache(pkg::Base.PkgId, path::String, internal_stderr::IO, internal_stdout::IO, ignore_loaded_modules::Bool)
   @ Base .\loading.jl:1466
 [3] compilecache(pkg::Base.PkgId, path::String)
   @ Base .\loading.jl:1410
 [4] _require(pkg::Base.PkgId)
   @ Base .\loading.jl:1120
 [5] require(uuidkey::Base.PkgId)
   @ Base .\loading.jl:1013
 [6] require(into::Module, mod::Symbol)
   @ Base .\loading.jl:997
```

docs/src/tutorials/parameter_identifiability.md
iliailmer: with this function the program takes longer and it's not ready to merge yet 😄

iliailmer: (basically, the output written right now in comments is wrong)

ChrisRackauckas:
Can you test the DSL-free form? This is the grand daddy where the whole thing should come together:

```julia
using StructuralIdentifiability, OrdinaryDiffEq
function f(du,u,p,t)
  b,c,a,beta,g,delta,sigma = p
  du[1] = -b * u[1] + 1/(c + u[4]),
  du[2] = a * u[1] - beta * u[2],
  du[3] = g * u[2] - delta * u[3],
  du[4] = sigma * u[4] * (g * u[2] - delta * u[3])/u[3]
end
prob = ODEProblem(f,ones(4),(0.0,1.0),ones(7))
us = states(sys)
measured_quantities = [y~us[1]]
@time global_id = assess_identifiability(ode, measured_quantities=measured_quantities)
                # 28.961573 seconds (88.92 M allocations: 5.541 GiB, 4.01% gc time)
                # Dict{Num, Symbol} with 7 entries:
                #   c     => :globally
                #   a     => :nonidentifiable
                #   g     => :nonidentifiable
                #   delta => :locally
                #   sigma => :globally
                #   beta  => :locally
                #   b     => :globally
```

Even though it's just a party trick, it's a rather useful party trick because now parameter identifiability (at least in a limited form) can be applied to any ODEProblem in Julia!!!

iliailmer:
> Beautiful. Thanks for seeing this through @iliailmer.
> 
> Trying to run it locally, does Singular.jl not work on Windows?

Only via Windows subsystem for Linux I think, they have instructions [here](https://oscar.computeralgebra.de/install/) for  installation

> Can you test the DSL-free form?

I just ran it and the problem is the non-ASCII characters such as x₁(t) that are not supported by Singular. Is there a way to convert them?

YingboMa:
You can try to normalize all names by the dictionary `Dict(t=>i for (i, t) in enumerate(names))`.

iliailmer:
Hey, @YingboMa, I'm still a bit confused about the conversion. I tried `Symbolics.variable` for state variables but this does not work with `parameters` of the mtk system.

YingboMa:
What do you mean by not work?

iliailmer:
Here is the code I have tried, fails for the same reason

```julia
using StructuralIdentifiability, OrdinaryDiffEq, ModelingToolkit
function f(du,u,p,t)
  b,c,a,beta,g,delta,sigma = p
  du[1] = -b * u[1] + 1/(c + u[4])
  du[2] = a * u[1] - beta * u[2]
  du[3] = g * u[2] - delta * u[3]
  du[4] = sigma * u[4] * (g * u[2] - delta * u[3])/u[3]
end
prob = ODEProblem(f,ones(4),(0.0,1.0),ones(7))
us = states(sys)
params = parameters(sys)
@variables t y(t)

# create new states/parameters
@variables x[1:length(us)](t) 
@parameters mu[1:length(params)]

# mapping to replace variables and states in system
states_map = Dict(s=>x[i] for (i, s) in enumerate(us))
param_map = Dict(p=>mu[i] for (i, p) in enumerate(params))

# new equations
eqs = substitute(substitute(equations(sys), states_map), param_map)

# new system
sys = ODESystem(eqs, t, name=:Name)
measured_quantities = [y~states_map(us[1])]

# error because of singular again
@time global_id = assess_identifiability(sys, measured_quantities=measured_quantities)
```

is there a way to call `@parameters` or `@variables` on a string?

ChrisRackauckas:
Let's merge and continue to improve. Thanks @iliailmer!

PR #1248, opened 2021-09-03
Update version for compat entry

bolognam:
Creating a version so f8a7f3b can be its own version of MTK. Looks like @shashi is taking v6.4.9 so I'll take v6.4.8.1. @YingboMa is responsible for anything that breaks :)


bolognam:
Actually, @shashi  can I take v6.4.9? CI doesn't like v6.4.8.1.

bolognam:
@shashi backported what I need in SymbolicUtils.jl, so I don't need the compat entry changed anymore.

PR #1251, opened 2021-09-06
Fix docstring

ChrisRackauckas:


ChrisRackauckas:
Failure on master @shashi ? https://github.com/SciML/ModelingToolkit.jl/pull/1251/checks?check_run_id=3527724844#step:6:522

PR #1258, opened 2021-09-10
Try out new Symbolics

shashi:
and SymbolicUtils.


shashi:
@paulflang SBMLToolkit is failing here. Let me know if you think there's something wrong with Symbolics that's causing it..

paulflang:
Hmm. I think the error is in this line https://github.com/SciML/SBMLToolkit.jl/blob/8c8f08cc4e06b46630b5a7d2971e282e266ef2c2/src/reactionsystem.jl#L235

My suspicion is that the new Symbolics either changes the species that are occurring in the kineticLaw `kl` or the `reactants` but not both. I am not in the picture of the changes in Symbolics. @anandijain : can you look into this?

anandijain:
I'm not positive, but it seems partially due to `(k1*s1)/s1` getting simplified in SU 0.13 whereas with this PR it doesn't seem to get reduced, even when I add a simplify call. 

What is the recommended way to do this rewrite? Do I just add my own rule?

PR #1260, opened 2021-09-11
Add autodetect kwarg that forces auto states detection

YingboMa:


YingboMa:
This is probably not that useful. Let's hold on to this though

YingboMa:
This doesn't seem to be needed any longer

PR #1268, opened 2021-09-16
add nicer discrete sys -> prob constructor

anandijain:
This makes it a bit easier to use DiscreteSystem/Problem, since you can now just do `DiscreteProblem(sys, [], tspan)` and it gets the defaults for you. 

A few things I was wondering about is how best to generalize the functions that internally hard-code `Differential`, such as 
`iv_from_nested_derivative`, `collect_vars!` and `var_from_nested_differential`. 

I was thinking maybe I should replace collect_vars! like so:

```julia 
function collect_vars!(states, parameters, expr, iv; T=Differential)
    if T == Differential
        f = var_from_nested_derivative 
    elseif T == Difference
        f = var_from_nested_difference
    end

    if expr isa Sym
        collect_var!(states, parameters, expr, iv)
    else
        for var in vars(expr)
            if istree(var) && operation(var) isa T
                var, _ = f(var)
            end
            collect_var!(states, parameters, var, iv)
        end
    end
    return nothing
end
```
which wouldn't be breaking and would avoid duplicated code.

I was hoping to do a loop + `@eval` for `iv_from_nested`, `var_from`, and a few others (`isdifferenceeq`, etc) like so:
```julia 
for T in [:Difference, :Differential]
    fname = lowercasefirst(string(T))
    f = Symbol(:iv_from_nested_, fname)
    f2 = Symbol(:var_from_nested_, fname)

    f3 = Symbol(:is, fname)
    f4 = Symbol(:is, fname, :eq)


    @eval begin 
        $f(x::Term) = operation(x) isa $T ? $f(arguments(x)[1]) : arguments(x)[1]
        $f(x::Sym) = x
        $f(x) = missing

        $f2(x::Term,i=0) = operation(x) isa $T ? $f2(arguments(x)[1], i + 1) : (x, i)
        $f2(x, i=0) = (missing, missing)
        $f2(x::Sym,i=0) = (x, i)

        $f3(expr) = istree(expr) && operation(expr) isa $T
        $f4(eq) = $f3(eq.lhs)
    end
end
```

I guess maybe this is worse than just having duplicate code, since actually doing the above would require a lot of annoying stuff

### DIFF
--- a/src/systems/discrete_system/discrete_system.jl
+++ b/src/systems/discrete_system/discrete_system.jl
@@ -45,22 +45,17 @@ struct DiscreteSystem <: AbstractTimeDependentSystem
     """
     systems::Vector{DiscreteSystem}
     """
-    default_u0: The default initial conditions to use when initial conditions
-    are not supplied in `DiscreteSystem`.
+    defaults: The default values to use when initial conditions and/or
+    parameters are not supplied in `DiscreteProblem`.
     """
-    default_u0::Dict
-    """
-    default_p: The default parameters to use when parameters are not supplied
-    in `DiscreteSystem`.
-    """
-    default_p::Dict
-    function DiscreteSystem(discreteEqs, iv, dvs, ps, var_to_name, ctrls, observed, name, systems, default_u0, default_p; checks::Bool = true)
+    defaults::Dict
+    function DiscreteSystem(discreteEqs, iv, dvs, ps, var_to_name, ctrls, observed, name, systems, defaults; checks::Bool = true)
         if checks
             check_variables(dvs, iv)
             check_parameters(ps, iv)
             all_dimensionless([dvs;ps;iv;ctrls]) ||check_units(discreteEqs)
         end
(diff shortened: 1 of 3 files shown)


src/systems/discrete_system/discrete_system.jl
```diff
@@ -118,16 +154,37 @@ function DiffEqBase.DiscreteProblem(sys::DiscreteSystem,u0map,tspan,
     ps = parameters(sys)
     eqs = equations(sys)
     eqs = linearize_eqs(sys, eqs)
-    # defs = defaults(sys)
-    t = get_iv(sys)
-    u0 = varmap_to_vars(u0map,dvs)
+    defs = defaults(sys)
+    iv = get_iv(sys)
+
+    if parammap isa Dict
```
anandijain, line 160: i ought to make this a utility function, since it gets used many places, but i don't think it is required for merge

anandijain:
don't really understand the test failures, they may be unrelated. i forget, am i able to turn off fast-failing?

YingboMa (commented):
Try to merge master back into this branch and see if that helps.

anandijain:
@YingboMa tests passing

PR #1272, opened 2021-09-21
Add `connection_type` to `DiscreteSystem`

ven-k:
- Required for defining `@connector` and `connect`ing multiple DiscreteSystems


YingboMa (approved):

ven-k:
@YingboMa with the new reduction.jl tests I expect this to pass all tests. Is there is any other parameter I should add to the DiscreteSystem?

PR #1278, opened 2021-09-24
Make the return behavior of modified_states consistent

isaacsas:
This makes it self-consistent, and consistent with `get_variables!`.


ChrisRackauckas:
@iliailmer can you look into that StructuralIdentifiability failure? I think it's because `a/b` used to transform into `a*inv(b)`, but I guess now it's `a/b` (@shashi's change), but now that seems to cause a Nemo issue.

iliailmer:
@ChrisRackauckas Fixed it, working on the new tag

ChrisRackauckas:
Awesome thanks 😊

PR #1289, opened 2021-10-08
add isvariable for Nums

isaacsas:
`isparameter(x::Num)` is defined, and I've found myself reaching for `isvariable(x::Num)` at times too...


isaacsas:
NeuralPDE failure seems unrelated.

PR #1306, opened 2021-10-19
switch LightGraphs to Graphs

isaacsas:
This would require a new breaking release I guess, so perhaps best to merge when ready to go to 7.0.


isaacsas:
@ChrisRackauckas can we merge this with the Symbolics 4 changes for the next release?

ChrisRackauckas:
Good idea.

PR #1318, opened 2021-10-31
[WIP] Update Symbolics and SymbolicUtils version

0x0f0f0f:
NOTE: Symbolics 4.0.0 has not yet been released.


0x0f0f0f:
Tests are all green locally with latest commit from https://github.com/JuliaSymbolics/Symbolics.jl/pull/421 
Will notify here when Symbolics.jl 4.0.0 gets registered.

0x0f0f0f:
That segfault in StructuralIdentifiability.jl looks like something I'm not able to fix myself, I'll need some help. @shashi @ChrisRackauckas

ChrisRackauckas:
Can you isolate the issue a bit?

0x0f0f0f:
> Can you isolate the issue a bit?

Testing locally.

0x0f0f0f:
> Can you isolate the issue a bit?
Cannot replicate on my machine.
StructuralIdentifiability.jl tests pass locally on my machine using this commit of MTK, latest MTK tag and StructuralIdentifiability.jl on latest master branch. Seems like the issue is due to `libgmp` in CI container.

```
 signal (11): Segmentation fault
in expression starting at /home/runner/work/ModelingToolkit.jl/ModelingToolkit.jl/downstream/test/identifiability.jl:1
__gmpn_addmul_1_coreisbr at /opt/hostedtoolcache/julia/1.6.3/x64/bin/../lib/julia/libgmp.so (unknown line)
Allocations: 196265728 (Pool: 196227157; Big: 38571); GC: 102
ERROR: LoadError: Package StructuralIdentifiability errored during testing (received signal: 11)
```

https://github.com/SciML/ModelingToolkit.jl/runs/4073261736?check_suite_focus=true

ChrisRackauckas:
What's the function call that causes a segfault?

0x0f0f0f:
> What's the function call that causes a segfault?

Doesn't show in the logs, only shows `Pkg` testing errors. Cannot find out since I cannot replicate locally
https://github.com/SciML/ModelingToolkit.jl/runs/4073261736?check_suite_focus=true#step:6:653

ChrisRackauckas:
What is the thing right before it in the tests, and after?

0x0f0f0f:
> What is the thing right before it in the tests, and after?

Right before:
```
[ Info: Assessing global identifiability using the coefficients of the io-equations
```
Third time that this line appears in the log. There's nothing after.

ChrisRackauckas:
No, what piece of code is ran that generates the segfault?

0x0f0f0f:
> What is the thing right before it in the tests, and after?

Judging from where the logs stop, the segfault should be happening here
https://github.com/SciML/StructuralIdentifiability.jl/blob/5dade8f547b3c68d527c7e67197f49a17bfa1948/src/global_identifiability.jl#L224-L225

ChrisRackauckas:
Isolate that to an MWE. I think it must be due to the conversion of the IR.

0x0f0f0f:
> Isolate that to an MWE. I think it must be due to the conversion of the IR.

Any suggestions on how to proceed knowing that the issue does not subsist on my local dev environment? I'll try replicating the downstream test in a docker container with https://github.com/nektos/act

ChrisRackauckas:
So when you run the tests it works?

0x0f0f0f:
> So when you run the tests it works?

Yes. Running on void linux. Using glibc. gmp v6.2.1_1 and gmpxx v6.2.1_1. Julia v1.6.1.

ChrisRackauckas:
Newer Julia? v1.6.3 is what the tests will use. Double check if the dependencies in the Symbolics stack are all similar.

0x0f0f0f:
@ChrisRackauckas passes locally on v1.6.3. Testing github action in docker container

0x0f0f0f:
Passes in locally when running the github action in a docker container!!! Can you try retriggering the workflow?

0x0f0f0f:
Dependency version differences from the Github CI and the GH action test in docker container on my local machine are:
ArrayInterface v3.1.38 => v3.1.39
Parsers v2.1.1 => Parsers v2.1.2
VectorizationBase v0.21.19 => v0.21.21

0x0f0f0f:
> Dependency version differences from the Github CI and the GH action test in docker container on my local machine are: ArrayInterface v3.1.38 => v3.1.39 Parsers v2.1.1 => Parsers v2.1.2 VectorizationBase v0.21.19 => v0.21.21

Test passed. probably related to one of those packages. Should bump compat? @ChrisRackauckas

ChrisRackauckas:
Yeah bump the lower bound to avoid the issue.

0x0f0f0f:
lgtm? @ChrisRackauckas

Project.toml
```diff
@@ -45,7 +45,7 @@ Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
 
 [compat]
 AbstractTrees = "0.3"
-ArrayInterface = "2.8, 3.0"
+ArrayInterface = "2.8, 3.1.39"
```
ChrisRackauckas, line 48:
```suggestion
ArrayInterface = "3.1.39"
```

0x0f0f0f: thx

shashi:
We should make a breaking release of MTK after this, so that further downstream updates can be more cautious.

Issue #1311, opened 2021-10-22
Input semantics concrete prototype


baggepinnen:
This script contains a suggestion for how the input interface could look and feel. Two different systems are implemented, one with an input specified and another without. The function `is_locally_balanced` can determine early if a component is locally balanced, i.e., the number of (non input) states equals the number of equations. `@variables` that are denoted `input = true` are not counted as states and are allowed to be unbound until the `ODEProblem` constructor, at which point they must have been either bound to a state, or provided as inputs to `ODEProblem`

The prototype implementation of `is_locally_balanced` below works for the first two tests. For `is_locally_balanced(connected)` it fails due to the extra algebraic equation, so part of the work done in `structural_simplify` would have to be done before the check `length(equations(sys)) == nx`.

@YingboMa what do you think about this, does it align with your understanding of how the inputs should work?
 
```julia
using ModelingToolkit

function is_locally_balanced(sys)
    nstates = count(!ModelingToolkit.isinput, states(sys))
    length(equations(sys)) == nstates
end

function decay_unbalanced(; name) 
    @parameters t a=1
    @variables x(t) = 0 f(t) = 0 # no input specificaton
    D = Differential(t)
    ODESystem([D(x) ~ -a * x + f]; name = name)
end
function decay(; name)
    @parameters t a=1
    @variables x(t) = 0 f(t) = 0 [input=true] # one state, one input, one equation, the number of equations equals the number of states
    D = Differential(t)
    ODESystem([D(x) ~ -a * x + f]; name = name)
end


dub = decay_unbalanced(; name=:d)
db  = decay(; name=:d)

# Only the system with an input specified is locally balanced at this stage.
@test !is_locally_balanced(dub) # two states, one equation => unbalanced
@test is_locally_balanced(db)  # one state, one input, one equation => unbalanced

# When we try to simulate the systems, all inputs must be either bound to states or provided to the ODEProblem constructor
@test_throws UnbalancedSystemError ODEProblem(dub, Pair[], (0.0, 100.0)) # Cannot simulate an unbalanced system, input not provided
@test_throws UnbalancedSystemError ODEProblem(db, Pair[], (0.0, 100.0)) # Still cannot simulate because the input is unbound
@test ODEProblem(db, Pair[], (0.0, 100.0), inputs=[db.f = 1]) isa ODEProblem # With the input specified, it can be simulated


# Compose two systems together
@named decay1 = decay()
@named decay2 = decay()

@parameters t
[...]

Issue #1216, opened 2021-08-18
Behavior of `output`


dannys4:
I was talking with @YingboMa regarding #1209 and the behavior of any variables labelled as `output=true`. We discussed what the appropriate way to use them were, especially for computing the Jacobian with respect to a variable labelled as `output`. The Jacobian can only be easily computed if an `output` variable is explicit, i.e. can be an observed variable. This can create confusion in the workflow of MTK if `output` variables. Since we usually only have observed variables after `structural_simplify`, this is problematic in the case of a partial model (e.g. #1211).

What behavior is appropriate here? One idea is that we `structural_simplify` the partial model, but that is undesirable since it wouldn't produce a valid model. @ChrisRackauckas

dannys4:
@cadojo @JTaets @jonniedie @baggepinnen @MuradTroll @ronisbr
Pinging some people who had thoughts on control systems in prior issues

cadojo:
Hm, is there anything wrong with _requiring_ a structurally simplified system to use features like output equations?

Maybe instead of a hard requirement, we could simply search the observed equations for the outputs with each output jacobian call? Most users could simply call structural simplify, and the output equations would be automatically found. If for some reason a user _doesn't_ want to call structural simplify, then they could specify the output equations in the `observed` keyword argument in the problem constructor. 

Those are my knee-jerk reactions anyways!

dannys4:
The problem we were coming across is more like, what if a user has `y` decorated as output, then makes it implicit in the system (let's say they have equation `y = f(u,p,t,y)` included). You can call `structural_simplify`, but that probably won't help, and there's no way you're getting any output jacobian even after calling `structural_simplify`. What's a person to do?

ValentinKaisermayer:
Couldn't you test for this and throw an exception that the `ODESystem` can not be converted into an explicit form?

ValentinKaisermayer:
From a control engineering perspective it would be nice if one could test if the `ODESystem` is already a linear `StateSpace` model, a `NonlinearStateSpace` or `AffineStateSpace` model, similar to what [Mathematica](https://reference.wolfram.com/language/guide/ControlSystems.html) provides. And convert between them. 
This would also be a nice entry point for other packages that provide automatic controller tuning or other classical control approaches.

dannys4:
[...]

Issue #1111, opened 2021-07-12
Input & output systems


YingboMa:
From the discussion with @ChrisRackauckas last Friday.

It seems like controls are a subproblem of input and output systems. The outputs of controllers are inputs of original systems, while closed-loop control can be handled by inputs of the controllers as well. Various matrices relevant for linearization can be formulated in this language as well.

cadojo:
Does this mean that the `control` syntax introduced in #1059 should be replaced with subsystems?

YingboMa:
Yes.

cadojo:
Gotcha. I hope y'all don't mind my commenting on these issues. I'm interested in helping out with features like this – I'm learning a lot by working through `ModelingToolkit` descriptions of systems I've worked on in my studies. 

If a controller is an input-output system by itself, is there a way to specify "no equations", so it's just a pass-through system? I ask because I think there's value in being able to represent a "whole system" (connected inputs, dynamics, and outputs) with symbolic control variables, and symbolic output variables. The only way I see accomplishing this is by writing the input and output subsystems as static gains.

```julia
using ModelingToolkit

# Input "system" (no equations, inputs are passed directly through this sub-system)
@variables δₑ, δₜ # elevator and thrust inputs to a longitudinal aircraft model
inputs = NonlinearSystem(ModelingToolkit.Equation[], [δₑ, δₜ], [])

# Dynamics (the traditional ODE system that describes the longitudinal aircraft dynamics)
@parameters t
@variables x[1:4](t)
d = Differential(t)

"""Multiplies each argument by a random number, and sums the result into a scalar expression."""
random_dynamics(syms::AbstractVector...) = sum(map(x -> rand() * x, collect(syms))
eqs = [
  d(x[1]) ~ randn(4) * transpose(Vector(x)) + randn(2) * transpose([inputs.δₑ, inputs.δₜ]),
  d(x[2]) ~ randn(4) * transpose(Vector(x)) + randn(2) * transpose([inputs.δₑ, inputs.δₜ]),
  d(x[3]) ~ randn(4) * transpose(Vector(x)) + randn(2) * transpose([inputs.δₑ, inputs.δₜ]),
  d(x[4]) ~ randn(4) * transpose(Vector(x)) + randn(2) * transpose([inputs.δₑ, inputs.δₜ]),
]

# Output "system" (let's assume no equations here, the outputs ~are~ the states)
[...]

PR #1322, opened 2021-11-03
add functions to handle inputs and outputs

baggepinnen:
This PR adds a number of functions to retrieve inputs and outputs, filtered on whether or not they are internally connected within a composite system.

In the language used , `r,d,n` below are `unbound_inputs` while `e, P.u` are `bound_inputs`.
`y` is an `unbound_output` while `z, controller.u` are `bound_output`. 
```
                          d│                 n│
                           │                  │
             ┌────────┐    │   ┌────────┐     │
    r      e │        │ u  ▼   │        │ z   ▼  y
  ───────+───►  PID   ├────+───►   P    ├─────+─┬───
        -▲   │        │        │        │       │
         │   └────────┘        └────────┘       │
         │                                      │
         └──────────────────────────────────────┘
```
A connection between two variables in different namespaces, where one namespace is internal to the other, only binds the inner variable and not the outer, e.g., `sys.u ~ sys.inner.u` will bind `sys.inner.u`, but `sys.u` remains an unbound, external signal. This reflects how components are typically connected in a block diagram.

The main thing to consider for a reviewer is if the strategy of determining the bound status of a signal based on connections to outer subsystems is sound.

This PR is the first in a series addressing
- #1311 
- #1111 
- #1088 
- #1216

### DIFF
--- a/src/inputoutput.jl
+++ b/src/inputoutput.jl
@@ -158,3 +158,91 @@ end
 
 has_var(ex, x) = x ∈ Set(get_variables(ex))
 
+# Build control function
+
+"""
+    (f_oop, f_ip), dvs, p = generate_control_function(sys::AbstractODESystem, dvs = states(sys), ps = parameters(sys); implicit_dae = false, ddvs = if implicit_dae
+
+For a system `sys` that has unbound inputs (as determined by [`unbound_inputs`](@ref)), generate a function with additional input argument `in`
+```
+f_oop : (u,in,p,t)      -> rhs
+f_ip  : (uout,u,in,p,t) -> nothing
+```
+The return values also include the remaining states and parameters, in the order they appear as arguments to `f`.
+
+# Example
+```
+using ModelingToolkit: generate_control_function, varmap_to_vars, defaults
+f, dvs, ps = generate_control_function(sys, expression=Val{false}, simplify=true)
+p = varmap_to_vars(defaults(sys), ps)
+x = varmap_to_vars(defaults(sys), dvs)
+t = 0
+f[1](x, inputs, p, t)
+```
+"""
+function generate_control_function(
+    sys::AbstractODESystem;
+    implicit_dae=false,
+    has_difference=false,
+    simplify=true,
+    kwargs...
+)
+
+    ctrls = unbound_inputs(sys)
+    if isempty(ctrls)
+        error("No unbound inputs were found in system.")
+    end
+
+    # One can either connect unbound inputs to new parameters and allow structural_simplify, but then the unbound inputs appear as states :( .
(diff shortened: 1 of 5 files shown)


baggepinnen:
Determining if outputs are bound present a bit more challenge since they might be moved to `observed` equations. 
In the following case
```julia
julia> observed(syss)
2-element Vector{Equation}:
 sys₊y(t) ~ sys₊x(t)
 y(t) ~ sys₊x(t)
```
`sys₊y(t)` is bound to the outer `y(t)` through the variable `sys₊x(t)` and should thus return `is_bound(sys₊y(t)) = true`. Tips on how to do this in a general way are welcome.

src/inputoutput.jl
```diff
@@ -0,0 +1,160 @@
...
+"""
+    has_var(eq, x)
+
+Determine whether or not an equation or expression contains variable `x`.
+"""
+function has_var(eq::Equation, x)
+    has_var(eq.rhs, x) || has_var(eq.lhs, x)
+end
+
+has_var(ex, x) = x ∈ Set(get_variables(ex))
```
YingboMa, line 159:
```suggestion
has_var(ex, x) = occursin(x, ex)
```

src/inputoutput.jl
```diff
@@ -0,0 +1,160 @@
...
+    is_bound(sys, u)
+
+Determine whether or not input/output variable `u` is "bound" within the system, i.e., if it's to be considered internal to `sys`.
+A variable/signal is considered bound if it appears in an equation together with variables from other subsystems.
+The typical usecase for this function is to determine whether the input to an IO component is connected to another component,
+or if it remains an external input that the user has to supply before simulating the system. 
+
+See also [`bound_inputs`](@ref), [`unbound_inputs`](@ref), [`bound_outputs`](@ref), [`unbound_outputs`](@ref)
+"""
+function is_bound(sys, u, stack=[])
```
YingboMa, line 69: We should think about an abstraction for viewing hierarchical models as a graph.

YingboMa (approved):

baggepinnen:
@YingboMa I'm looking into the test failure which appears after rebasing onto master. It appears as if there is a new field `sys.substitutions` that must be taken into account now? I guess it's this PR that introduced the changes
https://github.com/SciML/ModelingToolkit.jl/pull/1414/files

baggepinnen:
Tests are passing locally by adding 
```julia
pre, sol_states = get_substitutions_and_solved_states(sys)
f = build_function(rhss, args...; postprocess_fbody=pre, states=sol_states, kwargs...)
```
based on https://github.com/SciML/ModelingToolkit.jl/pull/1414/files#diff-c0933801181b7b4a7b3693c7bcfba2763a7dff4cf6e88356760c571e30692b24R270
If this was an okay move, the PR should be good to go after green CI

PR #1324, opened 2021-11-04
Refactor Alias Elimination some more

Keno:
This is a follow up to #1301 to finish refactoring the alias elimination
code. It mostly puts the result of AliasElimination into an `AliasGraph`
datastructure, but while we're at it, it also improves the asymptotics
of alias elimination slightly (as noted possible in #1301). There's further
asymptotics improvement possible that I've marked in the code, but don't
intend to pursue for the moment unless alias elimination starts showing
up in benchmarks. This also fixes a number of small bugs: For example,
the iteration to convergence in the back substitution was previously
an accidental infinite loop. Additionally, it should be fully able to
handle integer coefficients other than (1, -1), though this capability
is not currently enabled.

### DIFF
--- a/src/systems/alias_elimination.jl
+++ b/src/systems/alias_elimination.jl
@@ -7,36 +7,29 @@ include("compat/bareiss.jl")
 function alias_elimination(sys)
     sys = initialize_system_structure(sys; quick_cancel=true)
     s = structure(sys)
-    is_linear_equations, eadj, cadj = find_linear_equations(sys)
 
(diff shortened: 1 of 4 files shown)


src/systems/alias_elimination.jl
```diff
@@ -52,7 +45,7 @@ function alias_elimination(sys)
     newstates = []
     sts = states(sys)
     for j in eachindex(fullvars)
-        if isirreducible(v_types, j)
+        if !(j in keys(ag))
```
YingboMa, line 48: Ha, pretty cool design.

YingboMa (approved):

Issue #38, opened 2018-03-05
Support for events


AleMorales:
State and time-triggered events are common in dynamical models, so modelling frameworks that generate SciComp IR may want to pass this information along, rather than hook the callbacks after code generation. After all, events are part of a model definition even if they are stored separately. I don't how abstract the event support should be, maybe for now we could allow storing a `CallbackSet`?

ChrisRackauckas:
While the goal isn't to re-create Modia.jl at all (if it ever gets released), this is another spot where I think we should look to Modia/Modelica for inspiration, along with Tensorflow.

https://link.springer.com/chapter/10.1007%2F978-3-319-47169-3_15

They treat if statements specially and make people opt-out with a `NoEvent()`.

I have been thinking of doing something like `ContinuousEvent(x>y,rho = 3)` and one for `DiscreteEvent`. We could add to this repo a few `<:AbstractDiscreteEvent`s, like `PeriodicEvent` and `TimedEvent` (whose implementation in DiffEq would of course just be via the appropriate callbacks in DiffEqCallbacks.jl). They should allow for an initialization as well. Then we can add an `@continuous` etc. macro which essentially just expands to creating a parameter, creating a ContinuousEvent which changes the parameter (and initializes it appropriately), and if there are any equations in there it would make an `ifelse` operation. i.e.

```julia
@continuous if x > 5
   D*x == x*y
else
   D*x == x - 3
end
```

would expand to:

```julia
Parameter(:if1) # Some gensym name
ContinuousEvent(x>5,if1==1,if1==0,if1==x>5) # condition, upcrossing, downcrossing, initialization
ifelse(if1,D*x == x*y,D*x == x - 3)
```

It probably needs some bikeshedding to handle more combinations of models well, but you can see where this is going.

ChrisRackauckas:
Now that https://github.com/JuliaDiffEq/SciCompDSL.jl/issues/41 is through, it's clear that a `ContinuousEvent` should be defined via a rootfinding equation with `~`. But inequality equations are necessary and I'm not sure about the syntax.

ChrisRackauckas:
Welp, answered myself. 

https://github.com/JuliaDiffEq/SciCompDSL.jl/issues/43

And that's all we needed. Now all of the conditionals are `Operations` which specify equality or inequality equations. This means:

```julia
@continuous if x > 5
   D*x ~ x*y
else
   D*x ~ x - 3
end
```

gives

```julia
[...]

PR #1337, opened 2021-11-08
Include root-finding equations in `ODESystem`

baggepinnen:
This PR adds a vector of equations to an `ODESystem` that can be used for root-finding when solving an `ODEProblem`.
The intended usage is to specify such equations when an ODESystem is created rather than when the `ODEProblem` is created, for example, when a component in a component library contains discountinous dynamics etc.

The provided tests pass locally, but there are a number of points I would like to get feedback on
1. The `root_eqs` field was added to `ODESystem` only, how should all other `AbstractSystems` be treated?
2. I wasn't able to use the inplace update for `VectorContinuousCallback` [here](https://github.com/SciML/ModelingToolkit.jl/compare/master...baggepinnen:rootfuns?expand=1#diff-7195a73d9d71f7a993fe71086614d700629ebc22306be7b602f60fbc054bcdb7R184), can the `RuntimeGeneratedFunction` somehow be generated in an inplace form as well?

Fixes https://github.com/SciML/ModelingToolkit.jl/issues/38

### DIFF
--- a/src/systems/diffeqs/abstractodesystem.jl
+++ b/src/systems/diffeqs/abstractodesystem.jl
@@ -156,22 +156,15 @@ end
 function generate_rootfinding_callback(sys::ODESystem, dvs = states(sys), ps = parameters(sys); kwargs...)
     cbs = continuous_events(sys)
     isempty(cbs) && return nothing
-    callbacks = map(cbs) do eq_aff
-        generate_rootfinding_callback(eq_aff, sys, dvs, ps; kwargs...)
-    end
-    callbacks = filter(cb->cb !== nothing, callbacks)
-    if isempty(callbacks)
-        return nothing
-    elseif length(callbacks) == 1
-        return callbacks[]
-    else
-        CallbackSet(callbacks...)
-    end
+    generate_rootfinding_callback(cbs, sys, dvs, ps; kwargs...)
 end
 
-function generate_rootfinding_callback(eq_aff::SymbolicContinuousCallback, sys::ODESystem, dvs = states(sys), ps = parameters(sys); kwargs...)
-    eqs = eq_aff.eqs
+function generate_rootfinding_callback(cbs, sys::ODESystem, dvs = states(sys), ps = parameters(sys); kwargs...)
+    eqs = map(cb->cb.eqs, cbs)
+    num_eqs = length.(eqs)
     isempty(eqs) && return nothing
+    # fuse equations to create VectorContinuousCallback
+    eqs = reduce(vcat, eqs)
     # rewrite all equations as 0 ~ interesting stuff
     eqs = map(eqs) do eq
         isequal(eq.lhs, 0) && return eq
@@ -180,13 +173,16 @@ function generate_rootfinding_callback(eq_aff::SymbolicContinuousCallback, sys::
 
     rhss = map(x->x.rhs, eqs)
     root_eq_vars = unique(collect(Iterators.flatten(map(ModelingToolkit.vars, rhss))))
-
+    
     u = map(x->time_varying_as_func(value(x), sys), dvs)
     p = map(x->time_varying_as_func(value(x), sys), ps)
     t = get_iv(sys)
     rf_oop, rf_ip = build_function(rhss, u, p, t; expression=Val{false}, kwargs...)
-
-    affect = compile_affect(eq_aff, sys, dvs, ps; kwargs...)
+    
+    affect_functions = map(cbs) do cb # Keep affect function separate
+        eq_aff = affect_equations(cb)
+        affect = compile_affect(eq_aff, sys, dvs, ps; kwargs...)
+    end
 
     if length(eqs) == 1
         cond = function(u, t, integ)
@@ -198,10 +194,22 @@ function generate_rootfinding_callback(eq_aff::SymbolicContinuousCallback, sys::
                 rf_oop(u, integ.p, t)
             end
         end
-        ContinuousCallback(cond, affect)
+        ContinuousCallback(cond, affect_functions[])
     else
         cond = function(out, u, t, integ)
             rf_ip(out, u, integ.p, t)
+        end
+
+        # since there may be different number of conditions and affects,
+        # we build a map that translates the condition eq. number to the affect number
+        eq_ind2affect = reduce(vcat, [fill(i, num_eqs[i]) for i in eachindex(affect_functions)])
+        @assert length(eq_ind2affect) == length(eqs)
+        @assert maximum(eq_ind2affect) == length(affect_functions)
+
+        affect = let affect_functions=affect_functions, eq_ind2affect=eq_ind2affect
+            function(integ, eq_ind) # eq_ind refers to the equation index that triggered the event, each event has num_eqs[i] equations
+                affect_functions[eq_ind2affect[eq_ind]](integ)
+            end
         end
         VectorContinuousCallback(cond, affect, length(eqs))
     end
@@ -209,6 +217,12 @@ end
 
 compile_affect(cb::SymbolicContinuousCallback, args...; kwargs...) = compile_affect(affect_equations(cb), args...; kwargs...)
 
+"""
+    compile_affect(eqs::Vector{Equation}, sys, dvs, ps; kwargs...)
+    compile_affect(cb::SymbolicContinuousCallback, args...; kwargs...)
+
+Returns a function that takes an integrator as argument and modifies the state with the affect.
+"""
 function compile_affect(eqs::Vector{Equation}, sys, dvs, ps; kwargs...)
     if isempty(eqs)
         return (args...) -> () # We don't do anything in the callback, we're just after the event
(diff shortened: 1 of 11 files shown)


YingboMa:
RuntimeGeneratedFunction handles arbitrary functions without closures in-place or not. Try to read `build_function`'s implementation in Symbolics to see how in-place functions are expressed.

baggepinnen:
I changed the implementation to use `build_function` instead of `gen_nlsolve` and it appears to work well. The remaining problem is related to other system types, e.g., `NonlinearSystem` does not have any `root_eqs`. The code in MTK assumes that all system types have the same fields in many places, for instance here https://github.com/SciML/ModelingToolkit.jl/pull/1337/files#diff-287e017b6c265948a34862d11b63bade810cb38f11173644350028d4ba836661R989
One can either introduce the field `root_eqs` for other system types, or add the keyword argument to their constructors, check that the value of the argument is empty/nothing. I took the latter route in https://github.com/SciML/ModelingToolkit.jl/pull/1337/commits/ce89be96ecab10138cb895eafa79c4f3f2da2e74

baggepinnen:
After offline discussions, it was decided that we should add `affect!`s to the interface as well. The commits above does this, but requires the user to write the `affect!` function using the standard integrator interface, which is inconvenient. 

The remaining step is to allow symbolic affect expressions that are automatically compiled to an `affect!` function operator on the integrator interface.

baggepinnen:
One can now model events including effects like so
```julia
## Test bouncing ball with equation affect
@variables t x(t)=1 v(t)=0
D = Differential(t)

root_eqs = [x ~ 0]
affect   = [v ~ -v]

@named ball = ODESystem([
    D(x) ~ v
    D(v) ~ -9.8
], t, events = root_eqs => affect)

@test getfield(ball, :events)[] == EqAffectPair(Equation[x ~ 0], Equation[v ~ -v])
ball = structural_simplify(ball)

tspan = (0.0,5.0)
prob = ODEProblem(ball, Pair[], tspan)
sol = solve(prob,Tsit5())
@test 0 <= minimum(sol[x]) <= 1e-10 # the ball never went through the floor but got very close
# plot(sol)

**Comment by ChrisRackauckas** [comment]
While we're at it, to finish https://github.com/SciML/ModelingToolkit.jl/issues/38 we should have a way of designating an event condition in the discrete sense as well.

**Comment by isaacsas** [comment]
Could the symbolic conditions and affects be stored in the current continuous and discrete callbacks provided by DiffEqBase? It might be nice to reuse those structures since users are already familiar with them.

**Comment by isaacsas** [comment]
I meant that users crest and pass the symbolic callbacks using the same structures (but built with symbolic equations), and use `callback` as a kwarg to `ODESystem`s (as opposed to the `events` kwarg used currently and it’s format). This would keep the symbolic workflow the same as the DufferentialEquations.jl workflow from a user’s perspective.

**Comment by ChrisRackauckas** [comment]
@baggepinnen we should at least thing of the naming scheme so that way the next PR isn't immediately breaking. Is it `continuous_events` and `discrete_events`?

**Comment by baggepinnen** [comment]
`continuous_events` and `discrete_events` sound reasonable. The internal structure `EqAffectPair` can be renamed `SymbolicContinousCallback` as well to align more with the standard diffeq interface.

**Comment by ChrisRackauckas** [comment]
Sounds reasonable to me. Yeah, let's get that together and merge, and then follow up with the discrete.

File: src/systems/diffeqs/abstractodesystem.jl [hunk 1]
```diff
@@ -153,6 +153,90 @@ function generate_difference_cb(sys::ODESystem, dvs = states(sys), ps = paramete
...
+compile_affect(eqaff::EqAffectPair, args...; kwargs...) = compile_affect(affect_equations(eqaff), args...; kwargs...)
+
+function compile_affect(eqs::Vector{Equation}, sys, dvs, ps; kwargs...)
+    if isempty(eqs)
+        return (args...) -> () # We don't do anything in the callback, we're just after the event
+    else
+        rhss = map(x->x.rhs, eqs)
+        lhss = map(x->x.lhs, eqs)
+        update_vars = collect(Iterators.flatten(map(ModelingToolkit.vars, lhss))) # these are the ones we're chaning
+        length(update_vars) == length(update_vars) || error("affected variables not unique, each state can only be affected by one equation for a single `root_eqs => affects` pair.")
```
isaacsas, line 219: Should one of these `update_vars` be a `rhss`?

isaacsas:
Could you add a test with multiple affects for a given condition? i.e. changing more than one variable?

src/systems/diffeqs/abstractodesystem.jl
baggepinnen: Thanks, well spotted!

test/odesystem.jl
```diff
@@ -427,9 +427,9 @@ eqs = [
 # prob = ODEProblem(ODEFunction{false}(de),[1.0,1.0],(0.0,1.0),[1.5,1.0,3.0,1.0])
 
 prob = ODEProblem(de,[1.0,1.0],(0.0,1.0),[1.5,1.0,3.0,1.0], check_length=false)
-@test prob.kwargs[:difference_cb] isa ModelingToolkit.DiffEqCallbacks.DiscreteCallback
+@test prob.kwargs[:callback] isa ModelingToolkit.DiffEqCallbacks.DiscreteCallback
```
YingboMa, line 430: I think instead of relying on the exact symbol name here, we should have a function like `get_callback` to make it standard.

src/systems/diffeqs/abstractodesystem.jl
```diff
@@ -555,12 +640,24 @@ function DiffEqBase.ODEProblem{iip}(sys::AbstractODESystem,u0map,tspan,
...
+        event_cb = nothing
+    end
+    difference_cb = has_difference ? generate_difference_cb(sys; kwargs...) : nothing
+    cb = merge_cb(event_cb, difference_cb)
+
+    if cb === nothing
+        ODEProblem{iip}(f, u0, tspan, p; kwargs...)
     else
-        ODEProblem{iip}(f,u0,tspan,p;kwargs...)
+        ODEProblem{iip}(f, u0, tspan, p; callback=cb, kwargs...)
```
YingboMa, line 654: Let's check that `haskey(kwargs, :callback)` is false. Otherwise either user's input is silently ignored, or our generated callback is silently ignored.
```julia
julia> foo(;a) = a

julia> kwargs = (a = 100,)
(a = 100,)

julia> foo(;a=10, kwargs...)
100
```

baggepinnen: I merged user-provided callbacks with automatically generated ones

ChrisRackauckas:
How is this working when a system with an event is used as a subsystem in another system? Where is it appropriately renamespacing the events?

baggepinnen:
There is a test that the event equations have been renamespaced here
https://github.com/SciML/ModelingToolkit.jl/pull/1337/files#diff-4ff6b77a6fa947c70cfd3bcaf373b11d61ad04f2f1a74022ea9250e386459f74R133

and there is a method to renamespace events here
https://github.com/SciML/ModelingToolkit.jl/pull/1337/files#diff-287e017b6c265948a34862d11b63bade810cb38f11173644350028d4ba836661R188

ChrisRackauckas:
Missed it thanks.

src/systems/diffeqs/abstractodesystem.jl
```diff
@@ -153,6 +153,91 @@ function generate_difference_cb(sys::ODESystem, dvs = states(sys), ps = paramete
     PeriodicCallback(cb_affect!, first(dt))
 end
 
+function generate_rootfinding_callback(sys::ODESystem, dvs = states(sys), ps = parameters(sys); kwargs...)
+    cbs = continuous_events(sys)
+    isempty(cbs) && return nothing
+    callbacks = map(cbs) do eq_aff
+        generate_rootfinding_callback(eq_aff, sys, dvs, ps; kwargs...)
+    end
```
ChrisRackauckas, line 161: If you map here, won't it not fuse into a single VectorContinuousCallback?

baggepinnen: It only returns `VectorContinuousCallback` if there is an event with multiple equations. If there are several events, they are treated as separate callbacks. So this kind of event leads to `VectorContinuousCallback`
https://github.com/SciML/ModelingToolkit.jl/pull/1337/files#diff-4ff6b77a6fa947c70cfd3bcaf373b11d61ad04f2f1a74022ea9250e386459f74R194


while this kind
https://github.com/SciML/ModelingToolkit.jl/pull/1337/files#diff-4ff6b77a6fa947c70cfd3bcaf373b11d61ad04f2f1a74022ea9250e386459f74R132-R133
leads to a CallbackSet.

I kept them apart in order to make the following check 
https://github.com/SciML/ModelingToolkit.jl/pull/1337/files#diff-7195a73d9d71f7a993fe71086614d700629ebc22306be7b602f60fbc054bcdb7R219-R220

I couldn't prove for myself that there couldn't be bad outcomes from having one state affected by multiple equations in the same event since the ordering of the affect equations would then matter etc.

ChrisRackauckas: How do you make a VectorContinuousCallback which have a different affect based on the condition that is fired? Why not fuse anyways since it's more efficient?

ChrisRackauckas (approved):

ChrisRackauckas:
@YingboMa you got anything else?

YingboMa:
LGTM

tshort:
Cool stuff. Any chance it could be expanded to cover `affect_neg!`?

PR #1338, opened 2021-11-10
Refactor tearing code

Keno:
This refactors tearing to use a clean implementation of the BFGT15 Algorithm N,
together with the directed graph abstraction introduced in #1328. I believe
the asymptotics of the version I implemented in Graphs.jl are somewhat better
than the version we're replacing here from Modia, because the Modia versions
caches the level rather than recomputing it (thus potential performing
redundant updates) - however, this is probably unlikely to make a difference
in practice. That said, this version should also be significantly easier to
swap out for one of the more advanced cycle detection algorithms from the
literature. Algorithm N is tuned for dense graphs, which our MTK graphs are
unlikely to be, so we may want to switch to a sparse cycle detection algorithm
in the future.


Keno:
Guess, I should have mentioned in the commit message, this goes with https://github.com/JuliaGraphs/Graphs.jl/pull/36, so we'll need to get that merged first.

src/structural_transformation/bipartite_tearing/modia_tearing.jl
```diff
@@ -309,71 +23,39 @@ and so on). vTear must be selected, so that the equations eResidues are fulfille
...
 """
-function tearEquations!(td::TraverseDAG, Gsolvable, es::Vector{Int}, vs::Vector{Int};
+function tearEquations!(ict::IncrementalCycleTracker, Gsolvable, es::Vector{Int}, vs::Vector{Int};
                         eSolvedFixed::Vector{Int}=Int[], vSolvedFixed::Vector{Int}=Int[], vTearFixed::Vector{Int}=Int[])
-    vs2 = initAlgebraicSystem(td, es, vs, eSolvedFixed, vSolvedFixed, vTearFixed)
-    # eResidue = fill(0,0)
-    residue  = true
+    G = ict.graph
+    vActive = BitSet(vs)
+    vMatched = BitSet()
```
YingboMa, line 35: Since we already know the upperbound of the length of `vActive` and `vMatched`. Maybe just use a `BitVector`?

src/structural_transformation/bipartite_tearing/modia_tearing.jl
```diff
@@ -309,71 +23,39 @@ and so on). vTear must be selected, so that the equations eResidues are fulfille
...
-                    else
-                        # cycle; remove vj from DAG and undo its changes
-                        for vv in td.visitedStack
-                            td.level[vv] = td.lastlevel[vv]
-                        end
-                        td.assign[vj] = 0
-                        # continue with next variable in equation eq
-                    end
+            if !(vj in vMatched) && (vj in vActive)
+                r = add_edge_checked!(ict, Iterators.filter(!=(vj), G.graph.fadjlist[eq]), vj) do G
```
YingboMa, line 43: I often use `𝑠neighbors(graph, eq)`.

YingboMa (approved):

src/bipartite_graph.jl
```diff
@@ -280,50 +280,77 @@ end
...
+
+Graphs.inneighbors(g::DiCMOBiGraph{true}, v) = CMONeighbors{true}(g, v)
+function Base.iterate(c::CMONeighbors{true})
+    vsrc = c.g.matching[c.v]
+    vsrc === unassigned && return nothing
+    iterate(c, (c.g.graph.fadjlist[vsrc],))
+end
+function Base.iterate(c::CMONeighbors{true}, (l, state...))
+    while true
+        r = iterate(l, state...)
```
YingboMa, line 347: Is it safe to assume the implementation detail of `iterate(::Vector, ...)` in here?

src/structural_transformation/bipartite_tearing/modia_tearing.jl
Keno: `BitSet` is the correct abstraction here and `vActive` already gets the correct size from the iterator, but we can `sizehint!` the `vMatched` BitSet for a slight increase in efficiency.

src/structural_transformation/bipartite_tearing/modia_tearing.jl
Keno: Yes, that's better.

src/bipartite_graph.jl
Keno: Which internals are we relying on here?

src/structural_transformation/bipartite_tearing/modia_tearing.jl
Keno: Actually, on second thought `sizehint!` is probably bad, since we're likely to have only a couple of vertices actually active for small SCCs and if they happen to be next to each other, we don't want to allocate space for the whole vertex set.

src/bipartite_graph.jl
YingboMa: Ah, you are right. We are just forwarding the state here, but not relying on implementation details.

src/structural_transformation/bipartite_tearing/modia_tearing.jl
```diff
@@ -309,71 +23,39 @@ and so on). vTear must be selected, so that the equations eResidues are fulfille
...
-that can be explicitly solved in every equation without influencing the solution space
+Gsolvable defines the variables that can be explicitly solved in every equation without influencing the solution space
 (= rank preserving operation).
 
 eSolvedFixed/vSolvedFixed must be a DAG starting at eSolvedFixed/SolvedFixed[1]
 """
-function tearEquations!(td::TraverseDAG, Gsolvable, es::Vector{Int}, vs::Vector{Int};
+function tearEquations!(ict::IncrementalCycleTracker, Gsolvable, es::Vector{Int}, vs::Vector{Int};
                         eSolvedFixed::Vector{Int}=Int[], vSolvedFixed::Vector{Int}=Int[], vTearFixed::Vector{Int}=Int[])
-    vs2 = initAlgebraicSystem(td, es, vs, eSolvedFixed, vSolvedFixed, vTearFixed)
```
YingboMa, line 320: Seems like we are ignoring `vSolvedFixed`, currently. Could you include the ability to add pre-existing matchings? We should add an API to do that in https://github.com/JuliaGraphs/Graphs.jl/pull/36 as well.

YingboMa (approved):

Keno:
Updated to remove the `vSolvedFixed` etc keyword arguments which weren't used anywhere. I've also vendored the Graphs.jl change into this PR so that we can get this PR merged and make progress while we're finishing up the Graphs.jl PR, which may take a bit since the Graphs.jl re-vival is still not fully figured out.

PR #1343, opened 2021-11-15
add observed to var_to_name Dict

isaacsas:
Closes https://github.com/SciML/ModelingToolkit.jl/pull/1343


src/systems/diffeqs/odesystem.jl
```diff
@@ -137,6 +137,7 @@ function ODESystem(
     var_to_name = Dict()
     process_variables!(var_to_name, defaults, dvs′)
     process_variables!(var_to_name, defaults, ps′)
+    !isempty(observed) && collect_var_to_name!(var_to_name, (eq.lhs for eq in observed))
```
YingboMa, line 140:
```suggestion
    isempty(observed) || collect_var_to_name!(var_to_name, (eq.lhs for eq in observed))
```

YingboMa:
Do you have a use case for this? `observed` is not intended to be used by users.

isaacsas:
Certain bio file formats include "groups" which are observables of the states. Currently the ability to `@unpack` them does not propagate when converting `ReactionSystem`s to other system types (i.e. they don't unpack as the variable but as the full equation if I recall right).

isaacsas:
Examples include SBML, Antimony and BioNetGen.

YingboMa:
This is a breaking change, but sure, we can do that.

isaacsas:
OK, I can finish off the PR and leave it to merge for whenever the next breaking release is planned. I think it makes sense; one can unpack parameters and states, so it seems like one should be able to unpack observable variables too to work with them.

isaacsas:
@YingboMa I think this is good to go unless you have any other comments. Thanks!

YingboMa (approved):
LGTM

YingboMa:
Let's merge this in this next breaking release.

YingboMa:
Ah, merge error.

test/discretesystem.jl
```diff
@@ -112,3 +112,22 @@ linearized_eqs = [
     y(t - 2.0) ~ y(t)
 ]
 @test all(eqs2 .== linearized_eqs)
+
```
YingboMa, line 115: @isaacsas Sorry, this is a merge error. Could you delete this?

isaacsas: You mean to delete the extra test code the merge added? Sure, I’ll do that tonight when I have a bit of time free.

YingboMa:
Yes. Thanks!

isaacsas:
@YingboMa all removed. Should be good if tests pass.

PR #1347, opened 2021-11-17
BipartiteGraph: Clean up Graphs.jl integration

Keno:
Removes the `ALL` option for BipartiteGraph edges iteration.
This option makes little sense. The fadjlist and badjlist represent
the same edges just with two different index structures. It is up
to the BipartiteGraph to keep them internally consistent, but there's
no reason to allow iteration over both from outside - they should never
be inconsistent.

Also add a few more Graphs.jl integration to allow more interesting
graph algorithms to be run on our graph as we explore.

### DIFF
--- a/src/structural_transformation/bipartite_tearing/modia_tearing.jl
+++ b/src/structural_transformation/bipartite_tearing/modia_tearing.jl
@@ -35,6 +35,7 @@ function tearEquations!(ict::IncrementalCycleTracker, Gsolvable, es::Vector{Int}
             if G.matching[vj] === unassigned && (vj in vActive)
                 r = add_edge_checked!(ict, Iterators.filter(!=(vj), 𝑠neighbors(G.graph, eq)), vj) do G
                     G.matching[vj] = eq
+                    G.ne += length(𝑠neighbors(G.graph, eq)) - 1
                 end
                 r && break
             end
--- a/src/bipartite_graph.jl
+++ b/src/bipartite_graph.jl
@@ -3,7 +3,7 @@ module BipartiteGraphs
 export BipartiteEdge, BipartiteGraph, DiCMOBiGraph, Unassigned, unassigned
 
 export 𝑠vertices, 𝑑vertices, has_𝑠vertex, has_𝑑vertex, 𝑠neighbors, 𝑑neighbors,
-       𝑠edges, 𝑑edges, nsrcs, ndsts, SRC, DST
+       𝑠edges, 𝑑edges, nsrcs, ndsts, SRC, DST, set_neighbors!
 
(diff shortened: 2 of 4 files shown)


src/structural_transformation/bipartite_tearing/modia_tearing.jl
```diff
@@ -35,6 +35,7 @@ function tearEquations!(ict::IncrementalCycleTracker, Gsolvable, es::Vector{Int}
             if G.matching[vj] === unassigned && (vj in vActive)
                 r = add_edge_checked!(ict, Iterators.filter(!=(vj), 𝑠neighbors(G.graph, eq)), vj) do G
                     G.matching[vj] = eq
+                    G.ne += length(𝑠neighbors(G.graph, eq)) - 1
```
YingboMa, line 38: So one edge pointing to the dst is the matching. The rest of them point back to the src, and they are the edges of DiCMOBiGraph. It always takes me a bit to keep track of all these. Could you write some documentation on this? Maybe explain tearing a little bit as well, so all the information is here.

Keno: DiCMOBiGraph is documented here: https://github.com/SciML/ModelingToolkit.jl/blob/master/src/bipartite_graph.jl#L280. Perhaps you could suggest an improvement to the docstring - it's always hard to write good documentation for code you've written yourself.

YingboMa: I tend to think about tearing in terms of equations and variables. Maybe explain how tearing is about trying to maximize the number of matchings (solvable variables from the given tearing variable), and the the backward edges tell use the dependency among variables. No cycle means that assuming the tearing variables are known, then the rest of the variables can be computed by evaluation (i.e. no equation solving needed).

Keno: Sure, I'll write some doc strings for tearing, but I'll do that as a separate PR.

YingboMa: Sounds good!

baggepinnen: Something I'd appreciate from such a docstring, or elsewhere, is a brief explanation of the difference between alias elimination, tearing and index reduction. These three are all mentioned in the docs, and users are unlikely to understand what they mean, apart from perhaps index reduction

Keno: I think that'd probably be better for a whole section of the devdocs - it's a big topic.

Issue #1094, opened 2021-07-03
Allow passing kwargs to connect


ValentinKaisermayer:
https://github.com/SciML/ModelingToolkit.jl/blob/fc6926ee9c118176548fe575098fb20f5bcfd922/src/systems/abstractsystem.jl#L771

Since the `connect` method is essentially user defined it would be nice to allow for additional keyword arguments.

```julia
function connect(syss...; kwargs...)
    connect(promote_connect_type(map(get_connection_type, syss)...), syss...; kwargs...)
end
```

The reason is that if I compose a new component e.g. a `RealCapacitor` out of already existing components (`Resistor`, `Capacitor` and `Pin`) I need the connection to respect that the flow (i.e. current) from the pin of the `RealCapacitor`  and the flow from internal components have different sign conventions. This can be also found in the Modelica specification [here](https://specification.modelica.org/master/connectors-and-connections.html#inside-and-outside-connectors) and [here](https://specification.modelica.org/master/stream-connectors.html).

The way I solved this is to use a simple flag to indicate if the connector is inside or outside.
```julia
function connect(::Type{Pin}, connectors...; is_inside=fill(true, length(connectors))) # Which connector are inside connectors?
    eqs = [ 
        0 ~ sum([(i_ ? 1 : -1) * c.i for (c, i_) in zip(connectors, is_inside)]) # Flow balance
    ]
    for c in connectors[2:end] 
        push!(eqs, connectors[1].v ~ c.v) # Effort equality
    end
    return eqs
end
```

and in the definition of the `RealCapacitor` this can be used as follows:
```julia
function RealCapacitor(; name, C=1.0, R_esr=0.1, v0=0.0)
    C_val, R_esr_val = C, R_esr

    @named p = Pin()
    @named n = Pin()
    @named R_esr = Resistor(; R=R_esr_val)
    @named C = Capacitor(; C=C_val, v0=v0)
    connections = [
        connect(p, R_esr.p; is_inside=[false, true])
        connect(R_esr.n, C.p; is_inside=[true, true])
        connect(C.n, n; is_inside=[false, true])
    ]
    return ODESystem(connections, t, [], []; systems=[p, n, R_esr, C], name)
end
```
It is not really nice and a user is likely to forget about this but the results will be wrong otherwise.

ValentinKaisermayer:
From the Modelica spec: 

>...the pin includes the current, which is considered to be positive if it is flowing at the pin into the device.

PR #1348, opened 2021-11-17
Expand connections after model instantiation and add stream connectors

YingboMa:


ValentinKaisermayer:
I guess this closes #1094

PR #1350, opened 2021-11-18
Refactor (inv_)assign/{var/eq}assoc variables into proper structs

Keno:
These various assignments and associations look very similar, but
I can never remember which one goes in which direction and which one
is a matching and which one is a derivative ladder. Make this all
more clear by factoring into to types `Matching` for matching and
`DiffGraph` for derivative ladders. Both have a vector structure
and the latter also exposes a graph structure on the same vertex
set as the BipartiteGraph, which is useful for plotting (not part
of this PR). I've also renamed the variables to make the direction
clear.

### DIFF
--- a/src/bipartite_graph.jl
+++ b/src/bipartite_graph.jl
@@ -1,9 +1,11 @@
 module BipartiteGraphs
 
-export BipartiteEdge, BipartiteGraph, DiCMOBiGraph, Unassigned, unassigned
+export BipartiteEdge, BipartiteGraph, DiCMOBiGraph, Unassigned, unassigned,
+        Matching
 
 export 𝑠vertices, 𝑑vertices, has_𝑠vertex, has_𝑑vertex, 𝑠neighbors, 𝑑neighbors,
-       𝑠edges, 𝑑edges, nsrcs, ndsts, SRC, DST, set_neighbors!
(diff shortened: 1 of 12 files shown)


src/bipartite_graph.jl
```diff
@@ -17,6 +19,47 @@ struct Unassigned
...
+end
+
+function Base.push!(m::Matching, v::Union{Integer, Unassigned})
+    push!(m.match, v)
+    if v !== unassigned && m.inv_match !== nothing
+        m.inv_match[v] = length(m.match)
+    end
+end
+
+function complete(m::Matching)
```
YingboMa, line 48: Cool design.

src/bipartite_graph.jl
```diff
@@ -17,6 +19,47 @@ struct Unassigned
...
+    inv_match = Union{Unassigned, Int}[unassigned for _ = 1:length(m.match)]
+    for (i, eq) in enumerate(m.match)
+        eq === unassigned  && continue
+        inv_match[eq] = i
+    end
+    return Matching(collect(m.match), inv_match)
+end
+
+function invview(m::Matching)
+    m.inv_match === nothing && throw(ArgumentError("Backwards matching not defined. `complete` the matching first."))
```
YingboMa, line 59: Maybe add a `require_complete` method for matching as well.

Keno: I'll include this in my next PR

PR #1355, opened 2021-11-21
Refactor tearing

Keno:
The tearing code makes a number of assumptions about our particular method of tearing that do not need to be satisfied in general for all tearing algorithms. This starts refactoring in that direction with the expectation of more cleanly separating the particular choice of tearing algorithm from the abstract notion of tearing.


src/structural_transformation/StructuralTransformations.jl
```diff
@@ -44,6 +44,7 @@ include("utils.jl")
 include("pantelides.jl")
 include("bipartite_tearing/modia_tearing.jl")
 include("tearing.jl")
+include("symbolics_tearing.jl")
```
YingboMa, line 47: Maybe don't move the file just yet. It's going to be replaced anyway.

YingboMa:
`ModelingToolkitStandardLibrary` failed. I am not sure if it's related.

src/structural_transformation/symbolics_tearing.jl
```diff
@@ -0,0 +1,181 @@
+
```
YingboMa, line 1: I have some changes to this part of the code. I can move the file in another PR to avoid merge conflicts.

src/bipartite_graph.jl
```diff
@@ -396,21 +441,71 @@ function Base.iterate(c::CMONeighbors{true}, (l, state...))
...
+        g.ne = mapreduce(x->length(x.iter), +, _edges(g))
     end
     return g.ne
 end
 
 Graphs.has_edge(g::DiCMOBiGraph{true}, a, b) = a in inneighbors(g, b)
 Graphs.has_edge(g::DiCMOBiGraph{false}, a, b) = b in outneighbors(g, a)
 
+"""
+    struct ResidualCMOGraph
```
YingboMa, line 461: Is this used anywhere?

Keno:
> `ModelingToolkitStandardLibrary` failed. I am not sure if it's related.

No, that's the other PR that got merged. I'll rebase this.

src/bipartite_graph.jl
Keno: Not yet in MTK proper, but will be in the next PR.

Keno:
@YingboMa I've finished my tearing refactor and pushed it here. The highlight is that the output from tearing is now always a `Matching`, rather than `SystemPartition` since the latter doesn't really make sense as a persistent concept.

Keno:
The bipartite graph update at the end of tearing isn't quite right yet - but the rest should be ok. Will fix this shortly.

src/structural_transformation/symbolics_tearing.jl
```diff
@@ -3,185 +3,122 @@ function tearing_sub(expr, dict, s)
...
-            var = fullvars[iv]
-            rhs = value(solve_for(eq, var; simplify=simplify, check=false))
-            # if we don't simplify the rhs and the `eq` is not solved properly
-            (!simplify && occursin(rhs, var)) && (rhs = SymbolicUtils.polynormalize(rhs))
+    function solve_equation(ieq, iv)
+        var = fullvars[iv]
+        eq = eqs[ieq]
+        rhs = value(solve_for(eq, var; simplify=simplify, check=false))
+
+        if var in vars(rhs)
```
YingboMa, line 18: This should just throw an internal error.

Keno: Ok. Let's do that in a separate PR though, since that's a behavior change?

Keno:
> The bipartite graph update at the end of tearing isn't quite right yet - but the rest should be ok. Will fix this shortly.

This should be fixed, though we don't really test the post-tearing bipartite graph. I think it's fine for the moment though.

PR #1359, opened 2021-11-22
Drive by: Fix haystack/needle ordering

Keno:
Just a drive by I noticed while working on larger changes. I'm guessing we don't have a fallback for this
since it only happens when solve_for fails, but regardless, I'm pretty sure the other order was meant.

### DIFF
--- a/src/structural_transformation/symbolics_tearing.jl
+++ b/src/structural_transformation/symbolics_tearing.jl
@@ -29,7 +29,7 @@ function tearing_reassemble(sys; simplify=false)
             var = fullvars[iv]
             rhs = value(solve_for(eq, var; simplify=simplify, check=false))
             # if we don't simplify the rhs and the `eq` is not solved properly
-            (!simplify && occursin(rhs, var)) && (rhs = SymbolicUtils.polynormalize(rhs))
+            (!simplify && occursin(var, rhs)) && (rhs = SymbolicUtils.polynormalize(rhs))
             # Since we know `eq` is linear wrt `var`, so the round off must be a
             # linear term. We can correct the round off error by a linear
             # correction.


src/structural_transformation/symbolics_tearing.jl
```diff
@@ -29,7 +29,7 @@ function tearing_reassemble(sys; simplify=false)
             var = fullvars[iv]
             rhs = value(solve_for(eq, var; simplify=simplify, check=false))
             # if we don't simplify the rhs and the `eq` is not solved properly
-            (!simplify && occursin(rhs, var)) && (rhs = SymbolicUtils.polynormalize(rhs))
+            (!simplify && occursin(var, rhs)) && (rhs = SymbolicUtils.polynormalize(rhs))
             # Since we know `eq` is linear wrt `var`, so the round off must be a
             # linear term. We can correct the round off error by a linear
             # correction.
```
Keno, line 35: @YingboMa While we're here, why is the `rhs -= expand_derivatives(Differential(var)(rhs))*var`. Wouldn't we have to divide the rhs by `expand_derivatives(Differential(var)(rhs))` in order for this to be mathematically sound?

YingboMa: This case should never happen now. Symbolics used to use approaches with round-off errors when solving equations, so `solve(eq, x)` could return something silly like `y + eps() * x`, so we just subtract that away as a hack as the comment tries to explain.

Keno: Ok, so we should just delete that part?

YingboMa: I still want to keep the check. I can make a PR to do that.

Keno: sure, keeping the check makes sense. I'm working on this part anyway, I'll include it.

YingboMa: Yeah, if for some reason, it really happened, then we will get a stack overflow error.

Keno: @YingboMa Did you end up resolving this? Can we close the PR?

YingboMa: Yeah, this is no longer needed.

PR #1360, opened 2021-11-22
Add some comments in alias elimination

YingboMa:


src/systems/alias_elimination.jl
```diff
@@ -326,23 +345,32 @@ function locally_structure_simplify!(adj_row, pivot_col, ag, may_eliminate)
             continue
         end
         (coeff, alias_var) = alias
+        # `var = coeff * alias_var`, so we eliminate this var.
         adj_row[var] = 0
         if alias_var != 0
+            # val * var = (val * coeff) * alias_var
```
Keno, line 351:
```suggestion
            # val * var = val * (coeff * alias_var) = (val * coeff) * alias_var
```
Otherwise you're doing both the substitution and the re-association at once, which might be confusing.

Keno (approved):

PR #1368, opened 2021-11-23
Add array variable support for connect

YingboMa:


bolognam:
This seems to work for me, I'm good to merge this in.

PR #1370, opened 2021-11-24
Define isaffine methods for AbstractSystem

andreasnoack:
following the pattern of islinear.

### DIFF
--- a/docs/src/systems/NonlinearSystem.md
+++ b/docs/src/systems/NonlinearSystem.md
@@ -23,6 +23,7 @@ tearing
 ## Analyses
 
 ```@docs
+ModelingToolkit.isaffine
 ModelingToolkit.islinear
 ```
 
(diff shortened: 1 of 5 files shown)


baggepinnen:
Why are the lhs ignored? I get the definition for systems with only differential equations, but if the system contains algebraic equations like `x^2 ~ y` the result of `isaffine` (and probably `islinear`) might be misleading?

andreasnoack:
True but that applies to the current definition of `islinear` as well, right? So it might be better to consider the general issue in a separate issue.

ChrisRackauckas (approved):

ChrisRackauckas:
>Why are the lhs ignored? I get the definition for systems with only differential equations, but if the system contains algebraic equations like x^2 ~ y the result of isaffine (and probably islinear) might be misleading?

Yeah that's worth an issue. Right now the DAEs don't really handle non-trivial LHS expressions anyways, so yes we can make this generally more robust in the near future as something that is done post transformation (since the pull to the lhs would be done as part of the simplification part)

PR #1375, opened 2021-11-25
fix `collect_difference_variables`

baggepinnen:
used to always return the empty set

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -342,7 +342,7 @@ function collect_applied_operators(x, op)
     v = vars(x, op=op)
     filter(v) do x
         x isa Sym && return false
-        x isa Term && return operation(x) isa op
+        istree(x) && return operation(x) isa op
         false
     end
 end
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -1,3 +1,6 @@
+get_iv(D::Differential) = D.x
+get_iv(D::Difference) = D.t
+
 function make_operation(@nospecialize(op), args)
     if op === (*)
         args = filter(!_isone, args)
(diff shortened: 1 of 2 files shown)


src/utils.jl
```diff
@@ -282,7 +282,7 @@ function collect_operator_variables(sys, isop::Function)
     vars = Set()
     diffvars = Set()
     for eq in eqs
-        vars!(vars, eq)
+        isop === isdifferential ? vars!(vars, eq) : difference_vars!(vars, eq)
```
YingboMa, line 285: Instead of checking the function, we should just change the second argument of the function to take the type `Differential` or `Difference`

src/utils.jl
```diff
@@ -282,7 +282,7 @@ function collect_operator_variables(sys, isop::Function)
...
         for v in vars
             isop(v) || continue
```
YingboMa, line 287: This will be a `isa` check

sharanry (commented):
https://github.com/SciML/ModelingToolkit.jl/blob/6c21084c775eccfe36c05ab4683f283463b0d7b3/src/utils.jl#L294-L295

These lines will have to change too.

src/utils.jl
baggepinnen: Thanks! I apparently had an unpushed commit left on my machine, I modded it to take in the operator as well, which will be useful when there are more operators as well.

baggepinnen:
This is good to go, I have a few upcoming PRs making use of this so it would be great to have it merged if it looks okay

src/utils.jl
```diff
@@ -271,28 +297,55 @@ function vars!(vars, O; op=Differential)
...
+eq = D(y) ~ u
+ModelingToolkit.collect_applied_operators(eq, Differential) == Set([D(y)])
+```
+The difference compared to `collect_operator_variables` is that `collect_operator_variables` returns the variable without the operator applied.
+"""
+function collect_applied_operators(x, op)
+    v = vars(x, op=op)
+    filter(v) do x
+        x isa Sym && return false
+        x isa Term && return operation(x) isa op
```
YingboMa, line 345:
```suggestion
        istree(x) && return operation(x) isa op
```

Issue #1307, opened 2021-10-20
Time-shift operator for discrete-time control systems


baggepinnen:
MTK currently has the `Difference` operator that operates on `x(t)`, producing `x(t+dt) - x(t)`. Control-systems design and analysis in discrete time typically employs the shift operator, which operates on `x(t)` and produces `x(t+dt)`. The shift operator is the time-domain analog to the Z-transform operator, and allow you to write pulse-transfer functions like
```julia
TransferFunction{Discrete{Int64}, ControlSystems.SisoRational{Float64}}
1.5714702966431753z^3 + 0.5706479289308617z^2 - 0.3953765256425884z + 0.3354448928377647
----------------------------------------------------------------------------------------
       1.0z^3 + 0.5728489416081901z^2 - 0.1695044578033371z + 0.11243834527436224
```
in a natural way.

The easiest way to introduce this operator would be to define it in terms of the difference operator, i.e.,
```julia
struct TimeShift
    d::Difference
end
(q::TimeShift)(x) = q.d(x) + x
```
But that might conflict with assumptions such as the LHS of dynamics equations only containing a difference operator etc. (if such assumptions exist).

cf: https://github.com/SciML/ModelingToolkit.jl/issues/1180#issuecomment-896063856

ValentinKaisermayer:
Isn't there such a thing already?
https://github.com/JuliaSymbolics/Symbolics.jl/blob/c310aabc7854d64af4461ed64ef42018a63324d9/src/difference.jl#L33

baggepinnen:
Yes, it was hiding behind a name unfamiliar to me ^^
https://github.com/JuliaSymbolics/Symbolics.jl/pull/453 renames this operator to `Shift`

ValentinKaisermayer:
However, I think for `ODESystems`, `DiscreteUpdate` is really a ZOH (updates variables at certain discrete timestamps - as the name suggests):

https://github.com/SciML/ModelingToolkit.jl/blob/9cc6f21c1ffb1299d1a2d120c848f036d5f232c9/src/systems/diffeqs/abstractodesystem.jl#L129

Maybe it is different for `DiscreteSystem`. However, I still would like to have the update operator for ODESystems.

ValentinKaisermayer:
Maybe @YingboMa can clarify what he indented this operator for?

baggepinnen:
[...]

PR #1382, opened 2021-11-26
WIP: introduce time-domain operators

baggepinnen:
This PR introduces the operators
`Sample, Hold, Shift`
as well as functions to query their presence in expressions and equations. 

For background, see discussion in
- https://github.com/SciML/ModelingToolkit.jl/issues/1307

The intention is for this to serve as a foundation to improve the semantics of hybrid systems, systems with both continuous and discrete-time signals, similar to the Modelica Synchronous/Clocked module.

The `Sample` and `Hold` operators define the interface between the continuous domain and the discrete domain, while `Shift` is the time-shift operator that fills the same role for discrete systems as `Differential` does for continuous systems.

This PR depends on
- https://github.com/JuliaSymbolics/Symbolics.jl/pull/454

And supersedes
- https://github.com/SciML/ModelingToolkit.jl/pull/1381

Future work:
- Classify signals (states) as either continuous or discrete. 
- Check that time domains only mix through the interface `Sample/Hold`. Implicit sample and hold may be implemented at a later stage if the sample time can be inferred unambiguously.

### DIFF
--- a/src/ModelingToolkit.jl
+++ b/src/ModelingToolkit.jl
@@ -153,6 +153,9 @@ include("systems/alias_elimination.jl")
 include("structural_transformation/StructuralTransformations.jl")
 @reexport using .StructuralTransformations
 
+include("discretedomain.jl")
+include("timedomain.jl")
+
 for S in subtypes(ModelingToolkit.AbstractSystem)
     S = nameof(S)
     @eval convert_system(::Type{<:$S}, sys::$S) = sys
@@ -206,5 +209,10 @@ export build_function
 export modelingtoolkitize
 export @variables, @parameters
 export @named, @nonamespace, @namespace, extend, compose
+
+export Continuous, Discrete, sampletime, input_timedomain, output_timedomain
+export has_discrete_domain, has_continuous_domain, transitions_timedomain
+export is_discrete_domain, is_continuous_domain
+export Sample, Hold, Shift
 
 end # module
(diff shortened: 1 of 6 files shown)


ChrisRackauckas:
I don't quite understand the purpose of the Hold operator. Wouldn't it be better to just have the interpretation that `x(t+dt/n) == x(t)` for discrete processes with steps `dt`? That would be equivalent, match the way the plots are drawn, require less machinery, and be more explicit about the definition of the discrete values when used in a continuous context?

baggepinnen:
The hold operator is not required for `DiscreteSystem`s, it's only intended to define the interface from discrete to continuous time in hybrid systems. The modelica reference defines the `hold` operator as
> A discrete-time signal (such as ud) is converted to a continuous-time signal with the hold(..) operator.

The requirement for `hold` comes from discrete variables in modelica only being defined at discrete time points, and the hold operator extends the discrete time point to the interval between the point and the next time point. The diagram on p197 illustrates how it works in modelica
https://modelica.org/documents/ModelicaSpec34.pdf

ChrisRackauckas:
But you can already put discrete update operators in continuous ODESystems, and it has the interpretation that would match the hold operator, so that makes it seem redundant?

baggepinnen:
> But you can already put discrete update operators in continuous ODESystems, and it has the interpretation that would match the hold operator, so that makes it seem redundant?

The `DiscreteUpdate` operator of MTK only have the hold behavior if you have an additional equation that sets the derivative of the state you're holding to zero. Inbetween the "discrete updates", the variable may have any dynamics in general. This can certainly be useful at times, but it's not really the typical operator you'd use to model discrete systems.
```

Issue #894, opened 2021-03-15
Discrete-Time (Sub-)Systems


asprionj:
There is a discussion about how to implement hybrid system, i.e. a combination of a continuous system (in my domain of control-systems engineering often called the "plant" or the "physical system") with a discrete-time system ("controller"), [on Discourse](https://discourse.julialang.org/t/diffeqs-hybrid-continuous-discrete-system-periodic-callback/23791). The question came up whether MTK does already, will in the future, or should at all support discrete-time systems.

Please don't cosider this a "request to implement it", but a discussion on whether it makes sense, how it could be done, what effort it would take, etc. I know too little about the inner workings of MTK (and DiffEqs.jl) to judge any of this.

ChrisRackauckas:
It makes sense, and... close to duplicate but more so it's a combination of issues with a small cherry on top. And indeed, that Discourse issue describes why I don't want to waste any more time on DEDataArray: it's a dead end with some fundamental issues because of how the discrete data is "supposed" to propagate in broadcast, but that ends up making the propagation of controls information non-deterministic. We can make Tsit5 work with it and keep it tested, but it's not a good general solution. The general solution has to be MTK, with its observation equations, and all of that. So I want to focus the discussion on getting a real solution, and real workarounds until that's done.

(Don't feel bad about the duplicate issues though: it probably is good for populating the search term with different titles.) 

The issues that you're touching on here are:

Difference operator: https://github.com/JuliaSymbolics/Symbolics.jl/issues/32
Discrete system: https://github.com/SciML/ModelingToolkit.jl/issues/571
Events: https://github.com/SciML/ModelingToolkit.jl/issues/38
Promotion: https://github.com/SciML/ModelingToolkit.jl/issues/817

So there's a few steps to get there, at least to do it correctly. We want to make sure the whole symbolic system is directly supportive of it. You can hack it in today, but that won't mean you'll get great symbolics on it! So the steps are:

- Define a differencing operator so that we can symbolically represent a discrete-time equation. Define some symbolics for expanding and solving them.
- Define DiscreteSystem so that we have a symbolic analog to DiscreteProblem. Define all of the conversions. This is fairly routine and a great first issue BTW: it's just a bit more than a copy-paste of some ODESystem stuff.
- Define a way to symbolically specify and build events. @YingboMa is tackling that one, but it'll take a few months.
- Define a promotion system on subsystems, so you can put a NonlinearSystem inside of an ODESystem
[...]

PR #1383, opened 2021-11-26
WIP: Add `SampledTime` operator

baggepinnen:
This PR implements the `SampledTime` operator proposed by @jonniedie in https://github.com/SciML/ModelingToolkit.jl/issues/894#issuecomment-805893878

It provides a convenient interface between the continuous and discrete time domains, examplified by
```julia
julia> @variables t x(t);

julia> k = SampledTime(t, dt=0.1);

julia> x(k)
  Sample(t; dt=0.1)(x(t))

julia> x(k+1)
  Shift(t, 1; dt=0.1)(Sample(t; dt=0.1)(x(t)))
```
i.e., the semantics is that when you call a continuous-time signal with this operator it becomes sampled, and possibly shifted. If the indexed signal is already a discrete-time signal, the `Sample` should be a noop.

This PR builds on 
- #1382

### DIFF
--- a/src/timedomain.jl
+++ b/src/timedomain.jl
@@ -59,11 +59,11 @@ has_continuous_domain(x) = hasderiv(x) || hasdiff(x) || hassample(x) || hashold(
 
 
 """
-    transitions_timedomain(x)
+    is_hybrid_domain(x)
 
 true if `x` contains both discrete and continuous-domain signals. `x` may be an expression or equation.
 """
-transitions_timedomain(x) = has_discrete_domain(x) && has_continuous_domain(x)
+is_hybrid_domain(x) = has_discrete_domain(x) && has_continuous_domain(x)
 
 """
     is_discrete_domain(x)
--- a/src/timedomain.jl
+++ b/src/timedomain.jl
@@ -1,0 +1,82 @@
+using Symbolics: Differential, Difference
+using Symbolics: hasderiv, hasdiff
+
+abstract type TimeDomain end
+
+struct Continuous <: TimeDomain end
+struct Discrete <: TimeDomain end
+
+
+for op in [Differential, Difference]
+    @eval input_timedomain(::$op) = Continuous()
+    @eval output_timedomain(::$op) = Continuous()
+end
+
+
+"""
+    input_timedomain(op::Operator)
+
+Return the time-domain type (`Continuous()` or `Discrete()`) that `op` operates on. 
+"""
+input_timedomain(::Shift) = Discrete()
+
+"""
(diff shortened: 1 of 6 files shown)


ValentinKaisermayer:
If `x(k)` is the value of `x` at the current time step, I think any `x(k+n)` with `n>0` is not implementable. Modelica only defines the `previous(.)` operator.

baggepinnen:
> I think any x(k+n) with n>0 is not implementable

Being able to express a model is not the same as being able to implement or simulate it, I see no reason to limit the expressivity the way modelica does. The model
```
x(k+1) = x(k)
```
defines a constant system, which is trivially implementable. In general, for discrete-time systems you may transform all shifted variables by shifting them with the largest shift present and then you have a causal system you can implement, it doesn't matter if the way the model was written was not implementable.

Also, you often want to express non-causal / non-proper models which are later inverted, thus becoming causal/proper. This is common when you design model-based inverse filters for feedforward design.

ChrisRackauckas:
>If x(k) is the value of x at the current time step, I think any x(k+n) with n>0 is not implementable

That could actually be useful for PDE definitions.

src/timedomain.jl
```diff
@@ -0,0 +1,82 @@
+using Symbolics: Differential, Difference
+using Symbolics: hasderiv, hasdiff
+
+abstract type TimeDomain end
```
ChrisRackauckas, line 4: why would this be separate from the domain system?

baggepinnen: Is it this domain system you are referring to?
https://mtk.sciml.ai/dev/systems/PDESystem/#Domains-(WIP)

`TimeDomain` could possibly be integrated with that, but a variable can belong to both a domain in the sense of 
```julia
t ∈ (0.0,1.0)
```
and the `Continuous()` time domain at the same time so they don't really fill the same purpose. It would be nice if you could write
```julia
t ∈ (0.0, 1.0) # to get a Continuous() variable
k ∈ 0.0:0.1:1.0 # to get a clocked variable with clock Clock(k, 0.1)
```
I do however have a working WIP of automatic clock inference that would not require the user to specify the time domain of variables manually
https://github.com/SciML/ModelingToolkit.jl/issues/1307#issuecomment-982361244

baggepinnen:
> > If x(k) is the value of x at the current time step, I think any x(k+n) with n>0 is not implementable
> 
> That could actually be useful for PDE definitions.

Good point, perhaps `SampledTime` is a poor choice of name, maybe it should just be called `SampledVar` or something more generic to indicate that the independent variables does not have to be time.

Issue #988, opened 2021-04-25
State selection


YingboMa:
Without state selection, `ODAEProblem` w/ `Tsit5` gives terrible numerical drift on the pendulum model.
```julia
using ModelingToolkit, LinearAlgebra, OrdinaryDiffEq

function Pendulum(;name, g=9.81, r=1.)
    gval, rval = g, r
    @parameters g r
    @variables x(t) y(t) T(t)
    D2 = Differential(t)^2
    eqs = [
           D2(x) ~ T * x
           D2(y) ~ T * y - g
           x^2 + y^2 ~ r
          ]

    ODESystem(eqs, t, [x, y, T], [g, r]; name=name, defaults=(g=>gval, r=>rval, x=>1, y=>0, D(x)=>0, D(y)=>0.0, T=>0))
end

@parameters t
D = Differential(t)
@named sys = Pendulum(;g=9.81, r=1)
pendulum_sys = structural_simplify(ode_order_lowering(flatten(sys)))
prob = ODAEProblem(pendulum_sys, [], (0, 100.0))
sol = solve(prob, Tsit5())
x = @nonamespace sys.x
y = @nonamespace sys.y
@. sol[x]^2 + sol[y]^2
```
and
```julia
julia> using Plots

julia> plot(sol, vars=(x, y), xlab="x", ylab="y", dpi=300, title="Numerical drift")
```
gives
![dp_xy_drift](https://user-images.githubusercontent.com/17304743/116011023-adb53480-a5f0-11eb-8600-e9960e679a39.png)

anandijain:
could you post a reference? i still dont know what state selection is lol

kabdelhak:
> could you post a reference? i still dont know what state selection is lol

https://www.researchgate.net/publication/235324214_Index_Reduction_in_Differential-Algebraic_Equations_Using_Dummy_Derivatives

The dummy derivative method as a follow up for pantelides algorithm to resolve singular systems. Is something like this planned?

ChrisRackauckas:
Yes, definitely planned.

MarkusLohmayer:
The following code does not work:
```
using ModelingToolkit
using DifferentialEquations
@variables t
D = Differential(t)
ps = @parameters mass₊m = 1.0 spring₊c = 10.0
sts = @variables mass₊_p(t) spring₊_q(t)
eqs = [
    0 ~ (-mass₊_p) / mass₊m + D(spring₊_q),
[...]

PR #1384, opened 2021-11-27
Partial State Selection

Keno:
This implements the basic outline of partial state selection.
It is functional and resolves the example in #988, but we do not
currently implement any tearing priorities or handling of linear
subsystems, so it's pretty easy for state selection to end up
selecting states that are singular.

Note this currently includes commits from https://github.com/SciML/ModelingToolkit.jl/pull/1355, since
it builds on top, but I can't use that as a base branch because it's not on this repo. Once that is merged,
I'll rebase this. The relevant changes are in the final commit.


YingboMa:
@isaacsas It seems like Catalyst uses MTK internal `initialize_system_structure`. Could you update Catalyst so that it doesn't rely on it?

isaacsas:
> @isaacsas It seems like Catalyst uses MTK internal `initialize_system_structure`. Could you update Catalyst so that it doesn't rely on it?

Sure. Looks like we use it to determine if an equation is a diff eq (vs. algebraic). Is there a recommended alternative to check if an equation is a diff eq now?

YingboMa:
Maybe just iterate over `eq.lhs` and check all of them are in the form of `D(x)`?

YingboMa:
@isaacsas I will merge this PR for now and not release a new version of MTK until Catalyst updates accordingly.

isaacsas:
I’ll ping you once a new Catalyst release is out with the fix.

isaacsas:
New Catalyst release is out.

PR #1401, opened 2021-12-06
remove call to nonexisting function `polynormalize`

baggepinnen:
There appears to be no function `polynormalize` in the registered package ecosystem?
https://juliahub.com/ui/Search?q=polynormalize&type=symbols


YingboMa:
This was fixed on another branch.

PR #1429, opened 2022-01-20
Add preface for DiscreteSystem

sharanry:
@YingboMa


sharanry:
@YingboMa Can we merge and make a patch release?

PR #1438, opened 2022-01-27
Allow all events as part of ODEsystem as kwargs

sharanry:


baggepinnen:
Maybe there can be a mechanism that protects the states of the subsystem with a callback and makes sure the affect function is called with the correct state index?

sharanry:
> What if the system is a subsystem of a larger system? Then the state indices can be anything and this might be silently wrong?

Fair point. If we can embed callbacks as part of ODESystem type (as a separate callback field), the larger system could construct a callback which is the combination of all the sub-system callbacks. 

> Even if it's not a subsystem, simplification passes might permute the states and remove some states?

True, the simplification passes will then have to preserve the states/subsystems with such effects or it will need a mechanism to alter the `affect!` function to match the potentially new indices of the states (after preserving them).

PR #1449, opened 2022-02-08
move bareiss

oscardssmith:
Keno and I have discussed https://github.com/JuliaLang/julia/pull/42683, and believe it likely doesn't make sense for `Base` to have this added complexity, so this PR moves the bareiss code from `compat` to `systems`.


oscardssmith:
Can someone approve CI for this?

src/ModelingToolkit.jl
```diff
@@ -117,7 +117,7 @@ include("utils.jl")
 include("domains.jl")
 
 # Code that should eventually go elsewhere, but is here for fow
-include("compat/bareiss.jl")
+include("structural_transformation/bareiss.jl")
```
YingboMa, line 120: Why move the file?

oscardssmith: As I understand it, the point of the compat folder is for things that temporarily need to exist in MTK until a PR gets merged somewhere else. Since this code now isn't going in Base, it belongs somewhere permanent.

oscardssmith: moving the file isn't what messed up git blame. git blame gets messed up because removing the check for if `Base` has the functionality changes the indentation of the entire file.

YingboMa: Alright, cool!

YingboMa: CI fails though

oscardssmith: there was a typo. Should be good now.

oscardssmith:
Thanks!

PR #1450, opened 2022-02-09
remove compat incremental_cycles

oscardssmith:
Now that https://github.com/JuliaGraphs/Graphs.jl/pull/36 exists, we don't need to carry this code.


oscardssmith:
Note that this requires a new version of Graphs.jl to be tagged before this can be merged.

oscardssmith:
The new version of Graphs has been tagged. @YingboMa can you run CI on this?

ChrisRackauckas:
@KirillZubov can you look into https://github.com/SciML/ModelingToolkit.jl/runs/5136168032?check_suite_focus=true#step:6:844 ? How does a scalar get there?

But this looks good to go.

KirillZubov:
it is trouble with updating DiffEqFlux.FastDense 
https://github.com/SciML/NeuralPDE.jl/pull/485

PR #1455, opened 2022-02-14
add tests for bareiss, fix precompile

oscardssmith:
This adds direct tests for the bareiss code, and fixes a precompilation issue on Julia 1.8.

### DIFF
--- a/src/structural_transformation/bareiss.jl
+++ b/src/structural_transformation/bareiss.jl
@@ -8,105 +8,110 @@ macro swap(a, b)
     esc(:(($a, $b) = ($b, $a)))
 end
 
-function swaprows!(a::AbstractMatrix, i, j)
-    i == j && return
-    rows = axes(a,1)
-    @boundscheck i in rows || throw(BoundsError(a, (:,i)))
-    @boundscheck j in rows || throw(BoundsError(a, (:,j)))
-    for k in axes(a,2)
-        @inbounds a[i,k],a[j,k] = a[j,k],a[i,k]
+#added post Julia 1.7
+if isdefined(Base, :swaprows!)
+    import Base: swaprows!
+else
+    function swaprows!(a::AbstractMatrix, i, j)
+        i == j && return
+        rows = axes(a,1)
+        @boundscheck i in rows || throw(BoundsError(a, (:,i)))
+        @boundscheck j in rows || throw(BoundsError(a, (:,j)))
+        for k in axes(a,2)
+            @inbounds a[i,k],a[j,k] = a[j,k],a[i,k]
+        end
     end
-end
-function Base.circshift!(a::AbstractVector, shift::Integer)
-    n = length(a)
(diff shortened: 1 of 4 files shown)


oscardssmith:
I'm really confused why tests are failing since they pass locally. @ChrisRackauckas any ideas?

ChrisRackauckas:
"not defined": are you using a different version of Julia?

oscardssmith:
I think this is finally working.

oscardssmith:
@ChrisRackauckas can we merge?

src/structural_transformation/bareiss.jl
```diff
@@ -2,111 +2,116 @@
...
 
-function swaprows!(a::AbstractMatrix, i, j)
-    i == j && return
-    rows = axes(a,1)
-    @boundscheck i in rows || throw(BoundsError(a, (:,i)))
-    @boundscheck j in rows || throw(BoundsError(a, (:,j)))
-    for k in axes(a,2)
-        @inbounds a[i,k],a[j,k] = a[j,k],a[i,k]
+#added post Julia 1.7
+if isdefined(Base, :swaprows!)
```
sjkelly, line 12: In general `VERSION` seems to be the idiomatic check for such things.

oscardssmith: I did it this way so that 1.8 dev would work without me having to find the exact commit where it was added. Now that 1.8 beta1 is out, this isn't as necessary, but I still think it's a little cleaner.

sjkelly: You can `eval` into `Base` and break this trivially. `VERSION` is robust.
https://github.com/JuliaLang/Compat.jl/blob/fa274f7a5e980355c72ccc879e1406f1d6880444/src/Compat.jl#L398-L400

oscardssmith: technically you can just set the `VERSION` variable, but I agree it's better.

src/structural_transformation/bareiss.jl
```diff
@@ -2,111 +2,116 @@
...
 
-function swaprows!(a::AbstractMatrix, i, j)
-    i == j && return
-    rows = axes(a,1)
-    @boundscheck i in rows || throw(BoundsError(a, (:,i)))
-    @boundscheck j in rows || throw(BoundsError(a, (:,j)))
-    for k in axes(a,2)
-        @inbounds a[i,k],a[j,k] = a[j,k],a[i,k]
+# https://github.com/JuliaLang/julia/pull/42678
+if VERSION > v"1.8.0-DEV.762"
```
ChrisRackauckas, line 12:
```suggestion
@static if VERSION > v"1.8.0-DEV.762"
```

ChrisRackauckas:
@YingboMa check this one today?

oscardssmith:
Thanks!

PR #1465, opened 2022-02-22
Fix sparse jacobian from TearingState

YingboMa:


YingboMa:
CI failure is unrelated.

PR #1469, opened 2022-02-28
eltype promotion in varmap_tp_vars

baggepinnen:
The old approach took the first type it found (`example_val`), which is a brittle approach. This approach tries to promote all the element types to a common supertype. With this PR, one can simulate systems with uncertain coefficients etc., which previously failed if `example_val` by random happened to be `Float64` etc.

An alternative approach would perhaps be to allow the user to be explicit about the element type used in simulation. Currently, it's a bit inconvenient to simulating anything using, e.g., `Float32`, since all default values etc. would have to be provided in `Float32`.

### DIFF
--- a/src/variables.jl
+++ b/src/variables.jl
@@ -72,20 +72,15 @@ function _varmap_to_vars(varmap::Dict, varlist; defaults=Dict(), check=false, to
     varmap = merge(defaults, varmap) # prefers the `varmap`
     varmap = Dict(toterm(value(k))=>value(varmap[k]) for k in keys(varmap))
     # resolve symbolic parameter expressions
-    example_val = nothing
     for (p, v) in pairs(varmap)
         val = varmap[p] = fixpoint_sub(v, varmap)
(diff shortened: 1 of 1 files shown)


src/variables.jl
```diff
@@ -72,20 +72,15 @@ function _varmap_to_vars(varmap::Dict, varlist; defaults=Dict(), check=false, to
...
-        end
     end
     vs = values(varmap)
     T′ = eltype(vs)
     if Base.isconcretetype(T′)
         T = T′
     else
-        example_val === nothing && throw_missingvars(varlist)
-        T = float(typeof(example_val))
+        T = foldl((t, elem)->promote_type(t, typeof(elem)), vs; init=typeof(first(vs)))
```
ChrisRackauckas, line 83:
```suggestion
        T = foldl((t, elem)->promote_type(t, eltype(elem)), vs; init=typeof(first(vs)))
```

to support vectors?

YingboMa (approved):

PR #1477, opened 2022-03-02
Check against Identity with equality

### LINKED ISSUE #1476: ODEFunctonExpr(sys, sparse=true) Fails to generate sparse matrix or recognize identity
Generated matrix is not sparse, leading to extremely large generated code
```

@parameters t
@variables u[1:3](..) 
Dt = Differential(t)

eqs = [Dt(u[1](t)) ~ u[1](t), Dt(u[2](t)) ~ u[2](t), Dt(u[3](t)) ~ u[3](t)]
defaults = Dict([u[i](t) => 0. for i in 1:3])
alldepvars = Num[u[i](t) for i in 1:3]
ps = Num[]

sys = ODESystem(eqs, t, alldepvars, ps, defaults=defaults, name= :ODE_sparse_fail)

code = ODEFunctionExpr(sys, sparse=true)

open("code.jl", "a") do io
       println(io, code)
end
```
generates:

```
begin
    var"##f#306" = (ModelingToolkit.ODEFunctionClosure)(function (ˍ₋arg1, ˍ₋arg2, t)
                begin
                    var"u[1](t)" = @inbounds(ˍ₋arg1[1])
                    var"u[2](t)" = @inbounds(ˍ₋arg1[2])
                    var"u[3](t)" = @inbounds(ˍ₋arg1[3])
                    begin
                        (SymbolicUtils.Code.create_array)(typeof(ˍ₋arg1), nothing, Val{1}(), Val{(3,)}(), var"u[1](t)", var"u[2](t)", var"u[3](t)")
                    end
                end
            end, function (ˍ₋out, ˍ₋arg1, ˍ₋arg2, t)
                begin
                    var"u[1](t)" = @inbounds(ˍ₋arg1[1])
                    var"u[2](t)" = @inbounds(ˍ₋arg1[2])
                    var"u[3](t)" = @inbounds(ˍ₋arg1[3])
                    begin
                        @inbounds begin
                                ˍ₋out[1] = var"u[1](t)"
                                ˍ₋out[2] = var"u[2](t)"
                                ˍ₋out[3] = var"u[3](t)"
                                nothing
                            end
                    end
                end
            end)
    var"##tgrad#307" = nothing
    var"##jac#308" = nothing
    M = [1.0 0.0 0.0; 0.0 1.0 0.0; 0.0 0.0 1.0]
    ODEFunction{true}(var"##f#306", jac = var"##jac#308", tgrad = var"##tgrad#307", mass_matrix = M, jac_prototype = similar(Matrix{Num}(undef, 0, 0), Float64), syms = [Symbol("u[1](t)"), Symbol("u[2](t)"), Symbol("u[3](t)")], indepsym = :t)
end
```

Note that the mass matrix is sparse and further is the identity but is not recognized as such

xtalax:
Fixes #1476

### DIFF
--- a/src/systems/diffeqs/abstractodesystem.jl
+++ b/src/systems/diffeqs/abstractodesystem.jl
@@ -279,7 +279,7 @@ function calculate_massmatrix(sys::AbstractODESystem; simplify=false)
     end
     M = simplify ? ModelingToolkit.simplify.(M) : M
     # M should only contain concrete numbers
-    M === I ? I : M
+    M == I ? I : M
 end
 
 function jacobian_sparsity(sys::AbstractODESystem)


YingboMa:
Is the CI failure related?

xtalax:
Unfortunately it is related, I think because it is hoping that some operation with `I` will fail with a `DimensionMismatch` when of course it won't. There needs to be an explicit check that the system is invalid, not relying on a `DimensionMismatch`. 

I'll try to work out why this system should be invalid and throw another way though you could save me some time if you know this.

ChrisRackauckas:
> Solving this problem leads to the state without the equation always being equal to it's default, which seems like expected behaviour.

No, that would be an equation with a zero derivative. That's different from no equation. It probably should just error. @YingboMa do you recall why `check_length` was added to the ODEProblem constructor? It seems that should always just fail if you don't have a balanced system.

xtalax:
Could it assume that giving a state without an equation a default implicitly adds an equation `state ~ state0`?

YingboMa:
> do you recall why check_length was added to the ODEProblem constructor? It seems that should always just fail if you don't have a balanced system.

If we don't have the check, then if we have more equations that the number of states, we will be indexing out of bounds and potentially cause a segfault.

YingboMa:
Let's use squash commit next time please.

Issue #1484, opened 2022-03-18
DAEProblem fails 


ValentinKaisermayer:
Maybe I'm completely wrong but can it be that the following code is not correct and  `u0defs` and `pdefs`  should be switched?
Is the ordering of the `merge` correct? Since `defs = defaults(sys)`, the user input e.g. `u0map` will be always overruled. However, testing this it seems to work for e.g. `ODEProblem`.
https://github.com/SciML/ModelingToolkit.jl/blob/029addab54a82361ef4a3d97e732f4779cb2334f/src/systems/diffeqs/abstractodesystem.jl#L562-L579

Since `pdefs` is used for the parameters:
https://github.com/SciML/ModelingToolkit.jl/blob/029addab54a82361ef4a3d97e732f4779cb2334f/src/systems/diffeqs/abstractodesystem.jl#L589

This one errors (see example bellow). Should `defaults` be `du0defs` (which is not defined)?
https://github.com/SciML/ModelingToolkit.jl/blob/029addab54a82361ef4a3d97e732f4779cb2334f/src/systems/diffeqs/abstractodesystem.jl#L584

MWE:
```julia
using ModelingToolkit, Sundials
@parameters t
D = Differential(t)
@variables x[1:2](t) = zeros(2) 
@variables y(t) = 0.0
eqs= [
    D(x[1]) ~ x[2]
    D(x[2]) ~ -x[1] - 0.5 * x[2] + 1.0
    y ~ 0.9 * x[1] + x[2]
]
@named model = ODESystem(eqs, t, vcat(x, y), [])
prob = DAEProblem(sys, Pair[], Pair[], (0, 50))
sol = solve(prob, IDA())
```

```julia
ERROR: MethodError: no method matching iterate(::typeof(ModelingToolkit.defaults))
Closest candidates are:
  iterate(::Union{LinRange, StepRangeLen}) at \julia\base\range.jl:826
  iterate(::Union{LinRange, StepRangeLen}, ::Integer) at C\julia\base\range.jl:826
  iterate(::T) where T<:Union{Base.KeySet{<:Any, <:Dict}, Base.ValueIterator{<:Dict}} at \julia\base\dict.jl:695
  ...
Stacktrace:
 [1] isempty(itr::Function)
   @ Base .\essentials.jl:775
 [2] varmap_to_vars(varmap::Vector{Pair}, varlist::Vector{Term{Real, Nothing}}; defaults::Function, check::Bool, toterm::Function)
[...]

Issue #1475, opened 2022-03-02
parameter tuples having mixed numeric values are converted to the same type


isaacsas:
```julia
using ModelingToolkit, OrdinaryDiffEq
@parameters k1 k2
@variables t, A(t)
D = Differential(t)
eqs = [D(A) ~ -k1*k2*A]
@named osys = ODESystem(eqs,t)
u0map = [A => 1.0]
pmap = (k1 => 1.0, k2 => 1)
tspan = (0.0,1.0)
oprob = ODEProblem(osys, u0map, tspan, pmap)
oprob.p == (1.0,1.0)
```
Where as 
```julia
function f(du,u,p,t)
    du[1] = -p[1]*p[2]*u[1]
    nothing
end
oprob = ODEProblem(f, [uv[2] for uv in u0map], tspan, Tuple(pv[2] for pv in pmap))
oprob.p == (1.0,1)
```

A use case for having integer parameters is that in Catalyst we want to allow integer stoichiometry symbolic variables, which generate ODEs involving the factorial of these variables. Currently these then give an error when called with floating point parameters.

isaacsas:
Closed by https://github.com/SciML/ModelingToolkit.jl/pull/1487

PR #1487, opened 2022-03-19
Patch process_DEproblem

ValentinKaisermayer:
This one maybe?

### DIFF
--- a/src/systems/diffeqs/abstractodesystem.jl
+++ b/src/systems/diffeqs/abstractodesystem.jl
@@ -559,17 +559,6 @@ function process_DEProblem(constructor, sys::AbstractODESystem,u0map,parammap;
     ps = parameters(sys)
     defs = defaults(sys)
     iv = get_iv(sys)
-    function mergedefaults(defaults, varmap, vars)
-        defs = if varmap isa Dict
-            merge(defaults, varmap)
-        elseif eltype(varmap) <: Pair
-            merge(defaults, Dict(varmap))
-        elseif eltype(varmap) <: Number
-            merge(defaults, Dict(zip(vars, varmap)))
-        else
-            defaults
-        end
-    end
 
     pdefs = mergedefaults(defs, parammap, ps)
     u0defs = mergedefaults(defs, u0map, dvs)
--- a/src/systems/diffeqs/abstractodesystem.jl
+++ b/src/systems/diffeqs/abstractodesystem.jl
@@ -559,34 +559,31 @@ function process_DEProblem(constructor, sys::AbstractODESystem,u0map,parammap;
     ps = parameters(sys)
     defs = defaults(sys)
     iv = get_iv(sys)
-    if parammap isa Dict
-        u0defs = merge(parammap, defs)
-    elseif eltype(parammap) <: Pair
-        u0defs = merge(Dict(parammap), defs)
-    elseif eltype(parammap) <: Number
-        u0defs = merge(Dict(zip(ps, parammap)), defs)
-    else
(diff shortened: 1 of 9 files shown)


ValentinKaisermayer:
closes #1484

src/systems/diffeqs/abstractodesystem.jl
```diff
@@ -559,34 +559,31 @@ function process_DEProblem(constructor, sys::AbstractODESystem,u0map,parammap;
...
+        defs = if varmap isa Dict
+            merge(defaults, varmap)
+        elseif eltype(varmap) <: Pair
+            merge(defaults, Dict(varmap))
+        elseif eltype(varmap) <: Number
+            merge(defaults, Dict(zip(vars, varmap)))
+        else
+            defaults
+        end
     end
```
ChrisRackauckas, line 572: This should be moved to utils.

ChrisRackauckas (approved):

ChrisRackauckas:
Test failure looks real.

ValentinKaisermayer:
I'm on it 😄

ChrisRackauckas (approved):

ValentinKaisermayer:
Sry, for re-triggering the CI. I noticed #1475. With these changes it works and added a test for it.

src/systems/jumps/jumpsystem.jl
```diff
@@ -220,9 +220,17 @@ dprob = DiscreteProblem(js, u₀map, tspan, parammap)
...
                                     parammap=DiffEqBase.NullParameters(); checkbounds=false, kwargs...)
+    
+    dvs = states(sys)
+    ps = parameters(sys)
+    
     defs = defaults(sys)
-    u0 = varmap_to_vars(u0map, states(sys); defaults=defs)
-    p  = varmap_to_vars(parammap, parameters(sys); defaults=defs)
+    defs = mergedefaults(defs,parammap,ps)
+    defs = mergedefaults(defs,u0map,dvs)
```
isaacsas, line 229: @ValentinKaisermayer does this change mean that `merge` is now being called twice with `u0map` (since it is called again within `varmap_to_vars`)? If so, why do we need to merge the user maps into the defaults instead of letting `varmap_to_vars` handle merging (or vice-versa, should that merge be removed from `varmap_to_vars`)?

ValentinKaisermayer: The reason is, that initial conditions might contain symbolic parameters, hence `defaults` should contain states and parameters, updated with the provided values.

src/systems/jumps/jumpsystem.jl
```diff
@@ -220,9 +220,17 @@ dprob = DiscreteProblem(js, u₀map, tspan, parammap)
...
+    dvs = states(sys)
+    ps = parameters(sys)
+    
     defs = defaults(sys)
-    u0 = varmap_to_vars(u0map, states(sys); defaults=defs)
-    p  = varmap_to_vars(parammap, parameters(sys); defaults=defs)
+    defs = mergedefaults(defs,parammap,ps)
+    defs = mergedefaults(defs,u0map,dvs) 
+    
+    u0 = varmap_to_vars(u0map,dvs; defaults=defs)
```
ValentinKaisermayer, line 231: However, we can streamline this of course. At this point there is no nee information in `varmap`, that is not already in the `defs`.

src/systems/jumps/jumpsystem.jl
isaacsas: Thanks for clarifying!

PR #1490, opened 2022-03-22
Patch concrete init

ValentinKaisermayer:
Re-adds concrete type promotion for `u0`, otherwise solvers like `Rosenbrock23` throw an exception. This was present in `v8.5.2`. Parameters can have mixed types, though.

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -473,6 +473,7 @@ function mergedefaults(defaults, varmap, vars)
 end
 
 function promote_to_concrete(vs)
+    isemtpy(vs) && return vs
     T = eltype(vs)
     if Base.isconcretetype(T) # nothing to do
         vs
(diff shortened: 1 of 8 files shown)


src/utils.jl
```diff
@@ -470,4 +470,13 @@ function mergedefaults(defaults, varmap, vars)
...
         defaults
     end
+end
+
+function promote_to_concrete(vs::Vector{T}) where {T}
+    if Base.isconcretetype(T) # nothing to do
+        vs
+    else
+        C = foldl((t, elem)->promote_type(t, eltype(elem)), vs; init=typeof(first(vs)))
+        convert(Vector{C}, vs)
```
ChrisRackauckas, line 480: It's not always a vector, or at least, not always supposed to be a vector. I think you want to instead just broadcast a convert to C.

ValentinKaisermayer: I see, it could be a Tuple, this seems to be missing from the test then.

ChrisRackauckas (approved):

PR #1495, opened 2022-03-23
Fix minor writing error

jonschumacher:


YingboMa (commented):
I think iff is short for if and only if here.

PR #1508, opened 2022-04-04
Better error msg when independent variable detection fails

YingboMa:


ChrisRackauckas:
I don't see an error message written out 😅

YingboMa:
Yes, it prints somewhere else after a nothing check.

PR #1510, opened 2022-04-05
Added functionality for symbolically computing DAE jacobians

Abhishek-1Bhatt:
Aims to add `jac` and `sparse` implementation for `DAEProblem` and resolve #1482

PR #1519, opened 2022-04-08
Fixed Jacobian for DAEs and added a test for it

### LINKED ISSUE #1482: `jac` and `sparse` are not implemented for DAEProblem
```julia
using OrdinaryDiffEq, ModelingToolkit
using BenchmarkTools
using Plots
using Sundials

function build_system(n)

	@parameters t # or would @variables t be better!?
	@variables Tgas[1:n](t) Tsol[1:n](t) cR[1:n](t) mflux(t)
	@parameters params[1:3]
	D = Differential(t)


    dp, Tin, h = params

    dx = h / n

	############################################################
	# Build equations ##########################################
	############################################################

	eqns = Array{Equation}(undef, 3 * n + 1)
	count = 1

    # specification as function or intermediate variabel does not matter. I could also make this an
    # actual @variable, then I could retrieve it later.
    rR = 5.5e6 .* exp.(.-5680 ./(273 .+ Tsol)) .* (1 .- exp.(.-cR./50))
    q = 1e4 * (Tgas - Tsol)

    # gas energy balance

    eqns[count] = 0 ~  mflux * 1000 * (Tin - Tgas[1]) - q[1] * dx
    count += 1
	for i in 2:n
	        eqns[count] = 0 ~ mflux * 1000 * (Tgas[i-1] - Tgas[i]) - q[i] * dx
		count += 1
	end

	# solids energy balance
	for i in 1:n
        	eqns[count] = D(Tsol[i]) ~ q[i] / (1000 * 2000)
		count += 1
	end

	# gas momentum balance
	eqns[count] = 0 ~ - mflux[1] + 5e-3 * sqrt(dp)
	count += 1

	# reactant concentration equations
	for i in 1:n
        	eqns[count] = D(cR[i]) ~ -rR[i]
		count += 1
	end
    @named sys = ODESystem(eqns, t, vcat(Tgas, Tsol, cR, [mflux]), params)
    return sys

end


dp = 100e2
Tin = 200
h = 0.5
parameters = [dp, Tin, h]

N = 100
tspan = (0.0, 1500.0)
sys = build_system(N)

# quick and dirty definition of u0, should be Pairs instead
u0 = ones(length(sys.states))
u0[1:N] .= parameters[2]
u0[N+1 : 2*N] .= 20
u0[2*N+1 : 3*N] .= 3000
u0[3*N+1] = 0.5

du0 = 0 .* copy(u0)
prob3 = DAEProblem(sys, du0, u0, tspan, parameters, jac=true, sparse=true)
prob3.f.jac_prototype == nothing
prob3.f.jac == nothing
```

We might want to make this throw a warning until it's handled. 

https://discourse.julialang.org/t/modelingtoolkit-building-a-daeproblem-and-solving-it-with-ida-with-klu-jacobian-usage/77814

Abhishek-1Bhatt:
Continued from #1510 to resolve #1482

### DIFF
--- a/src/systems/diffeqs/abstractodesystem.jl
+++ b/src/systems/diffeqs/abstractodesystem.jl
@@ -19,7 +19,7 @@ end
 
 function calculate_jacobian(sys::AbstractODESystem;
                             sparse=false, simplify=false, dvs=states(sys))
-    
+
     if isequal(dvs, states(sys))
         cache = get_jac(sys)[]
         if cache isa Tuple && cache[2] == (sparse, simplify)
@@ -27,7 +27,7 @@ function calculate_jacobian(sys::AbstractODESystem;
         end
     end
 
-    rhs = [eq.rhs for eq ∈ full_equations(sys)]
+    rhs = [eq.rhs - eq.lhs for eq ∈ full_equations(sys)] #need du terms on rhs for differentiating wrt du
 
     iv = get_iv(sys)
 
--- a/test/dae_jacobian.jl
+++ b/test/dae_jacobian.jl
@@ -1,0 +1,56 @@
+using ModelingToolkit
+using Sundials, Test, SparseArrays
+
+# Comparing solution obtained by defining explicit Jacobian function with solution obtained from
+# symbolically generated Jacobian
+
+function testjac(res, du, u, p, t) #System of equations
+    res[1] = du[1] - 1.5 * u[1] + 1.0 * u[1] * u[2]
+    res[2] = du[2] + 3 * u[2] - u[1] * u[2]
+end
+
+function testjac_jac(J, du, u, p, gamma, t) #Explicit Jacobian
+    J[1, 1] = gamma - 1.5 + 1.0 * u[2]
+    J[1, 2] = 1.0 * u[1]
+    J[2, 1] = -1 * u[2]
+    J[2, 2] = gamma + 3 - u[1]
+    nothing
+end
+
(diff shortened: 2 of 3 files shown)


ChrisRackauckas:
Thanks!!!

src/systems/diffeqs/abstractodesystem.jl
```diff
@@ -19,15 +19,15 @@ end
...
+
     if isequal(dvs, states(sys))
         cache = get_jac(sys)[]
         if cache isa Tuple && cache[2] == (sparse, simplify)
             return cache[1]
         end
     end
 
-    rhs = [eq.rhs for eq ∈ full_equations(sys)]
+    rhs = [eq.rhs - eq.lhs for eq ∈ full_equations(sys)] #need du terms on rhs for differentiating wrt du
```
YingboMa, line 30: Where did you put the gamma?

Abhishek-1Bhatt: gamma is provided by the solver at runtime, so we just declare some symbolic gamma and use it in `generate_dae_jacobian` to generate a symbolic jacobian by doing `(jacobian wrt u) + (jacobian wrt du)*gamma`. Since its not part of the equation we won't need it here. Some details [here](https://diffeq.sciml.ai/stable/types/dae_types/#:~:text=For%20fully%20implicit,by%20the%20solver) . I also remember reading somewhere that gamma is just `dt`

PR #1534, opened 2022-04-22
Add but_ordered_incidence

YingboMa:


YingboMa:
Hi @pogudingleb and @iliailmer, I am pretty sure the test failure in StructuralIdentifiability is unrelated, but why does it segfault?

PR #1548, opened 2022-05-01
Promote to concrete type by default if it's an array

YingboMa:
`prob.u0 isa Vector{Real}` or `prob.p isa Vector{Real}` could lead to severe performance regression. Technically it's breaking, but I doubt anyone wants that.

CC: @ValentinKaisermayer

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -472,15 +472,31 @@ function mergedefaults(defaults, varmap, vars)
     end
 end
 
-function promote_to_concrete(vs, tofloat=true)
+function promote_to_concrete(vs; tofloat=true, use_union=false)
     if isempty(vs)
         return vs
     end
     T = eltype(vs)
-    if Base.isconcretetype(T) # nothing to do
+    if Base.isconcretetype(T) && (!tofloat || T === float(T)) # nothing to do
         vs
     else
-        C = foldl((t, elem)->promote_type(t, eltype(elem)), vs; init=typeof(first(vs)))
-        convert.(tofloat ? float(C) : C, vs)
+        C = typeof(first(vs))
+        has_int = false
+        I = Int8
+        for v in vs
+            E = eltype(v)
+            C = promote_type(C, E)
+            if E <: Integer
+                has_int = true
+                I = promote_type(I, E)
+            end
+        end
+        if tofloat
+            C = float(C)
+        elseif use_union && has_int && C !== I
+            C = Union{C, I}
+            return copyto!(similar(vs, C), vs)
+        end
+        convert.(C, vs)
     end
 end
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -472,8 +472,8 @@ function mergedefaults(defaults, varmap, vars)
     end
 end
 
-function promote_to_concrete(vs)
-    if isempty(vs) 
+function promote_to_concrete(vs, tofloat=true)
+    if isempty(vs)
         return vs
     end
     T = eltype(vs)
@@ -481,6 +481,6 @@ function promote_to_concrete(vs)
         vs
     else
         C = foldl((t, elem)->promote_type(t, eltype(elem)), vs; init=typeof(first(vs)))
-        convert.(C, vs)
+        convert.(tofloat ? float(C) : C, vs)
     end
 end
--- a/test/odesystem.jl
+++ b/test/odesystem.jl
@@ -641,7 +641,7 @@ end
 #issue 1475 (mixed numeric type for parameters)
 let
     @parameters k1 k2
-    @variables t, A(t)
+    @variables t A(t)
     D = Differential(t)
     eqs = [D(A) ~ -k1*k2*A]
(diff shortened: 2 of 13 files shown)


test/odesystem.jl
```diff
@@ -650,6 +650,11 @@ let
     tspan = (0.0,1.0)
     prob = ODEProblem(sys, u0map, tspan, pmap)
     @test prob.p === Tuple([(Dict(pmap))[k] for k in values(parameters(sys))])
+
+    pmap = [k1 => 1.0, k2 => 1]
```
ChrisRackauckas, line 654:
```suggestion
    pmap = [k1 => 1, k2 => 1]
```

to test the `tofloat`?

ChrisRackauckas:
We would need to make sure DiscreteSystem doesn't do this when building reaction systems, since using Ints isn't too uncommon there. @isaacsas

isaacsas:
I haven't looked through this PR closely, but if it simply promotes mixed arrays to a common type, leaving mixed tuples for parameters alone and pure integer arrays for initial conditions alone, it should be fine.

isaacsas:
(Even for ODEs we'd want parameter tuples to be left alone as Catalyst can generate models where you need integers due to factorials appearing.)

src/utils.jl
```diff
@@ -472,15 +472,15 @@ function mergedefaults(defaults, varmap, vars)
...
         return vs
     end
     T = eltype(vs)
     if Base.isconcretetype(T) # nothing to do
         vs
     else
         C = foldl((t, elem)->promote_type(t, eltype(elem)), vs; init=typeof(first(vs)))
-        convert.(C, vs)
+        convert.(tofloat ? float(C) : C, vs)
     end
```
ValentinKaisermayer, line 485: I think the whole if else case can be replaced:
```julia
C = foldl((t, elem)->promote_type(t, eltype(elem)), vs; init=typeof(first(vs)))
convert.(tofloat ? float(C) : C, vs)
```
That should resolve the test.

ValentinKaisermayer (changes requested):

isaacsas:
I take back my previous comment, this will be breaking for Catalyst per the example in https://github.com/SciML/ModelingToolkit.jl/issues/1475 if the parameter map is a `Dict` instead of tuple (since we have cases where we really want both integer and float inputs and a user might want to use a `Dict` to pass them). 

Could you use a `Union` type to support integers too?

YingboMa:
> I take back my previous comment, this will be breaking for Catalyst per the example in https://github.com/SciML/ModelingToolkit.jl/issues/1475 if the parameter map is a Dict instead of tuple (since we have cases where we really want both integer and float inputs and a user might want to use a Dict to pass them).

Is that problem specific?

> Could you use a Union type to support integers too?
Maybe, but only for parameters.

isaacsas:
It is actually an issue for parameters to any problem type Catalyst can generate, since factorial terms involving parameters can show up in converting to ODEs, SDEs or jumps. I guess having a mix of types could just lead to a parameter tuple being generated as another alternative?

Pure jump problems will usually have pure integer initial conditions (and states). So as long as initial condition tuples/dicts/arrays with a single concrete type lead to a numeric array of that same type there shouldn't be any issues there. Converting a mix of integer and float initial conditions to float should be ok.

YingboMa:
I added the `use_float` option.

isaacsas (commented):
The problem is that this is still going to sometimes lead to crashes by silently converting integer parameters to be floats as the default behavior (for example, if a user has specified a mix of types for default values). I guess the question is whether the default should be silently handling converting for performance reasons, assuming users just messed up in specifying mixed types for parameters or initial conditions, vs. assuming users made their choices intentially and trying to support them.

I guess in either case this is a subtle choice, but I'd say whichever is chosen there should consistently be a warning message printed when a mix of types is found. If we go with conversion as the default the warning should explain everything is getting converted to a float type, and say what should be done to avoid such a conversion. In this case a crash would presumably be clear to a user as occurring due to the parameter having an incorrect type. If the default is not to convert, one could give a warning that multiple types were detected and for performance reasons they should be converted to a single type if possible (or passed as a tuple for parameters).

YingboMa:
It's very common to have mixed values and Julia would promote them to Float64 by default anyway. Cannot Catalyst always use `use_union`?

isaacsas:
Otherwise I guess leave as is and we'll just have to update the Catalyst docs to more clearly explain this issue. Right now in the one tutorial where this issue could arise I made sure to use parameter tuples so it shouldn't break any of our examples I think.

YingboMa:
> Otherwise I guess leave as is and we'll just have to update the Catalyst docs to more clearly explain this issue. Right now in the one tutorial where this issue could arise I made sure to use parameter tuples so it shouldn't break any of our examples I think.

I think this is better, because functionalities like sensitivity analysis might not work very nicely when the eltype of parameters is `Union`. Letting the user to do conscious decisions would be better.

PR #1554, opened 2022-05-04
Optimization

YingboMa:
AbstractAlgebra
```julia
julia> M = netstoichmat(prnbng.rn); size(M)
(1122, 24388)

julia> aaM = matrix(AbstractAlgebra.Integers{Int}(), M');

julia> @time aaN = AbstractAlgebra.nullspace(aaM);
164.074442 seconds (17 allocations: 209.207 MiB, 0.01% gc time)

julia> iszero(M'aaN[2].entries)
true
```

MTK
```julia
julia> N = @time ModelingToolkit.nullspace(M');
  8.975512 seconds (9 allocations: 208.843 MiB, 1.23% gc time)

julia> iszero(M'N)
true
```

### DIFF
--- a/src/structural_transformation/bareiss.jl
+++ b/src/structural_transformation/bareiss.jl
@@ -114,20 +114,40 @@ else
     end
 end
 
+function bareiss_update!(zero!, M::StridedMatrix, k, swapto, pivot, prev_pivot::Base.BitInteger)
+    flag = zero(prev_pivot)
+    prev_pivot = Base.MultiplicativeInverses.SignedMultiplicativeInverse(prev_pivot)
+    @inbounds for i in k+1:size(M, 2)
+        Mki = M[k,i]
+        @simd ivdep for j in k+1:size(M, 1)
+            M[j,i], r = divrem(M[j,i]*pivot - M[j,k]*Mki, prev_pivot)
+            flag = flag | r
+        end
+    end
+    iszero(flag) || error("Overflow occurred")
+    zero!(M, k+1:size(M, 1), k)
+end
+
 function bareiss_update!(zero!, M::StridedMatrix, k, swapto, pivot, prev_pivot)
-    for i in k+1:size(M, 2), j in k+1:size(M, 1)
+    @inbounds for i in k+1:size(M, 2), j in k+1:size(M, 1)
         M[j,i] = exactdiv(M[j,i]*pivot - M[j,k]*M[k,i], prev_pivot)
     end
     zero!(M, k+1:size(M, 1), k)
 end
 
 @views function bareiss_update!(zero!, M::AbstractMatrix, k, swapto, pivot, prev_pivot)
+    if prev_pivot isa Base.BitInteger
+        prev_pivot = Base.MultiplicativeInverses.SignedMultiplicativeInverse(prev_pivot)
+    end
     V = M[k+1:end, k+1:end]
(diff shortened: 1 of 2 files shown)


isaacsas:
Is it possible to have nullspace also take an optional vector to fill with the indices of the pivot columns for the original matrix? I think we need this information too to decide which variables to keep and which to eliminate (using the nullspace matrix to construct the affine relationship between them).

YingboMa:
Do you want https://github.com/SciML/ModelingToolkit.jl/pull/1554/files#diff-f99d8e0b376f38a041d73526be91d1e846276c9b5d0241da8d37a835cf095723R211 to be an option?

isaacsas:
That link drops me at `is_linear_equations=...` in `aag_bareiss` is that what you meant? We are essentially doing a form of alias elimination by wanting to determine the algebraic relationship between the dependent (i.e. free) variables and the independent variables, as determined by the transpose of the net stoichiometry matrix.

If N is the conservation law matrix, then we know that if x is our symbolic chemical species vector then 

N*x = c

for an arbitrary vector of constants c (that depends on the initial conditions).

We want to use knowledge of the dependent variables (corresponding to free variables during row reduction) to reorder x to

x = [x_indep,x_dep]

and reorder/rescale N to

N = [-A I]

so that 

x_dep = A*x_indep + c

At that point I wanted to either move the x_dep variables over to be algebraic constraints in the `ReactionSystem` instead of species, or to explicitly eliminate them from the reactions by substitution. I'll probably offer both approaches since they can both be useful representations.

isaacsas:
Ahh, now it is taking me to the top of `aag_bareiss!`. I only glanced at that, but yes I guess we want something like that so we also know the solvable variables in the equation N*x=c I wrote above. Given that the nullspace calculation has that info from the echelon forms it seemed like it would be easy to add returning the independent column indices too.

YingboMa:
Ah, linking from a diff is a terrible idea. I meant to link https://github.com/SciML/ModelingToolkit.jl/blob/41cc6eaed7822dfe746b0d1e7d1c15b9e121d710/src/structural_transformation/bareiss.jl#L211

isaacsas:
That just stores the column permutations though right? We need to apply the inverse of the column pivots to B to pick out the columns in the original matrix that correspond to pivot columns. (Or alternatively we would need the complementary columns that correspond to free variables.) So one has to actually keep track / calculate the columns that have (or don't have) pivots at some point.

isaacsas:
Sorry, meant inverse of the column permutations.

isaacsas:
If it is easier I can just add a PR for this that adds an optional kwarg to `nullspace` to calculate and return this info. It would be turned off by default, so wouldn't have any impact on the performance usually.

YingboMa:
Do you mean https://github.com/SciML/ModelingToolkit.jl/blob/41cc6eaed7822dfe746b0d1e7d1c15b9e121d710/src/structural_transformation/bareiss.jl#L216 ? I already permute the variables back.

isaacsas:
If you apply the inverse column permutation to `B`'s columns, I then need the indices of the columns that contain a pivot (which there should be `rank` of).

YingboMa:
Yeah, feel free to open a PR. Seems like you just need to return `rank` and `apply_inv_pivot_rows!(collect(1:n), column_pivots)` as well.

isaacsas:
Sounds good. I'll put one together in the next day or two.

PR #1557, opened 2022-05-05
Remove `OptimizationFunction` and `OptimizationProblem` definitions and add `generate_jacobian(::OptimizationSystem)`

Vaibhavdixit02:
The motivation behind this is that the current interface doesn't match that for the other AD backends that are supported in GalacticOptim, this will be moved into `instantiate_function` in GalacticOptim https://github.com/SciML/GalacticOptim.jl/pull/237.

### DIFF
--- a/src/systems/optimization/optimizationsystem.jl
+++ b/src/systems/optimization/optimizationsystem.jl
@@ -78,13 +78,21 @@ function OptimizationSystem(op, states, ps;
     process_variables!(var_to_name, defaults, states)
     process_variables!(var_to_name, defaults, ps)
     isempty(observed) || collect_var_to_name!(var_to_name, (eq.lhs for eq in observed))
-
     OptimizationSystem(
                        value(op), states, ps, var_to_name,
                        observed,
                        equality_constraints, inequality_constraints,
-                       name, systems, defaults, checks = checks
+                       name, systems, defaults; checks = checks
                       )
+end
+
+function calculate_jacobian(sys::OptimizationSystem)
+    expand_derivatives.(jacobian(equations(sys), states(sys)))
+end
+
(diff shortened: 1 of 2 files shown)


src/systems/optimization/optimizationsystem.jl
```diff
@@ -121,153 +129,3 @@ equations(sys::OptimizationSystem) = isempty(get_systems(sys)) ? get_op(sys) : g
 namespace_expr(sys::OptimizationSystem) = namespace_expr(get_op(sys), sys)
 
 hessian_sparsity(sys::OptimizationSystem) = hessian_sparsity(get_op(sys), states(sys))
-
-struct AutoModelingToolkit <: DiffEqBase.AbstractADType end
-
-DiffEqBase.OptimizationProblem(sys::OptimizationSystem,args...;kwargs...) =
-    DiffEqBase.OptimizationProblem{true}(sys::OptimizationSystem,args...;kwargs...)
```
ChrisRackauckas, line 128: These dispatches are part of the ModelingToolkit standard system interface, so they should not be deleted.

ChrisRackauckas:
The deletes don't make sense. I agree with you that from GalacticOptim.jl, we should expose an interface that looks like the other GalacticOptim.jl OptimizationFunctions. However, ModelingToolkit's interface always has XXProblem and XXFunction dispatches for a system, so those should exist in ModelingToolkit. They are required if someone starts from a symbolic model, since otherwise, how do you turn a symbolic model into a numerical one?

Vaibhavdixit02:
Ah, I was missing that part. I'll revert that and adjust the PR accordingly.

YingboMa:
Maybe just open another PR?

PR #1563, opened 2022-05-08
Removes `OptimizationFunction` and add GalacticOptim as a dependency

Vaibhavdixit02:
Also, changes default for `grad` and `hess` to `true` in `OptimizationProblem` and adds `generate_jacobian` to use for constraint jacobian. Related to https://github.com/SciML/GalacticOptim.jl/pull/237.

### DIFF
--- a/src/ModelingToolkit.jl
+++ b/src/ModelingToolkit.jl
@@ -25,7 +25,6 @@ using DocStringExtensions
 using Base: RefValue
 using Combinatorics
 import IfElse
-using GalacticOptim
 import Distributions
 
 RuntimeGeneratedFunctions.init(@__MODULE__)
--- a/src/ModelingToolkit.jl
+++ b/src/ModelingToolkit.jl
@@ -25,7 +25,7 @@ using DocStringExtensions
 using Base: RefValue
 using Combinatorics
 import IfElse
-
+using GalacticOptim
 import Distributions
 
 RuntimeGeneratedFunctions.init(@__MODULE__)
--- a/src/ModelingToolkit.jl
+++ b/src/ModelingToolkit.jl
@@ -133,6 +133,7 @@ include("systems/diffeqs/basic_transformations.jl")
 include("systems/jumps/jumpsystem.jl")
 
 include("systems/nonlinear/nonlinearsystem.jl")
+include("systems/nonlinear/modelingtoolkitize.jl")
 
 include("systems/optimization/optimizationsystem.jl")
 
--- a/src/systems/nonlinear/nonlinearsystem.jl
+++ b/src/systems/nonlinear/nonlinearsystem.jl
@@ -137,6 +137,23 @@ function generate_jacobian(sys::NonlinearSystem, vs = states(sys), ps = paramete
     return build_function(jac, vs, ps; kwargs...)
 end
 
+function calculate_hessian(sys::NonlinearSystem; sparse=false, simplify=false)
+    rhs = [eq.rhs for eq ∈ equations(sys)]
+    vals = [dv for dv in states(sys)]
+    if sparse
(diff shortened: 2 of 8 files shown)


src/ModelingToolkit.jl
```diff
@@ -25,7 +25,7 @@ using DocStringExtensions
 using Base: RefValue
 using Combinatorics
 import IfElse
-
+using GalacticOptim
```
ChrisRackauckas, line 28: I don't see anything that requires this.

src/systems/optimization/optimizationsystem.jl
```diff
@@ -78,15 +78,23 @@ function OptimizationSystem(op, states, ps;
...
+                       name, systems, defaults; checks = checks
                       )
 end
 
+function calculate_jacobian(sys::OptimizationSystem)
+    expand_derivatives.(jacobian(equations(sys), states(sys)))
+end
+
+function generate_jacobian(sys::OptimizationSystem, vs = states(sys), ps = parameters(sys); kwargs...)
+    jac = calculate_jacobian(sys)
```
ChrisRackauckas, line 94: wait, the Jacobian should be on the constraints?

Vaibhavdixit02: The constraints are not currently used in any way right now, I think we would want to move to using them but for now this is used directly since `cons` is used to create an `OptimizationSystem` https://github.com/SciML/GalacticOptim.jl/blob/b1eacc11b87fd8e6d7cd5290b161377828cc5098/src/function/mtk.jl#L35 considering it as the objective

ChrisRackauckas: We shouldn't bake something incorrect like this... Just change that to NonlinearSystem

Vaibhavdixit02: Sure, makes sense. That would need a `generate_hessian` dispatch then

ChrisRackauckas: For second derivatives of a vector system?

Vaibhavdixit02: Yes

src/systems/optimization/optimizationsystem.jl
```diff
@@ -145,8 +142,8 @@ symbolically calculating numerical enhancements.
 function DiffEqBase.OptimizationProblem{iip}(sys::OptimizationSystem, u0map,
                                           parammap=DiffEqBase.NullParameters();
                                           lb=nothing, ub=nothing,
-                                          grad = false,
-                                          hess = false, sparse = false,
+                                          grad = true,
+                                          hess = true, sparse = false,
```
ChrisRackauckas, line 146: these should default to false for consistency

src/systems/nonlinear/nonlinearsystem.jl
```diff
@@ -137,6 +137,23 @@ function generate_jacobian(sys::NonlinearSystem, vs = states(sys), ps = paramete
     return build_function(jac, vs, ps; kwargs...)
 end
 
+function calculate_hessian(sys::NonlinearSystem; sparse=false, simplify=false)
+    rhs = [eq.rhs for eq ∈ equations(sys)]
+    vals = [dv for dv in states(sys)]
+    if sparse
+        hess = [sparsehessian(rhs[i], vals, simplify=simplify) for i in 1:length(rhs)]
```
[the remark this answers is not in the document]
ChrisRackauckas: That's good enough.

ChrisRackauckas:
I think this is good to go when the GalacticOptim dep is removed.

src/systems/nonlinear/nonlinearsystem.jl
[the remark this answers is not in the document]
YingboMa: Do they expect it in the vector of arrays form?

Issue #1537, opened 2022-04-25
structural_simplify add derivative equation automatically, UndefVar problem and equation form influence ODAEProblem.


YangYuebao:
I met three problems in my case.

1. structural_simplify add derivative equation automatically
2. UndefVar problem #1533
3. Form of equations has an inluence on the behaviour of ODAEProblem

I'm trying to do a transient pipe modeling, and met the same problem like this #1516
```
using ModelingToolkit, DifferentialEquations
#isclosed(x)=length(equations(x))-length(states(x))
@variables t
@variables begin
    p_1(t)
    p_2(t)
    rho_1(t)
    rho_2(t)
    rho_3(t)
    u_1(t)
    u_2(t)
    u_3(t)
    mo_1(t)
    mo_2(t)
    mo_3(t)
    Ek_1(t)
    Ek_2(t)
    Ek_3(t)
end

@parameters dx = 100 f = 0.3 pipe_D = 0.4

D = Differential(t)

eqs = [
    p_1 ~ 1.2e5
    p_2 ~ 1e5
    u_1 ~ 10
    mo_1 ~ u_1 * rho_1
    mo_2 ~ u_2 * rho_2
    mo_3 ~ u_3 * rho_3
    Ek_1 ~ rho_1 * u_1 * u_1
    Ek_2 ~ rho_2 * u_2 * u_2
    Ek_3 ~ rho_3 * u_3 * u_3
    rho_1 ~ p_1 / 273.11 / 300
    rho_2 ~ (p_1 + p_2) * 0.5 / 273.11 / 300
    rho_3 ~ p_2 / 273.11 / 300
    D(rho_2) ~ (mo_1 - mo_3) / dx
    D(mo_2) ~ (Ek_1 - Ek_3 + p_1 - p_2) / dx - f / 2 / pipe_D * u_2 * u_2
]

@named trans = ODESystem(eqs, t)

sys = structural_simplify(trans)

n = 3
u = 0 * ones(n)
rho = 1.2 * ones(n)

u0 = [
    p_1 => 1.2e5
    p_2 => 1e5
    u_1 => 0
    u_2 => 0.1
    u_3 => 0.2
    rho_1 => 1.1
    rho_2 => 1.2
    rho_3 => 1.3
    mo_1 => 0
    mo_2 => 1
    mo_3 => 2
]
prob = ODAEProblem(sys, u0, (0, 0.1))
[...]

Issue #1516, opened 2022-04-07
Key not found when calling ODAEProblem


LiborKudela:
Hello,

I'm a user of OpenModelica and I have only a little experience with Julia. I love ModelingToolkit.jl, and I am very excited to use it.
I have managed to use the acausal modeling to create some toys and even calculate the adjoint sensitivity which OpenModelica cannot do.

My last toy uses stream connectors:
I have created a few components, but I cannot create ODEAProblem. I'm not sure what I'm doing wrong and whether it is related to the stream connectors at all, but the error seems to come from a place I am unable to solve (MWE below). I will be grateful for any reactions/hints. (I'm using Julia 1.7.2)

Thank you

The error I get:
```julia
KeyError: key 7 not found
getindex at dict.jl:481 [inlined]
torn_system_with_nlsolve_jacobian_sparsity(state::TearingState{ODESystem}, var_eq_matching::ModelingToolkit.BipartiteGraphs.Matching{ModelingToolkit.BipartiteGraphs.Unassigned, Vector{Union{ModelingToolkit.BipartiteGraphs.Unassigned, Int64}}}, var_sccs::Vector{Vector{Int64}}, nlsolve_scc_idxs::Vector{Int64}, eqs_idxs::Vector{Int64}, states_idxs::Vector{Int64}) at codegen.jl:84
build_torn_function(sys::ODESystem; expression::Bool, jacobian_sparsity::Bool, checkbounds::Bool, max_inlining_size::Nothing, kw::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}}) at codegen.jl:346
build_torn_function at codegen.jl:238 [inlined]
ODAEProblem{true}(sys::ODESystem, u0map::Matrix{Pair{Num, Float64}}, tspan::Tuple{Float64, Float64}, parammap::Vector{Any}; callback::Nothing, kwargs::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}}) at codegen.jl:528
ODAEProblem{true}(sys::ODESystem, u0map::Matrix{Pair{Num, Float64}}, tspan::Tuple{Float64, Float64}, parammap::Vector{Any}) at codegen.jl:528
ODAEProblem(::ODESystem, ::Vararg{Any}; kw::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}}) at codegen.jl:506
[...]

PR #1575, opened 2022-05-14
Add the dummy derivative pass

YingboMa:
```julia
julia> u0 = [
         D(x)    => 0.0,
         D(y)    => 0.0,
         x       => sqrt(2)/2,
         y       => sqrt(2)/2,
         T       => 0.0
       ];

julia> p = [
           L => 1.0,
           g => 9.8
       ];

julia> equations(pendulum2)
3-element Vector{Equation}:
 Differential(t)(Differential(t)(x(t))) ~ T(t)*x(t)
 Differential(t)(Differential(t)(y(t))) ~ T(t)*y(t) - g
 0 ~ x(t)^2 + y(t)^2 - (L^2)

julia> prob_auto = DAEProblem(structural_simplify(pendulum2),zeros(length(u0)),u0,(0.0,10.0),p);

julia> sol = solve(prob_auto, DFBDF());
```

pendulum_dd_mtk

The solver would error out if I integrate it for longer. Since `x''(t) == d/dt x'(t)` may not hold if `x''` is a dummy derivative, we might have to implement manifold projection for DAE solvers to ensure that the dummy derivatives don't drift too much.

This PR fixes https://github.com/SciML/ModelingToolkit.jl/issues/1537 and https://github.com/SciML/ModelingToolkit.jl/issues/1516

### DIFF
--- a/docs/src/mtkitize_tutorials/modelingtoolkitize_index_reduction.md
+++ b/docs/src/mtkitize_tutorials/modelingtoolkitize_index_reduction.md
@@ -29,8 +29,8 @@ p = [9.8, 1]
 tspan = (0, 10.0)
 pendulum_prob = ODEProblem(pendulum_fun!, u0, tspan, p)
 traced_sys = modelingtoolkitize(pendulum_prob)
-pendulum_sys = structural_simplify(traced_sys)
-prob = ODAEProblem(pendulum_sys, Pair[], tspan)
+pendulum_sys = structural_simplify(dae_index_lowering(traced_sys))
+prob = ODAEProblem(pendulum_sys, [], tspan)
 sol = solve(prob, Tsit5(),abstol=1e-8,reltol=1e-8)
 plot(sol, vars=states(traced_sys))
 ```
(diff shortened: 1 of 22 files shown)


YingboMa:
The ModelingToolkitStandardLibrary test failure is actually a good thing. It's now detecting some ODEs are actually implicit DAEs so MTK rejects to build the wrong problem. We need better implicit DAE handling to make ModelingToolkitStandardLibrary pass again.

ChrisRackauckas:
> The ModelingToolkitStandardLibrary test failure is actually a good thing. It's now detecting some ODEs are actually implicit DAEs so MTK rejects to build the wrong problem. We need better implicit DAE handling to make ModelingToolkitStandardLibrary pass again.

How were they handled before?

YingboMa:
@ValentinKaisermayer @ChrisRackauckas can we mark the failing tests ModelingToolkitStandardLibrary for now? Writing implicit to semi-implicit pass will take a while.

ValentinKaisermayer:
> The ModelingToolkitStandardLibrary test failure is actually a good thing. It's now detecting some ODEs are actually implicit DAEs so MTK rejects to build the wrong problem. We need better implicit DAE handling to make ModelingToolkitStandardLibrary pass again.

Maybe such a test case should be part of MTK.

YingboMa:
> Maybe such a test case should be part of MTK.

Yeah, do you want to make a PR? You can put the definitions in examples, and add the tests in `test/state_selection.jl`

Issue #1576, opened 2022-05-15
8.11.1 changes to show break Catalyst and JumpSystems


isaacsas:
```julia
using ModelingToolkit, DiffEqJump
@parameters β γ
@variables t S(t) I(t) R(t)
rate₁   = β*S*I
affect₁ = [S ~ S - 1, I ~ I + 1]
rate₂   = γ*I
affect₂ = [I ~ I - 1, R ~ R + 1]
j₁      = ConstantRateJump(rate₁,affect₁)
j₂      = ConstantRateJump(rate₂,affect₂)
j₃      = MassActionJump(2*β+γ, [R => 1], [S => 1, R => -1])
@named js      = JumpSystem([j₁,j₂,j₃], t, [S,I,R], [β,γ])
```
gives
```julia
Error showing value of type JumpSystem{RecursiveArrayTools.ArrayPartition{Any, Tuple{Vector{MassActionJump}, Vector{ConstantRateJump}, Vector{VariableRateJump}}}}:
ERROR: type MassActionJump has no field lhs
Stacktrace:
  [1] getproperty
    @ ./Base.jl:38 [inlined]
  [2] (::ModelingToolkit.var"#139#140")(eq::MassActionJump{Num, Vector{Pair{Num, Int64}}, Vector{Pair{Num, Int64}}, Nothing})
    @ ModelingToolkit ~/.julia/packages/ModelingToolkit/RkmkM/src/systems/abstractsystem.jl:744
  [3] iterate
    @ ./generator.jl:47 [inlined]
  [4] _simple_count_helper(g::Base.Generator{RecursiveArrayTools.ArrayPartition{Any, Tuple{Vector{MassActionJump}, Vector{ConstantRateJump}, Vector{VariableRateJump}}}, ModelingToolkit.var"#139#140"}, init::Int64)
    @ Base ./reduce.jl:1293
  [5] _simple_count(pred::Function, itr::RecursiveArrayTools.ArrayPartition{Any, Tuple{Vector{MassActionJump}, Vector{ConstantRateJump}, Vector{VariableRateJump}}}, init::Int64)
    @ Base ./reduce.jl:1289
  [6] _count
    @ ./reducedim.jl:442 [inlined]
  [7] #count#767
    @ ./reducedim.jl:440 [inlined]
[...]

PR #1578, opened 2022-05-15
fix show crashes for non-equation systems

isaacsas:
Closes https://github.com/SciML/ModelingToolkit.jl/issues/1576


YingboMa:
Thanks!

Issue #1583, opened 2022-05-18
Warning: Unrecognized keyword arguments found.


isaacsas:
```julia
using Catalyst, OrdinaryDiffEq

rn = @reaction_network begin
    (a1,a2), C <--> 0
    (k1,k2), A + B <--> C    
    b1, 0 <-- B
end a1 a2 k1 k2 b1
ps = ones(5)
u0 = [10.0,20.0,0.0]
tspan = (0,100.0)
oprob1 = ODEProblem(rn, u0, tspan, ps)
sol1 = solve(oprob1, Tsit5())
```
gives
```
┌ Warning: Warning: Unrecognized keyword arguments found. Future versions
 will error.                                                             
│ The only allowed keyword arguments to `solve` are:                     
│ (:dense, :saveat, :save_idxs, :tstops, :d_discontinuities, :save_everys
tep, :save_on, :save_start, :save_end, :initialize_save, :adaptive, :abst
ol, :reltol, :dt, :dtmax, :dtmin, :force_dtmin, :internalnorm, :controlle
r, :gamma, :beta1, :beta2, :qmax, :qmin, :qsteady_min, :qsteady_max, :qol
dinit, :failfactor, :calck, :alias_u0, :maxiters, :callback, :isoutofdoma
in, :unstable_check, :verbose, :merge_callbacks, :progress, :progress_ste
ps, :progress_name, :progress_message, :timeseries_errors, :dense_errors,
 :calculate_errors, :initializealg, :alg, :save_noise, :delta, :seed, :al
g_hints, :kwargshandle, :trajectories, :batch_size, :sensealg, :advance_t
o_tstop, :stop_at_next_tstop, :default_set, :second_time)                
│ See https://diffeq.sciml.ai/stable/basics/common_solver_opts/ for more 
details.                                                                 
│                                                                        
│ Set kwargshandle=KeywordArgError for an error message and more details.
│ Set kwargshandle=KeywordArgSilent to ignore this message.              
└ @ DiffEqBase ~/.julia/packages/DiffEqBase/foWeO/src/solve.jl:281  
```
Not sure what is going on here as there aren't any keyword arguments...

Environment (1.8-rc3):
```
Status `~/.julia/environments/v1.8/Project.toml`
  [6e4b80f9] BenchmarkTools v1.3.1
  [5d742f6a] CSVFiles v1.0.1
  [479239e8] Catalyst v10.8.0
  [a93c6f00] DataFrames v1.3.4
[...]

PR #1584, opened 2022-05-18
Don't pass to DiffEqBase.ODEProblem invalid keyword check_length

isaacsas:
Fixes https://github.com/SciML/ModelingToolkit.jl/issues/1583

If this is ok, would be good to get a bug fix release once merged.


src/systems/diffeqs/abstractodesystem.jl
```diff
@@ -774,10 +776,10 @@ numerical enhancements.
 struct ODEProblemExpr{iip} end
 
 function ODEProblemExpr{iip}(sys::AbstractODESystem,u0map,tspan,
-                             parammap=DiffEqBase.NullParameters();
+                             parammap=DiffEqBase.NullParameters(); check_length=true,
```
YingboMa, line 779: Trailing whitespace.

YingboMa (commented):
There are quite a few trailing whitespaces.

isaacsas:
I had turned that off as I found vscode was having frequent pauses with it on unfortunately...

isaacsas:
@YingboMa ok to make a new patch release?

PR #1587, opened 2022-05-19
[RFC] Get depvars and indvars in the right type for MOL

xtalax:
This change reflects something that is done internally in MethodOfLines.jl, which may have more general applicability. 

Gets all dvs and ivs in their naked Symbolics.jl type, which is more useful for applying rules based on SymbolicUtils.jl.

What do you think of this @zoemcc @YingboMa?

I will add tests if this is decided to be a good change.

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -22,18 +22,18 @@ function detime_dvs(op)
     elseif operation(op) isa Sym
         Sym{Real}(nameof(operation(op)))
     else
-        similarterm(op, operation(op),detime_dvs.(arguments(op)))
+        similarterm(op, operation(op), detime_dvs.(arguments(op)))
     end
 end
 
-function retime_dvs(op::Sym,dvs,iv)
-    Sym{FnType{Tuple{symtype(iv)}, Real}}(nameof(op))(iv)
+function retime_dvs(op::Sym, dvs, iv)
+    Sym{FnType{Tuple{symtype(iv)},Real}}(nameof(op))(iv)
 end
 
 function retime_dvs(op, dvs, iv)
     istree(op) ?
-        similarterm(op, operation(op), retime_dvs.(arguments(op),(dvs,),(iv,))) :
-        op
+    similarterm(op, operation(op), retime_dvs.(arguments(op), (dvs,), (iv,))) :
+    op
 end
 
 modified_states!(mstates, e::Equation, statelist=nothing) = get_variables!(mstates, e.lhs, statelist)
@@ -49,7 +49,7 @@ macro showarr(x)
     end
 end
 
-@deprecate substitute_expr!(expr,s) substitute(expr,s)
+@deprecate substitute_expr!(expr, s) substitute(expr, s)
 
 function states_to_sym(states::Set)
     function _states_to_sym(O)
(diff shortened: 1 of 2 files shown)


src/utils.jl
```diff
@@ -501,13 +503,79 @@ function promote_to_concrete(vs; tofloat=true, use_union=false)
...
+        end
+    end
+    return filter(!iszero, orders)
+end
+
+"""
+find all the dependent variables given by depvar_ops in an expression
+"""
+function get_depvars(eq, depvar_ops)
+    S = Symbolics
```
YingboMa, line 560: These functions are already available in MTK.

src/utils.jl
```diff
@@ -501,13 +503,79 @@ function promote_to_concrete(vs; tofloat=true, use_union=false)
...
+    SU = SymbolicUtils
+    depvars = Set()
+    if eq isa Num
+        eq = eq.val
```
YingboMa, line 564: use `unwrap`

src/utils.jl
```diff
@@ -501,13 +503,79 @@ function promote_to_concrete(vs; tofloat=true, use_union=false)
...
+    end
+    if S.istree(eq)
+        if eq isa Term && any(u -> isequal(operation(eq), u), depvar_ops)
```
YingboMa, line 567: This is not generic.

src/utils.jl
```diff
@@ -501,13 +503,79 @@ function promote_to_concrete(vs; tofloat=true, use_union=false)
...
+            push!(depvars, eq)
+        else
+            for o in map(x -> get_depvars(x, depvar_ops), SU.arguments(eq))
```
YingboMa, line 570: You don't need to allocate the array if you are only iterating on it.

src/utils.jl
```diff
@@ -501,13 +503,79 @@ function promote_to_concrete(vs; tofloat=true, use_union=false)
...
+                union!(depvars, o)
+            end
+        end
+    end
+    return depvars
+end
+
+@inline function get_all_depvars(pdesys, depvar_ops)
+    pdeeqs = pdesys.eqs # Vector
```
YingboMa, line 579: `get_eqs`

src/utils.jl
```diff
@@ -501,13 +503,79 @@ function promote_to_concrete(vs; tofloat=true, use_union=false)
...
+function differential_order(eq, x::Symbolics.Symbolic)
+    S = Symbolics
+    SU = SymbolicUtils
+    orders = Set{Int}()
+    if S.istree(eq)
+        op = SU.operation(eq)
+        if op isa Differential
+            push!(orders, count_differentials(eq, x))
+        else
+            for o in map(ch -> differential_order(ch, x), SU.arguments(eq))
```
YingboMa, line 548: same thing here.

src/utils.jl
```diff
@@ -501,13 +503,79 @@ function promote_to_concrete(vs; tofloat=true, use_union=false)
...
+the order of a PDE.
+"""
+function count_differentials(term, x::Symbolics.Symbolic)
+    S = Symbolics
+    SU = SymbolicUtils
+    if !S.istree(term)
+        return 0
+    else
+        op = SU.operation(term)
+        count_children = sum(map(arg -> count_differentials(arg, x), SU.arguments(term)))
```
YingboMa, line 528: There's two arguments `sum`

YingboMa (commented):
Try not to change the style in the rest of the file. It makes PR harder to review.

PR #1590, opened 2022-05-21
Add .JuliaFormatter

YingboMa:
It's https://github.com/SciML/Sundials.jl/pull/357 with
```julia
normalize_line_endings = "unix"
```

### DIFF
--- a/src/bipartite_graph.jl
+++ b/src/bipartite_graph.jl
@@ -1,8 +1,8 @@
 module BipartiteGraphs
 
 export BipartiteEdge, BipartiteGraph, DiCMOBiGraph, Unassigned, unassigned,
-        Matching, ResidualCMOGraph, InducedCondensationGraph, maximal_matching,
-        construct_augmenting_path!, MatchedCondensationGraph
+       Matching, ResidualCMOGraph, InducedCondensationGraph, maximal_matching,
+       construct_augmenting_path!, MatchedCondensationGraph
 
 export 𝑠vertices, 𝑑vertices, has_𝑠vertex, has_𝑑vertex, 𝑠neighbors, 𝑑neighbors,
        𝑠edges, 𝑑edges, nsrcs, ndsts, SRC, DST, set_neighbors!, invview,
@@ -25,10 +25,9 @@ Base.size(u::Unassigned) = ()
 Base.iterate(u::Unassigned) = (unassigned, nothing)
 Base.iterate(u::Unassigned, state) = nothing
 
-Base.show(io::IO, ::Unassigned) =
-    printstyled(io, "u"; color=:light_black)
+Base.show(io::IO, ::Unassigned) = printstyled(io, "u"; color = :light_black)
 
-struct Matching{U #=> :Unassigned =#, V<:AbstractVector} <: AbstractVector{Union{U, Int}}
+struct Matching{U, V <: AbstractVector} <: AbstractVector{Union{U, Int}} #=> :Unassigned =#
     match::V
     inv_match::Union{Nothing, V}
 end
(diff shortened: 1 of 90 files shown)


src/bipartite_graph.jl
```diff
@@ -138,17 +140,26 @@ badjlist = [[1,2,5,6],[3,4,6]]
...
+BipartiteGraph(fadj::AbstractVector, badj::Union{AbstractVector,Integer} = maximum(maximum, fadj); metadata = nothing) = BipartiteGraph(mapreduce(length,
+                                                                                                                                                  +,
+                                                                                                                                                  fadj;
+                                                                                                                                                  init = 0),
+                                                                                                                                        fadj,
+                                                                                                                                        badj,
+                                                                                                                                        metadata)
```
ChrisRackauckas, line 159: oof I just don't like that it does this haha. That's ugly.

YingboMa:
https://github.com/domluna/JuliaFormatter.jl/issues/582

test/structural_transformation/runtests.jl
```diff
@@ -1,6 +1,14 @@
...
+end
+@safetestset "Index Reduction & SCC" begin
+    include("index_reduction.jl")
+end
+@safetestset "Tearing" begin
+    include("tearing.jl")
+end
+@safetestset "Bareiss" begin
+    include("bareiss.jl")
+end
```
ChrisRackauckas, line 14: @domluna is there a way to stop it from doing this to the runtests.jl scripts?

YingboMa: I think this is fine, but we could use something like
```julia
function p_begin(ds::SciMLStyle, cst::CSTParser.EXPR, s::State)
    style = getstyle(ds)
    t = FST(Begin, cst, nspaces(s))
    add_node!(t, pretty(style, cst[1], s), s)
    if length(cst) == 2
        add_node!(t, Whitespace(1), s)
        add_node!(t, pretty(style, cst[end], s), s, join_lines = true)
    else
        stmt_idxs = 2:length(cst)-1
        if length(stmt_idxs) == 1
            add_node!(t, Whitespace(1), s)
            add_node!(t, pretty(style, cst[2], s), s, join_lines = true)
            add_node!(t, Whitespace(1), s)
            add_node!(t, pretty(style, cst[end], s), s, join_lines = true)
        else
            nodes = CSTParser.EXPR[]
            for i = stmt_idxs
                push!(nodes, cst[i])
            end
            s.indent += s.opts.indent
            add_node!(t, p_block(style, nodes, s), s, max_padding = s.opts.indent)
            s.indent -= s.opts.indent
            add_node!(t, pretty(style, cst[end], s), s)
        end
    end
    t
end
```
which would result in
```julia
julia> print(format_text("""begin include("bareiss.jl") end"""))
begin include("bareiss.jl") end
julia> print(format_text("""begin include("bareiss.jl"); 1 end"""))
begin
    include("bareiss.jl")
    1
end
```

ChrisRackauckas: I like that.

domluna: You'd have to create an override like @YingboMa did. That looks like it works pretty well

src/systems/jumps/jumpsystem.jl
```diff
@@ -38,7 +38,7 @@ struct JumpSystem{U <: ArrayPartition} <: AbstractTimeDependentSystem
     """The parameters of the system. Must not contain the independent variable."""
     ps::Vector
     """Array variables."""
-    var_to_name
+    var_to_name::Any
```
isaacsas, line 41: Maybe it doesn't matter, but would it make more sense for this to be `Dict{Symbol,Any}` rather than `Any` as the formatter is selecting?

domluna: this can be disabled with https://github.com/domluna/JuliaFormatter.jl#annotate_untyped_fields_with_any

YingboMa: I prefer to have untyped fields to be any, but yes, in this case, it should be `Dict{Symbol,Any}`

src/systems/systemstructure.jl
```diff
@@ -6,23 +6,24 @@ using SymbolicUtils: istree, operation, arguments, Symbolic
...
-    similarterm=(x, f, args; kws...) -> similarterm(
-        x, f, args, SymbolicUtils.symtype(x);
-        metadata=SymbolicUtils.metadata(x), kws...
-    )
-)(expr)
+function quick_cancel_expr(expr)
+    Rewriters.Postwalk(quick_cancel,
+                       similarterm = (x, f, args; kws...) -> similarterm(x, f, args,
+                                                                         SymbolicUtils.symtype(x);
+                                                                         metadata = SymbolicUtils.metadata(x),
```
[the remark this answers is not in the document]
YingboMa: 92 character limit is a recommendation, not a strict rule.

YingboMa:
Thanks everyone for reviewing! Seems like there are only a few edge cases left that can be resolved by manual formatting. Let's track this in an issue.

src/structural_transformation/symbolics_tearing.jl
```diff
@@ -246,10 +258,13 @@ function tearing(state::TearingState; kwargs...)
     state.structure.solvable_graph === nothing && find_solvables!(state; kwargs...)
     complete!(state.structure)
     @unpack graph, solvable_graph = state.structure
-    algvars = BitSet(findall(v->isalgvar(state.structure, v), 1:ndsts(graph)))
+    algvars = BitSet(findall(v -> isalgvar(state.structure, v), 1:ndsts(graph)))
     aeqs = algeqs(state.structure)
     var_eq_matching = Matching{Union{Unassigned, SelectedState}}(tear_graph_modia(state.structure;
-        varfilter=var->var in algvars, eqfilter=eq->eq in aeqs))
+                                                                                  varfilter = var -> var in
```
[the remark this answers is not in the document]
YingboMa: It's easier to just write two lines in this case.

Issue #1506, opened 2022-04-01
Implement SemilinearODEProblem


RickDW:
Could be used as an alternative to DAEProblem/ODAEProblem/ODEProblem -> quoting Chris Rackauckas in the Julia slack

ChrisRackauckas:
@shashi added `Symbolics.semilinear_form` to take a function `f` and split it into `f = Au + f_n` i.e. split out the linear part efficiently. This could be used to generate a SplitODEProblem https://diffeq.sciml.ai/stable/types/split_ode_types/ which could be a much more efficient solver for some types of equations (such as partial differential equations). Also, it would reduce compile times in many models which are "mostly linear".

It would be nice to then also have a semiquadratic form `f = Au + u'Bu + f_n` as well. No ODE solvers that I know of make use of this information right now, so it would still just be a SplitODEProblem, but it could still have some pretty massive compile time benefits. This is because the only thing that would really have to compile would be the big `f_n`, but in many cases, like mass-action kinetics, have the majority of terms captured by the linear and quadratic terms. This would leave only the "truly nonlinear stuff" to `f_n`, making the function much smaller and much more managable. Future ODE solver research could improve the usage of this as well.

PR #1593, opened 2022-05-22
SplitODEProblem implementation

Abhishek-1Bhatt:
Implemented SplitODEProblem for #1506

### DIFF
--- a/src/systems/diffeqs/abstractodesystem.jl
+++ b/src/systems/diffeqs/abstractodesystem.jl
@@ -506,6 +506,65 @@ end
 
 """
 ```julia
+function SciMLBase.SplitFunction{iip}(sys::AbstractODESystem, dvs = states(sys),
+                                ps = parameters(sys);
+                                version = nothing, tgrad = false,
+                                jac = false,
+                                sparse = false,
+                                kwargs...) where {iip}
+```
+
+Create a `SplitFunction` from the [`ODESystem`](@ref). The arguments `dvs` and `ps`
+are used to set the order of the dependent variable and parameter vectors,
+respectively.
+"""
+function SciMLBase.SplitFunction{iip}(sys::AbstractODESystem, dvs = states(sys),
+                                      ps = parameters(sys), u0 = nothing;
+                                      version = nothing, tgrad=false,
+                                      jac = false,
+                                      eval_expression = true,
+                                      sparse = false, simplify = false,
+                                      eval_module = @__MODULE__,
+                                      steady_state = false,
+                                      checkbounds = false,
+                                      sparsity = false,
+                                      kwargs...) where {iip}
+    u = map(x->time_varying_as_func(value(x), sys), dvs)
+    p = map(x->time_varying_as_func(value(x), sys), ps)
+    t = get_iv(sys)
+    fs = [eq.rhs for eq in equations(sys)]
+    A, f_n = semilinear_form(fs, dvs)
+    A = DiffEqArrayOperator(A)
+    f_gen = build_function(f_n, A(u,p,t), u, p, t)#if failure try removing A
+    f_oop, f_iip = eval_expression ? (@RuntimeGeneratedFunction(eval_module, ex) for ex in f_gen) : f_gen
+    f(u,p,t) = f_oop(u,p,t)
+    f(du,u,p,t) = f_iip(du,u,p,t)
+
+    if jac
+        jac_gen = generate_jacobian(sys, dvs, ps;
+                                    simplify=simplify, sparse = sparse,
+                                    expression=Val{eval_expression}, expression_module=eval_module,
+                                    checkbounds=checkbounds, kwargs...)
+        jac_oop, jac_iip = eval_expression ? (@RuntimeGeneratedFunction(eval_module, ex) for ex in jac_gen) : jac_gen
+        _jac(u,p,t) = jac_oop(u,p,t)
+        _jac(J,u,p,t) = jac_iip(J,u,p,t)
+
+    else
+        _jac = nothing
+    end
+
+    SplitFunction{iip}(
+                        A,
+                        f,
+                        jac = _jac === nothing ? nothing : _jac,
+                        syms = Symbol.(states(sys)),
+                        indepsym = Symbol(get_iv(sys)),
+                        sparsity = sparsity ? jacobian_sparsity(sys) : nothing
+                      )
+end
+
+"""
+```julia
 function ODEFunctionExpr{iip}(sys::AbstractODESystem, dvs = states(sys),
                                      ps = parameters(sys);
                                      version = nothing, tgrad=false,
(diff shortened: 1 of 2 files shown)


Abhishek-1Bhatt:
```julia
cd(@__DIR__)
using Pkg
Pkg.activate("ModelingToolkit")

@parameters μ
@variables t x1(t) x2(t)
D = Differential(t)

eqs = [D(x1) ~ x2,
       D(x2) ~ μ*(1-x1^2)*x2 - x1]

@named sys = ODESystem(eqs)

u0 = [4.0,
      6.0]

p = [μ => 0.2]

tspan = (0.0, 80.0)
prob = SplitODEProblem(sys, u0, tspan, p, jac=true)
sol = solve(prob, Rodas5())
```
This test problem(Van Der Pol oscillator) works when SplitODEProblem{false} is built but when we build it in place, the solver errors out with

```
ERROR: MethodError: no method matching Float64(::ForwardDiff.Dual{ForwardDiff.Tag{OrdinaryDiffEq.OrdinaryDiffEqTag, Float64}, Float64, 2})
```

YingboMa:
Could you add tests in `test/odesystem.jl`? Also, please test the Jacobian generation as well.

ChrisRackauckas:
Can you test this on BCR 

https://benchmarks.sciml.ai/html/Bio/BCR.html

I'm curious if it has a significant compile time impact.

Abhishek-1Bhatt:
Sure

Abhishek-1Bhatt:
```julia
using ModelingToolkit
using Catalyst, OrdinaryDiffEq, ReactionNetworkImporters, Sundials, Plots, DiffEqDevTools, ODEInterface
using ODEInterfaceDiffEq, LSODA, TimerOutputs, LinearAlgebra, BenchmarkTools

datadir = joinpath(dirname(pathof(ReactionNetworkImporters)), "../data/bcr")
const to = TimerOutput()
tf       = 100000.0

#generate ModelingToolkit ODESystem
prnbng = loadrxnetwork(BNGNetwork(), joinpath(datadir, "bcr.net"))
rn = prnbng.rn
osys = convert(ODESystem, rn)

tspan = (0., tf)

@timeit to "ODEProb No Jac" oprob = ODEProblem(osys, Float64[], tspan, Float64[])
@timeit to "SplitODEProb No Jac" soprob = SplitODEProblem(osys, Float64[], tspan, Float64[])

@timeit to "ODEProb DenseJac" densejacprob = ODEProblem(osys, Float64[], tspan, Float64[], jac=true)
@timeit to "SplitODEProb Dense Jac" sdensejacprob = SplitODEProblem(osys, Float64[], tspan, Float64[], jac=true)

@timeit to "ODEProb SparseJac" sparsejacprob = ODEProblem(osys, Float64[], tspan, Float64[], jac=true, sparse=true)
@timeit to "SplitODEProb SparseJac" ssparsejacprob = SplitODEProblem(osys, Float64[], tspan, Float64[], jac=true, sparse=true)

show(to)
```
output:
```
────────────────────────────────────────────────────────────────────────────────
                                        Time                    Allocations
                               ───────────────────────   ────────────────────────
       Tot / % measured:             331s /  35.4%           16.1GiB /  59.5%

 Section               ncalls     time    %tot     avg     alloc    %tot      avg
 ────────────────────────────────────────────────────────────────────────────────
 ODEProb No Jac             1    65.9s   56.2%   65.9s   5.04GiB   52.6%  5.04GiB
 SplitODEProb No Jac        1    51.4s   43.8%   51.4s   4.53GiB   47.4%  4.53GiB
 ────────────────────────────────────────────────────────────────────────────────
```

Abhishek-1Bhatt:
For the one's with jacobian, it just doesn't fit in my system's ram(8 gigs). Due to using swap, it just doesn't finish executing.

ChrisRackauckas:
If `sparse=true` it should first find the sparsity pattern and then remove the linear operator via that into a sparse matrix. That should make the BCR case a lot faster.

Abhishek-1Bhatt:
```
───────────────────────────────────────────────────────────────────────────────────
                                           Time                    Allocations      
                                  ───────────────────────   ────────────────────────
         Tot / % measured:              386s /  70.0%           30.0GiB /  86.8%

 Section                  ncalls     time    %tot     avg     alloc    %tot      avg
 ───────────────────────────────────────────────────────────────────────────────────
 SplitODEProb SparseJac        1     231s   85.5%    231s   22.1GiB   85.0%  22.1GiB
 ODEProb SparseJac             1    39.1s   14.5%   39.1s   3.92GiB   15.0%  3.92GiB
 ───────────────────────────────────────────────────────────────────────────────────
```

src/systems/diffeqs/abstractodesystem.jl
```diff
@@ -504,6 +546,122 @@ function DiffEqBase.DAEFunction{iip}(sys::AbstractODESystem, dvs=states(sys),
...
+                                      eval_expression = true,
+                                      sparse = false, simplify = false,
+                                      eval_module = @__MODULE__,
+                                      steady_state = false,
+                                      checkbounds = false,
+                                      sparsity = false,
+                                      kwargs...) where {iip}
+    eqs = [eq for eq in equations(sys) if !isdifferenceeq(eq)]
+    rhs = [eq.rhs for eq in eqs]
+    A, f_n = semilinear_form(rhs, dvs)
```
YingboMa, line 576: Is `A` here sparse? We should add a config that switches between dense and sparse representations of `A`.

src/systems/diffeqs/abstractodesystem.jl
```diff
@@ -126,6 +146,22 @@ function generate_function(
...
+    )
+    u = map(x->time_varying_as_func(value(x), sys), dvs)
+    p = map(x->time_varying_as_func(value(x), sys), ps)
+    t = get_iv(sys)
+
+    A = DiffEqArrayOperator(A)
+
+    pre, sol_states = get_substitutions_and_solved_states(sys, no_postprocess = has_difference)
+
+    build_function(A(u,p,t), u, p, t; postprocess_fbody=pre, states=sol_states, kwargs...), build_function(f_n, u, p, t; postprocess_fbody=pre, states=sol_states, kwargs...)
```
YingboMa, line 162: I am pretty sure this is just unrolling `A` again.

Abhishek-1Bhatt:
```
───────────────────────────────────────────────────────────────────────────────────
                                           Time                    Allocations
                                  ───────────────────────   ────────────────────────
         Tot / % measured:             2079s /  66.6%            210GiB /  96.9%

 Section                  ncalls     time    %tot     avg     alloc    %tot      avg
 ───────────────────────────────────────────────────────────────────────────────────
 ODEProb SparseJac             1    1324s   95.6%   1324s    198GiB   97.2%   198GiB
 SplitODEProb SparseJac        1    60.5s    4.4%   60.5s   5.61GiB    2.8%  5.61GiB
 ───────────────────────────────────────────────────────────────────────────────────
```
Got these benchmarks in my system for sparse jacobian case on BCR.

src/systems/diffeqs/abstractodesystem.jl
Abhishek-1Bhatt: I think sparse representation of A is all we need. For jacobian computation, the final jacobian is the sum of the jacobian of linear split and that of the non-linear split(f_n). We calculate the jacobian of f_n as usual and then add A to it. If jacobian of f_n is dense adding A to it gives a dense final jacobian, and if `sparse=true` f_n's jacobian is sparse and we add sparse A to it and get a sparse jacobian as requested by the user.

Abhishek-1Bhatt: Oh, and A is the jacobian of the linear part

src/systems/diffeqs/abstractodesystem.jl
Abhishek-1Bhatt: The sparse `A` is converted to `DiffEqArrayOperator` , calling it as `A(u,p,t)` is like doing A*u (as described [here](https://diffeq.sciml.ai/stable/features/diffeq_operator/#:~:text=When%20defined%2C%20A%20has%20function%20calls%20A(u%2Cp%2Ct)%20and%20A(du%2Cu%2Cp%2Ct)%20that%20act%20like%20A*u)) which yields the symbolic expression for the linear part which is what build function needs exactly

Abhishek-1Bhatt:
`ODEProblem` might perform better than above though, in a system with ample ram, I guess.

src/systems/diffeqs/abstractodesystem.jl
ChrisRackauckas: There is no reason to `build_function` here. The semilinear part should be a `DiffEqArrayOperator` and not a function.

Abhishek-1Bhatt:
```
ODEProblem with uType Vector{Float64} and tType Float64. In-place: true
timespan: (0.0, 80.0)
u0: 2-element Vector{Float64}:
 4.0
 6.0
```
On building the SplitODEProblem it still returns this message. Where can we change it to say SplitODEProblem?

test/odesystem.jl
```diff
@@ -450,6 +450,43 @@ sol2 = solve(prob2, Tsit5(); callback = difference_cb,
...
+    densejac_oprob = ODEProblem(sys, u0, tspan, p, jac = true)
+    densejac_soprob = SplitODEProblem(sys, u0, tspan, p, jac = true)
+    sparsejac_oprob = ODEProblem(sys, u0, tspan, p, jac = true, sparse = true)
+    sparsejac_soprob = SplitODEProblem(sys, u0, tspan, p, jac = true, sparse = true)
+
+    #jacobian generation tests
+    u = states(sys)
+    t = 0.1
+    densejac_ode = densejac_oprob.f.jac(u, 0.2, t)
+    densejac_splitode = densejac_soprob.f.jac(u, 0.2, t)
```
Abhishek-1Bhatt, line 482: Is there something I've missed implementing for symbolic expression generation? It just returns `nothing` when calling `prob.f.jac` and thus calling it with arguments fails the test. Although it does pass the generated jacobian to the `SplitFunction` constructor  [here](https://github.com/Abhishek-1Bhatt/ModelingToolkit.jl/blob/e153aa270fa58b3adf394475ea3609d9a6f2d415/src/systems/diffeqs/abstractodesystem.jl#L695-L704)
I tested it by  causing a MethodError when the SplitODEProblem is built by passing an additional argument `nothing` at the end to the constructor above. The generated jacobian shows up correctly in the error message as being passed, but the SpltFunction which SplitODEProblem [receives](https://github.com/Abhishek-1Bhatt/ModelingToolkit.jl/blob/e153aa270fa58b3adf394475ea3609d9a6f2d415/src/systems/diffeqs/abstractodesystem.jl#L1088)  has the jac attribute set to `nothing`

xtalax:
Please ping me when this is merged :)

PR #1602, opened 2022-05-25
Add `hessian_sparsity` method for `NonlinearSystem`

Vaibhavdixit02:


ChrisRackauckas:
Hess should probably get cached like the Jacobians?

YingboMa:
Run `format(".")` on the repo.

YingboMa (changes requested):
Tests?

Project.toml
```diff
@@ -58,7 +58,7 @@ DocStringExtensions = "0.7, 0.8"
 DomainSets = "0.5"
 Graphs = "1.5.2"
 IfElse = "0.1"
-JuliaFormatter = "0.12, 0.13, 0.14, 0.15, 0.16, 0.17, 0.18, 0.19, 0.20, 0.21, 0.22, 0.23, 1"
+JuliaFormatter = "1"
```
Vaibhavdixit02, line 61: Since `"scimlstyle"` only works for v1 I think we should drop the older compat?

src/systems/optimization/optimizationsystem.jl
```diff
@@ -184,15 +184,14 @@ function DiffEqBase.OptimizationProblem{iip}(sys::OptimizationSystem, u0map,
...
-                                                                                            nothing,
-                                                                                            nothing,
-                                                                                            nothing)
+    _f = DiffEqBase.OptimizationFunction{iip}(f,
+                                              SciMLBase.NoAD(),
+                                              _grad,
+                                              _hess,
+                                              nothing,
+                                              nothing,
+                                              nothing,
```
ChrisRackauckas, line 193: If sparse is true, then this needs to give the prototype matrices.

ChrisRackauckas:
It's a little janky to just have a Hessian function without it being in the NonlinearProblem and NonlinearFunction constructors, but since there isn't a single nonlinear solver method right now that wants said Hessian I guess it's fine and can always be added later.

ChrisRackauckas:
Plus it fixes master.

ChrisRackauckas:
Formatter failed.

PR #1606, opened 2022-05-27
Pass hessian prototype to `OptimizationFunction`

Vaibhavdixit02:
https://github.com/SciML/ModelingToolkit.jl/pull/1602#discussion_r882570815 this got missed in #1602

### DIFF
--- a/src/systems/optimization/optimizationsystem.jl
+++ b/src/systems/optimization/optimizationsystem.jl
@@ -184,10 +184,20 @@ function DiffEqBase.OptimizationProblem{iip}(sys::OptimizationSystem, u0map,
         _hess = nothing
     end
 
+    if sparse
+        _hess_prototype = hessian_sparsity(sys)
+        hess_prototype = SparseArrays.sparse(findnz(_hess_prototype)[1],
+                                             findnz(_hess_prototype)[2],
+                                             zeros(nnz(_hess_prototype)))
+    else
+        hess_prototype = nothing
+    end
+
     _f = DiffEqBase.OptimizationFunction{iip}(f,
                                               SciMLBase.NoAD();
                                               grad = _grad,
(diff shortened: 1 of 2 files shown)


src/systems/optimization/optimizationsystem.jl
```diff
@@ -184,10 +184,20 @@ function DiffEqBase.OptimizationProblem{iip}(sys::OptimizationSystem, u0map,
         _hess = nothing
     end
 
+    if sparse
+        _hess_prototype = hessian_sparsity(sys)
+        hess_prototype = SparseArrays.sparse(findnz(_hess_prototype)[1],
+                                             findnz(_hess_prototype)[2],
+                                             zeros(nnz(_hess_prototype)))
```
ChrisRackauckas, line 191:
```suggestion
        hess_prototype = hessian_sparsity(sys)
```

src/systems/optimization/optimizationsystem.jl
```diff
@@ -251,6 +261,14 @@ function OptimizationProblemExpr{iip}(sys::OptimizationSystem, u0,
         _hess = :nothing
     end
 
+    if sparse
+        _hess_prototype = hessian_sparsity(sys)
+        hess_prototype = sparse(findnz(_hess_prototype)[1], findnz(_hess_prototype)[2],
+                                zeros(findnz(_hess_prototype)[3]))
```
ChrisRackauckas, line 267:
```suggestion
        hess_prototype = hessian_sparsity(sys)
```

ChrisRackauckas:
You need to type promote at the solver anyways, so this should just be a SparseMatrixCSC{Bool} which then you do `convert.(Float64,hess_sparsity)` or whatever (that's a much simpler piece of code than the `findnz` etc.)

PR #1609, opened 2022-05-27
Fix LaTeX show

ArnoStrouwen:
There is something missing in the LaTeX printing of sys
https://mtk.sciml.ai/stable/mtkitize_tutorials/sparse_jacobians/

Try this to fix it?
https://juliadocs.github.io/Documenter.jl/stable/man/latex/#Printing-LaTeX-from-Julia

### DIFF
--- a/src/systems/abstractsystem.jl
+++ b/src/systems/abstractsystem.jl
@@ -1021,7 +1021,7 @@ end
     return latexify(equations(sys))
 end
 
-Base.show(io::IO, ::MIME"text/latex", x::AbstractSystem) = print(io, latexify(x))
+Base.show(io::IO, ::MIME"text/latex", x::AbstractSystem) = print(io, "\$\$ "*latexify(x)*" \$\$")
 
 struct InvalidSystemException <: Exception
     msg::String
--- a/src/systems/abstractsystem.jl
+++ b/src/systems/abstractsystem.jl
@@ -1021,7 +1021,9 @@ end
     return latexify(equations(sys))
 end
 
-Base.show(io::IO, ::MIME"text/latex", x::AbstractSystem) = print(io, "\$\$ "*latexify(x)*" \$\$")
+function Base.show(io::IO, ::MIME"text/latex", x::AbstractSystem)
+    print(io, "\$\$ " * latexify(x) * " \$\$")
+end
 
 struct InvalidSystemException <: Exception
     msg::String


ChrisRackauckas:
How does this look in Pluto and Jupyter?

src/systems/abstractsystem.jl
```diff
@@ -1021,7 +1021,7 @@ end
     return latexify(equations(sys))
 end
 
-Base.show(io::IO, ::MIME"text/latex", x::AbstractSystem) = print(io, latexify(x))
+Base.show(io::IO, ::MIME"text/latex", x::AbstractSystem) = print(io, "\$\$ "*latexify(x)*" \$\$")
```
YingboMa, line 1024:
```suggestion
function Base.show(io::IO, ::MIME"text/latex", x::AbstractSystem)
    print(io, "\$\$ " * latexify(x) * " \$\$")
end
```

ArnoStrouwen:
Neither version works on Jupyter, but in different ways. In the old version it seems to show markdown that does not properly recognize latex and in the new version it just prints out a string. From my limited knowledge of Pluto, that also does not work properly.

ChrisRackauckas:
Pluto at least used to work properly.

@korsbo @stevengj @fonsp is there a way to correctly define the show outputs from Latexify so everything shows correctly on Pluto, Documenter, and Jupyter? It seems like there may be three different conventions, in terms of whether `$$` is required and such. Maybe we're overloading the wrong `show` methods?

stevengj:
See JuliaLang/IJulia.jl#66 for why a math delimiter (`$$` or similar) is required for text/latex math in Julia.

stevengj:
> Neither version works on Jupyter

I'm confused by this statement.  This works fine for me in Jupyter:
```jl
struct Foo; n end
Base.show(io::IO, ::MIME"text/latex", x::Foo) =
    print(io, "\$\$ x ^ {$(x.n)} \$\$")
```
image

YingboMa:
So this is OK to merge?

ChrisRackauckas:
I guess so? Sounds like it's the correct thing to do for Documenter and IJulia, and @stevengj links reasons for it so Pluto, if it's not on the same convention, should follow suit. So yes, let's go for it.

ArnoStrouwen:
>This works fine for me in Jupyter:
> 
> ```julia
> struct Foo; n end
> Base.show(io::IO, ::MIME"text/latex", x::Foo) =
>     print(io, "\$\$ x ^ {$(x.n)} \$\$")
> ```
I am not sure this is a complicated enough example, in the ModelingToolkit.jl documentations, equations are multi-line with stuff like  `\begin{align}`, `\\` and `=&` in there.

Capture

At least the docs look good since merging this in, though some of them load super slow.

ChrisRackauckas:
https://mtk.sciml.ai/dev/tutorials/acausal_components/

Doesn't look like most of the docs look good. It looks like this just handles the System printing, and the Symbolics printing needs a fix as well.

isaacsas:
It would be great if there is someone who understands all the Latex/show interactions and can get this working robustly again. There had also been longstanding issues with inconsistent printing within equations, where variables are in math italic font on one side of an equation but non-italic on the other side (perhaps in derivatives?). I think we had similar issues with how rates were being displayed within reactions.

It is all messed up enough that for Catalyst I now force plain text printing in tutorials like https://catalyst.sciml.ai/dev/tutorials/symbolic_stoich/. The Latex printing just wasn’t giving a good impression...

fonsp:
The cleanest might be to do a passthrough:

```julia
Base.show(io::IO, m::MIME"text/latex", x::AbstractSystem) = show(io, m, latexify(x))
```

stevengj:
> I am not sure this is a complicated enough example, in the ModelingToolkit.jl documentations, equations are multi-line with stuff like `\begin{align}`, `\\` and `=&` in there.

That works fine too:

image
It even works if you do this in *addition* to `$$ ... $$` delimiters, although this seems odd to me:

image

(Typically, I would think that you should *either* do `$$ ... $$` *or* `\begin{align} ... \end{align}`, but not both simultaneously, similar to how you would do it in a LaTeX document.)

isaacsas:
One problem that I'm seeing in Catalyst, and so perhaps this is related to the ModelingToolkit issues, is that Documenter seems to expect extra "\\" characters for the generated Latex compared to Jupyter and Pluto. i.e. 
```
\begin{align}
A &= B \\
C &= D
\end
```
as generated Latex looks good in Pluto and Jupyter, but in Documenter the line break is lost unless you use four forward slashes. I guess we need a way to detect when the Latex system conversion code is being called from Documenter vs. Pluto or Jupyter and correct for this.

fonsp:
@isaacsas That sounds like a Documenter bug, consider opening an issue

isaacsas:
@fonsp thanks! I think I've actually gotten all the Catalyst issues straightened out now by manually configuring the math engine in Documenter like 
```julia
mathengine = MathJax3(Dict(
    :loader => Dict("load" => ["[tex]/require","[tex]/mathtools"]),
    :tex => Dict(
        "inlineMath" => [["\$","\$"], ["\\(","\\)"]],
        "packages" => ["base", "ams", "autoload", "mathtools", "require"],
    ),
))
```
Last I checked on the PR where I'm adding this everything now looked ok on Documenter, Pluto and Jupyter (without needing double line breaks anymore).

PR #1610, opened 2022-05-27
Small changes to "Composing Ordinary Differential Equations" documentation

ArnoStrouwen:


docs/src/tutorials/ode_modeling.md
```diff
@@ -18,19 +18,17 @@ using ModelingToolkit
...
-                                  D(x) ~ RHS ])
+@named fol = ODESystem([ D(x)  ~ (1 - x)/τ])
 
 using DifferentialEquations: solve
 using Plots: plot
 
-prob = ODEProblem(structural_simplify(fol_separate), [x => 0.0], (0.0,10.0), [τ => 3.0])
+prob = ODEProblem(fol, [x => 0.0], (0.0,10.0), [τ => 3.0])
 sol = solve(prob)
-plot(sol, vars=[x,RHS])
```
ChrisRackauckas, line 29: we do want to show somewhere this way of subsetting a plot. Maybe make a second plot with just one thing?

ArnoStrouwen: If you scroll down a bit that plot is still there:
https://mtk.sciml.ai/stable/tutorials/ode_modeling/#Algebraic-relations-and-structural-simplification
I don't think the subsetting is needed in the very first MTK example.

Issue #1589, opened 2022-05-21
Implicit DAE observed indexing is broken


YingboMa:
https://github.com/SciML/ModelingToolkitStandardLibrary.jl/pull/56

@sharanry could you take a look?

ValentinKaisermayer:
Copy-paste example:
```julia
using ModelingToolkitStandardLibrary.Mechanical.Rotational, ModelingToolkit, OrdinaryDiffEq

@parameters t
D = Differential(t)

@named fixed = Fixed()
@named inertia1 = Inertia(J=2) # this one is fixed
@named spring = Spring(c=1e4)
@named damper = Damper(d=10)
@named inertia2 = Inertia(J=2, phi_start=pi/2)

connections = [
    connect(fixed.flange, inertia1.flange_b)
    connect(inertia1.flange_b, spring.flange_a, damper.flange_a)
    connect(spring.flange_b, damper.flange_b, inertia2.flange_a)
]

@named model = ODESystem(connections, t, systems=[fixed, inertia1, inertia2, spring, damper])
sys = structural_simplify(model)

prob = DAEProblem(sys, D.(states(sys)) .=> 0.0, Pair[], (0, 10.0))
sol = solve(prob, DFBDF())
```

```julia
julia> sol[inertia1.w]
ERROR: Indexing symbol inertia1₊w(t) is unknown.
Stacktrace:
  [1] error(s::String)
    @ Base .\error.jl:33
  [2] DEFAULT_OBSERVED(sym::Num, u::Vector{Float64}, p::Vector{Float64}, t::Float64)
    @ SciMLBase .julia\packages\SciMLBase\Vg9hW\src\scimlfunctions.jl:4
```

YingboMa:
https://github.com/SciML/ModelingToolkit.jl/pull/1616 is a good start. We might need interface changes to DAE observed functions from the DAEFunction side, so that observed functions can depend on derivatives as well.

ChrisRackauckas:
@oscardssmith this must've been solved given you rely on it?

oscardssmith:
yeah.

Issue #1436, opened 2022-01-24
Support for `observed` variables in `DAEFunction`


ggkountouras:
```Julia
using ModelingToolkit, Sundials, DifferentialEquations, Plots

@variables t
@connector function Pin(;name)
    sts = @variables v(t)=1.0 i(t)=1.0 [connect = Flow]
    ODESystem(Equation[], t, sts, []; name=name)
end

function Ground(;name)
    @named g = Pin()
    eqs = [g.v ~ 0]
    compose(ODESystem(eqs, t, [], []; name=name), g)
end

function OnePort(;name)
    @named p = Pin()
    @named n = Pin()
    sts = @variables v(t)=1.0 i(t)=1.0
    eqs = [
           v ~ p.v - n.v
           0 ~ p.i + n.i
           i ~ p.i
          ]
    compose(ODESystem(eqs, t, sts, []; name=name), p, n)
end

function Resistor(;name, R = 1.0)
    @named oneport = OnePort()
    @unpack v, i = oneport
    ps = @parameters R=R
    eqs = [
           v ~ i * R
          ]
    extend(ODESystem(eqs, t, [], ps; name=name), oneport)
end

function SineVSource(;name, V = 1.0)
    @named oneport = OnePort()
    @unpack v = oneport
    ps = @parameters V=V
    eqs = [
           v ~ V * sin(t)
          ]
    extend(ODESystem(eqs, t, [], ps; name=name), oneport)
end

function IdealVCCS(;name, alpha = 1.0)
    @named p1 = Pin()
    @named p2 = Pin()
    @named n1 = Pin()
    @named n2 = Pin()
    sts = @variables v1(t) v2(t) i1(t) i2(t)
    ps = @parameters alpha=alpha
    eqs = [
        v1 ~ p1.v - n1.v
        v2 ~ p2.v - n2.v
        0 ~ p1.i + n1.i
        0 ~ p2.i + n2.i
        i1 ~ p1.i
        i2 ~ p2.i
        i1 ~ 0
        i2 ~ - alpha * v1
    ]
    compose(ODESystem(eqs, t, sts, ps; name=name), p1, p2, n1, n2)
end

@named V1 = SineVSource(V=1.0)
@named R1 = Resistor(R=1.0)
@named VCCS = IdealVCCS(alpha=1.0)
@named R2 = Resistor(R=1.0)
@named GND = Ground()
ckt = [
       connect(V1.p, R1.p, VCCS.p1)
[...]

PR #1616, opened 2022-05-29
adds observedfun for DAEFunction

ValentinKaisermayer:


ValentinKaisermayer:
closes #1589, #1436

ChrisRackauckas:
Run the formatter

ValentinKaisermayer:
The stuff in MTKSL is no longer broken

src/systems/diffeqs/abstractodesystem.jl
```diff
@@ -509,6 +509,16 @@ function DiffEqBase.DAEFunction{iip}(sys::AbstractODESystem, dvs = states(sys),
         _jac = nothing
     end
 
+    obs = observed(sys)
+    observedfun = let sys = sys, dict = Dict()
+        function generated_observed(obsvar, u, p, t)
```
YingboMa, line 514: This would error if the observed variable depends on derivative variables.

YingboMa:
Thanks for be PR! But it doesn't fully solve the problem though.

ValentinKaisermayer:
You are welcome. 
Maybe a (broken) test should be added for such a general case?

PR #1621, opened 2022-06-02
Separate out pages

ChrisRackauckas:


ChrisRackauckas:
@YingboMa can you quickly get this formatted? I want to try and see what MTK looks like in the full docs build https://docs.sciml.ai/dev/

PR #1635, opened 2022-06-09
Strengthen alias elimination

Keno:
Previously, we would only alias eliminate variables that do not occur
differentiated. This was unnecessarily restricted for systems such as
```
x ~ 0
D(x) ~ x - y
```
where alias elimination is now able to determine that `x` and `y` are
aliases.

While this is the main point of the PR, this does improve alias elimination
in some of the tests, causing tearing order to change. Previously we had
a suboptimality in tearing where an equation assigned to an scc in which it
is not able to solve a variable would not be reconsidered in another scc.
Fix that also, to make sure the tests keep working.

### DIFF
--- a/src/structural_transformation/bipartite_tearing/modia_tearing.jl
+++ b/src/structural_transformation/bipartite_tearing/modia_tearing.jl
@@ -37,9 +37,38 @@ end
 
 function tear_graph_modia(structure::SystemStructure; varfilter = v -> true,
                           eqfilter = eq -> true)
+    # It would be possible here to simply iterate over all variables and attempt to
+    # use tearEquations! to produce a matching that greedily selects the minimal
+    # number of torn variables. However, we can do this process faster if we first
+    # compute the strongly connected components. In the absence of cycles and
+    # non-solvability, a maximal matching on the original graph will give us an
+    # optimal assignment. However, even with cycles, we can use the maximal matching
+    # to give us a good starting point for a good matching and then proceed to
+    # reverse edges in each scc to improve the solution. Note that it is possible
+    # to have optimal solutions that cannot be found by this process. We will not
+    # find them here [TODO: It would be good to have an explicit example of this.]
+
     @unpack graph, solvable_graph = structure
     var_eq_matching = complete(maximal_matching(graph, eqfilter, varfilter))
     var_sccs::Vector{Union{Vector{Int}, Int}} = find_var_sccs(graph, var_eq_matching)
+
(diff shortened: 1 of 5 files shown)


src/systems/alias_elimination.jl
```diff
@@ -287,11 +299,7 @@ function alias_eliminate_graph!(graph, var_to_diff, mm_orig::SparseMatrixCLIL)
     function lss!(ei::Integer)
         vi = pivots[ei]
         may_eliminate = true
-        for v in 𝑠neighbors(graph, mm.nzrows[ei])
-            # the differentiated variable cannot be eliminated
-            may_eliminate &= isnothing(diff_to_var[v]) && isnothing(var_to_diff[v])
-        end
-        locally_structure_simplify!((@view mm[ei, :]), vi, ag, may_eliminate)
+        locally_structure_simplify!((@view mm[ei, :]), vi, ag, var_to_diff)
```
YingboMa, line 302: Could you make sure that variables that are marked irreducible to not be eliminated away? See https://github.com/SciML/ModelingToolkit.jl/blob/083949f0e9c231593a1a4287c7525606f3b27616/src/structural_transformation/utils.jl#L173

YingboMa (commented):
Could you test the example
```julia
x ~ 0
D(x) ~ x - y
```
get eliminated as expected? What about
```julia
D(z) ~ 0
```
if we eliminate `D(z)` away, then we won't have `z` as an observed variable as no initial condition would be provided for it.

YingboMa:
BTW, remember to run `using JuliaFormatter, ModelingToolkit; format(joinpath(dirname(pathof(ModelingToolkit)), ".."))` before committing.

YingboMa:
The CI failures https://github.com/SciML/ModelingToolkit.jl/runs/6804925627?check_suite_focus=true#step:6:1540 look legitimate.

src/systems/alias_elimination.jl
Keno: irreducible values are not included in the linear subsystem, so they should not be eliminated.

src/structural_transformation/bipartite_tearing/modia_tearing.jl
```diff
@@ -37,17 +37,57 @@ end
 
 function tear_graph_modia(structure::SystemStructure; varfilter = v -> true,
                           eqfilter = eq -> true)
+    # It would be possible here to simply iterate over all variables and attempt to
```
YingboMa, line 40: Should we split this to another PR?

PR #1642, opened 2022-06-13
structural_simplify for SDESystem

YingboMa:
```julia
@variables t u[1:3](t)

params = @parameters ß=0.2 µ=0.1

D = Differential(t)

eqs = [D(u[1]) ~ -ß * u[1] * u[2],
    D(u[2]) ~ ß * u[1] * u[2] - µ * u[2],
    0 ~ u[1] + u[2] + u[3] - 1]

neqs = [-1 / 2 * u[1] * u[2] * u[3]
        u[1] * u[2] * u[3]
        0]

@named stochsys = SDESystem(eqs, neqs, t, collect(u), params)
sys = structural_simplify(stochsys)
prob = SDEProblem(sys, [u[1]=>0.2, u[2]=>0.3], (0.0, 100.0))
sol = solve(prob, SRIW1())
```
sdae

### DIFF
--- a/src/structural_transformation/symbolics_tearing.jl
+++ b/src/structural_transformation/symbolics_tearing.jl
@@ -63,16 +63,30 @@ function tearing_sub(expr, dict, s)
     s ? simplify(expr) : expr
 end
 
-function full_equations(sys::AbstractSystem; simplify = false)
-    empty_substitutions(sys) && return equations(sys)
-    substitutions = get_substitutions(sys)
-    substitutions.subed_eqs === nothing || return substitutions.subed_eqs
-    @unpack subs = substitutions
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
Should handle the case of non-diagonal noise as well, which requires a row of the noise_eqs is zero.

bjarthur:
looking forward to this being merged!

one question i have re. SDESystem interface, is why does it require that you input the variables and parameters when ODESystem does not.  isn't that the whole point of the `@variables` and `@parameters` macros, to obviate that and make the interface simple and not duplicate code?

YingboMa:
Superseded by #2208

PR #1643, opened 2022-06-14
missing var feature

bradcarman:
Work in progress... will add test


bradcarman:
Just realized this is a bad idea...

PR #1649, opened 2022-06-15
Periodic Callbacks with general affect function (PoC)

dodoplus:
This is a proof-of-concept of running general Julia code in event callbacks (in this case, periodic callbacks which are now supported alongside continuous callbacks).

Using equations to define an `affect!` is great when it works, but it can be limiting.

Defining a new event
```julia

function affect!(u,p, t, c)
    update!(c, p.R, p.T)
    p.R = some_complex_logic(p.R, t)
end

ctrl = MyControl()
ps = @parameters R, T
ODESystem(eqs, t, [], ps; name=name, periodic_events=(period, affect!, ctrl))

# ctrl was modified
```

### DIFF
--- a/src/systems/diffeqs/abstractodesystem.jl
+++ b/src/systems/diffeqs/abstractodesystem.jl
@@ -5,14 +5,6 @@ function filter_kwargs(kwargs)
     end
     pairs(NamedTuple(kwargs))
 end
-function gen_quoted_kwargs(kwargs)
-    kwargparam = Expr(:parameters)
-    for kw in kwargs
-        push!(kwargparam.args, Expr(:kw, kw[1], kw[2]))
-    end
-    kwargparam
-end
-
 function calculate_tgrad(sys::AbstractODESystem;
                          simplify = false)
     isempty(get_tgrad(sys)[]) || return get_tgrad(sys)[]  # use cached tgrad, if possible
@@ -319,7 +311,7 @@ function compile_affect(sys::ODESystem, cb::Function, cb_state, dvs, ps, args...
     # find indexes
     ind(sym, v) = findfirst(isequal(sym), v)
     inds(syms, v) = filter(!isnothing,map(sym -> ind(sym, v), syms)) # filter out eliminated symbols
-    tr_name(v) = Symbol(unnamespace(prefix, v))
+    tr_name(v) = Symbol(unnamespace(prefix, getname(v)))
 
     v_inds = inds(dvs, all_dvs)
     p_inds = inds(ps, all_ps)
@@ -929,14 +921,13 @@ function ODEProblemExpr{iip}(sys::AbstractODESystem, u0map, tspan,
                                  kwargs...)
     linenumbers = get(kwargs, :linenumbers, true)
     kwargs = filter_kwargs(kwargs)
-    kwarg_params = gen_quoted_kwargs(kwargs)
-    odep = Expr(:call, :ODEProblem, kwarg_params, :f, :u0, :tspan, :p)
+
     ex = quote
         f = $f
         u0 = $u0
         tspan = $tspan
         p = $p
-        $odep
+        ODEProblem(f, u0, tspan, p; $(kwargs...))
     end
     !linenumbers ? striplines(ex) : ex
 end
@@ -975,9 +966,7 @@ function DAEProblemExpr{iip}(sys::AbstractODESystem, du0map, u0map, tspan,
     sts = states(sys)
     differential_vars = map(Base.Fix2(in, diffvars), sts)
     kwargs = filter_kwargs(kwargs)
-    kwarg_params = gen_quoted_kwargs(kwargs)
-    push!(kwarg_params.args, Expr(:kw, :differential_vars, :differential_vars))
-    prob = Expr(:call, :(DAEProblem{$iip}), kwarg_params, :f, :du0, :u0, :tspan, :p)
+
     ex = quote
         f = $f
         u0 = $u0
@@ -985,7 +974,8 @@ function DAEProblemExpr{iip}(sys::AbstractODESystem, du0map, u0map, tspan,
         tspan = $tspan
         p = $p
         differential_vars = $differential_vars
-        $prob
+        DAEProblem{$iip}(f, du0, u0, tspan, p; differential_vars = differential_vars,
+                         $(kwargs...))
     end
     !linenumbers ? striplines(ex) : ex
 end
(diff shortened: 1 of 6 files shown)


test/periodiccb.jl
```diff
@@ -0,0 +1,112 @@
...
+
+
+function affect!(u,p, t, state)
+    p.R *= 2
+    append!(state, p.R)
+end
+
+period = 1.0
+
+@variables t
```
YingboMa, line 16: These are defined in the `example` folder. You can just include the file.

YingboMa (changes requested):
I am not sure about this design. Cannot we detect periodic callbacks via discrete operators?

ChrisRackauckas:
That would only allow update equations, not arbitrary `affect!` functions which is what I think is wanted here. I think we should really just make an arbitrary `affect!` support for the symbolic condition discrete callbacks though.

dodoplus:
> That would only allow update equations, not arbitrary `affect!` functions which is what I think is wanted here. I think we should really just make an arbitrary `affect!` support for the symbolic condition discrete callbacks though.

Absolutely. The point is less about periodic (and discrete in general) events, and more about a more general-purpose `affect!`.

I've tried to define a reasonable (I hope) interface for such `affect!`

test/periodiccb.jl
dodoplus: Here's a more simplified example/test.

The `affect!` function performs an update based in the state of the `controller`, which can perform a complex logic (here simulated by pseudo random number generator) as well as keep (and use) a history. This can be useful, e.g. when simulating hysteresis.

```julia
using ModelingToolkit, OrdinaryDiffEq, Test
using Random

struct MyController
    rng
    h
    MyController() = new(MersenneTwister(1234), Float32[])
end

function newdirection!(c::MyController, pars, t)
    # random update
    pars.a, = Random.rand(c.rng, 1)

    # use history for update
    pars.b = t + sum(c.h)
end

function updatehist!(c::MyController, sts)
    push!(c.h, sts.x)
end

function affect!(u, p, t, ctrl::MyController)
    updatehist!(ctrl, u)
    newdirection!(ctrl, p, t)
end

period = 1.0

@variables t, x(t)
@parameters a, b
D = Differential(t)

eqs = [ D(x) ~ a * x + b]

controller = MyController()

@named model = ODESystem(eqs, t, [x], [a, b], periodic_events=(1.0, affect!, controller))

prob = ODAEProblem(sys, [x => 1.0], (0, 2.0), [a => 1.0, b => 3.0])
sol = solve(prob, Tsit5())

@test length(controller.h) == 1
@test controller.h[1] ≈ 7.873127

```

src/systems/diffeqs/abstractodesystem.jl
```diff
@@ -253,6 +245,90 @@ function generate_rootfinding_callback(cbs, sys::ODESystem, dvs = states(sys),
...
+
+function compile_affect(sys::ODESystem, cb::Function, cb_state, dvs, ps, args...; kwargs...)
+    all_dvs = states(sys)
+    all_ps = parameters(sys)
+
+    prefix = common_namespace(vcat(dvs, ps))
+
+    # find indexes
+    ind(sym, v) = findfirst(isequal(sym), v)
+    inds(syms, v) = filter(!isnothing,map(sym -> ind(sym, v), syms)) # filter out eliminated symbols
```
YingboMa, line 313: To have better scaling, you can use `Dict(reverse(en) for en in enumerate(all_dvs))`

src/systems/diffeqs/abstractodesystem.jl
```diff
@@ -253,6 +245,90 @@ function generate_rootfinding_callback(cbs, sys::ODESystem, dvs = states(sys),
...
+    PeriodicCallback.(affect_functions, Δts)
+end
+
+function compile_affect(sys::ODESystem, cb::PeriodicEventCallback, args...; kwargs...)
+    compile_affect(sys, affect_function(cb), state(cb), states(cb), parameters(cb), args...; kwargs...)
+end
+
+# helper for affect
+struct VarDict
+    ₊v::Base.RefValue{Vector}
```
YingboMa, line 274: Do you need `Ref`? You can always `resize!` and `copyto!`.

YingboMa: Also, should it be just `Vector{Int}`?

YingboMa (commented):
Each callback requires at least one dictionary look up. Can we just generate affect functions such that it only does integer indexing? The symbol => index mapping is fixed after the model is compiled.

src/systems/diffeqs/abstractodesystem.jl
dodoplus: > Do you need `Ref`? You can always `resize!` and `copyto!`.

Not sure. I assume that each `affect!` only needs to access a small part of the overall variable array, so it _may_ be cheaper to use references to those instead of copying the whole thing.

dodoplus:
> Each callback requires at least one dictionary look up. Can we just generate affect functions such that it only does integer indexing? The symbol => index mapping is fixed after the model is compiled.

I'm not sure what you mean by that: `setvec!` which is called for each call of the callback function is constant-time. It just assigns a `Ref`:

```julia
function setvec!(vd::VarDict, v)
    vd.₊v[] = v
end
```

Within the `affect!` function, accessing a specific variable/parameter would require a dictionary lookup & a de-`Ref`:
```julia
function Base.getproperty(vd::VarDict, name::Symbol)
  ...
  ₊v[][₊d[name]]
end
```

As you say, the mapping is fixed (and is only computed once), so we *could* compile the `affect!` function, replacing e.g. `pars.a` with `integ.p[2]` (although I'm not sure *how* exactly. Rewriting the AST?). However, I think it's not worth it in this case.
What do you think?

dodoplus:
Hmm... maybe this?
https://www.stochasticlifestyle.com/zero-cost-abstractions-in-julia-indexing-vectors-by-name-with-labelledarrays/

ChrisRackauckas:
`LArray`s won't scale because there type structure gets bigger, so if you have an ODE of more than 100 variables this is going to be slow to compile. What @YingboMa means is that it should be lowered to a callback function which is just using normal Julia indexing, so the symbolics exists until the ODEProblem is being constructed, and then all of the states/parameters orderings are fixed so you can do codegen that has no symbolic naming anymore.

dodoplus:
> `LArray`s won't scale because there type structure gets bigger, so if you have an ODE of more than 100 variables this is going to be slow to compile. 

Is this a problem in this case, though? 
`PeriodicEventCallback` has `states` and `ps` for the list of state variables and parameters the user want to be "exposed" in the callback. By default (if these parameters are not set), only the states/parameters for the sub-system callback is part of are exposed (currently including all sub-systems - which may not be a good behavior by default).

I *think* it would be quite rare for a single component to have even tens of parameters/states.

> What @YingboMa means is that it should be lowered to a callback function which is just using normal Julia indexing, so the symbolics exists until the ODEProblem is being constructed, and then all of the states/parameters orderings are fixed so you can do codegen that has no symbolic naming anymore

I'm not sure what you mean by that. The dictionary in the original proposal was only calculated once (as part of the `ODEProblem` setup). At runtime, the code *should* only include a single lookup (which I think is what you're suggesting) and a de-`Ref`. The `Ref` can be avoided by using `@views`.

dodoplus:
@ChrisRackauckas @YingboMa I've switched back to an improved `VarDict` (no more `Ref`).
However, there's a `Dict{Symbol, Int}` which adds some overhead compared to `LArray`.

dodoplus:
will be replaced with a new pull request

PR #1657, opened 2022-06-19
Add pretty-printing for BipartiteGraph

Keno:
Helps with debugging. Needs https://github.com/JuliaLang/julia/pull/45751
for proper formatting, but works without it.


YingboMa:
Thanks!

PR #1661, opened 2022-06-22
Cleanup callbacks 

isaacsas:
The callback routines were kind of defined in a bunch of different places, so I factored them out to their own file. I also updated `compile_affect` to avoid having it generate a function that simply creates a view and then wraps another function, instead reusing how we generate affects for JumpSystems (which have the same affect function signature and was how @shashi originally built them to work with integrators). That allowed me to reuse `compile_affect` for jumps too now.

This should be non-breaking with respect to continuous callbacks -- all I actually changed was `compile_affect` and calls to `compile_affect` (adding one in JumpSystems).

I'll add `DiscreteCallbacks` in a followup PR.

### DIFF
--- a/src/systems/callbacks.jl
+++ b/src/systems/callbacks.jl
@@ -98,12 +98,14 @@ affect. The generated function has the signature `affect!(integrator)`.
 Notes
 - `expression = Val{true}`, causes the generated function to be returned as an expression.
   If  set to `Val{false}` a `RuntimeGeneratedFunction` will be returned.
-- `outputidxs`, a vector of indices of the output variables.
+- `outputidxs`, a vector of indices of the output variables which should correspond to
+  `states(sys)`. If provided checks that the LHS of affect equations are variables are
+  dropped, i.e. it is assumed these indices are correct and affect equations are
+  well-formed.
 - `kwargs` are passed through to `Symbolics.build_function`.
 """
 function compile_affect(eqs::Vector{Equation}, sys, dvs, ps; outputidxs = nothing,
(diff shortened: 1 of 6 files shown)


isaacsas:
@YingboMa would you prefer if I continue with `DiscreteCallbacks` in a followup, or is it ok to add here?

src/systems/callbacks.jl
```diff
@@ -0,0 +1,201 @@
...
+                        expression = Val{true},
+                        kwargs...)
+    if isempty(eqs)
+        if expression == Val{true}
+            return :((args...) -> ())
+        else
+            return (args...) -> () # We don't do anything in the callback, we're just after the event
+        end
+    else
+        rhss = map(x -> x.rhs, eqs)
```
YingboMa, line 114: Maybe also check `x.lhs` is in a valid form?

isaacsas: I added a check that each lhs `isvariable`. This is only in the case that a user does not give `outputidxs`, for which I've now documented that it is assumed the `outputidxs` correspond to the lhs variables and such checks are dropped. Only `JumpSystem`s use `outputidxs` as they can potentially get a lot of these being created, and can determine the output indices at a higher level.

isaacsas: (And probably we should add better validation checking at the JumpSystem construction level at some point...)

isaacsas:
@YingboMa I think this should be good now unless you have further comments.

YingboMa (approved):
LGTM. Thanks a lot!

PR #1662, opened 2022-06-23
upper bound Symbolics to 4.7.0

shashi:
changing the syntax as seen in deprecation causes ODESystem construction to fail. This seems to be a bug in trying to automatically figuring out the independent variable.


ChrisRackauckas:
You would need to do it for the whole registry of them. Might be better to just yank it?

shashi:
Wouldn't doing this in MTK prevent the others from getting the update?

shashi:
Might be better to revert and come back once we have released a fix in MTK though...

ChrisRackauckas:
> Wouldn't doing this in MTK prevent the others from getting the update?

No, because you would just go back one MTK version and get the Symbolics update. You have to yank it.

ChrisRackauckas:
But did you do the yank to General?

PR #1667, opened 2022-06-26
remove ControlSystem

ChrisRackauckas:
It was only half completed, and now it's been folded into ODESystem itself so it has no place anymore. Bye bye.

The only thing we should try to keep is the `runge_kutta_discretize`, which instead should be ODESystem -> OptimizationSystem transformation


baggepinnen:
Is anyone using this? It's a rather breaking change

ChrisRackauckas:
No one could be using it because it was never really completed. So I'd classify it as non-breaking. I assume you'd be the only one to even notice it 😅, but since we've pointed everyone towards using the ctrls in `ODESystem`, it has effectively been deprecated for a long time with no real functionality (no way to simulate it, no way to do anything really except build an OptimizationSystem).

stephans3:
Is it not possible (and useful) to reshape `OptimizationSystem()`  to handle optimal control problems?  So, one can compare the [Sensitivity example](https://sensitivity.sciml.ai/stable/optimal_control/optimal_control/) with the MTK example.

PR #1673, opened 2022-06-29
Func affect

dodoplus:
This includes all changes from https://github.com/SciML/ModelingToolkit.jl/pull/1663 and replaces https://github.com/SciML/ModelingToolkit.jl/pull/1649

Proposed changes were discussed here: https://discourse.julialang.org/t/generalized-affects-in-mtk-whats-a-good-interface/83390

There are several new feature here.
First, periodic and preset-time callbacks are now supported (not sure about the right interface for iterative):

A periodic (discrete) event is defined as:
```julia
ODESystem(eqs, t, sts, ps; name=name, 
            discrete_events=[1.0 => [ v ~ -v])
```

i.e. instead of a `condition` a period is passed - this will be passed to a `PeriodicCallback`

A preset-time event is similar:
```julia
ODESystem(eqs, t, sts, ps; name=name, 
            discrete_events=[[2.71, 3.14] => [ v ~ -v])
```

That is, a vector of preset times to perform the affect is passed (which will be passed to `PresetTimeCallback`)

In addition, `affect` can now be a regular Julia function.

```julia
ODESystem(eqs, t, sts, ps; name=name, 
            discrete_events=[[2.71, 3.14] => (affect!, [v => :my_v, p], [V], ctx)])
```
Here, `affect!` is a user-defined function, `[v => :my_v, p]` is the variable (states) to pass to the affect, `[V]` are the parameters and `ctx` is a context.

Here, `v` is passed as `:my_v` to the affect while `p` is passed as `:p` (`==getname(p)`). The context is passed as-is to the affect.

The affect function has the following signature: `affect!(t, u, p, ctx)` where `t` is the IV (time), `u` is a vector (actually a `LArray`) containing the states requested, `p` are the parameters, and `ctx` is the context.

One can then write code that accesses and modifies the variables and states:
```julia
function affect!(t, u, p, ctx)
    u[:my_v] = u.p * p[:V] # u, p are LArrays, so either u.X or u[:X] will work
    u[:p] = something_complex(ctx, t, u[:p])
 end
```

Note that the reason for the `v => :my_v` syntax is so the user doesn't have to deal with e.g. `resistor₊i`. A silly example:

```julia
function SpecialRC_Circuit(; name, R=1.0, C=1.0)
   function affect!(t, u, v, rng)
       p.C = u.i * randn(rng, 1)[1]
   end
   @named oneport = OnePort()
   @named resistor = Resistor(R=R)
   @named capacitor = Capacitor(C=C)
   eqs = [connect(resistor.p, capacitor.n), connect(oneport.p, resistor.n), connect(oneport.n, resistor.p)]
   extend(ODESystem(eqs, t, [], [], 
      discrete_events = [1.0 => (affect!, [resistor.i => :i], [capacitor.C => :C], Random.GLOBAL_RNG)]; name=name), oneport)
end
```


dodoplus:
@isaacsas @YingboMa should I continue in this direction or you have better ideas?

isaacsas:
@dodoplus I don't have a good suggestion for how to handle truly general `affect!` functions. I guess I'd ideally like an approach that generates code that just uses the correct indices for each species and parameter, without any types of dynamic look up (hopefully I haven't misunderstood what you are proposing). i.e. they look like what someone might directly code for DifferentialEquations.jl.

That is because I mainly come at this from the perspective of jump processes, where the condition and affect functions can get called at every step of the algorithm in each case, so keeping them as fast as possible is desirable.

YingboMa:
I'd prefer to expose the integrator interface and use something like https://discourse.julialang.org/t/generalized-affects-in-mtk-whats-a-good-interface/83390/5

dodoplus:
> I'd prefer to expose the integrator interface and use something like https://discourse.julialang.org/t/generalized-affects-in-mtk-whats-a-good-interface/83390/5

Why would exposing the integrator be a better option?

dodoplus:
> I guess I'd ideally like an approach that generates code that just uses the correct indices for each species and parameter, without any types of dynamic look up (hopefully I haven't misunderstood what you are proposing). i.e. they look like what someone might directly code for DifferentialEquations.jl.
> 
> That is because I mainly come at this from the perspective of jump processes, where the condition and affect functions can get called at every step of the algorithm in each case, so keeping them as fast as possible is desirable.

I *think* my implementation does that.
The relevant code is in `compile_user_affect`:

```julia
function compile_user_affect(affect::FunctionalAffect, sys, dvs, ps; kwargs...)
    ind(sym, v) = findfirst(isequal(sym), v)
    inds(syms, v) = map(sym -> ind(sym, v), syms)
    v_inds = inds(states(affect), dvs)
    p_inds = inds(parameters(affect), ps)

    # HACK: filter out eliminated symbols. Not clear this is the right thing to do
    # (MTK should keep these symbols)
    v = filter(x -> !isnothing(x[1]), collect(zip(v_inds, states_syms(affect))))
    v_inds = [x[1] for x in v]
    v_syms = Tuple([x[2] for x in v])
    p = filter(x -> !isnothing(x[1]), collect(zip(p_inds, parameters_syms(affect))))
    p_inds = [x[1] for x in p]
    p_syms = Tuple([x[2] for x in p])

    let v_inds=v_inds, p_inds=p_inds, v_syms=v_syms, p_syms=p_syms, user_affect=func(affect), ctx = context(affect)
        function (integ)
            uv = @views integ.u[v_inds]
            pv = @views integ.p[p_inds]

            u = LArray{v_syms}(uv)
            p = LArray{p_syms}(pv)

            user_affect(integ.t, u, p, ctx)
        end
    end
end
```

The runtime overhead is `@views` followed by `LArray`, which I think are both resolved at compile-time.

Within the `user_affect` itself, access to a parameter should be zero-cost:
```julia
   u[:C] = 1
```
should translate (at compile time) to an index access into the `u` array:
```julia
   u[4] = 1 # assuming :C is at location 4 in u
```

which itself is just a view into the `integ.u` array:
```julia
   u[14] = 1 # assuming @views maps 14 to 4
```

DD

YingboMa:
> Why would exposing the integrator be a better option?

Because `integrator` interface is already well defined, and we shouldn't go out of our way to encapsulate. Users could write the affect exactly like normal callbacks including being able to get the temporary cache from the integrator.

dodoplus:
> > Why would exposing the integrator be a better option?
> 
> Because `integrator` interface is already well defined, and we shouldn't go out of our way to encapsulate. Users could write the affect exactly like normal callbacks including being able to get the temporary cache from the integrator.

OK. I'll implement your suggestion

dodoplus:
@YingboMa one small issue: there's no guarantee the same symbol is not used as both `state` and `parameter`. Correct?

This means we will have to check for clashes when passing the states and parameters for the affect:
```julia
ODESystem(eqs, t, sts, ps; name=name, 
            discrete_events=[[2.71, 3.14] => (affect!, [v => :V, p], [V], ctx)])

# should `error`: :V appears as a state and as a parameter
```

dodoplus:
> > one small issue: there's no guarantee the same symbol is not used as both state and parameter. Correct?
> 
> The affect function can be
> 
> ```julia
> function affect2!(integ, ctx, (u,), (up,))
> ...
> end
> ```
> 
> which won't rely on names.

I think this is not much of a problem, after all: as both parameters and variables are Julia symbols, it requires some effort to get them to clash. Regardless, my implementation allows for renaming of parameters:

```julia
ODESystem(... discrete_events=[[t1, t2]=>(affect!, [resistor.v => :rv, capacitor.v => :cv], [ball.v => :bv], ctx)])
```
The original idea was to avoid having to use `resistor₊v` in the affect function. That's the behavior without renaming, though:

```julia
ODESystem(... discrete_events=[1.0=>(affect!, [resistor.v, capacitor.v], [ball.v], ctx)])

function affect!(integ, ctx; resistor₊v, capacitor₊v, ball₊v)
   integ.u[resistor₊v] = integ.p[ball₊v]*integ.u[capacitor₊v]
end
```
An error is raised if a symbol repeats:
```julia
ODESystem(... discrete_events=[1.0=>(affect!, [u], [a=> :u], ctx))
# error: "All symbols for variables & parameters must be unique."
```

Let me know if this is acceptable.
If it is - how do you want me to proceed? Note that besides the `affect` stuff, this PI also attempts to support periodic and preset-times events. This part depends on https://github.com/SciML/ModelingToolkit.jl/pull/1663:

```julia
ODESystem(... discrete_events=1.0=>[v ~ -v])
# affect (v ~ -v) will trigger at 1.0, 2.0, 3.0, ...

ODESystem(... discrete_events=[1.0, 1.5]=>[v ~ -v])
# affect (v ~ -v) will trigger at 1.0 and 1.5
```

DD

YingboMa:
The issue there is that states and parameters are not separated. I wonder if we could do
```julia
function affect!(integ, ctx, indices)
    integ.u[indices.u.v] += ctx.mul * integ.p[indices.p.C]
end
```
where `indices` is a NamedTuple of NamedTuples.

dodoplus:
> The issue there is that states and parameters are not separated. I wonder if we could do
> 
> ```julia
> function affect!(integ, ctx, indices)
>     integ.u[indices.u.v] += ctx.mul * integ.p[indices.p.C]
> end
> ```
> 
> where `indices` is a NamedTuple of NamedTuples.

This seems complex, and one can still mix up the indices:
```julia
function affect!(integ, ctx, indices)
     integ.u[indices.p.C] += ctx.mul * integ.p[indices.u.v]
end
```

I think that `NamedTuple`s as you suggested earlier is preferable:
```julia
function affect!(integ, ctx, u, p)
     integ.u[u.v] += ctx.mul * integ.p[p.C]
end
```
although this also doesn't solve the mixing indices problem.

YingboMa:
Nothing is impervious. I think that’s good enough.

dodoplus:
@YingboMa made the changes as discussed.
It's now:
```julia
function affect!(integ, ctx, u, p)
     integ.u[u.v] += ctx.mul * integ.p[p.C]
end
```

Please let me know how do you want me to proceed.

YingboMa:
That looks quite good.

ChrisRackauckas:
Test failures look real.

dodoplus:
At least some of the failures seem to be related to https://github.com/SciML/ModelingToolkit.jl/pull/1663.
The functional affect stuff is not dependant on discrete events - so I'll revert all the discrete events code and commit it later when [isaacsas](https://github.com/isaacsas) is done.

isaacsas:
https://github.com/SciML/ModelingToolkit.jl/pull/1663 passes tests so I’m not sure what you are referring to.

dodoplus:
my bad - you're right.
I was referring to this:
```
Test Summary:       | Pass  Total
DiscreteSystem Test |   17     17
ODESystem Test: Test Failed at /home/runner/work/ModelingToolkit.jl/ModelingToolkit.jl/test/odesystem.jl:426
  Expression: prob.kwargs[:callback] isa ModelingToolkit.DiffEqCallbacks.DiscreteCallback
```

but this seems to pass on https://github.com/SciML/ModelingToolkit.jl/pull/1663

Will take a closer look

ChrisRackauckas:
Test failure looks real.

isaacsas:
Note also that this has pretty minimal testing of the pure symbolic discrete callbacks too, since that part is just taken from https://github.com/SciML/ModelingToolkit.jl/pull/1663 and I was waiting on some feedback before adding more there.

YingboMa (changes requested):
Could you also test this with RC circuit on the component level? Just to make sure that it works hierarchically as well.

dodoplus:
> Could you also test this with RC circuit on the component level? Just to make sure that it works hierarchically as well.

Added a couple of tests.

dodoplus:
> That's not hierarchical. You are adding it on the top level.

there are a couple of tests.
In the last test, the event is handled by `Capacitor`:

```julia
function affect6!(integ, u,p,ctx)
    @test integ.u[u.v] ≈ 0.3
    integ.p[p.C] *= 200
end

function Capacitor(; name, C = 1.0)
    @named oneport = OnePort()
    @unpack v, i = oneport
    ps = @parameters C = C
    D = Differential(t)
    eqs = [
        D(v) ~ i / C,
    ]
    extend(ODESystem(eqs, t, [], ps; name = name, continuous_events=[[v ~ 0.3]=>(affect6!, [v], [C], nothing)]), oneport)
end

@named capacitor = Capacitor(C = C)
@named rc_model = ODESystem(rc_eqs, t)
rc_model = compose(rc_model, [resistor, capacitor, source, ground])

sys = structural_simplify(rc_model)
u0 = [capacitor.v => 0.0
      capacitor.p.i => 0.0
      resistor.v => 0.0]

prob = ODEProblem(sys, u0, (0, 10.0))
sol = solve(prob, Rodas4())
@test all(sol[rc_model.capacitor.v] .< 0.4)
```

YingboMa:
Note that some files are still not formatted.

dodoplus:
@isaacsas I've added some documentation. This includes periodic-events and set-timed events, but I've left a "TBD" for the general discrete-events support.

dodoplus:
@YingboMa, I think this is ready for review.

src/systems/callbacks.jl
```diff
@@ -3,24 +3,94 @@ get_continuous_events(sys::AbstractSystem) = Equation[]
...
+end
+
+struct FunctionalAffect
+    f::Function
+    sts::Vector
+    sts_syms::Vector{Symbol}
+    pars::Vector
+    pars_syms::Vector{Symbol}
+    ctx::Any
+    function FunctionalAffect(f, sts, sts_syms, pars, pars_syms, ctx = nothing)
```
YingboMa, line 19: Don't use inner constructor unless necessary.

src/systems/callbacks.jl
```diff
@@ -3,24 +3,94 @@ get_continuous_events(sys::AbstractSystem) = Equation[]
...
 has_continuous_events(sys::AbstractSystem) = isdefined(sys, :continuous_events)
 
+has_discrete_events(sys::AbstractSystem) = isdefined(sys, :discrete_events)
+function get_discrete_events(sys::AbstractSystem)
+    has_discrete_events(sys) || return SymbolicDiscreteCallback[]
+    getfield(sys, :discrete_events)
+end
+
+struct FunctionalAffect
+    f::Function
```
YingboMa, line 13: Function shouldn't be typed.

dodoplus: if I leave it untyped, JuliaFormatter will just stick an `Any` there, which would be wrong.
Or am I missing something?

src/systems/callbacks.jl
```diff
@@ -208,3 +412,87 @@ function generate_rootfinding_callback(cbs, sys::AbstractODESystem, dvs = states
         VectorContinuousCallback(cond, affect, length(eqs))
     end
 end
+
+function compile_user_affect(affect::FunctionalAffect, sys, dvs, ps; kwargs...)
+    ind(sym, v) = findfirst(isequal(sym), v)
+    inds(syms, v) = map(sym -> ind(sym, v), syms)
```
YingboMa, line 415: O(m * n) again.

YingboMa (commented):
This PR looks pretty good. Just a few minor things to address.

src/systems/callbacks.jl
YingboMa: Yeah, use `::Any`

src/systems/callbacks.jl
isaacsas: Maybe it would be better to just calculate these mappings once at the system level where callbacks get called into, and then pass them down? Or perhaps in `process_events`?

YingboMa: Yeah, in compute the mapping in `process_events` would be ideal.

src/systems/callbacks.jl
```diff
@@ -316,11 +316,11 @@ function compile_affect(eqs::Vector{Equation}, sys, dvs, ps; outputidxs = nothin
             alleq = all(isequal(isparameter(first(update_vars))),
                         Iterators.map(isparameter, update_vars))
             if !isparameter(first(lhss)) && alleq
-                stateind(sym) = findfirst(isequal(sym), dvs)
-                update_inds = stateind.(update_vars)
+                stateind = Dict(map(a -> reverse(a), enumerate(dvs)))
```
YingboMa, line 319: This creates unnecessary allocations. You could write `Dict(reverse(en) for en in enumerate(dvs))`

YingboMa: Also https://docs.julialang.org/en/v1/manual/style-guide/#Do-not-write-x-f(x)

YingboMa:
Oops format failed again.

dodoplus:
> Oops format failed again.

apparently JuliaFormatter got an update :(

YingboMa:
Let's try the CI again.

dodoplus:
Seems test fails at `optimizationsystem.jl:58`:
```
   Expression: sol.minimum < 1.0
   Evaluated: NaN < 1.0
```

However, I get the same error on SciML:master.
As `Optimization` was bumped from 3.7.1 to 3.8.0 a couple of days ago, it may be the culprit.

ChrisRackauckas:
@Vaibhavdixit02 I assume that an issue with the choice of IIP vs OOP

PR #1688, opened 2022-07-13
add description metadata to variables

baggepinnen:
This PR adds the option to attach a descriptive string to variables and parameters. If a variable has a description attached, it will be printed after the variable when the system is shown in the terminal, example:

```julia
@parameters t
@variables u(t) [description = "A short description of u"]
@parameters p   [description = "A description of p"]
@named sys = ODESystem([u ~ p], t)

Model sys with 1 equations
States (1):
  u(t): A short description of u
Parameters (1):
  p: A description of p
```

This will hopefully make it easier to work with large models and component libraries.

@jvaverka We spoke about this possibility. This PR currently does nothing to any docstring. We could potentially write a version of the `Base.Docs.@doc` macro that would look for variable descriptions and add them to the docstring of a function or object automatically, I'll have to think a bit more about that one though.

---

Edit: The PR
- https://github.com/JuliaSymbolics/Symbolics.jl/pull/649

Adds metadata to the docstring of a variable

### DIFF
--- a/docs/src/basics/Variable_metadata.md
+++ b/docs/src/basics/Variable_metadata.md
@@ -1,6 +1,7 @@
 # Symbolic metadata
-It is possible to add metadata to symbolic variables. The following
-information can be added (note, it's possible to extend this to user-defined metadata as well)
+It is possible to add metadata to symbolic variables, the metadata will be displayed when calling help on a variable.
+
+The following information can be added (note, it's possible to extend this to user-defined metadata as well)
 
 ## Variable descriptions
 Descriptive strings can be attached to variables using the `[description = "descriptive string"]` syntax:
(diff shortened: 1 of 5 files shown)


jvaverka:
Love this feature. We also spoke about units and default values. How might we best extend this syntax to capture those (while still making them optional if that is the desire). Maybe something as seen below?

```julia

@variables u(t) [description = "A short description of u"]
@parameters begin
    p = 0.0 [description = "A description of p", unit = u"m"]
    q = 1.0 [description = "A description of q", unit = u"cm"]
end
```

baggepinnen:
Units are already supported that way 😊
https://mtk.sciml.ai/dev/basics/ContextualVariables/#Variable-metadata-[Experimental/TODO]

PR #1689, opened 2022-07-13
Clarify usage intent of `tunable` metadata

baggepinnen:
@SebastianM-C I hope this clarification is in line with your usage?


SebastianM-C:
Yes, thanks.

PR #1690, opened 2022-07-13
optimize calculation of input jacobian

baggepinnen:
This PR builds upon
- #1680

### DIFF
--- a/src/systems/abstractsystem.jl
+++ b/src/systems/abstractsystem.jl
@@ -1032,6 +1032,10 @@ function linearization_function(sys::AbstractSystem, inputs,
     lin_fun = let fun = fun,
         h = ModelingToolkit.build_explicit_observed_function(sys, outputs)
 
+        ps = parameters(sys)
+        fg_u = zeros(length(sts), length(input_idxs))
+        colorvec = eachindex(ps)
+
         function (u, p, t)
             if u !== nothing # Handle systems without states
                 length(sts) == length(u) ||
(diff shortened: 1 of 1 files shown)


src/systems/abstractsystem.jl
```diff
@@ -1040,8 +1044,19 @@ function linearization_function(sys::AbstractSystem, inputs,
                 fg_xz = ForwardDiff.jacobian(uf, u)
                 h_xz = ForwardDiff.jacobian(xz -> h(xz, p, t), u)
                 pf = SciMLBase.ParamJacobianWrapper(fun, t, u)
-                # TODO: this is very inefficient, p contains all parameters of the system
-                fg_u = ForwardDiff.jacobian(pf, p)[:, input_idxs]
+
+                p_closure = function (p_small_inner)
```
YingboMa, line 1048: You need a `let` to make sure that it's type stable.

baggepinnen:
I don't get why the format check is failing, I get
```julia
julia> using JuliaFormatter, ModelingToolkit; format(joinpath(dirname(pathof(ModelingToolkit)), ".."))
true
```
locally.

src/systems/abstractsystem.jl
```diff
@@ -1040,8 +1044,21 @@ function linearization_function(sys::AbstractSystem, inputs,
                 fg_xz = ForwardDiff.jacobian(uf, u)
                 h_xz = ForwardDiff.jacobian(xz -> h(xz, p, t), u)
                 pf = SciMLBase.ParamJacobianWrapper(fun, t, u)
-                # TODO: this is very inefficient, p contains all parameters of the system
-                fg_u = ForwardDiff.jacobian(pf, p)[:, input_idxs]
+
+                p_closure = let pf = pf
+                    function (p_small_inner::T) where {T}
+                        p_big[input_idxs] .= p_small_inner
+                        pf(p_big)::T
```
YingboMa, line 1051: Is this true? The parameters could be `Vector{Int}` while the observed variables are `Vector{Float64}`

src/systems/abstractsystem.jl
```diff
@@ -1040,8 +1044,21 @@ function linearization_function(sys::AbstractSystem, inputs,
                 fg_xz = ForwardDiff.jacobian(uf, u)
                 h_xz = ForwardDiff.jacobian(xz -> h(xz, p, t), u)
                 pf = SciMLBase.ParamJacobianWrapper(fun, t, u)
-                # TODO: this is very inefficient, p contains all parameters of the system
-                fg_u = ForwardDiff.jacobian(pf, p)[:, input_idxs]
+
+                p_closure = let pf = pf
```
YingboMa, line 1048: `p_big` and `input_idxs` are also enclosed.

YingboMa (commented):
I think we should add this functionality in SparseDiffTools.jl

YingboMa:
Superseded by #1724

PR #1691, opened 2022-07-13
Fix dimension mismatch and wrong equation in linearize

baggepinnen:
The referenced paper is likely wrong in this case, derivation below

IMG_20220713_172453

our tests currently don't catch this bug since it happens in the DAE branch, which is currently not supported

### DIFF
--- a/src/systems/abstractsystem.jl
+++ b/src/systems/abstractsystem.jl
@@ -1210,11 +1210,11 @@ function linearize(sys, lin_fun; t = 0.0, op = Dict(), allow_input_derivatives =
         C = [
         h_x h_z
 ]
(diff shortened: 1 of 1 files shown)


YingboMa:
It will be supported once https://github.com/SciML/ModelingToolkit.jl/pull/1685 merges.

YingboMa:
BTW, how do you handle du/dt?

baggepinnen:
> BTW, how do you handle du/dt?

https://github.com/SciML/ModelingToolkit.jl/blob/master/src/systems/abstractsystem.jl#L1111

YingboMa (approved):

PR #1699, opened 2022-07-20
Add jump dependency graph test

isaacsas:
Realized there isn't a good test on these (i.e. one where fadjlist and badjlist are actually distinct), so this adds a test for a network where they disagree to make sure we get the right input to JumpProcesses.


isaacsas:
Failure is unrelated.

Issue #1717, opened 2022-08-01
`SymbolicDiscreteCallback` fails for non-timebased events


wsphillips:
Setting an equation condition for a discrete callback, like `[x ~ 10.0]`, will fail. Tests in [`test/funcaffect.jl`](https://github.com/SciML/ModelingToolkit.jl/blob/master/test/funcaffect.jl) only validate that time-based events successfully run. 

```julia
using ModelingToolkit, OrdinaryDiffEq

@parameters t
@variables u(t)
D = Differential(t)

eqs = [D(u) ~ -u]

affect1!(integ, u, p, ctx) = integ.u[u.u] += 10

# set condition to be an equation instead of a timestop
@named sys = ODESystem(eqs, t, [u], [],
                       discrete_events = [[t ~ 4.0] => (affect1!, [u], [], nothing)])
prob = ODEProblem(sys, [u => 10.0], (0, 10.0)) # fails
sol = solve(prob, Tsit5())
```
Comparison conditions for discrete events are also not yet supported, but probably should be (e.g. `x > y`). Passing a bare `Num` condition `SymbolicDiscreteCallback( x > y, [ x ~ 0.0])` does the right thing, but I'm not sure that was intentional given the [comments here](https://github.com/SciML/ModelingToolkit.jl/blob/46778e31ec2d095d493c558761c141c63d709574/src/systems/callbacks.jl#L154) and because `Base.hash(::SymbolicDiscreteCallback, s)` [assumes conditions are packed into a vector](https://github.com/SciML/ModelingToolkit.jl/blob/46778e31ec2d095d493c558761c141c63d709574/src/systems/callbacks.jl#L203-L206).

isaacsas:
You're too quick on picking up new features -- we haven't even gotten that into the docs yet and I only completed the tests with a PR that was merged today! :) 

A symbolic condition should evaluate to true or false, so you need to say `t == 4.0` as the condition:
```julia
@named sys = ODESystem(eqs, t, [u], [], discrete_events = [(t == 4.0) => (affect1!, [u], [], nothing)])
prob = ODEProblem(sys, [u => 10.0], (0, 10.0)) 
sol = solve(prob, Tsit5(); tstops = [4.0])
```
Note also that with a generic condition like this you need to include `tstops` with the time to stop at in calling solve -- this is the usual DifferentialEquations generic `DiscreteCallback` condition to make sure the ODE solver steps to that point. 

Also note this example could be done completely symbolically:
[...]

PR #1719, opened 2022-08-02
hashing for discrete cb conditions

wsphillips:
bugfix for hashing the condition of `SymbolicDiscreteCallback`, which can be a `Num`. Fixes https://github.com/SciML/ModelingToolkit.jl/issues/1717#issuecomment-1201942870


isaacsas:
Shouldn’t we just unwrap the `Num` when storing it in the SymbolicDiscreteCallback? That is what my original PR did, and seems more consistent with how ModelingToolkit usually handles symbolics expressions internally? Maybe that was changed in the functional affects PR and no longer works?

YingboMa (approved):
LGTM

PR #1764, opened 2022-08-18
implement coerce_units

### LINKED ISSUE #1763: selectively disabling unit validation on specific equations is awkward
In some cases I would like to be able to work with equations where the LHS units are intentionally _not_ equivalent to the RHS units. For example, an equation that produces a `NoUnits` activation as a function of some physical state in the system. An equation like this might be interspersed among many other equations for which I _do_ want normal unit checking behavior.

At present, the solution is to write an entire RHS first as a normal Julia function, then `@register_symbolic`, overload `ModelingToolkit.get_unit` and then proceed to build an equation. That's a lot of boilerplate and it may not be desirable to disable normal tracing just to satisfy units. Orchestrating this for the user is doable but not very practical since it involves global method definitions.

A simpler solution could be to permit opt-in disabling of unit validation for select equations. I have a solution that seems to work on my end that I'll PR for this.

wsphillips:
fixes #1763

Introduces a non-exported utility called `coerce_units` that allows applying LHS units of an equation _directly_ onto the RHS term. This allows short circuiting `safe_get_unit` if unit metadata is detected prior to checking units through the whole operation.

@isaacsas assuming this passes tests, what do you think?

### DIFF
--- a/src/systems/validation.jl
+++ b/src/systems/validation.jl
@@ -152,9 +152,8 @@ end
 
 "Coerce the units of an equation to match, overriding the right-hand side."
 function coerce_units(eq::Equation)
-    (; lhs, rhs) = eq
-    rhs = setmetadata(rhs, VariableUnit, safe_get_unit(lhs, "left"))
-    return lhs ~ rhs
+    new_rhs = setmetadata(eq.rhs, VariableUnit, safe_get_unit(eq.lhs, "left"))
+    return eq.lhs ~ new_rhs
 end
 
 function _validate(terms::Vector, labels::Vector{String}; info::String = "")
--- a/src/systems/validation.jl
+++ b/src/systems/validation.jl
@@ -132,6 +132,7 @@ end
 
 "Get unit of term, returning nothing & showing warning instead of throwing errors."
 function safe_get_unit(term, info)
+    hasmetadata(term, VariableUnit) && return getmetadata(term, VariableUnit)
     side = nothing
     try
         side = get_unit(term)
@@ -147,6 +148,13 @@ function safe_get_unit(term, info)
         end
     end
     side
+end
+
+"Coerce the units of an equation to match, overriding the right-hand side."
+function coerce_units(eq::Equation)
+    (; lhs, rhs) = eq
+    rhs = setmetadata(rhs, VariableUnit, safe_get_unit(lhs, "left"))
+    return lhs ~ rhs
 end
 
 function _validate(terms::Vector, labels::Vector{String}; info::String = "")
--- a/src/systems/validation.jl
+++ b/src/systems/validation.jl
@@ -132,7 +132,7 @@ end
 
 "Get unit of term, returning nothing & showing warning instead of throwing errors."
 function safe_get_unit(term, info)
-    if term isa Union{Num,Complex{Num},Symbolic} && hasmetadata(term, VariableUnit)
+    if term isa Union{Num, Complex{Num}, Symbolic} && hasmetadata(term, VariableUnit)
         return getmetadata(term, VariableUnit)
     end
     side = nothing
--- a/src/systems/validation.jl
+++ b/src/systems/validation.jl
@@ -132,7 +132,9 @@ end
 
 "Get unit of term, returning nothing & showing warning instead of throwing errors."
 function safe_get_unit(term, info)
-    hasmetadata(term, VariableUnit) && return getmetadata(term, VariableUnit)
+    if term isa Union{Num,Complex{Num},Symbolic} && hasmetadata(term, VariableUnit)
+        return getmetadata(term, VariableUnit)
+    end
     side = nothing
     try
         side = get_unit(term)
--- a/test/units.jl
+++ b/test/units.jl
@@ -47,6 +47,12 @@ eqs = [D(E) ~ P - E / τ
 @test !MT.validate(D(D(E)) ~ P)
 @test !MT.validate(0 ~ P + E * τ)
 
+# Equation unit coercion
+@variables X
+eq1 = D(X) ~ 3P - 5
+eq2 = MT.coerce_units(eq1)
+@test !MT.validate(eq1) && MT.validate(eq2)
+
 # Array variables
 @variables t [unit = u"s"] x(t)[1:3] [unit = u"m"]
 @parameters v[1:3]=[1, 2, 3] [unit = u"m/s"]


src/systems/validation.jl
```diff
@@ -149,6 +150,13 @@ function safe_get_unit(term, info)
     side
 end
 
+"Coerce the units of an equation to match, overriding the right-hand side."
+function coerce_units(eq::Equation)
+    (; lhs, rhs) = eq
```
wsphillips, line 155: syntax is too new. I'll make it backwards compatible

isaacsas:
@wsphillips I'm not that familiar with the unit code -- I think @lamorton was the one who put it together. I have to read through to understand the impact of the early return you added to `safe_get_unit`. I'll try to take some time to read through that code later today.

lamorton:
This might work. There are automatic transformations that might get applied which would end up replacing the RHS `Term` with an equivalent but not identical one.  

I'd also be concerned about the potential for this mechanism to unintentionally bypass unit checking in a situation where it was actually desired/needed to avoid bugs.  I think it should be OK thought. I can't imagine a scenario where a composite term was somehow generated with unit metadata that was inconsistent with the sub-terms, other than what @wsphillips is proposing to do with `coerce_units`.  

I still don't understand the use-case though.

wsphillips:
> This might work. There are automatic transformations that might get applied which would end up replacing the RHS Term with an equivalent but not identical one. That would prevent coerce_units from having any effect.

I considered that, but (currently) unit validation only happens on initial system construction, so it winds up being a non-issue.

> I can't imagine a scenario where a composite term was somehow generated with unit metadata that was inconsistent with the sub-terms

As it stands now, I believe this is correct--`VariableUnit` is only defined in places where you would end up by working through the methods of `get_unit`. `coerce_units` would be the only exception. If there's ever an edge case that produces a bug, I would propose adding a unique metadata context to query instead of borrowing `VariableUnit`.

> I still don't understand the use-case though.

It can happen if you have state-dependent surrogates within your model where the LHS represents a quantity or weight of arbitrary units. The only way to solve this is either by selectively turning off unit validation for a particular equation or registering new methods for the RHS of each such equation. The later is not practical or desirable to do in all cases.

As a concrete example, the Hodgkin-Huxley neuron model:

image

The first equation for current is physical and has units that match. The state variables `m`,`h`, and `n` are unitless, representing relative activation of voltage- and time-dependent ion channels. The LHS is ms^-1, while the RHS is a function of Voltage that isn't naturally cancelled out.

wsphillips:
The only fail here is coveralls, which I'm not sure how to fix.

Is there anything else blocking this PR? I'm happy to make revisions or change the implementation/approach to get a fix in.

isaacsas:
I think it makes sense to just add a separate metadata type like `ForceUnitsVariable` or such instead, but I’d defer to @lamorton and @YingboMa for what should be done.

lamorton:
We should find a different way to solve the underlying issue that @wsphillips has. I can see 6 options at the moment:

1. Give `αₙ` etc consistent units by normalizing the arguments & denormalizing the return values: 
```
@params τ₀ V₀
eqs = [D(n) ~ αₙ(Vₘ/V₀)/τ₀]
```
2. Absorb those coefficients into the definitions of those functions & register the functions.
3. Represent the functions as equations directly in the system, including the parameters.
4. Create a subsystem to encapsulate those equations & their normalization parameters, then compose the full system.
5. Drop the units on all the variables of the system.
6. Construct the system with `checks = False`.

The issues with these approaches are:
1. Possibly repetitious and doesn't match how the equations are written in the textbooks.
2. Inconvenient for manual set-up, maybe impractical for automation.
3. Repetitious.
4. Not well-documented, but probably the best.
5. Doesn't take advantage of validation for the subset of equations that have units.
6. Not recommended - it bypasses more than just unit validation.

Slight tangent: I wish dealing with [Untiful Literals](https://mtk.sciml.ai/dev/basics/Validation/#Unitful-Literals) were more convenient, but fundamentally I don't see how it can be. In fact, there are potential correctness bugs lurking: registration assumes that a Julia function behaves like a mathematical function, but this isn't true if there are methods that specialize on units. That's why I had to disallow variables with units of degrees, but it could happen more generally.

lamorton:
To flesh out what I'm talking about with the subsystem approach:
```
function HHneuron(;name,τ₀, V₀, Cₘ, gₖ)
    pars = @params τ₀=τ₀ V₀=V₀ Cₘ gₖ
    sts = @variables Vₘ(t) n(t) I(t) Vₖ(t)
    eqs = [D(n) ~ αₙ(Vₘ/V₀)/τ₀ - βₙ(Vₘ/V₀)*n/τ₀,
    I ~ CₘD(Vₘ)+gₖ*n^4*(Vₘ - Vₖ)]
    ODESystem(eqs,t,sts,pars;name=name)
end
```
Then you can make copies of the system with different/same parameters:
```
@parameters τ₀ V₀ Cₘ gₖ
@named n1 = HHneuron(τ₀=τ₀,V₀=V₀, Cₘ=Cₘ, gₖ= gₖ)
@named n2 = HHneuron(τ₀=τ₀,V₀=V₀, Cₘ= 0.5*Cₘ, gₖ=2*gₖ)
eqs = [connect(n1.Vₘ, n2. Vₘ)]
@named sys = ODESystem(eqs,sts,pars,systems=[n1,n2])
```

wsphillips:
@lamorton In my case, the alpha and beta functions are user-defined symbolic expressions that are already expanded.

Starting from user input, I have tried automatically building Julia functions, registering and rewriting the RHS + adding methods for `get_unit` and `Symbolics.derivative` but thats obviously not runtime friendly. 

I thought of using dummy parameters to cancel the units every time, but that adds potentially very many extraneous parameters to the final system? Unlike literals, I don't think they would get reduced/eliminated?

>     5. Drop the units on all the variables of the system.
This is what I currently do. In `Conductor.jl` I lightly check and then drop all units before lowering to `ODESystem`. However, its a parallel implementation that isn't robust.

>     6. Construct the system with `checks = False`.
@lamorton @YingboMa What about adding some granularity to the `checks` flag? For example, perhaps an enum:
``` julia
@enum SystemChecks
    None = 0
    Components = 1 # alternatively a flag for each: variables, parameters, equations
    Units = 2
# ... more as needed with bit shift
end
```
where default `checks = Components | Units` and subsequently:

```julia
if (checks & Components) > 0
    check_variables(dvs, iv)
    check_parameters(ps, iv)
    check_equations(deqs, iv)
    check_equations(equations(cevents), iv)
end
if !all_dimensionless([dvs; ps; iv]) && (checks & Units) > 0
    check_units(deqs)
end
```
 In `Conductor.jl` I use my own `<:AbstractSystem` that gets lowered to `ODESystem`. Turning off unit checking would let me specialize `ModelingToolkit.check_units(sys::MyConductorSystem)` to be called before the vanilla MTK constructor...?

wsphillips:
I've implemented my last suggestion at #1789 which sidesteps the issue by allowing users to select which system checks get run instead of turning them all on or off.

wsphillips:
Closing since #1789 is merged.

Issue #1772, opened 2022-08-25
Automated sparse Jacobian webpage = hard to navigate


Datseris:
My puny laptop cannot actually navigate this page https://mtk.sciml.ai/stable/mtkitize_tutorials/sparse_jacobians/  due to the rendering of thousands (millions?) of latex entries.

Datseris:
Related with https://github.com/SciML/ModelingToolkit.jl/issues/1658

isaacsas:
That tutorial should just not display the output of using `modelingtoolkitize`. Even if it rendered correctly it wouldn’t be useful to read given the large number of equations

ChrisRackauckas:
Agreed. It needs to fix the Latex for other pages, but this page should put a `;`

PR #1774, opened 2022-08-25
Enable MathJax to fix the Latex prints

ChrisRackauckas:
Fixes https://github.com/SciML/ModelingToolkit.jl/issues/1772

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -2,6 +2,16 @@ using Documenter, ModelingToolkit
 
 include("pages.jl")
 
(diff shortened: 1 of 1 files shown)


Datseris:
Hm, doesn't seem to affect the page: https://mtk.sciml.ai/previews/PR1774/mtkitize_tutorials/sparse_jacobians/

xtalax:
@Datseris 
I just discovered, entirely by accident that this does actually work, it just takes >10mins to render a Latex expression that long

ChrisRackauckas:
both are true.

PR #1775, opened 2022-08-26
Don't show the monster sparse jacobian

ChrisRackauckas:
grr arrrrg

### DIFF
--- a/docs/src/mtkitize_tutorials/sparse_jacobians.md
+++ b/docs/src/mtkitize_tutorials/sparse_jacobians.md
@@ -50,7 +50,7 @@ prob = ODEProblem(brusselator_2d_loop,u0,(0.,11.5),p)
 Now let's use `modelingtoolkitize` to generate the symbolic version:
 
 ```@example sparsejac
-sys = modelingtoolkitize(prob)
+sys = modelingtoolkitize(prob);
 ```
 
 Now we regenerate the problem using `jac=true` for the analytical Jacobian


docs/src/mtkitize_tutorials/sparse_jacobians.md
```diff
@@ -50,7 +50,7 @@ prob = ODEProblem(brusselator_2d_loop,u0,(0.,11.5),p)
 Now let's use `modelingtoolkitize` to generate the symbolic version:
 
 ```@example sparsejac
-sys = modelingtoolkitize(prob)
+sys = modelingtoolkitize(prob);
 ```
```
baggepinnen, line 54: I think you need to end the cell with 
```julia
nothing # hide
```

ChrisRackauckas: If you do `;` it should return nothing. Looks like it worked: https://mtk.sciml.ai/dev/mtkitize_tutorials/sparse_jacobians/

baggepinnen: `;` does not change the return value. 
> A semicolon ; at the end of the last line has no effect.

https://juliadocs.github.io/Documenter.jl/stable/man/syntax/#@example-block

It worked because [it's fixed here](https://github.com/SciML/ModelingToolkit.jl/blob/master/docs/src/mtkitize_tutorials/sparse_jacobians.md)

PR #1776, opened 2022-08-26
Minor refactor: Be consistent about alias_eliminate_graph! return type

Keno:
Generally, it's bad style to change the return type in a corner cases,
because a lot of the time (as I happed to), users will mostly test
the common case and then when the corner case (in this case no linear
equations in the system) comes up, things break. Simplify things
by being consistent about the return types.


src/systems/alias_elimination.jl
```diff
@@ -4,7 +4,7 @@ const KEEP = typemin(Int)
 
 function alias_eliminate_graph!(state::TransformationState)
     mm = linear_subsys_adjmat(state)
-    size(mm, 1) == 0 && return nothing, mm, BitSet() # No linear subsystems
+    size(mm, 1) == 0 && return AliasGraph(ndsts(state.structure.graph)), mm, BitSet() # No linear subsystems
```
YingboMa, line 7: This creates unnecessary allocation though

Keno: I don't think it's worth worrying about.

PR #1783, opened 2022-08-28
set default parallel kwarg to nothing instead of SerialForm

### LINKED ISSUE #1782: don't override Symbolics.jl sharding heuristics
When parallel form is not specified, Symbolics.jl will automatically shard large systems:

https://github.com/JuliaSymbolics/Symbolics.jl/blob/06cdbf6dbee7af3a102d1941639bb820177be916/src/build_function.jl#L261-L263

...and otherwise falls back to just `SerialForm`:

https://github.com/JuliaSymbolics/Symbolics.jl/blob/06cdbf6dbee7af3a102d1941639bb820177be916/src/build_function.jl#L302-L309

Here in MTK the default is `parallel = SerialForm()`: 

https://github.com/SciML/ModelingToolkit.jl/blob/d417a2cf5158537905eb93d6a3aa8d6e44bd9888/src/systems/diffeqs/abstractodesystem.jl#L524-L530

...which causes automatic sharding to be skipped in `Symbolics.build_function`. This will break large systems which could otherwise run (even if slowly) in single-thread sharded form.

tl;dr we should probably set `parallel = nothing` in `process_DEProblem` to match `Symbolics.jl` assumptions and heuristics.

wsphillips:
should fix #1782

### DIFF
--- a/src/systems/diffeqs/abstractodesystem.jl
+++ b/src/systems/diffeqs/abstractodesystem.jl
@@ -527,7 +527,7 @@ function process_DEProblem(constructor, sys::AbstractODESystem, u0map, parammap;
                            jac = false,
                            checkbounds = false, sparse = false,
                            simplify = false,
-                           linenumbers = true, parallel = SerialForm(),
+                           linenumbers = true, parallel = nothing,
                            eval_expression = true,
                            use_union = false,
                            kwargs...)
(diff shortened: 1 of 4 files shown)


wsphillips:
The failed downstream test in DataDrivenDiffEq.jl seems to be unrelated to this PR (looks like recent CI runs in other PRs are also hitting it).

CC: @AlCap23  @YingboMa

AlCap23:
Hey, I've been on vacation and currently attending a workshop. Feel free to merge from my side. The next release of DataDriven will be based on (this PR)[https://github.com/SciML/DataDrivenDiffEq.jl/pull/371] .

YingboMa (approved):

PR #1784, opened 2022-08-28
Use `paramsyms` from new `DiffEqFunction`s

### LINKED ISSUE #1777: support symbolic indexing integrators with parameters
See https://github.com/SciML/SciMLBase.jl/issues/231

Upstream modifications would require updating all `<:AbstractDiffEqFunction` constructor calls to include an additional `paramsyms=Symbol.(ps)` keyword argument. Following the existing pattern for `syms` and `indepsym` seen in this example:

https://github.com/SciML/ModelingToolkit.jl/blob/d417a2cf5158537905eb93d6a3aa8d6e44bd9888/src/systems/diffeqs/abstractodesystem.jl#L341-L350

AayushSabharwal:
Closes #1777 using https://github.com/SciML/SciMLBase.jl/pull/253


AayushSabharwal:
Tests here will fail since the dependency on SciMLBase is updated to a version that hasn't released yet (the one in https://github.com/SciML/SciMLBase.jl/pull/232)

YingboMa (commented):
What about other systems?

AayushSabharwal:
The SciMLBase stuff this depends on has some bugs, which I'll get to fixing soon. I'll update the rest of the systems once that goes through.

ChrisRackauckas:
Looks like this is still omitting PDESystem and OptimizationSystem?

AayushSabharwal:
I haven't touched `OptimizationSystem` since the issue (https://github.com/SciML/SciMLBase.jl/issues/231) only suggested changes to `AbstractDiffEqFunction`. I may be missing something, but I don't see `PDESystem` explicitly calling an `AbstractDiffEqFunction` anywhere.

ChrisRackauckas:
But why would this only be in DiffEqFunctions? That makes no sense to me.

AayushSabharwal:
I'll make the changes everywhere then

AayushSabharwal:
I'm not sure what the invalidations error is, the stack trace doesn't even have MTK in it. I'm guessing codecov is either being flaky or those uncovered lines will be covered by downstream tests from SciMLBase.jl once this gets merged?

Issue #1785, opened 2022-08-28
Constraints in nested systems are ignored


ValentinKaisermayer:
https://github.com/SciML/ModelingToolkit.jl/blob/d417a2cf5158537905eb93d6a3aa8d6e44bd9888/src/systems/optimization/optimizationsystem.jl#L214

MWE:
```julia
using ModelingToolkit

@variables x y
o1 = (x - 1)^2
o2 = (y - 1/2)^2
c1 = [
    x ~ 1
]
c2 = [
    y ~ 1
]
sys1 = OptimizationSystem(o1, [x, y], [], name = :sys1, constraints=c1)
sys2 = OptimizationSystem(o2, [x, y], [], name = :sys2, constraints=c2)
sys = OptimizationSystem(0, [], []; name = :sys, systems=[sys1, sys2], checks=false)
prob = OptimizationProblem(sys, [0.0, 0.0])

julia> isnothing(prob.f.cons_expr)
true
```

I think there needs to be a getter function `constraints(sys)` similar to `equations(sys)`.

Maybe like this:
```julia
namespace_constraint(eq::Equation, sys) = Equation(
    namespace_expr(eq.lhs, sys), 
    namespace_expr(eq.rhs, sys),
)
namespace_constraints(sys::OptimizationSystem) = namespace_constraint.(get_constraints(sys), Ref(sys))
get_constraints(sys) = sys.constraints
function constraints(sys::OptimizationSystem)
    isempty(get_systems(sys)) ? value(get_constraints(sys)) :
    vcat(value(get_constraints(sys)), reduce(vcat, value.(namespace_constraints.(get_systems(sys)))))
end
```

```julia
julia> equations(sys)
(sys1₊x - 1)^2 + (sys2₊y - 0.5)^2

julia> constraints(sys)
2-element Vector{Any}:
 sys1₊x ~ 1
 sys2₊y ~ 1
```

PR #1787, opened 2022-08-28
Improves OptimizationSystem

ValentinKaisermayer:


ValentinKaisermayer:
closes #1785

src/systems/optimization/optimizationsystem.jl
```diff
@@ -26,25 +26,22 @@ struct OptimizationSystem <: AbstractTimeIndependentSystem
...
+    The default values to use when initial guess and/or
+    parameters are not supplied in `OptimizationProblem`.
     """
     defaults::Dict
     function OptimizationSystem(op, states, ps, var_to_name, observed,
                                 constraints, name, systems, defaults;
                                 checks::Bool = true)
         if checks
-            check_units(op)
+            op isa Num && check_units(op)
```
ValentinKaisermayer, line 44: I'm not sure about this. The intended use-case is when one system has no objective function i.e. just `0` but sub-systems do have one.

YingboMa: MTK should never work with Num. You should always unwrap it and then check if it's Symbolic.

YingboMa (approved):

src/systems/abstractsystem.jl
```diff
@@ -179,8 +179,6 @@ for prop in [:eqs
              :systems
              :structure
              :op
-             :equality_constraints
-             :inequality_constraints
```
YingboMa, line 183: `:constraints` should be here.

src/systems/optimization/optimizationsystem.jl
```diff
@@ -124,6 +126,25 @@ function equations(sys::OptimizationSystem)
...
+end
+# namespace_constraint(ineq::Inequality, sys) = Inequality(
+#     namespace_expr(ineq.lhs, sys), 
+#     namespace_expr(ineq.rhs, sys),
+#     ineq.relational_op,
+# )
+function namespace_constraints(sys::OptimizationSystem)
+    namespace_constraint.(get_constraints(sys), Ref(sys))
+end
+get_constraints(sys::OptimizationSystem) = value(sys.constraints)
```
YingboMa, line 141: `sys.x` means getting the variable named `x` from `sys`.

YingboMa (changes requested):
`sys.constraints` is deprecated.

src/systems/optimization/optimizationsystem.jl
```diff
@@ -119,10 +121,38 @@ function generate_function(sys::OptimizationSystem, vs = states(sys), ps = param
...
+    if isempty(systems) 
+        op
+    else
+        op + reduce(+, map(sys_ -> namespace_expr.(get_op(sys_), sys_), systems))
+    end
+end
+
+namespace_constraint(eq::Equation, sys) = namespace_equation(eq, sys)
+
+# namespace_constraint(ineq::Inequality, sys) = namespace_inequality(ineq, sys)
```
YingboMa, line 135: Why are they removed?

ValentinKaisermayer: They are not removed but rather I put them in and commented them out. Needs https://github.com/JuliaSymbolics/Symbolics.jl/pull/717 first

YingboMa:
The CI failure looks real.

YingboMa:
Format

ValentinKaisermayer:
I swear I did that:
```julia
(@v1.8) pkg> st
Status .julia\environments\v1.8\Project.toml`
  [70b36510] AutomaticDocstrings v1.0.5
  [98e50ef6] JuliaFormatter v1.0.9
  [295af30f] Revise v3.4.0

julia> format(".")
true
```

ChrisRackauckas:
using the SciML format?

ValentinKaisermayer:
Isn't that picked up automatically from the `.JuliaFormatter.toml`?

ValentinKaisermayer:
Just to make sure
```
julia> JuliaFormatter.format("."; style=JuliaFormatter.SciMLStyle)
true
```

YingboMa:
The format failure is due to https://github.com/SciML/ModelingToolkit.jl/commit/56ce7e5aae0e72786cf6666532173d5f455f8d3f . @baggepinnen please don't commit on master directly, and if you did it by mistake, please revert it if the CI fails.

baggepinnen:
Sorry about that :( I was trying to figure out how I managed to commit straight to master, and I think it's because edits in the web interface defaults to committing instead of opening a PR
image
is there some way to make opening the PR the default option, like a branch-protection rule? I hadn't even noticed that I already committed :/

PR #1789, opened 2022-08-28
allow bit flagging for system checks

wsphillips:
alternative fix for #1763

Instead of altering unit validation itself I added support for bit flags to selectively disable unit checking. It's written such that the old defaults continue to work, or one can use the newly defined constants to turn checks on or off.

This would let me solve #1763 since I can then specialize the application of `check_units`/`validate` etc ad hoc, and disable further unit checks (but not system checks) when lowering to an `ODESystem`

CC: @lamorton and @YingboMa


YingboMa (approved):
Very nice! LGTM

PR #1798, opened 2022-09-01
Adds binary and integer options to variable metadata

ValentinKaisermayer:


src/variables.jl
```diff
@@ -286,3 +286,37 @@ end
...
+Symbolics.option_to_metadata_type(::Val{:integer}) = VariableInteger
+
+isinteger(x::Num) = isinteger(Symbolics.unwrap(x))
+
+"""
+    isinteger(x)
+
+Determine if a variable is integer.
+"""
+function isinteger(x)
```
ValentinKaisermayer, line 318: `Base` also exports this.

YingboMa: Maybe use `isintegervar`?

YingboMa (approved):

Issue #1786, opened 2022-08-28
Loose ends from OptimizationSystem


ValentinKaisermayer:
- [x] Handling of nested constraints, see #1785 
- [x] Proper inequality constraint handling, see discussion in #1660, see #1799 and https://github.com/JuliaSymbolics/Symbolics.jl/pull/717
The definition of inequality constraints is decoupled from the definition of the actual bounds. Also one has to specify an equality constraint first even if it ends up as inequality. As much as possible should be specified during the modelling phase.
```julia
@variables x y
@parameters a b
loss = (a - x)^2 + b * (y - x^2)^2
cons = [
    x^2 + y^2 ~ 0, 
    y * sin(x) - x ~ 0
]
sys = OptimizationSystem(loss, [x, y], [a, b], name = :sys, constraints = cons)
prob = OptimizationProblem(sys, [x => 0.0, y => 0.0], [a => 1.0, b => 100.0],
                           lcons = [-1.0, -1.0], ucons = [500.0, 500.0])
```
- [x] Proper handling of variable bounds, see  #1799
Again this is decoupled from the system modelling and is specified during construction of the `OptimizationProblem`. However, MTK has already a way of specifiying box constraints on variables via e.g., `@variables x [bounds=(-1.0, 1.0)]`. 
- [x] Handling of discrete and logical constraints, e.g., `@variables x [binary=true]`. see #1798 and https://github.com/SciML/SciMLBase.jl/issues/270
- [x] Support for `MathOptInterface.AbstractOptimizer` that do not support `MathOptInterface.NLPBlock`. If a problem is linear or quadratic why use an NLP solver? However, I see this more of an issue for `OptimizationMOI.jl`. This is important for MPCs which are typically QPs. https://github.com/SciML/Optimization.jl/issues/369
- [x] Convenient accessing solutions. For `ODEProblem` one can do `sol[x]`.

ChrisRackauckas:
Thanks, this is a nice list.

ValentinKaisermayer:
Yeah!

PR #1799, opened 2022-09-01
Adds support for inequalities

ValentinKaisermayer:
Depends on https://github.com/JuliaSymbolics/Symbolics.jl/pull/717

https://github.com/SciML/ModelingToolkit.jl/issues/1786


ValentinKaisermayer:
Tests will fail

src/systems/optimization/optimizationsystem.jl
```diff
@@ -247,15 +256,21 @@ function DiffEqBase.OptimizationProblem{iip}(sys::OptimizationSystem, u0map,
         hess_prototype = nothing
     end
 
+    lb = first.(getbounds.(dvs)) # TODO: handle case where bounds are parameters
+    ub = last.(getbounds.(dvs))
+    BT = promote_type(eltype(lb), eltype(ub))
+
     if length(cstr) > 0
-        @named cons_sys = NonlinearSystem(cstr, dvs, ps)
+        @named cons_sys = ConstraintsSystem(cstr, dvs, ps)
```
ValentinKaisermayer, line 264: Would be nice to have a simplify step here that reduces the system of equalities and inequalities to a minimal form.

test/optimizationsystem.jl
```diff
@@ -52,41 +52,57 @@ sol = solve(prob, BFGS(initial_stepnorm = 0.0001), allow_f_increases = true)
...
-@test sol.minimum < 1.0
-prob.f.cons(out, sol.minimizer, [1.0, 1.0])
-@test out ≈ [1.0]
+@testset "inequality constraint" begin
+    @variables x y
+    @parameters a b
+    loss = (a - x)^2 + b * (y - x^2)^2
+    cons = [
+        -1 ≲ x^2 + y^2,
+        x^2 + y^2 ≲ 500,
```
ValentinKaisermayer, line 61: A nicer syntax can come later: `-1 ≲ x^2 + y^2 ≲ 500` 😄

YingboMa:
Try to use a different solver in the tests

ValentinKaisermayer:
Should be good to go now.

src/systems/optimization/optimizationsystem.jl
```diff
@@ -207,29 +209,50 @@ function DiffEqBase.OptimizationProblem{iip}(sys::OptimizationSystem, u0map,
                                              linenumbers = true, parallel = SerialForm(),
                                              use_union = false,
                                              kwargs...) where {iip}
+    if !(isnothing(lb) && isnothing(ub))
+        Base.depwarn("`lb` and `ub` are deprecated. Use the `bounds` metadata for the variables instead.",
```
Vaibhavdixit02, line 213: Let's not deprecate, just remove from docs. I don't want to create a mismatch for the constructor, it would just cause confusions that's not worth handling

Vaibhavdixit02: Both bounds and constraints

ValentinKaisermayer: But what should happen if a user supplies values for these?

src/systems/optimization/optimizationsystem.jl
```diff
@@ -207,29 +209,50 @@ function DiffEqBase.OptimizationProblem{iip}(sys::OptimizationSystem, u0map,
...
+    if haskey(kwargs, :lcons) || haskey(kwargs, :ucons)
+        Base.depwarn("`lcons` and `ucons` are deprecated. Specify constraints directly instead.",
+                     :OptimizationProblem, force = true)
+    end
+
     dvs = states(sys)
     ps = parameters(sys)
     cstr = constraints(sys)
 
+    lb = first.(getbounds.(dvs))
```
Vaibhavdixit02, line 226:
```suggestion
    lb = lb === nothing ? first(getbounds.(dvs)) : lb
```

src/systems/optimization/optimizationsystem.jl
```diff
@@ -207,29 +209,50 @@ function DiffEqBase.OptimizationProblem{iip}(sys::OptimizationSystem, u0map,
...
 
     u0 = varmap_to_vars(u0map, dvs; defaults = defs, tofloat = false)
     p = varmap_to_vars(parammap, ps; defaults = defs, tofloat = false, use_union)
-    lb = varmap_to_vars(lb, dvs; check = false, tofloat = false, use_union)
-    ub = varmap_to_vars(ub, dvs; check = false, tofloat = false, use_union)
+    lb = varmap_to_vars(dvs .=> lb, dvs; defaults = defs, tofloat = false, use_union)
+    ub = varmap_to_vars(dvs .=> ub, dvs; defaults = defs, tofloat = false, use_union)
+
+    if all(lb .== -Inf) && all(ub .== Inf)
+        lb = nothing
```
Vaibhavdixit02, line 242:
```suggestion
        lb = lb
```

src/systems/optimization/optimizationsystem.jl
```diff
@@ -207,29 +209,50 @@ function DiffEqBase.OptimizationProblem{iip}(sys::OptimizationSystem, u0map,
...
+        ub = nothing
```
Vaibhavdixit02, line 243:
```suggestion
        ub = ub
```

src/systems/optimization/optimizationsystem.jl
```diff
@@ -409,7 +457,10 @@ function OptimizationProblemExpr{iip}(sys::OptimizationSystem, u0map,
             hess = $_hess
             lb = $lb
             ub = $ub
-            cons = $cons
+            int = $int
+            cons = $cons[1]
+            lbcons = $lbcons
```
Vaibhavdixit02, line 462: This should be `lcons` i think?

src/systems/optimization/optimizationsystem.jl
ValentinKaisermayer: Actually no, I think. Otherwise some solvers that do not support box constraints will complain.

src/systems/optimization/optimizationsystem.jl
ValentinKaisermayer: Oh, yes. My bad.

src/systems/optimization/optimizationsystem.jl
Vaibhavdixit02: It would still be nothing right, since the default for the argument is `nothing`?

Vaibhavdixit02: A combination of https://github.com/SciML/ModelingToolkit.jl/pull/1799#discussion_r1010953949 should allow this to allow the user to pass in bounds through the kwarg as well as use the symbolic interface for it. It preserves backward compatibility and keeps the constructor consistent.

ValentinKaisermayer: It would be possible for the box constraints but I think the effort for the constraints is not worth it. It would be very hard to keep in sync. For example if a user specifies `ucons[i] != lcons[i]` for some `i` that corresponds to an equality constraint, we either have to forbid it or we would have to convert it to two inequalities. 

BTW what do you mean with consistent? With what?

ValentinKaisermayer: I think the comfort of writing the constraints directly, like in any good algebraic modelling language like JuMP or GAMS, is worth the depreciation and breaking change.

src/systems/optimization/optimizationsystem.jl
ValentinKaisermayer: Unfortunately this does not work

src/systems/optimization/optimizationsystem.jl
YingboMa: We should avoid unnecessary breaking changes.

YingboMa (commented):
The doc failure looks real. Could you merge master back to your PR?

src/systems/optimization/optimizationsystem.jl
```diff
@@ -354,8 +365,41 @@ function OptimizationProblemExpr{iip}(sys::OptimizationSystem, u0map,
...
+    defs = defaults(sys)
+    defs = mergedefaults(defs, parammap, ps)
+    defs = mergedefaults(defs, u0map, dvs)
+
+    u0 = varmap_to_vars(u0map, dvs; defaults = defs, tofloat = false)
+    p = varmap_to_vars(parammap, ps; defaults = defs, tofloat = false, use_union)
+    lb = varmap_to_vars(dvs .=> lb, dvs; defaults = defs, tofloat = false, use_union)
+    ub = varmap_to_vars(dvs .=> ub, dvs; defaults = defs, tofloat = false, use_union)
+
+    if !isnothing(lb) && all(lb .== -Inf) && !isnothing(ub) && all(ub .== Inf)
```
Vaibhavdixit02, line 398: Do it separately for `lb ` and `ub`

Vaibhavdixit02:
```suggestion
    if !isnothing(lb) && all(isinf, lb)
    
```

Vaibhavdixit02: I changed `!isnothing` to `isnothing`

src/systems/optimization/optimizationsystem.jl
```diff
@@ -354,8 +365,41 @@ function OptimizationProblemExpr{iip}(sys::OptimizationSystem, u0map,
                                       linenumbers = false, parallel = SerialForm(),
                                       use_union = false,
                                       kwargs...) where {iip}
+    if haskey(kwargs, :lcons) || haskey(kwargs, :ucons)
```
Vaibhavdixit02, line 368: I think we can still allow it @ValentinKaisermayer, the only reason I am being pedantic about these is that I want to avoid divergences between the symbolic interface here and the regular one. That creates a maintenance burden. It would just require the same kind of check as https://github.com/SciML/ModelingToolkit.jl/pull/1799/files#diff-b5294724a3c67d3b2d440b7dba97870040e1a699da417e7b3ac519f91ed61c22R398 after https://github.com/SciML/ModelingToolkit.jl/pull/1799/files#diff-b5294724a3c67d3b2d440b7dba97870040e1a699da417e7b3ac519f91ed61c22R438

ValentinKaisermayer: Somehow the links do not work as intendent - at least for me.

Vaibhavdixit02:
Looks like you need to run the formatter

docs/src/systems/OptimizationSystem.md
```diff
@@ -8,9 +8,10 @@ OptimizationSystem
 
 ## Composition and Accessor Functions
 
-- `get_eqs(sys)` or `equations(sys)`: The equation to be minimized.
+- `get_op(sys)` or `objective(sys)`: The objective to be minimized.
```
ChrisRackauckas, line 11: What the heck is this for? It ruins the uniformity and then bloats the namespace downstream. It serves no purpose other than to make things less consistent and break downstream codes. Let's revert this change. Just call the equations equations. In fact, there can be more than one, so it's not even the objective, so it's not even a correct description!

PR #1771, opened 2022-08-24
Strong alias elimination

YingboMa:

PR #1805, opened 2022-09-03
Update non-linear equations in `graph` after alias elimination

Keno:
We already update the linear equations, but of course aliasing
variables has an effect non the non-linear equations also, so
go through the graph and update it appropriately.


Keno:
Some of this was done in #1771, so I've updated this to only do the missing part, which is to update the solvable graph also.

YingboMa:
Handled in #1870

PR #1808, opened 2022-09-03
fixes input output handling for array variables

ValentinKaisermayer:


ValentinKaisermayer:
closes #1808

YingboMa (changes requested):
The CI failure is real.

ValentinKaisermayer:
Looks real but at first glance unrelated, #1806 has the same problem.

YingboMa (approved):

PR #1816, opened 2022-09-09
fix matrix sizes and input order in linearize

baggepinnen:
This PR
- Fixes the sizes of matrices when there are input derivatives present.
- Ensures that the input parameters are in the correct order as specified by the user, so that subsequent matrix operations make sense.
- The modelica paper used as reference contained multiple errors which this PR fixes, the code now links to a derivation in a previous PR. 

My attempt at testing this resulted in 
```
ERROR: Input p₊x(t) is differentiated!
```

I have a complicated model which is supposed to be proper (integrating), but currently still requires input differentiation, likely due to some non-optimality in a simplification pass. When explicitly creating a model that is supposed to require input differentiation (an inverse model), I get an error before `linearize`. I understand why this error appears, but I'm not sure how to think about the difference between this case and the case where [this matrix in linearize](https://github.com/SciML/ModelingToolkit.jl/blob/master/src/systems/abstractsystem.jl#L1215) is non-zero.
```julia
julia> lsys, ssys = linearize(cl, [p.x], [f.u])
ERROR: Input p₊x(t) is differentiated!
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] inputs_to_parameters!(state::TearingState{ODESystem}, check_bound::Bool)
   @ ModelingToolkit ~/.julia/dev/ModelingToolkit/src/inputoutput.jl:234
 [3] structural_simplify(sys::ODESystem, io::Tuple{Vector{Num}, Vector{Num}}; simplify::Bool, kwargs::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
   @ ModelingToolkit ~/.julia/dev/ModelingToolkit/src/systems/abstractsystem.jl:968
 [4] io_preprocessing(sys::ODESystem, inputs::Vector{Num}, outputs::Vector{Num}; simplify::Bool, kwargs::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
   @ ModelingToolkit ~/.julia/dev/ModelingToolkit/src/systems/abstractsystem.jl:984
 [5] linearization_function(sys::ODESystem, inputs::Vector{Num}, outputs::Vector{Num}; simplify::Bool, kwargs::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
   @ ModelingToolkit ~/.julia/dev/ModelingToolkit/src/systems/abstractsystem.jl:1024
 [6] linearization_function
   @ ~/.julia/dev/ModelingToolkit/src/systems/abstractsystem.jl:1021 [inlined]
 [7] linearize(sys::ODESystem, inputs::Vector{Num}, outputs::Vector{Num}; op::Dict{Any, Any}, t::Float64, allow_input_derivatives::Bool, kwargs::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
   @ ModelingToolkit ~/.julia/dev/ModelingToolkit/src/systems/abstractsystem.jl:1236
 [8] linearize(sys::ODESystem, inputs::Vector{Num}, outputs::Vector{Num})
   @ ModelingToolkit ~/.julia/dev/ModelingToolkit/src/systems/abstractsystem.jl:1233
 [9] top-level scope
   @ REPL[92]:1
```

### DIFF
--- a/src/inputoutput.jl
+++ b/src/inputoutput.jl
@@ -278,10 +278,12 @@ function inputs_to_parameters!(state::TransformationState, io)
 
     if io !== nothing
         # Change order of new parameters to correspond to user-provided order in argument `inputs`
-        param_permutation = map(io.inputs) do inp
-            findfirst(isequal(inp), new_parameters)
+        d = Dict{Any, Int}()
+        for (i, inp) in enumerate(new_parameters)
+            d[inp] = i
         end
-        new_parameters = new_parameters[param_permutation]
+        permutation = [d[i] for i in io.inputs]
+        new_parameters = new_parameters[permutation]
     end
 
     @set! sys.ps = [ps; new_parameters]
--- a/src/inputoutput.jl
+++ b/src/inputoutput.jl
@@ -218,7 +218,8 @@ function generate_control_function(sys::AbstractODESystem, inputs = unbound_inpu
     f, dvs, ps
 end
 
-function inputs_to_parameters!(state::TransformationState, check_bound = true)
+function inputs_to_parameters!(state::TransformationState, io)
+    check_bound = io === nothing
     @unpack structure, fullvars, sys = state
     @unpack var_to_diff, graph, solvable_graph = structure
     @assert solvable_graph === nothing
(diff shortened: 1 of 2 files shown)


src/inputoutput.jl
```diff
@@ -274,6 +275,15 @@ function inputs_to_parameters!(state::TransformationState, check_bound = true)
     @set! sys.eqs = map(Base.Fix2(substitute, input_to_parameters), equations(sys))
     @set! sys.states = setdiff(states(sys), keys(input_to_parameters))
     ps = parameters(sys)
+
+    if io !== nothing
+        # Change order of new parameters to correspond to user-provided order in argument `inputs`
+        param_permutation = map(io.inputs) do inp
+            findfirst(isequal(inp), new_parameters)
```
YingboMa, line 282: Dict?

baggepinnen: Fixed

YingboMa (approved):

PR #1819, opened 2022-09-14
Reorder metadata and pass kwarg

YingboMa:


xtalax:
Thanks for this, I'm sure I had it in a very similar state to what you pushed at one point, but got a lot of test failures. Don't know what went on there.

PR #1824, opened 2022-09-19
fix `getvar` access of state

baggepinnen:
The result of 
```julia
i = findfirst(x -> getname(x) == name, sts)
```
right above the added code was never used, causing some states not to be found. I'm no sure if the added code is correct, but something is at least missing here

### DIFF
--- a/src/systems/abstractsystem.jl
+++ b/src/systems/abstractsystem.jl
@@ -304,6 +304,9 @@ function getvar(sys::AbstractSystem, name::Symbol; namespace = false)
 
     sts = get_states(sys)
     i = findfirst(x -> getname(x) == name, sts)
+    if i !== nothing
+        return namespace ? renamespace(sys, sts[i]) : sts[i]
+    end
 
     if has_observed(sys)
         obs = get_observed(sys)


src/systems/abstractsystem.jl
```diff
@@ -304,6 +304,9 @@ function getvar(sys::AbstractSystem, name::Symbol; namespace = false)
 
     sts = get_states(sys)
```
YingboMa, line 305: These are already handled above. We should just remove this block.

YingboMa (changes requested):
We should remove the later state index check.

src/systems/abstractsystem.jl
baggepinnen: I had an example that failed without this change but not with it. It's from the analysis point PR.
It might be related to [how I added states to the system here](https://github.com/SciML/ModelingToolkitStandardLibrary.jl/pull/110/files#diff-15d92341d21323574736195c285cbb0a69060f96b926cce44b433d09c93232e0R226)?

YingboMa: Ah, yeah, we should remember to update the `var_to_name` map, when vars are updated.

YingboMa:
Let's merge this for now, but we should work on a more efficient solution later.

PR #1829, opened 2022-09-22
Add verbose REPL printing for SystemStructure

Keno:
This adds REPL printing for SystemStructure that prints out the details of graph, solvable_graph, var_to_diff and eq_to_diff in a compact, but comprehensive dump. Because this is a lot of information, it is not provided by default. Instead, the user may opt into it by setting `:limit` or `:mtk_limit` to false in the IOContext. E.g. to enable the verbose printing for all structures shown in the REPL, use

```
Base.active_repl.options.iocontext[:mtk_limit] = false
```
Screenshot from 2022-09-22 02-41-03

### DIFF
--- a/src/systems/systemstructure.jl
+++ b/src/systems/systemstructure.jl
@@ -403,11 +403,93 @@ function linear_subsys_adjmat(state::TransformationState)
                             linear_equations, eadj, cadj)
 end
 
+using .BipartiteGraphs: Label, BipartiteAdjacencyList
+struct SystemStructurePrintMatrix <:
+       AbstractMatrix{Union{Label, Int, BipartiteAdjacencyList}}
+    bpg::BipartiteGraph
+    highlight_graph::BipartiteGraph
+    var_to_diff::DiffGraph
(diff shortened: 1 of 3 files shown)


src/systems/systemstructure.jl
```diff
@@ -403,11 +403,91 @@ function linear_subsys_adjmat(state::TransformationState)
...
+        @assert false
+    end
+end
+
 function Base.show(io::IO, mime::MIME"text/plain", s::SystemStructure)
-    @unpack graph = s
-    S = incidence_matrix(graph, Num(Sym{Real}(:×)))
-    print(io, "Incidence matrix:")
-    show(io, mime, S)
+    @unpack graph, solvable_graph, var_to_diff, eq_to_diff = s
```
YingboMa, line 459: Should we run `complete!` here?

Keno: Yeah, I did it for the other call site below, but forgot it here. Feel free to fix that up, or I can do it in a few hours.

PR #1834, opened 2022-09-23
Add clock inference

baggepinnen:
This pull request is the start towards taking discrete-time systems seriously. This implements clock inference, which can be described as the assignment of equations and variables to clock partitions. A discrete-time clock partition contains all equations that trigger upon the tick of a particular clock, while the continuous partition contains the entire continuous-time subsystem.

For readers that want to familiarize themselves with this PR, I suggest starting at the [hybrid systems test](https://github.com/SciML/ModelingToolkit.jl/pull/1834/files#diff-33f8e68d1bb215cfb4f58a4d4095db9092ac27462119c70ab9c2ca7d7ac2d422R178) (at line 178 in `test_clocks.jl`, the link will go stale when the PR is updated) where a simple system with one discrete-time partition (a controller) and one continuous time partition (the plant) is modeled.

- [x] Introduce `Shift` and `Hold` operators
- [x] Difference order lowering
- [x] Implement clock inference
- [x] Multi-rate systems (more than one clock)
- [ ] Code generation for discrete-time partitions (separate PR)

The example below illustrates basic user-facing interface. `Sample(t, dt)` creates an operator that when applied to a continuous-time signal produces a discrete-time signal, while the `Hold` operator takes a discrete-time signal and holds it constant to produce a continuous-time signal.
```julia
    dt = 0.1
    @variables t x(t) y(t) u(t) yd(t) ud(t) r(t)
    @parameters kp
    D = Differential(t)

    eqs = [
           # controller (time discrete part `dt=0.1`)
           yd ~ Sample(t, dt)(y)
           ud ~ kp * (r - yd)

           # plant (time continuous part)
           u ~ Hold(ud)
           D(x) ~ -x + u
           y ~ x]

    d = Clock(t, dt)
    @test get_eq_domain(eqs[1]) == d
    @test get_eq_domain(eqs[3]) == Continuous()
    @test get_eq_domain(eqs[4]) == Continuous()
```
A similar setup in modelica is shown below ([ref](https://specification.modelica.org/maint/3.4/Ch16.html)).

image

The modelica concept of `previous` is in this PR handled by the `Shift` operator. To make the user-facing interface nicer, we also introduce the `ShiftIndex` operator that automatically shifts the desired number of time steps using the syntax
```julia
k = ShiftIndex(t, 0.1)
xk = x(k)
xk1 = x(k + 1)
```
so the modelica code `previous(xd)` would be written `xd(k-1)`.


src/clocks.jl
```diff
@@ -0,0 +1,563 @@
...
+function get_eq_domain(eq, domain_above = Inferred())
+    ovars = collect_applied_operators(eq, Operator)
+    for op in ovars
+        while isoperator(op, Operator) # handle nested operators
+            odom = output_timedomain(op)
+            domain_above = merge_domains(domain_above, odom, eq)
+            changes_domain(op) && break
+            op = arguments(op)[1] # unwrap operator in case there are wrapped operators with a more specific output
+        end
+        if op isa Num
```
YingboMa, line 447: Doing anything with `Num` other than `unwrap` is almost certainly wrong.

src/clocks.jl
```diff
@@ -0,0 +1,563 @@
...
+    x isa InferredDiscrete && !(y isa Continuous) && return y
+    y isa InferredDiscrete && !(x isa Continuous) && return x
+    x == y || throw(ClockInferenceException("Cannot merge $x and $y"))
+    x
+end
+
+# nothing cases
+merge_domains(::Nothing, ::Nothing, args...) = nothing
+merge_domains(x::Any, ::Nothing, args...) = x
+merge_domains(::Nothing, x::Any, args...) = x
```
YingboMa, line 110: This is just something.

src/clocks.jl
```diff
@@ -0,0 +1,563 @@
...
+    end
+
+    throw(ClockInferenceException(emsg))
+end
+
+"""
+    preprocess_hybrid_equations(eqs, states)
+
+1. Perform clock inference using [`clock_inference`](@ref).
+2. Normalize `Shift` operations so that the largest positive shift becomes 1, and is the only term on the LHS of the equation. This uses [`normalize_shifts`](@ref).
```
YingboMa, line 141: Maybe don't do this step, because `Shift(10)` and `Shift(9)` would have a different interpretation than `Shift(1)` and `Shift(0)`.

src/clocks.jl
```diff
@@ -0,0 +1,563 @@
...
+3. For each clock partition, expand shift equations to companion form where the only remaining shifts are 1 and 0. This introduces `n` new states and equations where `n` = max_shift - min_shift - 1. 
+4. Rewrite Shift(1) to the continuous-time ` [`DiscreteUpdate`](@ref) operator and strip away hybrid and discrete operators from equations (domains are known from the clock inference). Uses [`strip_operator`](@ref)
```
YingboMa, line 143: Maybe just keep the shifts. It's more flexible this way.

YingboMa: We should normalize to `Shift(0)`

src/clocks.jl
```diff
@@ -0,0 +1,563 @@
...
+    r = AbortablePrewalk(PassThrough(r1), cond = isoperator(Operator)) # We should not insert a shift inside another operator
+    r(eq.lhs) ~ r(eq.rhs)
+end
+
+"""
+    new_eqs, new_vars = shift_order_lowering(eqs::Vector{Equation}, clock)
+
+Takes a vector of equations, all belonging to the same discrete clock partition, and introduces new variables for terms that are shifted by more than -1. For example
+```
+u(k+1) ~ u(k) + u(k-1) + u(k-2)
```
YingboMa, line 273: We should normalize everything to 0.

src/clocks.jl
```diff
@@ -0,0 +1,563 @@
...
+strip_operator(ex, operator) = strip_operator(ex ~ 0, operator).lhs
+
+"""
+    is_discrete_algebraic(eq)
+
+Returns true if the equation is a discrete algebraic equation. This function assumes that the equation has been shift normalized using [`normalize_shifts`](@ref).
+"""
+function is_discrete_algebraic(eq)
+    t = eq.lhs
+    t isa Term && operation(t) isa Shift && operation(t).steps == 0
```
YingboMa, line 353: -1

PR #1835, opened 2022-09-23
Remake with symbolic map

xtalax:
Depends on https://github.com/SciML/SciMLBase.jl/pull/262

Allows remaking with a list of pairs or dict, solves the ordering problem.


xtalax:
@YingboMa This PR is stuck in wrapper type hell at the moment, working for parameters but now getting `ERROR: ArgumentError: SymbolicUtils.Symbolic{Real}[y₁(t), y₂(t), y₃(t)] are missing from the variable map.` for `u0`.

src/variables.jl
```diff
@@ -46,7 +46,8 @@ applicable.
 function varmap_to_vars(varmap, varlist; defaults = Dict(), check = true,
                         toterm = Symbolics.diff2term, promotetoconcrete = nothing,
                         tofloat = true, use_union = false)
-    varlist = map(unwrap, varlist)
+    varlist = collect(map(unwrap, varlist))
```
YingboMa, line 49: Should this be `map(x->collect(unwrap(x)), varlist)`?

src/variables.jl
```diff
@@ -97,7 +98,7 @@ function _varmap_to_vars(varmap::Dict, varlist; defaults = Dict(), check = false
         varmap[p] = fixpoint_sub(v, varmap)
     end
 
-    missingvars = setdiff(varlist, keys(varmap))
+    missingvars = setdiff(varlist, collect(keys(varmap)))
```
YingboMa, line 101: If you are trying handle array variables, could you add a test?

xtalax: That wasn't my intention, I think due to a change in Symbolics this was giving me a lazy broadcast object, which was erroring on this line unless I collected the result.

ChrisRackauckas:
@YingboMa what's needed to finish this?

YingboMa (approved):

PR #1841, opened 2022-09-27
Show selected states in MatchedSystemStructure

Keno:
In the printing for `MatchedSystemStructure`, add a little blue `∫` symbol to indicate variables that are set to `SelectedState`, thus making the printing suitable for matchings produced from state selection (not just pantelides).


YingboMa (commented):
Could you add a test?

PR #1843, opened 2022-09-27
Compute `complete_ag` that contains irreducible alias information

YingboMa:
```julia
julia> @variables t x(t) y(t) z(t);

julia> D = Differential(t)
(::Differential) (generic function with 2 methods)

julia> eqs = [x - y + 2 ~ 0; D(y) ~ 0; z - D(x) ~ 0]; @named sys = ODESystem(eqs, t);

julia> alias_elimination(sys);
┌ Info:
└   vv => aa = Differential(t)(y(t)) => 0
┌ Info:
└   vv => aa = z(t) => Differential(t)(x(t))
```

### DIFF
--- a/src/systems/alias_elimination.jl
+++ b/src/systems/alias_elimination.jl
@@ -8,7 +8,7 @@ function alias_eliminate_graph!(state::TransformationState)
     mm = linear_subsys_adjmat(state)
     if size(mm, 1) == 0
         ag = AliasGraph(ndsts(state.structure.graph))
-        return ag, ag, mm, BitSet() # No linear subsystems
+        return ag, ag, mm, mm, BitSet() # No linear subsystems
     end
 
     @unpack graph, var_to_diff = state.structure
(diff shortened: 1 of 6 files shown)


src/systems/alias_elimination.jl
```diff
@@ -6,7 +6,10 @@ const KEEP = typemin(Int)
 
 function alias_eliminate_graph!(state::TransformationState)
     mm = linear_subsys_adjmat(state)
-    size(mm, 1) == 0 && return AliasGraph(ndsts(state.structure.graph)), mm, BitSet() # No linear subsystems
+    if size(mm, 1) == 0
+        ag = AliasGraph(ndsts(state.structure.graph))
+        return ag, ag, mm, mm, BitSet() # No linear subsystems
```
Keno, line 11:
```suggestion
        return ag, mm, ag, mm, BitSet() # No linear subsystems
```
to match the updated order of return values

PR #1844, opened 2022-09-28
Add `state_priority` to prefer some states in dummy derivative

YingboMa:


oscardssmith:
Is this good to go other than a reformat?

PR #1865, opened 2022-10-06
pass through force_SA to build_function

baggepinnen:
`build_function` has a documented keyword arg `force_SA` that I would find useful. However, it does no appear to be implemented at the moment so here's a companion PR to implement it
- https://github.com/JuliaSymbolics/Symbolics.jl/pull/756

### DIFF
--- a/src/inputoutput.jl
+++ b/src/inputoutput.jl
@@ -192,6 +192,7 @@ function generate_control_function(sys::AbstractODESystem, inputs = unbound_inpu
                                    disturbance_inputs = disturbances(sys);
                                    implicit_dae = false,
                                    simplify = false,
+                                   force_SA = false,
                                    kwargs...)
     if isempty(inputs)
         error("No unbound inputs were found in system.")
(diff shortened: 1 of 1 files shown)


baggepinnen:
CI failure seems unrelated
```
 Info: Reference Test for "10.tex" failed.
```

bowenszhu:
We don’t need to explicitly add this optional argument, as it can be passed through `kwargs`.

baggepinnen:
No, `kwargs...` is passed in in more than one location, and you'll thus get an error from the method that does not take the `force_SA` argument

YingboMa (commented):
Should we just make `io_preprocessing` take unused kwargs?

baggepinnen:
I'm not sure what you mean, it would perhaps be good to allow passing any kwargs to both functions

YingboMa:
Yes

YingboMa:
Does https://github.com/SciML/ModelingToolkit.jl/pull/1868 make this PR obsolete?

bowenszhu:
> Does #1868 make this PR obsolete?

Yes. I think so.

PR #1875, opened 2022-10-09
`ODAEProblem` doesn't pass `paramsyms`

AayushSabharwal:
`ODAEProblem` didn't pass `paramsyms` and `indepsym` to the underlying `ODEFunction` constructor, so symbolic parameter indexing didn't work.


ChrisRackauckas:
add a test?

ChrisRackauckas (approved):

PR #1878, opened 2022-10-12
expose tofloat for ODEProblem

bradcarman:
Attempting to construct an ODEProblem with a parameter type of `Vector{Vector{Float64}}` will cause an error because `tofloat` defaults to true and `Vector{Float64}` cannot be converted to `Float64`, therefore now with this change one can set `tofloat=false`

### DIFF
--- a/src/systems/diffeqs/abstractodesystem.jl
+++ b/src/systems/diffeqs/abstractodesystem.jl
@@ -550,6 +550,7 @@ function process_DEProblem(constructor, sys::AbstractODESystem, u0map, parammap;
                            linenumbers = true, parallel = nothing,
                            eval_expression = true,
                            use_union = false,
+                           tofloat = !use_union,
                            kwargs...)
     eqs = equations(sys)
     dvs = states(sys)
@@ -560,8 +561,8 @@ function process_DEProblem(constructor, sys::AbstractODESystem, u0map, parammap;
     defs = mergedefaults(defs, parammap, ps)
     defs = mergedefaults(defs, u0map, dvs)
 
-    u0 = varmap_to_vars(u0map, dvs; defaults = defs, tofloat = true)
-    p = varmap_to_vars(parammap, ps; defaults = defs, tofloat = !use_union, use_union)
+    u0 = varmap_to_vars(u0map, dvs; defaults = defs, tofloat)
+    p = varmap_to_vars(parammap, ps; defaults = defs, tofloat, use_union)
     p = p === nothing ? SciMLBase.NullParameters() : p
 
     if implicit_dae && du0map !== nothing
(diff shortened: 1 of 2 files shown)


src/systems/diffeqs/abstractodesystem.jl
```diff
@@ -560,8 +561,8 @@ function process_DEProblem(constructor, sys::AbstractODESystem, u0map, parammap;
     defs = mergedefaults(defs, parammap, ps)
     defs = mergedefaults(defs, u0map, dvs)
 
-    u0 = varmap_to_vars(u0map, dvs; defaults = defs, tofloat = true)
-    p = varmap_to_vars(parammap, ps; defaults = defs, tofloat = !use_union, use_union)
+    u0 = varmap_to_vars(u0map, dvs; defaults = defs, tofloat)
```
YingboMa, line 564: Is there a case where `tofloat` shouldn't be true for states?

ChrisRackauckas: JumpProblem and DiscreteProblem

YingboMa: They are don't call DEProblem.

bradcarman:
I'm not sure there are that many use cases, but the following example attempts to use `Int` for the `u` type.  In this case we need the option `tofloat=false` to construct the ODEProblem correctly

```julia
using ModelingToolkit
using OrdinaryDiffEq

@variables t
sts = @variables x(t)
D = Differential(t)
eqs = [
    D(x) ~ 1
]
@named sys = ODESystem(eqs, t, sts, [])
u0 = [1]
prob = ODEProblem(sys, u0, (0, 10), tofloat=false)
solve(prob, ImplicitEuler(), dt=1, adaptive=false, initializealg=NoInit())
```

This gives `prob` as...
```julia
julia> prob
ODEProblem with uType Vector{Int64} and tType Int64. In-place: true
timespan: (0, 10)
u0: 1-element Vector{Int64}:
 1
```

The solution though is converted to Float64...
```julia
retcode: Success
Interpolation: 3rd order Hermite
t: 11-element Vector{Float64}:
  0.0
  1.0
  2.0
  3.0
  4.0
  5.0
  6.0
  7.0
  8.0
  9.0
 10.0
u: 11-element Vector{Vector{Float64}}:
 [1.0]
 [2.0]
 [3.0]
 [4.0]
 [5.0]
 [6.0]
 [7.0]
 [8.0]
 [9.0]
 [10.0]
 [11.0]
```

YingboMa:
Should we have separate control of `tofloat` for states and parameters?

bradcarman:
I'm OK with the 1 flag, such that the meanings are:
- `tofloat=true` means convert everything (parameters and states) to Float64's
- `tofloat=false` means the user is responsible for supplying the desired types for parameters and states

YingboMa:
I think ODE solvers don't really work with non-float state. So we should always set `tofloat = true` for states.

ChrisRackauckas:
yes, non-discrete problems require continuous states, so it will always fail if you use an int. If this piece isn't hit by DiscreteProblem/JumpProblem, then it shouldn't even be an option for states.

PR #1879, opened 2022-10-12
dummy_derivatives: Make robust against present, but unused derivatives

Keno:
Like #1842, but for dummy_derivatives.


YingboMa:
Superseded by #1890

PR #1886, opened 2022-10-17
Add sparsity option to ODEFunctionExpr

YingboMa:


bradcarman:
The added `sparsity` keyword does not appear to be implemented

PR #1887, opened 2022-10-18
Fix ODEFunction QuoteNode bug and implement iip_config

bradcarman:


YingboMa (changes requested):
`kwargs...` already passes the `iip_config`

YingboMa (approved):

PR #1891, opened 2022-10-19
Handle self-alias in a differentiation chain more carefully

YingboMa:


src/systems/alias_elimination.jl
```diff
@@ -901,10 +902,14 @@ function alias_eliminate_graph!(graph, var_to_diff, mm_orig::SparseMatrixCLIL)
     # RHS or its derivaitves must still exist in the system to be valid aliases.
     needs_update = false
     function contains_v_or_dv(var_to_diff, graph, v)
+        counter = 0
         while true
             isempty(𝑑neighbors(graph, v)) || return true
             v = var_to_diff[v]
             v === nothing && return false
+            counter += 1
+            counter > 10_000 &&
```
oscardssmith, line 911: would it be better to do real loop detection?

YingboMa: This should really never happen, so this check is here only to make sure that MTK terminates.

PR #1898, opened 2022-10-21
Add tspan to *System

YingboMa:


isaacsas:
Any chance to support a `tspan` given in terms of symbolic parameters?

YingboMa:
Could you give me an example?

isaacsas:
Maybe something like
```julia
@parameters T1 = 0.0 T2 = 1.0
tspan = (T1, T2)
@named sys = ODESystem(eqs, t, [x], ps; defaults = [default_u0; default_p], tspan)
...

# uses the default values or a user parameter map to set the actual numeric tspan
# via varmap_to_vars
prob = ODEProblem(sys)  
```

The idea would be to keep allowing as much of the system to be specified symbolically as possible by allowing parameters in `tspan`. 

I've thought that it might be useful to do something similar with letting events store symbolic tstop parameters (or maybe that makes more sense in a system), since right now discrete events require users to still manually submit a list of numeric tstops to `solve`.

YingboMa:
Yeah, it should be pretty straightforward to implement. Do you want to make a PR?

PR #1903, opened 2022-10-23
Fix wrong variable name in `OptimizationProblemExpr`

AayushSabharwal:


src/systems/optimization/optimizationsystem.jl
```diff
@@ -366,7 +366,7 @@ function OptimizationProblemExpr{iip}(sys::OptimizationSystem, u0,
 
     defs = defaults(sys)
     defs = mergedefaults(defs, parammap, ps)
-    defs = mergedefaults(defs, u0map, dvs)
+    defs = mergedefaults(defs, u0, dvs)
```
ChrisRackauckas, line 369: I'd think, change the name above?

PR #1920, opened 2022-11-04
Fix symbolic remake

ValentinKaisermayer:
Depends on: https://github.com/SciML/SciMLBase.jl/pull/299


ValentinKaisermayer:
I think this should not work:
https://github.com/SciML/DiffEqBase.jl/blob/7f565c60fffc18587e3b9c2a029866db679c2d5a/test/remake_tests.jl#L86-L94
Why should you be allowed to change from `NullParameters` to something else?

ValentinKaisermayer:
Closing this in favour of #1941

PR #1929, opened 2022-11-08
Add unit validation for Connections

contradict:


YingboMa:
format

src/systems/validation.jl
```diff
@@ -170,6 +170,36 @@ function _validate(terms::Vector, labels::Vector{String}; info::String = "")
     valid
 end
 
+function _validate(conn::Connection; info::String = "")
+    valid = true
+    syss = get_systems(conn)
+    sys = first(syss)
+    st = states(sys)
+    for s in syss[2:end]
```
YingboMa, line 178:
```suggestion
    for s in @view syss[2:end]
```

YingboMa:
Wow, there are so many glaring problem with `src/systems/validation.jl` both style-wise and performance. I most regret that it is merged.

contradict:
Oh dear, anything I can clean up with this PR?

YingboMa:
There are two major problems after the `const`s fix.
1. all the functions in https://github.com/SciML/ModelingToolkit.jl/blob/2eaf863a223155248b5b7bddb82e03d41bf98424/src/systems/validation.jl#L173-L223 allocate unnecessarily. They should either use tuples or iterators to avoid allocation, and call `_validate_terms`.
2. https://github.com/SciML/ModelingToolkit.jl/blob/2eaf863a223155248b5b7bddb82e03d41bf98424/src/systems/validation.jl#L134-L150 `safe_get_unit` shouldn't exist. We should compare dimensions manually like
```julia
julia> dimension(u"K") == dimension(u"°C")
true

julia> dimension(u"K") == dimension(u"s")
false
```
and fast return in `get_unit` instead of throwing exceptions then catching it.

PR #1934, opened 2022-11-09
propagate force_SA to build function

baggepinnen:
depends on a new version (unreleased) of Symbolics


YingboMa:
The reason we splat kwarg is to not do this.

PR #1943, opened 2022-11-13
Makes cons_h and cons_j optional

ValentinKaisermayer:
Not all solvers need it and it seems to take a LOT of time for larger systems.


ChrisRackauckas:
Docs failure is related. Please fix the doc example by adding the `cons_h` and `cons_j` true options.

PR #1951, opened 2022-11-17
Add DelayParentScope scoping utility

jamesjscully:
from slack: I made a scoping utility DelayParentScope for MTK and I am happy to document it and submit a PR if it would be a welcome addition. It was intended to solve the following use case:
I have this model structure: current < cell < cellgroup < network and I need the parameters that control equations at the current-level to be shared across all cells at the celltype level, but each current should have its own namepace within that the cellgroup namespace. This allows me to construct many instances of Identical Cells.

One could easily combine ParentScope and DelayParentScope, but I don't think it really matters.

### DIFF
--- a/src/systems/abstractsystem.jl
+++ b/src/systems/abstractsystem.jl
@@ -363,15 +363,15 @@ function ParentScope(sym::Union{Num, Symbolic})
     setmetadata(sym, SymScope, ParentScope(getmetadata(value(sym), SymScope, LocalScope())))
 end
 
-struct DelayParentScope <: SymScope 
+struct DelayParentScope <: SymScope
     parent::SymScope
     N::Int
 end
 function DelayParentScope(sym::Union{Num, Symbolic}, N)
     setmetadata(sym, SymScope,
-        DelayParentScope(getmetadata(value(sym), SymScope, LocalScope()),N))
+                DelayParentScope(getmetadata(value(sym), SymScope, LocalScope()), N))
 end
-DelayParentScope(sym::Union{Num, Symbolic}) = DelayParentScope(sym,1)
+DelayParentScope(sym::Union{Num, Symbolic}) = DelayParentScope(sym, 1)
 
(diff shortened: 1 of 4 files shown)


YingboMa:
Could you run the formatter? See https://github.com/SciML/ModelingToolkit.jl/blob/master/CONTRIBUTING.md

src/systems/abstractsystem.jl
```diff
@@ -363,6 +363,16 @@ function ParentScope(sym::Union{Num, Symbolic})
     setmetadata(sym, SymScope, ParentScope(getmetadata(value(sym), SymScope, LocalScope())))
 end
 
+struct DelayParentScope <: SymScope
```
YingboMa, line 366: Missing an export like `ParentScope`.

YingboMa (commented):
The test failure looks real

jamesjscully:
I fixed the tests, but with nested ParentScope, and with DelayParentScope, there is strange behavior, although I'm not sure if it is intended. 

```julia
@parameters t a b
p = [a
     ParentScope(b)
     ParentScope(ParentScope(c))
]

level0 = ODESystem(Equation[], t, [], p; name = :level0)
level1 = ODESystem(Equation[], t, [], []; name = :level1) ∘ level0
level2 = ODESystem(Equation[], t, [], []; name = :level2) ∘ level1

ps = parameters(level2)
p_names = ModelingToolkit.getname.(ps)
@test isequal(p_names[1], :level1₊level0₊a)
@test isequal(p_names[2], :level1₊b)
@test isequal(ps[1], level1.level0.a)
@test isequal(ps[2], level1.b) # this fails
```

It seems that you can't use getproperty to access these properties. If this behavior is intended I will update the docs, otherwise maybe I should open an issue separately?

YingboMa:
That looks like a bug indeed. Do you mind to file a PR to fix and test this?

YingboMa (approved):

PR #1963, opened 2022-11-24
Prevent repeated call of substituter(rules) in substitute.

evetion:
By calling `map` we now repeatedly calculate `substituter(input_to_parameters)` inside Symbolics, while `input_to_parameters` stays the same. 

We can use [this substitute call](https://github.com/JuliaSymbolics/Symbolics.jl/blob/61bf8a9015749d7869b5e26f08210933c55d2c98/src/equations.jl#L62) instead to prevent this.

@visr


YingboMa (changes requested):
Just call `fast_substitute`

YingboMa:
We can check if `input_to_parameters` is empty as well.

evetion:
Ok, done so. Thanks for the feedback 👍🏻

PR #1974, opened 2022-11-27
Add MethodOfLines to Downstream

ChrisRackauckas:


ChrisRackauckas:
@xtalax what is this failure?

xtalax:
Haven't seen this before, this passed on the previous CI run. I will investigate, looks like time independent being used to index a time dependant solution

xtalax:
as of MTK v8.29.1 this error is not present
...
And I do not reproduce as of 8.30, something in its dependencies is limiting MTK to this version

xtalax:
@ChrisRackauckas please cycle this and I will fix what comes up

xtalax:
please also add the burgers group as this fails due to a regression from upstream

PR #1975, opened 2022-11-27
Try to avoid violating AliasGraph invariants

Keno:
In https://github.com/SciML/ModelingToolkit.jl/commit/f0ec2982d991b1dc8f1c4e0b70497159e6a1f5f7, AliasGraph gained cases where x is aliased to `-x`. This happens when there a side branch of the differentiation tree that was deeper than the stem. Currently alias elimination attempts to mutate-in-place the var-to-diff relationships in order to move the side branch onto the main stem. This is problematic for two reasons:

1. It assumes that TransformationState is mutable and that var_to_diff relationships may be arbitrarily updated. This is not necessarily the case.
2. It abuses the AliasGraph data structure. Ordinarily an x -> -x alias would force these to be equal, i.e. imply `x->0`. However, here it is used as an in-place marker of this branch transplant.

This attempts to fix both of these issues by allowing alias_elimination to introduce extra derivative variables on the main stem (similar to pantelides introducing extra derivative variables).

This seems to work for what I need, but currently breaks MTK tests, so some more work is required here.

### DIFF
--- a/src/ModelingToolkit.jl
+++ b/src/ModelingToolkit.jl
@@ -122,6 +122,8 @@ function parameters end
 # this has to be included early to deal with depency issues
 include("structural_transformation/bareiss.jl")
 function complete end
+function var_derivative! end
+function var_derivative_graph! end
 include("bipartite_graph.jl")
 using .BipartiteGraphs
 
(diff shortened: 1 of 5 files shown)


Keno:
❤️ Thanks for finishing this up!

PR #1980, opened 2022-11-29
Add Conductor.jl as a downstream test

ChrisRackauckas:


anandijain:
it'd probably be fine to skip conductor on LTS/1.6, is that possible?

ChrisRackauckas:
@wsphillips what's the issue with v1.6? It would be difficult to section that out.

anandijain:
its just kwargs `ERROR: LoadError: LoadError: syntax: invalid assignment location "; interval, duration, delay, npulses, offset, amplitude" around /home/runner/work/ModelingToolkit.jl/ModelingToolkit.jl/downstream/src/stimulus.jl:24`

https://github.com/SciML/ModelingToolkit.jl/actions/runs/3578134519/jobs/6017939231

wsphillips:
I'm using destructuring syntax. I guess I can go back to use `@unpack` or similar

Issue #1902, opened 2022-10-22
`modelingtoolkitize` ignores constraints


ValentinKaisermayer:
https://github.com/SciML/ModelingToolkit.jl/blob/2905e79270a73ae389e949779684fb96c78776c1/src/systems/diffeqs/modelingtoolkitize.jl#L224

PR #1983, opened 2022-11-30
Handle constraints inside `modelingtoolkitize` of `OptimizationSystem`

Vaibhavdixit02:
This will enable the `AutoModelingToolkit` in Optimization.jl to switch over to using `modelingtoolkitize`. https://github.com/SciML/Optimization.jl/pull/432

### DIFF
--- a/src/systems/optimization/modelingtoolkitize.jl
+++ b/src/systems/optimization/modelingtoolkitize.jl
@@ -1,0 +1,39 @@
+"""
+$(TYPEDSIGNATURES)
+
+Generate `OptimizationSystem`, dependent variables, and parameters from an `OptimizationProblem`.
+"""
+function modelingtoolkitize(prob::DiffEqBase.OptimizationProblem; num_cons = 0, kwargs...)
+    if prob.p isa Tuple || prob.p isa NamedTuple
+        p = [x for x in prob.p]
+    else
+        p = prob.p
+    end
+
+    vars = reshape([variable(:x, i) for i in eachindex(prob.u0)], size(prob.u0))
+    params = p isa DiffEqBase.NullParameters ? [] :
+             reshape([variable(:α, i) for i in eachindex(p)], size(Array(p)))
+
+    eqs = prob.f(vars, params)
+
+    if DiffEqBase.isinplace(prob) && !isnothing(prob.f.cons)
+        lhs = Array{Num}(undef, num_cons)
+        prob.f.cons(lhs, vars, params)
+
+        if !isnothing(prob.lcons) && !isnothing(prob.ucons)
+            cons = prob.lcons .≲ lhs .≲ prob.ucons
+        else
+            cons = lhs .~ 0.0
+        end
+    elseif !isnothing(prob.f.cons)
+        cons = prob.f.cons(vars, params)
+    else
+        cons = []
+    end
+
+    de = OptimizationSystem(eqs, vec(vars), vec(params);
+                            name = gensym(:MTKizedOpt),
+                            constraints = cons,
+                            kwargs...)
+    de
+end
--- a/src/systems/optimization/modelingtoolkitize.jl
+++ b/src/systems/optimization/modelingtoolkitize.jl
@@ -19,10 +19,23 @@ function modelingtoolkitize(prob::DiffEqBase.OptimizationProblem; num_cons = 0, 
     if DiffEqBase.isinplace(prob) && !isnothing(prob.f.cons)
         lhs = Array{Num}(undef, num_cons)
         prob.f.cons(lhs, vars, params)
+        cons = Union{Equation,Inequality}[]
 
-        if !isnothing(prob.lcons) && !isnothing(prob.ucons)
-            cons = prob.lcons .≲ lhs .≲ prob.ucons
-        else
+        if !isnothing(prob.lcons)
+            for i in 1:num_cons
+                !isinf(prob.lcons[i]) && prob.lcons[i] !== prob.ucons[i] && push!(cons, prob.lcons[i] ≲ lhs[i])
+            end
+        end
+
+        if !isnothing(prob.ucons)
+            for i in 1:num_cons
+                if !isinf(prob.ucons[i])
+                    prob.lcons[i] !== prob.ucons[i] ? push!(cons, lhs[i] ~ prob.ucons[i]) : push!(cons, lhs[i] ≲ prob.ucons[i])
+                end
+            end
+        end
+
+        if (isnothing(prob.lcons) || all(isinf.(prob.lcons))) && (isnothing(prob.ucons) || all(isinf.(prob.ucons)))
             cons = lhs .~ 0.0
         end
     elseif !isnothing(prob.f.cons)
--- a/src/systems/optimization/modelingtoolkitize.jl
+++ b/src/systems/optimization/modelingtoolkitize.jl
@@ -11,9 +11,10 @@ function modelingtoolkitize(prob::DiffEqBase.OptimizationProblem; kwargs...)
         p = prob.p
     end
 
-    vars = reshape([variable(:x, i) for i in eachindex(prob.u0)], size(prob.u0))
+    vars = ArrayInterfaceCore.restructure(prob.u0,
+                                          [variable(:x, i) for i in eachindex(prob.u0)])
     params = p isa DiffEqBase.NullParameters ? [] :
-             reshape([variable(:α, i) for i in eachindex(p)], size(Array(p)))
+             ArrayInterfaceCore.restructure(p, [variable(:α, i) for i in eachindex(p)])
 
     eqs = prob.f(vars, params)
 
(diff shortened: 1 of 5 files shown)


src/systems/optimization/optimizationsystem.jl
```diff
@@ -209,9 +208,9 @@ function DiffEqBase.OptimizationProblem{iip}(sys::OptimizationSystem, u0map,
                                              parammap = DiffEqBase.NullParameters();
                                              lb = nothing, ub = nothing,
                                              grad = false,
-                                             hess = false, sparse = false,
+                                             hess = false, obj_sparse = false,
                                              cons_j = false, cons_h = false,
-                                             checkbounds = false,
+                                             cons_sparse, checkbounds = false,
```
ChrisRackauckas, line 213: no default here?

ChrisRackauckas:
Needs tests

src/systems/optimization/optimizationsystem.jl
Vaibhavdixit02: Missed it, I just noticed it as well

ValentinKaisermayer:
closes #1902

src/systems/optimization/modelingtoolkitize.jl
```diff
@@ -19,10 +19,23 @@ function modelingtoolkitize(prob::DiffEqBase.OptimizationProblem; num_cons = 0,
...
+
+        if !isnothing(prob.ucons)
+            for i in 1:num_cons
+                if !isinf(prob.ucons[i])
+                    prob.lcons[i] !== prob.ucons[i] ? push!(cons, lhs[i] ~ prob.ucons[i]) : push!(cons, lhs[i] ≲ prob.ucons[i])
+                end
+            end
+        end
+
+        if (isnothing(prob.lcons) || all(isinf.(prob.lcons))) && (isnothing(prob.ucons) || all(isinf.(prob.ucons)))
```
ValentinKaisermayer, line 38: Why is this the behaviour for equality contraints? It seems wrong.

Vaibhavdixit02: Why? Btw did you see line 33 as well?

src/systems/optimization/modelingtoolkitize.jl
```diff
@@ -19,10 +19,23 @@ function modelingtoolkitize(prob::DiffEqBase.OptimizationProblem; num_cons = 0,
     if DiffEqBase.isinplace(prob) && !isnothing(prob.f.cons)
         lhs = Array{Num}(undef, num_cons)
         prob.f.cons(lhs, vars, params)
+        cons = Union{Equation,Inequality}[]
 
-        if !isnothing(prob.lcons) && !isnothing(prob.ucons)
-            cons = prob.lcons .≲ lhs .≲ prob.ucons
-        else
+        if !isnothing(prob.lcons)
```
ValentinKaisermayer, line 24: I think this can be refactored, since ucons and lcons are either both nothing or both not.

src/systems/optimization/modelingtoolkitize.jl
```diff
@@ -19,10 +19,23 @@ function modelingtoolkitize(prob::DiffEqBase.OptimizationProblem; num_cons = 0,
...
+            for i in 1:num_cons
+                !isinf(prob.lcons[i]) && prob.lcons[i] !== prob.ucons[i] && push!(cons, prob.lcons[i] ≲ lhs[i])
+            end
+        end
+
+        if !isnothing(prob.ucons)
+            for i in 1:num_cons
+                if !isinf(prob.ucons[i])
+                    prob.lcons[i] !== prob.ucons[i] ? push!(cons, lhs[i] ~ prob.ucons[i]) : push!(cons, lhs[i] ≲ prob.ucons[i])
```
ValentinKaisermayer, line 33: Isn't it the other way around? If ucons and lcons are equal, an equality has to be added.

src/systems/optimization/modelingtoolkitize.jl
ValentinKaisermayer: I meant that -inf <= cons <= inf yields cons ~ 0

src/systems/optimization/modelingtoolkitize.jl
Vaibhavdixit02: Oh shit. Yes it's flipped, thanks for catching that, how did this not get caught by the tests though I'll check that

src/systems/optimization/modelingtoolkitize.jl
Vaibhavdixit02: I see what you are saying, it might be better to no assume that if user is not passing equal constraints they don't intend it to be an equality constraint. I was assuming that case to be equivalent to having a equality to 0, it might not always be true

ValentinKaisermayer: I think if a user supplies such a pair of (-inf,inf) it is either a modelling error or the intension is to deactivate the constraint. So in any case no constraint should be added. We might even cry. I wonder how JuMP handles such a case.

src/systems/optimization/modelingtoolkitize.jl
Vaibhavdixit02: That holds true for `ub` and `lb` but not for these. We could change it in SciMLBase, but I have left it as it is for correct behavior now and will make the change there subsequently.

src/systems/optimization/modelingtoolkitize.jl
Vaibhavdixit02: Added an error in this case

src/systems/optimization/modelingtoolkitize.jl
```diff
@@ -0,0 +1,52 @@
...
+        if (isnothing(prob.lcons) || all(isinf.(prob.lcons))) && (isnothing(prob.ucons) || all(isinf.(prob.ucons)))
+            cons = lhs .~ 0.0
+        end
+    elseif !isnothing(prob.f.cons)
+        cons = prob.f.cons(vars, params)
+    else
+        cons = []
+    end
+
+    de = OptimizationSystem(eqs, vec(vars), vec(params);
```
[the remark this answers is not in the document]
Vaibhavdixit02: I am not sure I understood what you meant here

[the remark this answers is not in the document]
ValentinKaisermayer: You are generating the parameters via `variable(:α, i)`

src/systems/optimization/modelingtoolkitize.jl
```diff
@@ -0,0 +1,62 @@
...
+            Ensure you pass equal bounds (the scalar that the constraint should evaluate to) for equality constraints
+            or pass the lower and upper bounds for inequality constraints."))
+        end
+    elseif !isnothing(prob.f.cons)
+        cons = prob.f.cons(vars, params)
+    else
+        cons = []
+    end
+
+    de = OptimizationSystem(eqs, vec(vars), vec(params);
```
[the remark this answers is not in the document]
ValentinKaisermayer: 👍

Vaibhavdixit02:
Should be ready now

src/systems/optimization/modelingtoolkitize.jl
```diff
@@ -0,0 +1,62 @@
...
+"""
+function modelingtoolkitize(prob::DiffEqBase.OptimizationProblem; kwargs...)
+    num_cons = isnothing(prob.lcons) ? 0 : length(prob.lcons)
+    if prob.p isa Tuple || prob.p isa NamedTuple
+        p = [x for x in prob.p]
+    else
+        p = prob.p
+    end
+
+    vars = reshape([variable(:x, i) for i in eachindex(prob.u0)], size(prob.u0))
```
ChrisRackauckas, line 14: This should be a `restructure`. Use the ODE one as a template, i.e. https://github.com/SciML/ModelingToolkit.jl/pull/1983/files#diff-5c721fb74d2a713d8da5efa082f9b6927958cbedf859232a859c252869193196R20-R21

ChrisRackauckas (approved):

test/optimizationsystem.jl
```diff
@@ -27,7 +27,7 @@ using ModelingToolkit: get_metadata
     generate_gradient(combinedsys)
     generate_hessian(combinedsys)
     hess_sparsity = ModelingToolkit.hessian_sparsity(sys1)
-    sparse_prob = OptimizationProblem(sys1, [x, y], [a, b], grad = true, sparse = true)
+    sparse_prob = OptimizationProblem(sys1, [x, y], [a, b], grad = true, obj_sparse = true)
```
ChrisRackauckas, line 30: what's `obj_sparse` and how is it different from `sparse`?

Vaibhavdixit02: In this PR I have removed `sparse` in favor of `obj_sparse` (objective's sparse) and `cons_sparse` (constraints' sparse). This is similar to the `AutoModelingToolkit` struct. A single kwarg is not sufficient since either could be sparse

ChrisRackauckas: I'd keep `obj_sparse` as `sparse`. Consistency matters.

Vaibhavdixit02: We can keep it and deprecate and interpret it as `obj_sparse`, but also keep these two for consistency with Optimization.jl?

ChrisRackauckas: No, not deprecate it. Why would an ODE use `obj_sparse` instead of `sparse`?

PR #1988, opened 2022-12-03
Symbolic indexing rework

AayushSabharwal:
Is this interface appropriate?


ChrisRackauckas:
Those look good. Is that all that's required for the full interface?

AayushSabharwal:
I just realised this is a bit of a circular dependency problem. SciMLBase can't use these functions, because MTK depends on it and not the other way around.

AayushSabharwal:
Do I move `AbstractSystem` to SciMLBase too? Or just let the functions be `Any`-typed

ChrisRackauckas:
Define stub functions and use them in the interface, and they get extended downstream.

ChrisRackauckas:
Then we should define the interface in RecursiveArrayTools as something about `DiffEqArrays` and `AbstractVectorOfArray`, and bubble it up.

AayushSabharwal:
DiffEqArray doesn't seem to support indexing with multiple symbols. The following doesn't work:

```julia
julia> dx = DiffEqArray([[f(x) f2(x)] for x in t], t, [:a,:b])
julia> dx[[:a,:b]]
ERROR: DimensionMismatch: arrays could not be broadcast to a common size; got a dimension with lengths 11 and 2
...
```

Is this intentional?

AayushSabharwal:
I've messed about with this a lot (going the `SymbolCache` route) and it seems `independent_variables` at the very least needs to be hoisted to RecursiveArrayTools since `getindepsym` won't work [here](https://github.com/AayushSabharwal/SciMLBase.jl/blob/1c59f99a8e51ad27e78c9c0b1e0fdc463b0230eb/src/solutions/ode_solutions.jl#L115-L132)

ChrisRackauckas:
At that point it's a whole package. SymbolicIndexingInterface.jl? Document it well enough and it can be a thing.

AayushSabharwal:
That sounds like a good idea. It would be pretty small to start: `states`, `parameters`, `independent_variables`, the new interface described here, and maybe `SymbolCache`. But I have a feeling we'll end up chucking a lot more in there eventually. Should I just make this work for now (hoisting to RecursiveArrayTools) and refactor into the new package later, or just create the repo now and see where we get.

ChrisRackauckas:
Make it work, get it tested, and then in a last step refactor to a new package and we move it to SciML to register.

AayushSabharwal:
SciMLBase tests pass if the following is done:
- Make this branch of MTK use [`AayushSabharwal/RecursiveArrayTools#symbolcache`](https://github.com/AayushSabharwal/RecursiveArrayTools.jl/tree/symbolcache)
- Make SciMLBase use the above branch of RecursiveArrayTools
- Make SciMLBase's `test/downstream` use this branch of MTK and the above RecursiveArrayTools

ChrisRackauckas:
Alright cool, seems like a good time to make that extra package then.

AayushSabharwal:
Everything that needs to be in the package is in `src/symbolic_indexing_interface.jl` in RecursiveArrayTools. Should be simple enough to move it over.

AayushSabharwal:
Should I tag a release for SymbolicIndexingInterface, so the other packages can depend on it?

ChrisRackauckas:
I set it to go.

AayushSabharwal:
This _should_ work. For some reason I can't install SymbolicIndexingInterface locally yet. It won't work in CI, however, since it doesn't know to depend on the specific RecursiveArrayTools branch without Manifest.toml

AayushSabharwal:
Can I rename the other `states` functions (which aren't intended to implement the interface) to `_states` (such as the one in `src/systems/connectors.jl:164`)? Or do I just overload it and ignore the fact that it's not the intended use of the function? Fairly certain we need one or the other, since the interface `states` method is used a lot and adding `SymbolicIndexingInterface.` to all those places seems a bit redundant. It's also why precompilation will fail.

ChrisRackauckas:
> Can I rename the other states functions (which aren't intended to implement the interface) to _states (such as the one in src/systems/connectors.jl:164)?

yes

AayushSabharwal:
Is there a case where `states(sys)` has two variables, but `size(sol, 1) == 1`? It's happening for `test/components.jl:100`. `states(sys2) == [capacitor.v, capacitor.i]` but `size(sol2) == (1, 15)`

AayushSabharwal:
Everything should work now, bar the tests mentioned above, if the relevant branches of the other repositories are used. I haven't been able to run the distributed tests since Distributed tends to segfault on my system for reasons I haven't figured out yet.

AayushSabharwal:
Should this PR still pass `syms`/`indepsym`/`paramsyms` to SciMLFunctions? It doesn't currently, but I realised it probably should so it can be merged without SciMLBase, then after SciMLBase gets merged they can be removed in a separate PR to make 100% sure everything works.

ChrisRackauckas:
Yes, and then do an upper bound in the second PR. That way it's all done safely.

ChrisRackauckas (approved):

AayushSabharwal:
Once this merges, SciMLBase can depend on it, so those changes can be tested and then merged

PR #1991, opened 2022-12-06
Update index.md

jake484:
add Ai4EComponentLib.jl as a downstream test


ChrisRackauckas:
I meant add it to https://github.com/SciML/ModelingToolkit.jl/blob/master/.github/workflows/Downstream.yml

PR #1992, opened 2022-12-06
Update Downstream.yml

jake484:
add Ai4EComponentLib.jl as a downstream test


jake484:
> The tests in the downstream repo are all commented out. https://github.com/ai4energy/Ai4EComponentLib.jl/blob/8ba8f67bebde70756e856d7ea00b4401842b7af4/test/runtests.jl

thank for your suggestion!

PR #1996, opened 2022-12-12
pass analytic to ODEFunction

xtalax:
Where is the list of allowed kwargs to ODEFunction? Would be better to do this for all.


xtalax:
@ChrisRackauckas This should fix the benchmarks

YingboMa (approved):

PR #1997, opened 2022-12-12
bump SymbolicIndexingInterface to 0.2

oscardssmith:
required for https://github.com/SciML/SciMLBase.jl/pull/342


ChrisRackauckas:
Test failure

oscardssmith:
interestingly it's a test failure that occurred even though the version bump I allowed didn't happen (because it hasn't picked up the new version of RecursiveArrayTools)

AayushSabharwal:
This error seems to be because the `ODEFunction` is passed a `Symbol` for indepsym, which is passed as the `indepsym` to DiffEqArray. Then, while indexing it tries to compare the `Symbol` to a `Symbolics.Num` and fails, thus searching for `t` in `observed`

AayushSabharwal:
The same test seems to have passed in #1988 because it used RecursiveArrayTools v2.32.3, just before the change there

PR #2003, opened 2022-12-15
option to set all dummy derivatives to zero in `linearize`

baggepinnen:


YingboMa (commented):
Does Julia 1.6 not like ControlSystemsMTK?

YingboMa (commented):
Could you change the tests such that ControlSystemsMTK isn't needed?

baggepinnen:
I don't think that `ControlSystemsMTK` is the problem, since `ControlSystemsMTK` tests pass on julia v1.6 without problems. I'd also like to keep it as kind-of a downstream test that the simplification in `named_ss` when input derivatives are present is working as expected.

Running it on v1.6 locally indicates that `using ControlSystemsMTK` runs, and it gets stuck on 
```julia
@named link1 = Link(; m = 0.2, l = 10, I = 1, g = -9.807)
```
If I put `@show` in front of 
```julia
@show @parameters t
@show D = Differential(t)

@show @named link1 = Link(; m = 0.2, l = 10, I = 1, g = -9.807)
@show @named cart = Translational.Mass(; m = 1, s_0 = 0)
@show @named fixed = Fixed()
@show @named force = Force()
```
I get this printed before it gets stuck
```julia
#= /home/fredrikb/.julia/dev/ModelingToolkit/test/linearize.jl:181 =# @parameters(t) = Symbolics.Num[t]
D = Differential(t) = Differential(t)
```

baggepinnen:
I disabled those tests for Julia v1.6, so the tests are passing now. We could open a separate issue with the freezing of 
```julia
@named link1 = Link(; m = 0.2, l = 10, I = 1, g = -9.807)
```

YingboMa (approved):

PR #2009, opened 2022-12-24
Fix pantelides-with-ag corner case

Keno:
Fixes the corner case mentioned in the doc comment. In partiuclar, when we have an incomplete alias chain, the alias can change which variable needs to be considered the highest-differentiated. While we're at it, also take care of the todo in the same place.


src/structural_transformation/pantelides.jl
```diff
@@ -69,6 +69,86 @@ function pantelides_reassemble(state::TearingState, var_eq_matching)
...
+                stem == 0 && continue
+                # Ascend the stem
+                while isempty(𝑑neighbors(graph, var))
+                    var′ = invview(var_to_diff)[var]
+                    var′ === nothing && break
+                    stem′ = invview(var_to_diff)[var]
+                    # Invariant from alias elimination: Stem is chosen to have
+                    # the highest differentiation order.
+                    @assert stem′ !== nothing
+                    if !haskey(ag, var′) || ag[var′][2] != stem′
```
YingboMa, line 123: Could you add an ASCII diagram to show why this is true and why the clean up step is needed in comments?

Keno: ASCII diagram explaining why which condition is true?

YingboMa: Yeah

Keno: No, I'm asking which condition you want explained.

YingboMa: Maybe just explain the general process and why the clean up at the end is necessary?

Keno: The cleanup is just addressing the `# TODO: remove lower diff vars from whitelist` that was already there, but I can expand the comments a bit.

YingboMa (approved):

PR #2016, opened 2023-01-03
Add `unknown_states`

YingboMa:
Fix the CI failure of https://github.com/SciML/SciMLBase.jl/pull/342

### DIFF
--- a/src/systems/diffeqs/odesystem.jl
+++ b/src/systems/diffeqs/odesystem.jl
@@ -131,16 +131,17 @@ struct ODESystem <: AbstractODESystem
     """
     discrete_subsystems::Any
     """
-    solver_states: a list of actual solver states. Only used for ODAEProblem.
+    unknown_states: a list of actual states needed to be solved by solvers. Only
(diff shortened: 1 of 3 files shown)


src/systems/diffeqs/odesystem.jl
```diff
@@ -127,16 +127,21 @@ struct ODESystem <: AbstractODESystem
...
     complete::Bool
     """
-    discrete_subsystems: a list of discrete subsystems
+    discrete_subsystems: a list of discrete subsystems.
     """
     discrete_subsystems::Any
+    """
+    solver_states: a list of actual solver states. Only used for ODAEProblem.
+    """
+    solver_states::Union{Nothing, Vector{Any}}
```
ChrisRackauckas, line 136: That isn't a great name given how save_idxs would need to work.

YingboMa: How about unknown states?

ChrisRackauckas: that works

PR #2044, opened 2023-01-20
Apply scope recursively

### LINKED ISSUE #2043: ParentScope Bug
The following causes an error

```julia
julia> @parameters x=1
julia> ParentScope(1/x)
ERROR: MethodError: no method matching SymbolicUtils.Div(::Int64, ::Sym{Real, Base.ImmutableDict{DataType, Any}}, ::Bool, ::Base.ImmutableDict{DataType, Any})
```

This is caused by the division it seems, all other operations are OK: `ParentScope(1*x)`, `ParentScope(1 + x)`, `ParentScope(x^2)`

Also, the inverse will fail: `ParentScope(x^-1)`, but using a float is OK: `ParentScope(x^1.0)`

This is relevant when attempting to apply a parameter keyword with an expression

```julia
using ModelingToolkitStandardLibrary.Mechanical.Translational
@parameters x = 2
@named m = Mass(; m = 1/x)
```

YingboMa:
Fix #2043

### DIFF
--- a/test/structural_transformation/bareiss.jl
+++ b/test/structural_transformation/bareiss.jl
@@ -16,12 +16,13 @@ function det_bareiss!(M)
 end
 
 @testset "bareiss tests" begin
-# copy gives a dense matrix
-@testset "bareiss tests: $T" for T in (copy, sparse)
-    # matrix determinent pairs
-    for (M, d) in ((BigInt[9 1 8 0; 0 0 8 7; 7 6 8 3; 2 9 7 7], -1),
-                   (BigInt[1 big(2)^65+1; 3 4], 4 - 3 * (big(2)^65 + 1)))
-        # test that the determinent was correctly computed
-        @test det_bareiss!(T(M)) == d
+    # copy gives a dense matrix
+    @testset "bareiss tests: $T" for T in (copy, sparse)
+        # matrix determinent pairs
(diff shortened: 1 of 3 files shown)


test/structural_transformation/bareiss.jl
```diff
@@ -16,12 +16,13 @@ function det_bareiss!(M)
...
-@testset "bareiss tests: $T" for T in (copy, sparse)
-    # matrix determinent pairs
-    for (M, d) in ((BigInt[9 1 8 0; 0 0 8 7; 7 6 8 3; 2 9 7 7], -1),
-                   (BigInt[1 big(2)^65+1; 3 4], 4 - 3 * (big(2)^65 + 1)))
-        # test that the determinent was correctly computed
-        @test det_bareiss!(T(M)) == d
+    # copy gives a dense matrix
+    @testset "bareiss tests: $T" for T in (copy, sparse)
+        # matrix determinent pairs
+        for (M, d) in ((BigInt[9 1 8 0; 0 0 8 7; 7 6 8 3; 2 9 7 7], -1),
```
ChrisRackauckas, line 22: yeah the formatter update is really annoying lol

YingboMa: It's doublely annoying when I have another PR https://github.com/SciML/ModelingToolkit.jl/pull/2045/files#diff-fa6087bf9fde7edad13448c5fb462cd9da11b412d765f493658be9926598933b

Issue #2000, opened 2022-12-13
How to prevent breakage from changes to `structural_simplify`


baggepinnen:
I very often run into problems where tests or documentation builds suddenly start failing because a change has been made somewhere in the simplification pipeline that changes what variables I need to provide initial conditions for. As an example, I just started getting the error
```
SymbolicUtils.Symbolic{Real}[link1₊x1ˍtt(t), link1₊y_cmˍtt(t)] are missing from the variable map.
```
in a documentation example that has previously been running just fine. I'm now being requested to provide initial conditions for the second derivatives of some of the variables in my system. Technically, any change to `structural_simplify` that would force the user to provide such initial conditions where they were not previously required is a breaking change. Is there a good way for the user to guard themselves from such changes to not have tests broken every now and then?

ChrisRackauckas:
Make your tests be downstream tests to MTK?

> Technically, any change to structural_simplify that would force the user to provide such initial conditions where they were not previously required is a breaking change.

In theory, you only need the minimal set, and it should be able to deduce some of them from the others. In practice that part isn't implemented yet so it's just waiting in the wings.

finmod:
Along the same issue, this DAE model of constrained dynamics used to work before Julia v.1.8.4 and some modfifications in MTK in December. Now, after `structural_simplify` it breaks down on states[A1_ode] and later using ODAEProblem it gives the error mentioned by  @baggepinnen 

`SymbolicUtils.Symbolic{Real}[link1₊x1ˍtt(t), link1₊y_cmˍtt(t)] are missing from the variable map.` 

```Julia
using DifferentialEquations
using OrdinaryDiffEq
using ModelingToolkit
using Plots; gr()

ps = @parameters α β dp̂ γ K₀ l₀ L̂ Mᶠ₀ Mʰ₀ M̂ʰ p₀ S₀ Ŝ w₀ μ₁ μ₂ μ₃ μ₄ μ₅ μ₆ μ₇ μ₈ μ₉ μ₁₀ 𝛌₁ 𝛌₂ 𝛌₃

@variables t 
states = @variables  c(t) K(t) l(t) Mᶠ(t) Mʰ(t) p(t) S(t) w(t) k(t) mᶠ(t) mʰ(t) s(t)
[...]

PR #2048, opened 2023-01-24
add option to set state priority as argument to `structural_simplify`…

### LINKED ISSUE #2001: Set state priority from outside of component definition
The `state_priority` property can be set when creating a variable, but I'm not aware of any good way to select state priorities when, e.g., linearizing a system that is composed out of MTKstdlib components. Certain applications would benefit from control over what states are selected, for example optimal control, MPC and linearization. It is also inconvenient that the states that are selected by MTK changes between releases, since the varmap that has to be supplied changes. 

It would also be nice if MTK would figure out the initial conditions of some variables based on the conditions provided for other variables if possible.

baggepinnen:
… and friends

Closes #2001 
- #2001 

And somewhat mitigates (but doesn't solve)
- https://github.com/SciML/ModelingToolkit.jl/issues/2000

### DIFF
--- a/src/systems/abstractsystem.jl
+++ b/src/systems/abstractsystem.jl
@@ -1285,6 +1285,16 @@ function markio!(state, orig_inputs, inputs, outputs; check = true)
     state, orig_inputs
 end
 
+function set_priorities!(state, priorities)
+    fullvars = state.fullvars
+    for (i, v) in enumerate(fullvars)
+        if v in keys(priorities)
+            v = setmetadata(v, VariableStatePriority, priorities[v])
+            fullvars[i] = v
+        end
+    end
+end
+
 """
     (; A, B, C, D), simplified_sys = linearize(sys, inputs, outputs;    t=0.0, op = Dict(), allow_input_derivatives = false, kwargs...)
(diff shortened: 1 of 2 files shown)


src/systems/abstractsystem.jl
```diff
@@ -1285,6 +1285,16 @@ function markio!(state, orig_inputs, inputs, outputs; check = true)
     state, orig_inputs
 end
 
+function set_priorities!(state, priorities)
+    fullvars = state.fullvars
+    for (i, v) in enumerate(fullvars)
+        if v in keys(priorities)
+            v = setmetadata(v, VariableStatePriority, priorities[v])
```
YingboMa, line 1292: `v in keys(priorities)` and `priorities[v]` requires two dictionary look ups. Always use `get(dict, key, nothing)` if possible.

YingboMa (commented):
Still needs tests.

baggepinnen:
I added the system from https://discourse.julialang.org/t/why-is-modelingtoolkit-simplifying-this-way/96872 as test, but the fix in this PR does not actually help, the problem appears to have occurred already in the alias-elimination step, and state priorities are not considered until tearing.

PR #2049, opened 2023-01-24
avoid n^2 scaling in expand_instream

visr:
Before this patch, the connection set that belongs to an instream expression was found by looping over all of them. With n expressions and n connection sets, the time this took for large systems became very large (at least a day for me). This fixes that by creating a Dict that maps directly to the required connection set. This Dict is created by looping over all connection sets once.

### DIFF
--- a/src/systems/connectors.jl
+++ b/src/systems/connectors.jl
@@ -378,8 +378,29 @@ function expand_instream(csets::AbstractVector{<:ConnectionSet}, sys::AbstractSy
         end
     end
 
+    if !isempty(instream_exprs)
+        # map from a namespaced stream variable to a ConnectionSet
+        expr_cset = Dict()
+        for cset in csets
+            crep = first(cset.set)
+            current = namespace == crep.sys.namespace
+            for v in cset.set
+                if (current || !v.isouter)
+                    expr_cset[namespaced_var(v)] = cset.set
+                end
+            end
+        end
+    end
+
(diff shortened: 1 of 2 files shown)


src/systems/connectors.jl
```diff
@@ -378,8 +378,29 @@ function expand_instream(csets::AbstractVector{<:ConnectionSet}, sys::AbstractSy
...
+
     for ex in instream_exprs
-        cset, idx_in_set, sv = get_cset_sv(namespace, ex, csets)
+        ns_sv = only(arguments(ex))
+        full_name_sv = renamespace(namespace, ns_sv)
+        if haskey(expr_cset, full_name_sv)
+            cset = expr_cset[full_name_sv]
+        else
+            error("$ns_sv is not a variable inside stream connectors")
+        end
```
YingboMa, line 402:
```suggestion
        cset = get(expr_cset, full_name_sv, nothing)
        cset === nothing && error("$ns_sv is not a variable inside stream connectors")
```

YingboMa:
When adding the tests, did you make sure that they return the same expressions as before?

src/systems/connectors.jl
YingboMa: This minimized the number of dictionary lookups

visr:
Yes I ran these tests both with and without the changes, both passed.

PR #2051, opened 2023-01-25
Add `System` that handles ODESystem and SDESystem

YingboMa:


YingboMa:
@frankschae it looks like the `Girsanov_transform` fails. Could you take a look?

frankschae:
Yup, I'll have a look. Could it be that the default parameters are not used in
https://github.com/SciML/ModelingToolkit.jl/blob/af025524ba90cf51f8e5354eb1d0d940c6b06618/src/systems/diffeqs/sdesystem.jl#L354
after structural simplify is applied somewhere else in this PR?

src/systems/diffeqs/sdesystem.jl
```diff
@@ -545,7 +558,7 @@ function DiffEqBase.SDEProblem{iip}(sys::SDESystem, u0map = [], tspan = get_tspa
     sparsenoise === nothing && (sparsenoise = get(kwargs, :sparse, false))
 
     noiseeqs = get_noiseeqs(sys)
-    if noiseeqs isa AbstractVector
+    if noiseeqs isa AbstractVector || size(noiseeqs, 2) == 1
```
frankschae, line 561:
```suggestion
    if noiseeqs isa AbstractVector
```

frankschae (commented):
The default noise behavior is diagonal noise if the `noiseeqs` are a vector. Thus, it's important to keep the noise rate prototype for scalar noise processes (with states of dim > 1). In the Girsanov example, one adds an additional component (likelihood term) to a one-diagonal SDE driven by a single Wiener process.

ChrisRackauckas:
@YingboMa you forgot tests

YingboMa:
Yeah, it was accidentally removed and added back in https://github.com/SciML/ModelingToolkit.jl/pull/2120

PR #2054, opened 2023-01-26
Add substitute for ODEProblem

YingboMa:


isaacsas:
Should there be a warning that this only supports substituting within the equations? For example, it doesn’t handle events?

baggepinnen:
There should probably be a warning in the docstring, and a warning could be thrown if the system has events (otherwise not)

Issue #2061, opened 2023-02-01
Add a misc metadata field which allows for taking Any


ChrisRackauckas:
This is good to be used as a passthrough for upstream libraries to dump metadata that they don't know how to handle.

xtalax:
Isn't this handled already by the metadata fields?

YingboMa:
Yeah https://github.com/SciML/ModelingToolkit.jl/blob/16c1e8ff118cbd8336da74f8cf8c615230f5b9ce/src/systems/diffeqs/odesystem.jl#L113-L116

xtalax:
I'd take this a step further and offer a `Ref{Any}` metadata field to all the SciMLBase types that can be used after creation to save any misc metadata, an issue for elsewhere though

ChrisRackauckas:
`@variable x [misc = some_weird_type]`?

YingboMa:
> I'd take this a step further and offer a Ref{Any} metadata field to all the SciMLBase types that can be used after creation to save any misc metadata, an issue for elsewhere though

You can just do `@set! sys.metadata = something`

YingboMa:
> @variable x [misc = some_weird_type]?

Ah, you are asking for variables.

ChrisRackauckas:
Yes for variables, not systems.

PR #2062, opened 2023-02-01
Add misc metadata for variables/parameters

YingboMa:
Fix https://github.com/SciML/ModelingToolkit.jl/issues/2061

### DIFF
--- a/src/variables.jl
+++ b/src/variables.jl
@@ -5,6 +5,7 @@ struct VariableInput end
 struct VariableOutput end
 struct VariableIrreducible end
 struct VariableStatePriority end
+struct VariableMisc end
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
@jpfairbanks this feature request is done.

src/variables.jl
```diff
@@ -5,13 +5,15 @@ struct VariableInput end
 struct VariableOutput end
 struct VariableIrreducible end
 struct VariableStatePriority end
+struct VariableMisc end
```
baggepinnen, line 8: What's the motivation for having this quite convoluted way of attaching metadata to variables, rather than just having a `Dict{Symbol, Any}` or something?

YingboMa: Just to have something that passes through MTK and will be handled downstream.

Issue #2053, opened 2023-01-26
Simple model apparently runs out of memory


sh2pro:
Please see https://github.com/JuliaLang/julia/issues/48418

The machine is an ml.m5.large AWS instance with 8GB of memory.
Should be more than enough for a model with a couple of dozens equations

ChrisRackauckas:
Closing since this is almost certainly a Base issue with parallel precompilation.

sh2pro:
can you explain? what is the workaround?

ChrisRackauckas:
Open Julia with less threads and see if that fixes it.

sh2pro:
It doesn't


```
(base) jovyan@julia-ml-m5-large:~/work$ julia --threads=1 test.jl
  Activating project at `~/work`
after using
here we go
Killed
(base) jovyan@julia-ml-m5-large:~/work$
```

baggepinnen:
I had problems about a week ago where all CI runs using MTK failed to precompile due to running out of memory. It has gone away by now, so maybe it will for you as well if you update all packages?

sh2pro:
Seems that `ODAEProblem()` is running out of memory, so I don't think it's related to precompilation

ChrisRackauckas:
Sounds like it's a core dump related to codegen then. Can you share the core dump? Write it to a file before the terminal crashes?

ChrisRackauckas:
What's the reproducible example? I ran the code and get that `Segment is not defined` at `@named s = S()`, and indeed `Segment` is not defined in this example and not in mtkstdlib.

ChrisRackauckas:
@YingboMa able to recreate at all?

sh2pro:
FYI, here's the environment:
```shell
(base) jovyan@julia-ml-t3-medium:~/work$ julia
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.8.5 (2023-01-08)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

(@v1.8) pkg> activate .
  Activating project at `~/work`

(work) pkg> st
Status `~/work/Project.toml`
  [1b08a953] Dash v1.1.2
  [0c46a032] DifferentialEquations v7.6.0
[...]

Issue #2015, opened 2023-01-02
OOM on cyclical network


evetion:
Over at https://github.com/Deltares/Ribasim.jl, we're using MTK to build large hydrological models. What we see is that on some models, especially those that form a cyclic graph (but not on all of them), we run out of memory in the following code:

https://github.com/SciML/ModelingToolkit.jl/blob/a9dc1fc8d26b02e183ce5cc67e7471aae019d14b/src/structural_transformation/codegen.jl#L116...L121

The size of `next_assignments` keeps growing for each loop, hitting 1e9 for a system with 1290 states. Smaller models, even those with many cyclic parts converge much quicker. This behavior has also improved in the last months, now only running out of memory on the largest models.

With this [serialized structurally simplified sim here to deserialize](https://filesender.surf.nl/?s=download&token=ff506aef-d243-4e22-b7f1-64d4ec022763), we have the following MWE for the current MTK master branch, Julia 1.9.

```julia
using Serialization
using ModelingToolkit
using DataInterpolations
sim = deserialize("cached_simplified_sim.jls")
tspan = (1.5463008e9, 1.5513984e9)
prob = ODAEProblem(sim, [], tspan; sparse = true)  # OOM
```

@visr

evetion:
An overview of the model, with names of the nodes. Nodes with a black background form a cycle, of which there are two.

 <img width="1357" alt="Screenshot 2023-01-03 at 09 15 06" src="https://user-images.githubusercontent.com/8655030/210320578-fb8db5f0-c1da-44f0-9e36-35d3d4f9c64e.png">

ChrisRackauckas:
We now have our first benchmarks out which shows JuliaSimCompiler solves the problem:

https://docs.sciml.ai/SciMLBenchmarksOutput/dev/ModelingToolkit/ThermalFluid/

So closing this as it's just inherent to the MTK compiler's form, with trade-offs between the two. I recommend using the IR system for large problems.

It would be nice to get this as a benchmark, though we would need to not have a serialized version if we're to test the codegen elements as well. I would entertain that over at SciMLBenchmarks.jl: we're looking for large models to test!

visr:
Thanks for the update. We don't have any recent large models from https://github.com/Deltares/Ribasim since we moved to directly using OrdinaryDiffEq.jl, since JuliaSimCompiler wasn't an option for us.

PR #2068, opened 2023-02-06
Fix a cycle in dependency chasing

YingboMa:
Fix https://github.com/SciML/ModelingToolkit.jl/issues/2053


evetion:
Probably also fixes #2015

YingboMa:
I will tag a new release when this is merged. Please let me know if this fixes the problem. Also, I recommend to carefully benchmark ODEProblem and ODAEProblem before just using ODAEProblem. It is sometimes slower.

PR #2070, opened 2023-02-06
fix unbound arg

sjkelly:
I am not sure why this original syntax works.
This bounds the argument, which should avoid
some excess precompile/inference time.


YingboMa:
I think it was written like that to avoid a Julia compiler bug. The CI failed.

PR #2088, opened 2023-02-17
Implement in-place observed

### LINKED ISSUE #2083: Use `observed` to generate in-place functions for arrays

YingboMa:
Fix #2083

I will add tests later.


xtalax:
Do we want to merge this before or after the sym save_idxs changes, as they will likely both edit this code?

PR #2089, opened 2023-02-17
Fix Barreiss algorithm

oscardssmith:
This has a number of changes.
The first is that we were using ÷ instead of exact_div in bareiss_update_virtual_colswap_mtk which led to undetected overflows. We also had a number of bugs that arise when using BigInt precsion in bareiss (e.g. checking `!==0` vs `!=0`. This PR makes it so we first try to bareiss with the regular Int64 matrix, but if that gets an overflow, we fallback to the BigInt version.
@YingboMa I'm not 100% convinced that the `exact_div` is a sufficient check. Is it obvious that it always suffices to check the remainder or should we instead be checking for overflow on the multiplications?

### DIFF
--- a/src/structural_transformation/bareiss.jl
+++ b/src/structural_transformation/bareiss.jl
@@ -124,8 +124,8 @@ function bareiss_update!(zero!, M::StridedMatrix, k, swapto, pivot,
             M[j, i], r = divrem(M[j, i] * pivot - M[j, k] * Mki, prev_pivot)
             flag = flag | r
         end
-        iszero(flag) || error("Overflow occurred")
     end
+    iszero(flag) || error("Overflow occurred")
     zero!(M, (k + 1):size(M, 1), k)
 end
 
--- a/src/structural_transformation/bareiss.jl
+++ b/src/structural_transformation/bareiss.jl
@@ -124,8 +124,8 @@ function bareiss_update!(zero!, M::StridedMatrix, k, swapto, pivot,
             M[j, i], r = divrem(M[j, i] * pivot - M[j, k] * Mki, prev_pivot)
             flag = flag | r
         end
+        iszero(flag) || error("Overflow occurred")
     end
-    iszero(flag) || error("Overflow occurred")
     zero!(M, (k + 1):size(M, 1), k)
 end
 
--- a/src/systems/alias_elimination.jl
+++ b/src/systems/alias_elimination.jl
@@ -557,7 +557,8 @@ function aag_bareiss!(graph, var_to_diff, mm_orig::SparseMatrixCLIL)
     local bar
     try
         bar = do_bareiss!(mm, mm_orig, is_linear_variables)
-    catch
+    catch e
+        e isa OverflowError || rethrow(e)
         mm = changetype(BigInt, mm_orig)
         bar = do_bareiss!(mm, mm_orig, is_linear_variables)
     end
--- a/src/systems/alias_elimination.jl
+++ b/src/systems/alias_elimination.jl
@@ -224,7 +224,6 @@ the `constraint`.
         if constraint(length(vertices))
             for (j, v) in enumerate(vertices)
                 if (mask === nothing || mask[v])
-                    iszero(M.row_vals[i][j]) && error("zero found in sparse matrix")
                     return (CartesianIndex(i, v), M.row_vals[i][j])
                 end
             end
@@ -240,7 +239,6 @@ end
     for i in range
         row = @view M[i, :]
         if constraint(count(!iszero, row))
-            iszero(val) && continue
             for (v, val) in enumerate(row)
                 if mask === nothing || mask[v]
                     return CartesianIndex(i, v), val
--- a/src/systems/alias_elimination.jl
+++ b/src/systems/alias_elimination.jl
@@ -1,8 +1,6 @@
 using SymbolicUtils: Rewriters
 using Graphs.Experimental.Traversals
 
-const KEEP = typemin(Int)
-
 function alias_eliminate_graph!(state::TransformationState; kwargs...)
     mm = linear_subsys_adjmat!(state; kwargs...)
     if size(mm, 1) == 0
@@ -225,8 +223,10 @@ the `constraint`.
         vertices = eadj[i]
         if constraint(length(vertices))
             for (j, v) in enumerate(vertices)
-                (mask === nothing || mask[v]) &&
+                if (mask === nothing || mask[v])
+                    iszero(M.row_vals[i][j]) && error("zero found in sparse matrix")
                     return (CartesianIndex(i, v), M.row_vals[i][j])
+                end
             end
         end
     end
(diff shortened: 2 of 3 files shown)


src/systems/alias_elimination.jl
```diff
@@ -996,7 +1005,7 @@ function locally_structure_simplify!(adj_row, pivot_var, ag)
         alias = get(ag, var, nothing)
         if alias === nothing
             nirreducible += 1
-            alias_candidate = val => var
```
YingboMa, line 999: How do we know `val` can always be stored in `Int`?

src/systems/sparsematrixclil.jl
```diff
@@ -29,11 +29,20 @@ function Base.copy(S::SparseMatrixCLIL{T, Ti}) where {T, Ti}
                      map(copy, S.row_vals))
 end
 function swaprows!(S::SparseMatrixCLIL, i, j)
+    i == j && return
     swap!(S.nzrows, i, j)
     swap!(S.row_cols, i, j)
     swap!(S.row_vals, i, j)
 end
 
+function changetype(T, S::SparseMatrixCLIL)
```
YingboMa, line 38: `convert`?

YingboMa: Or `SparseMatrixCLIL{T}`

src/systems/sparsematrixclil.jl
```diff
@@ -189,13 +199,14 @@ function bareiss_update_virtual_colswap_mtk!(zero!, M::SparseMatrixCLIL, k, swap
             v == vpivot && continue
             ck = getcoeff(kvars, kcoeffs, v)
             ci = getcoeff(ivars, icoeffs, v)
-            ci = (pivot * ci - coeff * ck) ÷ last_pivot
-            if ci !== 0
+            p1 = Base.Checked.checked_mul(pivot, ci)
+            p2 = Base.Checked.checked_mul(coeff, ck)
+            ci = exactdiv(p1 - p2, last_pivot)
+            if ci != 0
```
YingboMa, line 205:
```suggestion
            if !iszero(ci)
```

YingboMa:
> Is it obvious that it always suffices to check the remainder or should we instead be checking for overflow on the multiplications?

Yes, we should check all integer ops.

src/systems/alias_elimination.jl
oscardssmith: I don't, but if I didn't convert it was complaining.

YingboMa: See the type assert above

oscardssmith: should I just remove the type assert?

YingboMa: Maybe? But could you write the correct type assert according to the eltype?

src/systems/sparsematrixclil.jl
```diff
@@ -189,13 +199,14 @@ function bareiss_update_virtual_colswap_mtk!(zero!, M::SparseMatrixCLIL, k, swap
             v == vpivot && continue
             ck = getcoeff(kvars, kcoeffs, v)
             ci = getcoeff(ivars, icoeffs, v)
-            ci = (pivot * ci - coeff * ck) ÷ last_pivot
-            if ci !== 0
+            p1 = Base.Checked.checked_mul(pivot, ci)
+            p2 = Base.Checked.checked_mul(coeff, ck)
+            ci = exactdiv(p1 - p2, last_pivot)
+            if !iszero(ci)
```
pepijndevos, line 205: why not checked_subtract and checked_div rather than exactdiv?

oscardssmith: div can't overflow, and checking exactdiv is a check that the subtraction doesn't overflow.

YingboMa: > checking exactdiv is a check that the subtraction doesn't overflow.

I am not sure if that's strong enough though. We probably want to use `checked_subtract`.

Keno: Yes, the subtraction should be checked. `exactdiv` is an algorithmic assertion though. If that's wrong, that error should propagate.

src/systems/alias_elimination.jl
```diff
@@ -554,7 +554,15 @@ function aag_bareiss!(graph, var_to_diff, mm_orig::SparseMatrixCLIL)
     end
     solvable_variables = findall(is_linear_variables)
 
-    return mm, solvable_variables, do_bareiss!(mm, mm_orig, is_linear_variables)
+    local bar
+    try
+        bar = do_bareiss!(mm, mm_orig, is_linear_variables)
+    catch
```
Keno, line 560: This should rethrow anything that's not an `OverflowError`.

src/structural_transformation/bareiss.jl
```diff
@@ -124,8 +124,8 @@ function bareiss_update!(zero!, M::StridedMatrix, k, swapto, pivot,
             M[j, i], r = divrem(M[j, i] * pivot - M[j, k] * Mki, prev_pivot)
             flag = flag | r
         end
+        iszero(flag) || error("Overflow occurred")
```
Keno, line 127: why this change?

src/systems/alias_elimination.jl
```diff
@@ -240,8 +240,8 @@ end
     for i in range
         row = @view M[i, :]
         if constraint(count(!iszero, row))
+            iszero(val) && continue
```
Keno, line 243: val is not defined here

src/structural_transformation/bareiss.jl
oscardssmith: it allows for early exit on overflow while keeping vectorization

Keno: We should assume that overflow is unlikely. With the check outside the loop, the compiler has more opportunities for pipelining. In any case, ideally these kinds of changes need to be their own commit rather than being lumped in with a bugfix.

src/systems/alias_elimination.jl
oscardssmith: the problem here is that `AliasGraph` is currently an `AbstractDict{Int, Pair{Int, Int}}` and changing that to  `AbstractDict{Int, Pair{BigInt, Int}}` seems like a pretty dramatic change.

Keno: We only support+-1 aliases at the moment anyway, so this should be fine

oscardssmith:
Ok. I think this is ready to merge.

oscardssmith:
I don't understand why format is failing here. JuliaFormatting isn't changing it.

Keno:
format-check fails on master with the same error. I assume it's a JuliaFormatter version difference.

Keno (approved):

YingboMa:
> format-check fails on master with the same error. I assume it's a JuliaFormatter version difference.

It's because the formatting on master is broken because of https://github.com/SciML/ModelingToolkit.jl/commit/a5f4e1e6abf941a6b2c2885228a8c659cef3094a @ChrisRackauckas

PR #2101, opened 2023-02-23
Change `show` format for `MatchedSystemStructure` and add DeepDiffs.jl support

topolarity:
This PR adds a "diff" view for `MatchedSystemStructure`, and modifies `show` to rely less on text coloring.

#### `Base.show(::IO, ::MatchedSystemStructure)`

  - **var ⇔ eq matches** - marked with (parentheses) (previously, yellow)
  - **(un)solvable vars/eqs** - printed in dark gray and white, respectively (previously, white/green)
  - **selected states (∫)** - the ∫ symbol is printed without coloring and positioned more prominently
  - **legend** - A one-line glossary for how to read the table is provided at the bottom.

The var/eq table also have an added separator, and the indices are repeated.

#### `deepdiff(::MatchedSystemStructure, ::MatchedSystemStructure)`

- Color-highlights any added/removed variables and equations
- Color-highlights changes to the dependency graph and the var ⇔ eq matching graph
- Does **not** highlight any differences in the solvable sub-graph (`old.structure.solvable_graph`)

### DIFF
--- a/src/bipartite_graph.jl
+++ b/src/bipartite_graph.jl
@@ -213,12 +213,14 @@ end
 struct HighlightInt
     i::Int
     highlight::Symbol
-    underline::Bool
+    match::Bool
 end
 Base.typeinfo_implicit(::Type{HighlightInt}) = true
 function Base.show(io::IO, hi::HighlightInt)
-    if hi.underline
-        printstyled(io, hi.i, color = hi.highlight, underline = true)
+    if hi.match
+        printstyled(io, "(", color = hi.highlight)
+        printstyled(io, hi.i, color = hi.highlight)
+        printstyled(io, ")", color = hi.highlight)
     else
         printstyled(io, hi.i, color = hi.highlight)
     end
@@ -227,6 +229,8 @@ end
 function Base.show(io::IO, l::BipartiteAdjacencyList)
     if l.match === true
         printstyled(io, "∫ ")
+    else
+        printstyled(io, "  ")
     end
     if l.u === nothing
         printstyled(io, '⋅', color = :light_black)
--- a/src/bipartite_graph.jl
+++ b/src/bipartite_graph.jl
@@ -212,21 +212,21 @@ end
 
 struct HighlightInt
     i::Int
-    highlight::Union{Symbol, Nothing}
+    highlight::Symbol
+    underline::Bool
 end
 Base.typeinfo_implicit(::Type{HighlightInt}) = true
-
 function Base.show(io::IO, hi::HighlightInt)
-    if hi.highlight !== nothing
-        printstyled(io, hi.i, color = hi.highlight)
+    if hi.underline
+        printstyled(io, hi.i, color = hi.highlight, underline = true)
     else
-        print(io, hi.i)
+        printstyled(io, hi.i, color = hi.highlight)
     end
 end
 
 function Base.show(io::IO, l::BipartiteAdjacencyList)
     if l.match === true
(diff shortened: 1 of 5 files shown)


src/systems/systemstructure.jl
```diff
@@ -475,6 +497,43 @@ function Base.show(io::IO, mime::MIME"text/plain", ms::MatchedSystemStructure)
...
+
+Prints a unified "diff" view of the provided `MatchedSystemStructure` objects.
+
+Color-highlights any added/removed variables and equations, as well as changes
+to their dependency graph and the var ⇔ eq matching graph.
+
+Note: This will not highlight any differences in the solvable sub-graph
+(`old.solvable_graph`).
+"""
+function show_with_compare(io::IO, new::MatchedSystemStructure, old::MatchedSystemStructure)
```
topolarity, line 511: Bikeshedding over this name is encouraged.

src/bipartite_graph.jl
```diff
@@ -210,23 +210,24 @@ function BipartiteAdjacencyList(u::Union{Vector{Int}, Nothing})
...
 function Base.show(io::IO, hi::HighlightInt)
-    if hi.highlight !== nothing
-        printstyled(io, hi.i, color = hi.highlight)
-    else
-        print(io, hi.i)
-    end
+    printstyled(io, hi.i, color = hi.highlight, underline = hi.underline)
 end
 
 function Base.show(io::IO, l::BipartiteAdjacencyList)
```
topolarity, line 228: Technically, this implementation could be replaced with just `Base.show(io::IO, l::BipartiteAdjacencyList) = Base.show(io, BipartiteAdjacencyListDiff(l, l))`

However, I've kept it for now, since that might be confusing.

Keno:
I liked the coloring in the non-diff view, because it lets you quickly scan for some important properties (how many selected states are there, which equations don't have a matching, etc). Perhaps we can keep the colors, but just free up green (I think I'm ok with the light gray change you did for that one, but I'll have to play with it). The diff view seems nice. I wonder whether this should just integrate with DeepDiffs though rather than invent a new generic function.

topolarity:
> I liked the coloring in the non-diff view, because it lets you quickly scan for some important properties (how many selected states are there, which equations don't have a matching, etc). Perhaps we can keep the colors, but just free up green (I think I'm ok with the light gray change you did for that one, but I'll have to play with it).

Maybe we should keep the yellow highlighting along with the underlining. Something like: image

That way, `show` gets highlighting, and the diff can still rely on the underline to show changes to matching. As for selected states, I find the ∫ fairly easy to scan for visually since it's always left-aligned, but I'm open to feedback on that.

> The diff view seems nice. I wonder whether this should just integrate with DeepDiffs though rather than invent a new generic function.

Sounds like a good idea to me.

ChrisRackauckas:
Needs docs.

staticfloat:
This looks good to me overall (it's quite beautiful!), I think it might be worthwhile to add a test to ensure that printing isn't broken?  So e.g. add `DeepDiff` to the [`[extras]` section of `Project.toml`](https://github.com/SciML/ModelingToolkit.jl/blob/master/Project.toml#L87-L108), and [the `tests` target](https://github.com/SciML/ModelingToolkit.jl/blob/master/Project.toml#L111), then add a few tests to take a matched system structure and diff it out to an `IOBuffer` with the [appropriate context set](https://docs.julialang.org/en/v1/base/io-network/#Base.IOContext-Tuple{IO,%20Pair})?

YingboMa (approved):

PR #2102, opened 2023-02-24
Optimize `linearize` by avoiding function generation

baggepinnen:
`process_DEProblem` generates the dynamics function, which is extremely costly. `linearize` only needed the `u0` and `p` parameters, so the function to compute those was factored out from `process_DEProblem` and reused in `linearize`.

This improves my benchmark for repeated linearization from
```julia
3.449409 seconds (36.37 M allocations: 1.803 GiB, 5.74% gc time) # before
0.743150 seconds (7.88 M allocations: 360.666 MiB, 8.23% gc time) # don't generate DE function
```

### DIFF
--- a/src/systems/diffeqs/abstractodesystem.jl
+++ b/src/systems/diffeqs/abstractodesystem.jl
@@ -570,6 +570,22 @@ function ODEFunctionExpr{iip}(sys::AbstractODESystem, dvs = states(sys),
     !linenumbers ? striplines(ex) : ex
 end
 
+function get_u0_p(sys, u0map, parammap; tofloat, use_union)
+    eqs = equations(sys)
+    dvs = states(sys)
+    ps = parameters(sys)
+    iv = get_iv(sys)
+
+    defs = defaults(sys)
+    defs = mergedefaults(defs, parammap, ps)
+    defs = mergedefaults(defs, u0map, dvs)
+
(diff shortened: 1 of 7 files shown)


src/systems/diffeqs/abstractodesystem.jl
```diff
@@ -570,6 +570,22 @@ function ODEFunctionExpr{iip}(sys::AbstractODESystem, dvs = states(sys),
     !linenumbers ? striplines(ex) : ex
 end
 
+function get_u0_p(sys, u0map, parammap; tofloat, use_union)
```
ChrisRackauckas, line 573: we should probably make this work on all sys?

baggepinnen: I'm not sure I understand what you mean? `OptimizationSystems` etc.?

ChrisRackauckas: Yes, they all do something similar in the same spot, so it would be good to keep the uniformity, and I think this would just be a uinformly useful utility.

baggepinnen: Yep, it seems the same code was run for Nonlinear and Discrete systems as well. I've updated to use the new function there as well

ChrisRackauckas: Add add it to the doc pages?

YingboMa (approved):

PR #2104, opened 2023-02-28
Add `at component`

YingboMa:


baggepinnen:
Maybe the docs should mention *why* `@component` is added to all those function definitions?

PR #2105, opened 2023-03-01
fix: drop the closure in `component_post_processing`

ven-k:
- This allows to define components with `f` as arg

Fixes this error:
```julia
julia> @component function A(; name, f)
           ODESystem(Equation[], t, [], []; name=name)
       end
ERROR: syntax: cannot add method to function argument f
Stacktrace:
 [1] top-level scope
   @ REPL[3]:1
```

### DIFF
--- a/src/systems/abstractsystem.jl
+++ b/src/systems/abstractsystem.jl
@@ -1130,10 +1130,7 @@ function component_post_processing(expr, isconnector)
 
     quote
         function $fname($(args...))
-            function f()
-                $body
-            end
-            res = f()
+            res = $body
             if $isdefined(res, :gui_metadata) && $getfield(res, :gui_metadata) === nothing
                 name = $(Meta.quot(fname))
                 if $isconnector


baggepinnen:
You could probably use `gensym` which is intended for exactly this?

src/systems/abstractsystem.jl
```diff
@@ -1128,12 +1128,14 @@ function component_post_processing(expr, isconnector)
...
+    rand_f = Symbol(randstring())
+
     quote
         function $fname($(args...))
-            function f()
+            function $rand_f()
                 $body
             end
-            res = f()
+            res = $rand_f()
```
ChrisRackauckas, line 1138: why is this good? Won't this cause recompilation when applied to the same model?

ven-k: Didn't consider that. I wanted it to pick a name that is unlikely to be passed as an argument.
gensym would be a better bet for that. But I'm wondering - that would run into same recompilation issue.

Perhaps `modelingtoolkit_f` is unlikely enough?

ven-k: Switched to using `gensym` to avoid any name clash.

src/systems/abstractsystem.jl
```diff
@@ -1130,10 +1130,10 @@ function component_post_processing(expr, isconnector)
 
     quote
         function $fname($(args...))
-            function f()
+            function component_f_nCX19jvoVRROsIPs()
                 $body
```
ChrisRackauckas, line 1134: wait, why is a closure needed at all?

ven-k: Actually yeah 😅

YingboMa:
> wait, why is a closure needed at all?

Because the body could contain explicit `return`

YingboMa (changes requested):
A closure is necessary for correctness

YingboMa:
superseded by #2107

PR #2107, opened 2023-03-04
Use an anonymous function to avoid naming conflict in macro

YingboMa:


ChrisRackauckas (approved):
much more sane.

YingboMa:
It has the bonus of adding a lot of punctuations.

ChrisRackauckas:
There are almost enough parenthesis to make it valid Lisp code.

YingboMa:
Test failure unrelated.

PR #2113, opened 2023-03-08
Typo, slight rephrase

KronosTheLate:


docs/src/examples/higher_order.md
```diff
@@ -26,15 +26,15 @@ eqs = [D(D(x)) ~ σ * (y - x),
...
+`D = Differential(t)^2` and then `D(x)` would be the second derivative,
 and this syntax extends to `N`-th order. Also, we can use `*` or `∘` to compose
 `Differential`s, like `Differential(t) * Differential(x)`.
 
 Now let's transform this into the `ODESystem` of first order components.
 We do this by simply calling `ode_order_lowering`:
 
 ```@example orderlowering
-sys = structural_simplify(sys)
+sys = ode_order_lowering(sys)
```
ChrisRackauckas, line 37: no you should just use `structural_simplify` for this these days

docs/src/examples/higher_order.md
```diff
@@ -26,15 +26,15 @@ eqs = [D(D(x)) ~ σ * (y - x),
 ```
 
 Note that we could've used an alternative syntax for 2nd order, i.e.
-`D = Differential(t)^2` and then `E(x)` would be the second derivative,
+`D = Differential(t)^2` and then `D(x)` would be the second derivative,
 and this syntax extends to `N`-th order. Also, we can use `*` or `∘` to compose
 `Differential`s, like `Differential(t) * Differential(x)`.
 
 Now let's transform this into the `ODESystem` of first order components.
 We do this by simply calling `ode_order_lowering`:
```
YingboMa, line 34: We should change this instead :p

docs/src/examples/higher_order.md
KronosTheLate: done

docs/src/examples/higher_order.md
KronosTheLate: I don't really get what you mean... 😅

KronosTheLate:
As a sidenote, if `sys = structural_simplify(sys)` is as common as it seems, perhaps it would make sense to make a mutating version `structural_simplify!`.

PR #2119, opened 2023-03-14
add symbolic linearization function

baggepinnen:
For simple systems, it's possible to perform linearization symbolically, this can be super handy at times.

This PR adds the function `linearize_symbolic` that behaves very similar to `linearize`, but returns symbolic matrices instead. For now, it has some known limitations, it can not handle `ifelse` statements appearing in the dynamics function. This can likely be fixed in the future by either
- allowing code gen to output code with `ifelse` statements intact rather than inserting branches
- Not relying on `generate_function` and instead generalizing the internals of `generate_function`  sufficiently to support this use case.

### DIFF
--- a/src/systems/abstractsystem.jl
+++ b/src/systems/abstractsystem.jl
@@ -1287,6 +1287,28 @@ function linearization_function(sys::AbstractSystem, inputs,
     return lin_fun, sys
 end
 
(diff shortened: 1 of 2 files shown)


baggepinnen:
The test failures seem unrelated to me

YingboMa (approved):

PR #2124, opened 2023-03-17
set tension correctly

oscardssmith:
This was causing the dae init to fail silently (which turns into a loud failure with https://github.com/SciML/OrdinaryDiffEq.jl/pull/1909).


oscardssmith:
Further investigation shows that this isn't MTKs fault but rather the fault of dae initialization.

PR #2125, opened 2023-03-20
Use bounded Ipopt_jll from AmplNLWriter

Vaibhavdixit02:


Project.toml
```diff
@@ -63,6 +63,7 @@ ForwardDiff = "0.10.3"
 FunctionWrappersWrappers = "0.1"
 Graphs = "1.5.2"
 IfElse = "0.1"
+Opopt_jll = "=300.1400.400"
```
Vaibhavdixit02, line 66:
```suggestion
Ipopt_jll = "=300.1400.400"
```

Project.toml
```diff
@@ -63,6 +63,7 @@ ForwardDiff = "0.10.3"
 FunctionWrappersWrappers = "0.1"
 Graphs = "1.5.2"
 IfElse = "0.1"
+Ipopt_jll = "=300.1400.400"
```
ChrisRackauckas, line 66: Why is this bound here in this repo?

Vaibhavdixit02: Because it is a test dependency here as well https://github.com/SciML/ModelingToolkit.jl/blob/master/Project.toml#L93

YingboMa: Can we just not use `Ipopt_jll` in the test?

Vaibhavdixit02: https://github.com/SciML/Optimization.jl/commit/88fc980a385c795cb0b7456e46151bdef9ecbafe I thought so too, but the tests were still failing here
https://github.com/SciML/ModelingToolkit.jl/actions/runs/4439811295/jobs/7852885734#step:6:70
https://github.com/SciML/ModelingToolkit.jl/actions/runs/4439811295/jobs/7852885734#step:6:92

ChrisRackauckas: Is it using the right version?

Vaibhavdixit02: https://github.com/SciML/ModelingToolkit.jl/actions/runs/4472354692/jobs/7858412913#step:6:331 no it's not

Issue #1266, opened 2021-09-16
export Interval, infimum, supremum


KirillZubov:
Add `export Interval, infimum, supremum`  in order to not import explicitly.

relate to https://github.com/SciML/NeuralPDE.jl/issues/319

PR #2128, opened 2023-03-22
fixes issue #1266

AstitvaAggarwal:
added exports


YingboMa (changes requested):
It doesn't feel like MTK should export those symbols

AstitvaAggarwal:
Are the exports faulty or not needed?

YingboMa:
Not needed.

PR #2137, opened 2023-04-05
fix edge cases in pantiledes

oscardssmith:
fix edge cases in pantiledes where there are edge cases like self loops in the stem. These occurred in systems that had aliases like
```
x->D(x)
```
or
```
x->y
D(x)->D(y)
x->D(y)
```
and would lead to the `@assert stem′ !== nothing` failing.

Mostly written by @keno

### DIFF
--- a/src/structural_transformation/pantelides.jl
+++ b/src/structural_transformation/pantelides.jl
@@ -126,10 +126,20 @@ function computed_highest_diff_variables(structure, ag::Union{AliasGraph, Nothin
                     var′ = invview(var_to_diff)[var]
                     var′ === nothing && break
                     stem′ = invview(var_to_diff)[stem]
+                    avar′ = haskey(ag, var′) ? ag[var′][2] : nothing
+                    if avar′ == stem || var′ == stem
+                        # If we have a self-loop in the stem, we could have the
+                        # var′ also alias to the original stem. In that case, the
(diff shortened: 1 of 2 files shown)


src/structural_transformation/pantelides.jl
```diff
@@ -126,10 +126,20 @@ function computed_highest_diff_variables(structure, ag::Union{AliasGraph, Nothin
                     var′ = invview(var_to_diff)[var]
                     var′ === nothing && break
                     stem′ = invview(var_to_diff)[stem]
+                    avar′ = haskey(ag, var′) ? ag[var′][2] : nothing
+                    if avar′ == stem || var′ == stem
```
YingboMa, line 130: Is this check sufficient to detect any loop?

oscardssmith: I've thought about this a lot, and it's not clear to me that it does, but I haven't yet been able to trip it up.

Keno: Upon further reflection, I don't believe this to be sufficient. This only catches first order loops, but higher order loops are possible.

YingboMa:
format

src/structural_transformation/pantelides.jl
YingboMa: It's probably not hard to check any loop.

PR #2149, opened 2023-04-24
fix all loops in alias graph rather than just first order loops

oscardssmith:
This is the followup to #2137 that is able to detect loops of higher orders.

### DIFF
--- a/src/structural_transformation/pantelides.jl
+++ b/src/structural_transformation/pantelides.jl
@@ -127,7 +127,7 @@ function computed_highest_diff_variables(structure, ag::Union{AliasGraph, Nothin
                 loop_found = false
                 var′ = invview(var_to_diff)[var]
                 while var′ !== nothing
-                    if var′ === stem || haskey(ag, var′) && ag[var′][2] == stem
+                    if var′ === stem || (haskey(ag, var′) && ag[var′][2] == stem)
                         dstem = var_to_diff[stem]
                         @assert dstem !== nothing
                         varwhitelist[dstem] = true
--- a/src/structural_transformation/pantelides.jl
+++ b/src/structural_transformation/pantelides.jl
@@ -127,7 +127,7 @@ function computed_highest_diff_variables(structure, ag::Union{AliasGraph, Nothin
                 loop_found = false
                 var′ = invview(var_to_diff)[var]
                 while var′ !== nothing
-                    if var′ === stem || (haskey(ag, var′) && ag[var′][2] == stem)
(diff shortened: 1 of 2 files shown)


src/structural_transformation/pantelides.jl
```diff
@@ -121,25 +121,33 @@ function computed_highest_diff_variables(structure, ag::Union{AliasGraph, Nothin
...
-                        # If we have a self-loop in the stem, we could have the
-                        # var′ also alias to the original stem. In that case, the
-                        # derivative of the stem is highest differentiated, because of the loop
+                # If we have a self-loop in the stem, we could have the
+                # var′ also alias to the original stem. In that case, the
+                # derivative of the stem is highest differentiated, because of the loop
+                loop_found = false
+                var′ = invview(var_to_diff)[var]
+                while var′ !== nothing
+                    if var′ === stem || haskey(ag, var′) && ag[var′][2] == stem
```
YingboMa, line 130: Maybe add a pair of parenthesis here. Mix || and && is pretty confusion.

YingboMa (commented):
Do you mean `ÿ` instead of ` ̈y`? It should be `y\ddot<tab>`.

src/structural_transformation/pantelides.jl
```diff
@@ -121,25 +121,33 @@ function computed_highest_diff_variables(structure, ag::Union{AliasGraph, Nothin
...
-                        # If we have a self-loop in the stem, we could have the
-                        # var′ also alias to the original stem. In that case, the
-                        # derivative of the stem is highest differentiated, because of the loop
+                # If we have a self-loop in the stem, we could have the
+                # var′ also alias to the original stem. In that case, the
+                # derivative of the stem is highest differentiated, because of the loop
+                loop_found = false
+                var′ = invview(var_to_diff)[var]
+                while var′ !== nothing
+                    if var′ === stem || (haskey(ag, var′) && ag[var′][2] == stem)
```
YingboMa, line 130: It should also be `var′ === stem` right?

oscardssmith: I don't think it matters. They're both `Int`, right?

YingboMa: It's a good way to get silent bugs when refactoring.

oscardssmith: fair enough.

YingboMa, line 130:
```suggestion
                    if var′ == stem || (haskey(ag, var′) && ag[var′][2] == stem)
```

PR #2152, opened 2023-04-25
missing variable defaults function

bradcarman:


src/systems/abstractsystem.jl
```diff
@@ -1756,3 +1756,30 @@ Base.:(∘)(sys1::AbstractSystem, sys2::AbstractSystem) = compose(sys1, sys2)
...
+    missingvars = setdiff(states(sys), keys(varmap))
+    ds = Pair[]
+
+    n = length(missingvars)
+
+    defaults = if default isa Vector
+        @assert length(default)==n "`default` size ($(length(default))) should match the number of missing variables: $n"
+        default
+    else
+        default * ones(typeof(default), n)
```
YingboMa, line 1777: I think we should add the branch inside the loop so that we can avoid this allocation.

YingboMa (approved):

PR #2160, opened 2023-05-04
Provide logical type guarantees for PDESystem fields

xtalax:


xtalax:
I don't like this solution, but I rely far too much on the assumption of `operation` giving me a unique component in MOL, and being able to call this object with any arguments such that doing this properly would be a complete rewrite.

Is there an obvious way to address this that I have missed?

EDIT: This is to allow `@variables u(..)[1:n]` to work vs `@variables u[1:n](..)`

xtalax:
Sorry about the format, I have no idea why it edited so many lines

ChrisRackauckas:
I don't think we should flatten here. If a solver doesn't support the structure, then it should flatten. But we shouldn't flatten for all possible solvers just because some don't support it right now.

xtalax:
Thing is this will soon be a requirement for much of the common parsing, and leaving it doesn't give us any important structure. I could move this to PDEBase though

xtalax:
Swerving to just provide the type guarantees for system elements

ChrisRackauckas:
> and leaving it doesn't give us any important structure

it should?

I mean, if the common parser needs to do this, then the common parser should just do it. Though we should work to retain structure when possible.

ChrisRackauckas (approved):

ChrisRackauckas:
Tests fail?

xtalax:
These failures are not related,  they don't even use PDESystem

YingboMa:
The format check doesn't pass either

YingboMa:
Maybe it's easier to start over now.

PR #2167, opened 2023-05-11
Work around Julia 1.9.0 regression

YingboMa:
Ref: https://github.com/JuliaLang/julia/issues/49766


src/structural_transformation/bareiss.jl
```diff
@@ -143,6 +143,9 @@ end
     V = M[(k + 1):end, (k + 1):end]
     V .= exactdiv.(V .* pivot .- M[(k + 1):end, k] * M[k, (k + 1):end]', prev_pivot)
     zero!(M, (k + 1):size(M, 1), k)
+    if M isa AbstractSparseMatrixCSC
+        dropzeros!(M)
```
Keno, line 147: Why is this required are we relying on structural rather than value comparisons somewhere?

YingboMa: There's another fun bug https://github.com/JuliaLang/julia/issues/49771

PR #2168, opened 2023-05-16
Alias fix

staticfloat:


staticfloat:
I'm converting this to a draft, as it's currently based on top of https://github.com/SciML/ModelingToolkit.jl/pull/2161 and it also needs tests.

src/systems/alias_elimination.jl
```diff
@@ -986,6 +986,12 @@ function locally_structure_simplify!(adj_row, pivot_var, ag)
     if pivot_var === nothing
         if iszero(nirreducible)
             zero!(adj_row)
+        elseif nirreducible == 1
+            zero!(adj_row)
+            ks = [k for (k, v) in pairs(nonzerosmap(adj_row)) if !iszero(v)]
```
YingboMa, line 991: Doesn't `zero!(adj_row)` make `ks` always empty? Also, note that there's a `dropzeros!(adj_row)` in the next branch. We could move the `nirreducible == 1 && length(SparseArrays.nonzeroinds(adj_row)) == 1` there for better efficiency.

staticfloat: Now that I understand this better, I think I can vastly simplify this.  I've pushed a new commit, see if you like that better?

src/systems/alias_elimination.jl
```diff
@@ -984,7 +984,7 @@ function locally_structure_simplify!(adj_row, pivot_var, ag)
     end
 
     if pivot_var === nothing
-        if iszero(nirreducible)
+        if iszero(nirreducible) || nirreducible == 1
```
YingboMa, line 987: I am less convinced that this is correct now. If you read the previous logic, the pivot variable has non-zero coefficient, and `nirreducible` is the number of non-trivial variables other than the pivot variable, so won't the code below handle this naturally? We should have a test case and make sure that this solves the problem.

staticfloat: Creating the testcase was significantly more annoying than I had hoped, but it is now integrated into `test/alias.jl`, and asserts that both `mm` results from `alias_eliminate_graph!()` maintain the property that all single-value-rows are properly eliminated to zero.  Removing the fix from here shows that the tests fail.

I'm more than happy to figure out a different way of solving this, hopefully with the test case reduced like this it is now much easier to fix.

oscardssmith:
Now that https://github.com/SciML/ModelingToolkit.jl/pull/2172 is merged, this can be closed, right?

PR #2175, opened 2023-05-30
Fix `structural_transformation` for optimization systems

YingboMa:
Fix https://github.com/SciML/Optimization.jl/pull/539

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -846,12 +846,13 @@ end
 # Symbolics needs to call unwrap on the substitution rules, but most of the time
 # we don't want to do that in MTK.
 const Eq = Union{Equation, Inequality}
-function fast_substitute(eq::Inequality, subs)
-    Inequality(fast_substitute(eq.lhs, subs), fast_substitute(eq.rhs, subs),
-               eq.relational_op)
-end
-function fast_substitute(eq::Equation, subs)
-    Equation(fast_substitute(eq.lhs, subs), fast_substitute(eq.rhs, subs))
+function fast_substitute(eq::Eq, subs)
+    if eq isa Inequality
+        Inequality(fast_substitute(eq.lhs, subs), fast_substitute(eq.rhs, subs),
+                   eq.relational_op)
+    else
+        Equation(fast_substitute(eq.lhs, subs), fast_substitute(eq.rhs, subs))
+    end
 end
 function fast_substitute(eq::T, subs::Pair) where {T <: Eq}
     T(fast_substitute(eq.lhs, subs), fast_substitute(eq.rhs, subs))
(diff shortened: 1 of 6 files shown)


src/utils.jl
```diff
@@ -845,13 +845,14 @@ end
 
 # Symbolics needs to call unwrap on the substitution rules, but most of the time
 # we don't want to do that in MTK.
-function fast_substitute(eq::Equation, subs)
-    fast_substitute(eq.lhs, subs) ~ fast_substitute(eq.rhs, subs)
+const Eq = Union{Equation, Inequality}
+function fast_substitute(eq::T, subs) where {T <: Eq}
+    T(fast_substitute(eq.lhs, subs), fast_substitute(eq.rhs, subs))
```
Vaibhavdixit02, line 850: This needs the `relational_op` argument as well in case of `Inequality`

Vaibhavdixit02:
I was able to use this branch (after making the change for `Inequality` in `fast_subsititutes`) on a MTK version https://github.com/Vaibhavdixit02/rosetta-opf/blob/main/mtk.jl of the opf problem

src/utils.jl
YingboMa: Can you suggest the change?

YingboMa: Also, why should we have `relational_op`? Shouldn't everything normalize to `≲` and define `a ≳ b` as `b ≲ a` due to reflexivity?

YingboMa: Are there any reasons to not enforce it? At least, we can normalize it in OptimizationProblem to save a lot of hassle.

Vaibhavdixit02: No, I don't see any reason not to do it

YingboMa: Cool. I will merge this PR if the tests pass. Do you mind to work on the normalization PR?

Vaibhavdixit02: So after looking through the constraints code a bit better it looks like we already convert to canonical form https://github.com/SciML/ModelingToolkit.jl/blob/e1a9a06ca1fd26679b792771ecf04e7c11dbe927/src/systems/optimization/constraints_system.jl#L110 and https://github.com/SciML/ModelingToolkit.jl/blob/e1a9a06ca1fd26679b792771ecf04e7c11dbe927/src/systems/optimization/constraints_system.jl#L180-L181 so I don't think we need to it

Vaibhavdixit02: What can `subs` be other than `Pair`?

YingboMa: Dict or Vector of pairs

YingboMa: OK, so undo that commit?

PR #2176, opened 2023-05-30
Enforce Vector inputs for PDESystem

xtalax:


src/systems/pde/pdesystem.jl
```diff
@@ -101,6 +101,15 @@ struct PDESystem <: ModelingToolkit.AbstractMultivariateSystem
         end
 
         eqs = eqs isa Vector ? eqs : [eqs]
+		bcs = bcs isa Vector ? bcs : [bcs]
```
ChrisRackauckas, line 104: formatting

PR #2187, opened 2023-06-07
Add back rank2 (highest diffed variables) in alias elimination

YingboMa:


YingboMa:
We want to pivot on highest differentiated variables to eliminate/simplify any equation that is not structurally visible.

staticfloat:
This looks good in our tests!

staticfloat:
@YingboMa we've got the MSL tests working more or less with this branch, so we're pretty happy with it.  It looks like there's a failure on a `dummy_derivatives()` test in `linearize.jl`, though.

YingboMa:
How do you know it's a dummy_derivatives test?

staticfloat:
> How do you know it's a dummy_derivatives test?

Because it was [this test](https://github.com/SciML/ModelingToolkit.jl/blob/master/test/linearize.jl#L229), and at the top of that file, it [mentions dummy_derivatives](https://github.com/SciML/ModelingToolkit.jl/blob/master/test/linearize.jl#L182).

YingboMa:
Ah, OK. I thought I missed something. It's a pretty complicated test, and failed for a different reason.

PR #2188, opened 2023-06-08
Support for observable serialization

bradcarman:


bradcarman:
Something seems to be wrong with the formatting, will close and try again...

PR #2192, opened 2023-06-11
Revive `pss_graph_modia!`

Keno:
When I originally wrote this code, I didn't really understand how it was supposed to work, simply copying the algorithm from modia. As a result, the algorithm was somewhat incomplete, and we ended up switching to dummy_derivative_graph! instead. However, we are now seeing issues where the state selection from `dummy_derivative_graph!` (ddg) is resulting in state selections that are numerically sub-optimal because ddg does not respect solvability. With the benefit of hindsight, fix pss_graph_modia! to be actually correct.

The keen eyed reader will notice that the changes to pss_graph_modia! are essentially exactly what ddg does right now, except that I don't currently implement the linear case. I think it would make sense to further unify this and fully switch over MTK to the new `pss_graph_modia!`, but past experience has shown the MTK/Downstream tests to be quite sensitive to the exact state selection, so this does not attempt to make that change yet.

### DIFF
--- a/src/structural_transformation/partial_state_selection.jl
+++ b/src/structural_transformation/partial_state_selection.jl
@@ -27,36 +27,40 @@ function ascend_dg_all(xs, dg, level, maxlevel)
     return r
 end
 
-function pss_graph_modia!(structure::SystemStructure, var_eq_matching, varlevel,
+function pss_graph_modia!(structure::SystemStructure, maximal_top_matching, varlevel,
     inv_varlevel, inv_eqlevel)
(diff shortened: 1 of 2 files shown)


YingboMa:
Can we add this as an option in ddg? It seems like a different priority for nonlinear equations.

Keno:
Yes, we should unify these. However, I'd like to do that in a separate PR. This function still needs support for the linear exact jacobian that ddg has, so I'd just like to do that all at once.

YingboMa (approved):

PR #2196, opened 2023-06-15
Pass arguments (including hierarchal ones) to `Model`s

ven-k:
- The kwargs and args can be specified for a Model.
- It checks the `@components` for respective kwargs/args and sets up necessary ways to access them with a `.`
This works:
```julia
@model A(; cval=1, kval) begin                       
                  @components begin                         
                      mb = B(; t3 = 1, t4 = t4val)          
                  end                                       
              @parameters begin                             
                     c(t) = cval                            
                     k(t) = kval                            
              end                                           
       end

 @named ma = A(; mb.t4 = 100, kval = 20)
```

-----
While here, fixes following conditions:
```
@variables begin
    a(t), [description = ""]
    b(t) = bval, [description = ""]
    c(t) = 1, [description = ""]
end
```

### DIFF
--- a/src/systems/model_parsing.jl
+++ b/src/systems/model_parsing.jl
@@ -48,68 +48,36 @@ function connector_macro(mod, name, body)
 end
 
 function parse_variable_def!(dict, mod, arg, varclass)
+    arg isa LineNumberNode && return
     MLStyle.@match arg begin
-        ::Symbol => generate_var!(dict, arg, varclass)
-        Expr(:call, a, b) => generate_var!(dict, a, b, varclass)
+        ::Symbol => (generate_var!(dict, arg, varclass), nothing)
+        Expr(:call, a, b) => (generate_var!(dict, a, b, varclass), nothing)
         Expr(:(=), a, b) => begin
-            var = parse_variable_def!(dict, mod, a, varclass)
+            var, _ = parse_variable_def!(dict, mod, a, varclass)
             def, meta = parse_default(mod, b)
             dict[varclass][getname(var)][:default] = def
-            var = setdefault(var, def)
+            if typeof(def) != Symbol
+                var = setdefault(var, def)
+                def = nothing
+            end
             if !isnothing(meta)
                 if (ct = get(meta, VariableConnectType, nothing)) !== nothing
                     dict[varclass][getname(var)][:connection_type] = nameof(ct)
                 end
                 var = set_var_metadata(var, meta)
             end
-            var
+            (var, def)
         end
         Expr(:tuple, a, b) => begin
-            var = parse_variable_def!(dict, mod, a, varclass)
+            var, _ = parse_variable_def!(dict, mod, a, varclass)
             meta = parse_metadata(mod, b)
             if (ct = get(meta, VariableConnectType, nothing)) !== nothing
                 dict[varclass][getname(var)][:connection_type] = nameof(ct)
             end
-            set_var_metadata(var, meta)
+            (set_var_metadata(var, meta), nothing)
         end
         _ => error("$arg cannot be parsed")
     end
-end
-
-function parse_variables_with_kw!(exprs, var, dict, mod, body, varexpr, varclass, kwargs)
-    for arg in body.args
-        arg isa LineNumberNode && continue
-        MLStyle.@match arg begin
-            ::Symbol || Expr(:tuple, a, b) || Expr(:call, a, b) || Expr(:(=), a, b::Number) => begin
-                parse_variables!(exprs, var, dict, mod, arg, varclass)
-            end
-            Expr(:(=), a, b::Symbol) => begin
-                isdefined(mod, b) ?
-                parse_variables!(exprs, var, dict, mod, arg, varclass) :
-                push!(varexpr.args[end].args[end].args, arg)
-            end
-            Expr(:(=), a, b) => begin
-                def = Base.remove_linenums!(b)
-                MLStyle.@match def begin
-                    Expr(:tuple, x::Symbol, y) => begin
-                        isdefined(mod, x) ?
-                        parse_variables!(exprs, var, dict, mod, arg, varclass) :
-                        push!(varexpr.args[end].args[end].args, arg)
-                    end
-                    ::Symbol => push!(varexpr.args[end].args[end].args,
-                        :($a = $(def.args[end])))
-                    ::Number || Expr(:tuple, x::Number, y) => begin
-                        parse_variables!(exprs, var, dict, mod, arg, varclass)
-                    end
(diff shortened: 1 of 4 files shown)


src/systems/model_parsing.jl
```diff
@@ -111,68 +177,101 @@ function parse_default(mod, a)
...
-    vs = Symbol[]
-    ps = Symbol[]
     eqs = Expr[]
     icon = Ref{Union{String, URI}}()
+    vs = []
+    ps = []
+    parexpr = :(pss = @parameters begin
+        end)
+    varexpr = :(vss = @variables begin
+        end)
```
YingboMa, line 230: Why should we generate more macros? We are already parsing them in the variable and parameter parsing part.

src/systems/model_parsing.jl
```diff
@@ -204,6 +296,44 @@ function parse_components!(exprs, cs, dict, body)
     dict[:components] = comps
 end
 
+function _rename(compname, varname)
+    compname = Symbol(compname, :__, varname)
+end
+
+function component_args!(a, b, expr, kwargs)
+    start = b.head == :parameters ? 1 : 2
```
YingboMa, line 304: Could you add a comment on this?

src/systems/model_parsing.jl
```diff
@@ -132,47 +191,77 @@ macro model(name::Symbol, expr)
...
 
-function model_macro(mod, name, expr)
+@inline is_kwarg(::Symbol) = false
+@inline is_kwarg(e::Expr) = (e.head == :parameters)
+
+macro model(fcall::Expr, expr)
+    fcall.head == :call || "Couldn't comprehend the model $arg"
+
+    arglist, kwargs = if lastindex(fcall.args) > 1 && is_kwarg(fcall.args[2])
+        (lastindex(fcall.args) > 2 ? (@info 1; Set(fcall.args[3:end])) : Set()), Set(fcall.args[2].args)
```
YingboMa, line 201: Extra debug statement.

test/model_parsing.jl
```diff
@@ -10,12 +10,13 @@ end
 @connector RealOutput begin
     u(t), [output = true]
 end
-@model Constant begin
+
+@model Constant(; k = 1) begin
```
YingboMa, line 14: Should we have explicit parameters like this? I think `@parameters` and `@variables` should automatically generate keyword arguments. So,
```
@parameters begin
    k = 1
end
```
alone should generate the keyword argument `k = 1` in `Constant.`

src/systems/model_parsing.jl
ven-k: For cases like `k(t) = kval` (where `kval` is passed via kwargs), it is kept as an expression, and evaluated while executing.
```
@model A(; kval = 1) begin
    @variables begin
       k(t) = kval
    end
end
```
Otherwise here, the default value of `k(t)` would be set to 1 and passed `kval` would have no effect.

`parexpr` and `varexpr` are populated with the parameters and variables that expect a value from kwargs. (The rest are part of vs and ps)

test/model_parsing.jl
ven-k: Till this commit: https://github.com/SciML/ModelingToolkit.jl/pull/2196/commits/39845a8c7c4c7f6f8ebf015601f74a7f43fff55f for something like `k = kval`, kval would be automatically promoted to a kwarg; it is not hard to add a step to parse `k = 1`  as expect k (with default value 1) in kwargs.

However, I feel it is very restrictive.
```
@model A(; k1 = 1, k2 = 2, T) begin
    begin
        kval = k1 + k2
    end

    @parameters begin
        k = kval
        R = rho * l / A # Electrical resistance of eddy current path
        # https://github.com/SciML/ModelingToolkitStandardLibrary.jl/blob/main/src/Magnetic/FluxTubes/basic.jl#L146
    end
end
```
`R = rho * l / A` could be parsed for symbols and promote them as kwargs; however `k = kval`-like equations are tricky.


And for cases like this[^1] where `eqs = [ D(y) ~ max(min((u - y) / Td, rising), falling) ]` # intricate equation with std vars, we'll have parse it and figure out if it is from a component/@-variables or simply a const input.

[^1]:https://github.com/SciML/ModelingToolkitStandardLibrary.jl/blob/main/src/Blocks/nonlinear.jl#L102

src/systems/model_parsing.jl
ven-k: Resolved with: https://github.com/SciML/ModelingToolkit.jl/pull/2196/files#diff-e4267563028a9bba382f312bc5bdff2b5d737c02f2141cfe4b358b1c1396109aR298-R299

src/systems/model_parsing.jl
ven-k: Resolved with: https://github.com/SciML/ModelingToolkit.jl/pull/2196/files#diff-e4267563028a9bba382f312bc5bdff2b5d737c02f2141cfe4b358b1c1396109aR194

src/systems/model_parsing.jl
ven-k: By setting defaults on the fly (whenever they are passed by kwargs) here: https://github.com/SciML/ModelingToolkit.jl/pull/2196/commits/9b2599337b67d2c96c06b04758869f9b92f1633a generating more macros is avoided now.

test/model_parsing.jl
```diff
@@ -110,10 +113,10 @@ end
 
 @model RC begin
```
ChrisRackauckas, line 114: This wouldn't actually be an `@model`, but I assume this is just for testing purposes.

test/model_parsing.jl
```diff
@@ -137,3 +144,40 @@ end
...
+@test hasmetadata(model.j, VariableDescription)
+@test hasmetadata(model.k, VariableDescription)
+
+@test getdefault(model.cval) == 1
+@test getdefault(model.c) == 2
+@test getdefault(model.d) == 2
+@test_throws KeyError getdefault(model.e)
+@test getdefault(model.f) == 3
+@test getdefault(model.i) == 4
+@test getdefault(model.j) == :jval
```
YingboMa, line 182: This isn't a valid default. Shouldn't it be an actual symbolic variable?

test/model_parsing.jl
```diff
@@ -137,3 +144,40 @@ end
...
+@test getdefault(model.k) == kval
```
YingboMa, line 183: The default should be symbolic here as well.

test/model_parsing.jl
YingboMa: Yeah

YingboMa (approved):

PR #2199, opened 2023-06-26
add input signals to `build_explicit_observed_function`

baggepinnen:
The PR adds the possibility to specify input signals when building an observed function, this is useful for control systems. The resulting function signature will be
```math
y = h(x, u, p, t)
```
where $u$ are the inputs.

If this looks reasonable, I can add a few tests.

### DIFF
--- a/src/systems/diffeqs/odesystem.jl
+++ b/src/systems/diffeqs/odesystem.jl
@@ -304,6 +304,7 @@ Build the observed function assuming the observed equations are all explicit,
 i.e. there are no cycles.
 """
 function build_explicit_observed_function(sys, ts;
+    inputs = nothing,
     expression = false,
     output_type = Array,
     checkbounds = true,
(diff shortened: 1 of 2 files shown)


src/systems/diffeqs/odesystem.jl
```diff
@@ -304,6 +304,7 @@ Build the observed function assuming the observed equations are all explicit,
 i.e. there are no cycles.
 """
 function build_explicit_observed_function(sys, ts;
+    inputs = nothing,
```
ChrisRackauckas, line 307: why would we not include them by default?

baggepinnen: Only control systems have inputs in the sense implied here, so this functionality would only be relevant when you're building observed functions for things like the measurement/performance equation for MPC controllers etc. 

When you simulate a model the standard way, it will always be an autonomous system without inputs

ChrisRackauckas:
Looks reasonable. Add a few tests.

baggepinnen:
The tests are failing on v1.6 due to RGF emitting an opaque closure. I deactivated the added tests on 1.6
```
LoadError: The function body AST defined by this @generated function is not pure. This likely means it contains a closure or comprehension.
  Stacktrace:
    [1] (::RuntimeGeneratedFunctions.RuntimeGeneratedFunction{(Symbol("##arg#15198614511220004104"), Symbol("##arg#9481781105307072587"), Symbol("##arg#1087180715767745052"), :t), ModelingToolkit.var"#_RGF_ModTag", ModelingToolkit.var"#_RGF_ModTag", (0x01a3dfe2, 0xf37029e8, 0xc2df25d1, 0xb0e6e8c4, 0x919d8e05), Expr})(::Vector{Float64}, ::Vector{Float64}, ::Vector{Real}, ::Int64)
      @ RuntimeGeneratedFunctions ~/.julia/packages/RuntimeGeneratedFunctions/b1u1T/src/RuntimeGeneratedFunctions.jl:128
    [2] top-level scope
      @ ~/work/ModelingToolkit.jl/ModelingToolkit.jl/test/input_output_handling.jl:148
```

YingboMa:
Why would it generate a closure at all?

baggepinnen:
No idea, but it did generate a closure that accessed the input. The generated code is working correctly though.

PR #2204, opened 2023-07-04
Add `isconnector` to `Model` type + add all metadata to `Model.structure` 

ven-k:
- While here, changes, the typeof `Model.structure[:components]` as Vector{Symbols} to match the rest.
- Adds the test to default value of an extended component (here Capacitor)

### DIFF
--- a/test/model_parsing.jl
+++ b/test/model_parsing.jl
@@ -33,6 +33,7 @@ end
 
 @named p = Pin(; v = π)
 @test getdefault(p.v) == π
+@test Pin.isconnector == true
 
 @mtkmodel OnePort begin
     @components begin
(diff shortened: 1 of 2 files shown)


test/model_parsing.jl
```diff
@@ -212,3 +216,20 @@ getdefault(a.b.k) == 1
 getdefault(a.b.i) == 20
 getdefault(a.b.j) == 30
 getdefault(a.b.k) == 40
+
+metadata = Dict(:description => "Variable to test metadata in the Model.structure",
+    :input => true, :bounds => :((-1, 1)), :connection_type => :Flow, :integer => true,
+    :binary => false, :tunable => false, :disturbance => true, :dist => :(Normal(1, 1)))
+
+@connector MockMeta begin
+    function m(t)
```
YingboMa, line 225: Should a connector ever take a function?

YingboMa:
Format

YingboMa (approved):

PR #2220, opened 2023-08-04
ODEProblem Parameter type for driving models with data

bradcarman:
Currently using the parameter space for a mix of single values and vectors is not so easy using ModelingToolkit.  See the `SampleData` component for example (https://docs.sciml.ai/ModelingToolkitStandardLibrary/stable/tutorials/input_component/).  The code for this component needs to explicitly convert the parameter vector to the `Parameter` type for the `SampleData` component and remake the ODEProblem...

```julia
    p = Parameter.(ModelingToolkit.varmap_to_vars(defs, parameters(sys); tofloat = false))
    remake(prob; p)
```

This code is now simply...

```julia
prob = ODEProblem(sys, [], (0.0, t_end))
```

To achieve this the `Parameter{T}` type is moved from ModelingToolkitStandardLibrary.jl to ModelingToolkit.jl

### DIFF
--- a/src/parameters.jl
+++ b/src/parameters.jl
@@ -118,9 +118,9 @@ Base.isless(y::Number, x::Parameter) = Base.isless(y, x.ref)
 
 Base.copy(x::Parameter{T}) where {T} = Parameter{T}(copy(x.data), x.ref)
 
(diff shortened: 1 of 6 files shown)


YingboMa:
This is superseded by #2231

PR #2222, opened 2023-08-08
Recompile Invalidations

xtalax:
Use PrecompileTools to recompile invalidations


YingboMa:
We really should not merge any PR with a format-check CI failure. It makes master not formatted properly, so all other PR's format check will fail as well.

YingboMa:
Well, in this instance, the formatter fails because of a trailing space, and it should be fixed. https://github.com/SciML/ModelingToolkit.jl/actions/runs/6146959683/job/16677371720#step:5:22

MTK's format CI always passes before this PR, and there's not much reason to break that.

PR #2226, opened 2023-08-14
Symbolic defaults for sub-components and base system kwargs + set base sys kwargs as component's kwargs 

ven-k:
1.  Kwargs of sub-components and base  systems (in `extend`) can now have symbolic defaults. This allows scenarios like:
```julia
@mtkmodel A begin
    @parameters begin
            a1 = 1
    end
    @components begin
           bb = B(; b1 = a1)
    end
end
```
Earlier, this would be `A(; a1 = nothing, B.b1 = a1)`, where b1 would have default value set as a1 arguement.
Now, it is `A(; a1 = nothing, B.b1 = nothing)` and internally, default value of b1 is set to parameter a1.

2. While **extending** components, the kwargs of base system are promoted as kwargs of the main sys.
```
@mtkmodel D begin
     @extend e1, e2 = ee = E(; ek)
end
```
Earlier D would have a kwarg `ee__ek` (as if `ee` was a subcomponent). Now D will have `ek` as the kwarg. This avoids having to access partial components of a model to set parameters. 

The model-parsing tests are updated to cover both the above changes.

### DIFF
--- a/src/systems/model_parsing.jl
+++ b/src/systems/model_parsing.jl
@@ -236,7 +236,8 @@ end
 
 function parse_components!(exprs, cs, dict, body, kwargs)
     expr = Expr(:block)
-    push!(exprs, expr)
+    varexpr = Expr(:block)
+    push!(exprs, varexpr)
     comps = Vector{Symbol}[]
     for arg in body.args
         arg isa LineNumberNode && continue
@@ -247,15 +248,17 @@ function parse_components!(exprs, cs, dict, body, kwargs)
                 arg = deepcopy(arg)
                 b = deepcopy(arg.args[2])
 
-                component_args!(a, b, expr, kwargs)
+                component_args!(a, b, dict, expr, varexpr, kwargs)
 
-                push!(b.args, Expr(:kw, :name, Meta.quot(a)))
                 arg.args[2] = b
                 push!(expr.args, arg)
             end
             _ => error("`@components` only takes assignment expressions. Got $arg")
         end
     end
+
+    push!(exprs, :(@named $expr))
+
     dict[:components] = comps
 end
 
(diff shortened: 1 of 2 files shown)


ven-k:
The Ai4EComponentLib, StructuralIdentifiability integration test failures are unrelated to the changes introduced here.

MTK test failure and MSL integration test failure is fixed by https://github.com/SciML/ModelingToolkitStandardLibrary.jl/pull/209. However that needs this to be merged and released.

This is ready for review @YingboMa @ChrisRackauckas.

baggepinnen:
In your first example, could the default value for `b1` be an expression (not just a parameter), like
```julia
B(; b1 = a1^2)
```
?

Thanks for this PR, I have been missing this feature :)

ven-k:
Expressions without conditional part work. (I'm working on the latter in a follow up PR)

So `B(; b1 = a1^2)` gives:
```julia
julia> @named aa = A()
Model aa with 0 equations
States (0):
Parameters (2):
  a1 [defaults to 1]
  bb₊b1 [defaults to a1^2]
```

Something like `B(; b1 = true ? a1 : 1)` doesn't work yet.

ven-k:
Additional to above changes, a new `@structural_parameters` is added to `@mtkmodel`. This allows arguments that aren't either parameters or variables, aka boolean flags, functions etc...
```
@mtkmodel A begin
    @structural_parameters begin
         a = 1
         b
    end
    @parameters begin
        c = 0
    end
end
```

`A`'s arg list will be: `A(; name, a = 1, b, c = nothing)`.

ven-k:
- And note that any test involving MSL fail for now and will need https://github.com/SciML/ModelingToolkitStandardLibrary.jl/pull/209 (which updates MSL with the changes introduced by this PR)

YingboMa (approved):

PR #2237, opened 2023-08-22
Promote `vs` to a `Union{Bool, ...}` whenever a Boolean value is present

ven-k:
- With MTKModel macro all input vars are parameters. This introduces, boolean values to `vs`.

Ref: https://github.com/SciML/ModelingToolkitStandardLibrary.jl/issues/213


YingboMa:
The root issue is fixed by #2231 .

Issue #2239, opened 2023-08-25
Namespace issues in `remake`


martincornejo:
Related to https://github.com/SciML/ModelingToolkit.jl/issues/1746

`remake`ing the problem using the wrong indexing (`sys.x` instead of `x`) leads to udesired behavior. In this case, it's also problematic because it doesn't throw an error, but silently continues with the defaults...

MWE:
```julia
@variables t
D = Differential(t)
@variables x(t) = 1.0
eqs = [D(x) ~ -x]
@named sys = ODESystem(eqs, t)

tspan = (0.0, 1.0)
prob = ODEProblem(sys, u0, tspan)

prob.u0
# 1-element Vector{Float64}:
#  1.0

## incorrectly change initial conditions
u0 = [sys.x => 2.0]
prob = remake(prob; u0)

prob.u0
# 1-element Vector{Float64}:
#  1.0

## correctly change initial conditions
u0 = [x => 2.0]
prob = remake(prob; u0)

prob.u0
# 1-element Vector{Float64}:
#  2.0
```

ChrisRackauckas:
@TorkelE

TorkelE:
I think this is a known issue? A while ago we fixed quite a few of these cases, but don't think we ver got to remake. It is something I'm happy to see fixed/fix though, as I agree it would be useful to have it working.

TorkelE:
At it. The problem is that currently, `remake` calls `ODEProblem`, which creaets a new `ODEProblem` using the input. However, `ODEProblem`s does not support the notation `ODEProblem(sys, [sys.x => 3.0], tspan).u0`. We could change this, or make `remake` throw an error if the input is of a certain type.

TorkelE:
The MTK PR have tests now as well.

For this example, basically what happens here is that when one calls `sys.x` one basically get a `Num` of the form `sys₊x(t)`, which is not recognised as one of the variables in `sys` (which only have a variable `x(t)`. The resulting phenomenon is the same as if you'd had used
```julia
@variables y
u0 = [y => 5.0]
prob = remake(prob; u0)
prob.u0
```
(with the exception that should be clear to the user that it does not work)

PR #2240, opened 2023-08-25
Enable stuff like :x and sys.x in ODEProblem

TorkelE:
This shoudl all work now:
```julia
using ModelingToolkit, SciMLBase
@variables t
D = Differential(t)
@variables x(t) = 1.0
eqs = [D(x) ~ -x]
@named sys = ODESystem(eqs, t)

@variables y(t)

tspan = (0.0, 1.0)
prob = ODEProblem(sys, [x => 3.0], tspan)

ODEProblem(sys, [sys.x => 3.0], tspan).u0    # Did not work previously
ODEProblem(sys, [:x => 3.0], tspan).u0         # Did not work previously
ODEProblem(sys, [x => 3.0], tspan).u0
```
This might not solve some cases of `remake` though, as I need to check how it reroutes through `ODEProblem`. I probably should also add some additional tests.

### DIFF
--- a/src/systems/diffeqs/abstractodesystem.jl
+++ b/src/systems/diffeqs/abstractodesystem.jl
@@ -896,6 +896,8 @@ function DiffEqBase.DAEProblem{iip}(sys::AbstractODESystem, du0map, u0map, tspan
     parammap = DiffEqBase.NullParameters();
     check_length = true, kwargs...) where {iip}
     has_difference = any(isdifferenceeq, equations(sys))
+    u0map = symmap_to_varmap(sys, u0map)
+    parammap = symmap_to_varmap(sys, parammap)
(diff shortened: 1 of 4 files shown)


TorkelE:
I tried a formating run, but it changes like 100 files, so I decided not to use it...

TorkelE:
This is intended to start fixing issues as seen in https://github.com/SciML/ModelingToolkit.jl/issues/2239. Tests now added, should sort out the `XProblem` part of the isseu.

YingboMa (changes requested):
I don't think we should go with this approach. Using `complete` is the right fix. If this is a frontend issue, we should fix the frontend instead of trying to duct tape every single function that can possibly take a symbolic variable.

PR #2246, opened 2023-08-31
Add support for parameters, equations... for `@connector`s

ven-k:
- `@connector` now supports `@parameters`, `@equations`, `@extend`, `@variables`, `@components` and a `begin...end` block along with variables specified without a begin block.
- This is useful for connectors like `HydraulicFluid`[^1]
- Replaces internal functions `connect_macro` and `mtkmodel_macro` with `_model_macro`, which handles both based on the `isconnector` flag
- Adds `"Connector with parameters, equations..." ` for testing the enhancement.

[^1]:https://github.com/SciML/ModelingToolkitStandardLibrary.jl/blob/main/src/Hydraulic/IsothermalCompressible/utils.jl#L53-L75

### DIFF
--- a/src/systems/model_parsing.jl
+++ b/src/systems/model_parsing.jl
@@ -6,58 +6,64 @@ end
 (m::Model)(args...; kw...) = m.f(args...; kw...)
 
 for f in (:connector, :mtkmodel)
+    isconnector = f == :connector ? true : false
     @eval begin
         macro $f(name::Symbol, body)
-            esc($(Symbol(f, :_macro))(__module__, name, body))
+            esc($(:_model_macro)(__module__, name, body, $isconnector))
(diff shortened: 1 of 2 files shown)


YingboMa (approved):

ven-k:
Additionally, generated function `<Model>.f` will have an identifiable  name `__<Model>__`. This will make it easier to identify the components in the stack trace.

```julia
julia> @mtkmodel A begin
       end
ModelingToolkit.Model{typeof(__A__), Dict{Symbol, Any}}(__A__, Dict{Symbol, Any}(:kwargs => Dict{Symbol, Any}(), :independent_variable => t), false)

julia> A.f
__A__ (generic function with 1 method)
```

PR #2250, opened 2023-09-06
Add a check that user-provided initial conditions are respected

baggepinnen:
Previously, initial conditions provided by the user were silently ignored if they applied to variables that were moved to observed. This PR throws an error in that case, informing the user that their IC cannot be met.

In the future, it would be good if MTK
1. Distinguished between default IC and user-provided IC
2. Could solve for an initial condition that respects IC provided despite some variables with specified IC having been moved to observed variables


baggepinnen:
@YingboMa the failing test indicates that my usage of the `observed` function is not correct in all circumstances. Could you spot what's wrong with the way I use it?

PR #2257, opened 2023-09-12
Implement `substitute` for `AbstractSystem`s

YingboMa:


baggepinnen:
How would this work if I want to substitute a subsystem that is deeper than level one? For example, if I want to substitute 
```
addPI = Blocks.Product() # My new component
substitute(model, Dict(pi_controller.addPI => addPI))
```

YingboMa:
Just `:addPI => addPI` because it's recursive.

baggepinnen:
So if there's more than one PI controller in the model, it will change all of them? What if I only want to change one of them?

YingboMa:
This function is not created for that use case.

baggepinnen:
In that case, there should probably by something more than just name that identifies the components to substitute since using the name alone could easily lead to problems if the replaced name is used more than once, for potentially different component types, by multiple different component authors.

ChrisRackauckas:
The name cannot be used more than once, and component authors don't specify the name. They are set when defining the instance (via `@named`) and its local to the user. Indeed if there's name collusions we should (and do?) error.

I think what you're getting at though is that we should have a form of typing on components, so we can identify all of the components that are the same type. That's a wholly separate thing though which would be nice to have for other reasons as well.

ChrisRackauckas:
That's a good point to clarify.

YingboMa:
Yes, those are valid concerns, but they are just not what this function does. `substitute` works recursively, and if you want to only replace one specific model in a specific level, we need to implement a different function.

PR #2258, opened 2023-09-14
Remove `@named` explanation

YingboMa:


ChrisRackauckas:
Add a note on how naming is done to the internals documentation?

YingboMa:
It's in `@named` macro docs already.

PR #2259, opened 2023-09-14
add module to RuntimeGeneratedFunctions call

xtalax:
Again I think my formatter is broken as it keeps trying to format your stuff, don't know what's going on here

### DIFF
--- a/src/systems/pde/pdesystem.jl
+++ b/src/systems/pde/pdesystem.jl
@@ -94,6 +94,7 @@ struct PDESystem <: ModelingToolkit.AbstractMultivariateSystem
         analytic = nothing,
         analytic_func = nothing,
         gui_metadata = nothing,
+		eval_module = @__MODULE__,
         checks::Union{Bool, Int} = true,
         name)
         if checks == true || (checks & CheckUnits) > 0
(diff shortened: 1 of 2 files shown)


ChrisRackauckas (approved):

src/systems/pde/pdesystem.jl
```diff
@@ -108,16 +108,16 @@ struct PDESystem <: ModelingToolkit.AbstractMultivariateSystem
...
-                end
-            end
-        end
+			if isnothing(analytic_func)
+				analytic_func = map(analytic) do eq
+					args = arguments(eq.lhs)
+					p = ps isa SciMLBase.NullParameters ? [] : map(a -> a.first, ps)
+					args = vcat(DestructuredArgs(p), args)
+					ex = Func(args, [], eq.rhs) |> toexpr
+					eq.lhs => drop_expr(RuntimeGeneratedFunction(@__MODULE__, @__MODULE__, ex))
```
YingboMa, line 117: This should not hardcoded. Also, I think only the first argument is necessary for correct precompilation caching.

xtalax: This actually doesn't fix the bug I have, I may need to pass down @__MODULE__ from PDESysLibrary

YingboMa:
How do you run your formatter? Also, do you have a forced CRLF setting if you are using Windows?

src/systems/pde/pdesystem.jl
YingboMa: You can use the `@RuntimeGeneratedFunction ` macro which only needs two.

xtalax:
It appears to run automatically, I am on MacOS

YingboMa:
Have you tried to run it in a vanilla Julia REPL that you start from a terminal?

PR #2270, opened 2023-09-18
skip ci.yml for doc pr

ArnoStrouwen:
Is it worth adding these lines into all the SciML repos @ChrisRackauckas ?
If so, I'll add them to all the Documenter 1.0 upgrades.


ChrisRackauckas:
Alright, that sounds good.

PR #2279, opened 2023-09-22
Refactor SDEProblem constructor

ErikQQY:
Now the SDE constructor is simply ```SDEProblem(SDEFunction(f, g), u0, tspan)``` or ```SDEProblem(f, g, u0, tspan)```

https://github.com/SciML/SciMLBase.jl/pull/489


ChrisRackauckas:
There's deadlock so I'm going break the deadlock and see in the StochasticDiffEq.jl tests what happens.

PR #2281, opened 2023-09-25
Fix SDEProblem constructor issue

ErikQQY:
Should fix the problems in StochasticDiffEq

### DIFF
--- a/src/systems/diffeqs/sdesystem.jl
+++ b/src/systems/diffeqs/sdesystem.jl
@@ -583,7 +583,7 @@ function DiffEqBase.SDEProblem{iip}(sys::SDESystem, u0map = [], tspan = get_tspa
         noise_rate_prototype = zeros(eltype(u0), size(noiseeqs))
     end
 
-    SDEProblem{iip}(f, u0, tspan, p; callback = cbs,
+    SDEProblem{iip}(f, f.g, u0, tspan, p; callback = cbs,
         noise_rate_prototype = noise_rate_prototype, kwargs...)
 end
 
@@ -648,7 +648,7 @@ function SDEProblemExpr{iip}(sys::SDESystem, u0map, tspan,
         tspan = $tspan
         p = $p
         noise_rate_prototype = $noise_rate_prototype
-        SDEProblem(f, u0, tspan, p; noise_rate_prototype = noise_rate_prototype,
+        SDEProblem(f, f.g, u0, tspan, p; noise_rate_prototype = noise_rate_prototype,
             $(kwargs...))
     end
     !linenumbers ? striplines(ex) : ex


src/systems/diffeqs/sdesystem.jl
```diff
@@ -583,7 +583,7 @@ function DiffEqBase.SDEProblem{iip}(sys::SDESystem, u0map = [], tspan = get_tspa
         noise_rate_prototype = zeros(eltype(u0), size(noiseeqs))
     end
 
-    SDEProblem{iip}(f, u0, tspan, p; callback = cbs,
+    SDEProblem{iip}(f, f.g, u0, tspan, p; callback = cbs,
```
ChrisRackauckas, line 586: wasn't this removed?

ErikQQY: Yeah, but from testing locally
```julia
f, u0, p = process_DEProblem(SDEFunction{iip}, sys, u0map, parammap; check_length, kwargs...)
SDEProbelm(f, u0, tspan, p; ...)
```
would cause the same error in CI
but if we changed to
```julia
f, u0, p = process_DEProblem(SDEFunction{iip}, sys, u0map, parammap; check_length, kwargs...)
SDEProbelm(f, f.g u0, tspan, p; ...) # or SDEProbelm(f.f, f.g u0, tspan, p; ...)
```
working fine

ChrisRackauckas: But that's the constructor you deleted? Why not just use the `SDEProblem(::SDEFunction, u0, tspan, p)` dispatch?

ErikQQY: What is the ```f``` generated from ```process_DEProblem``` here? I think it should be working fine with the v2.0 SciMLBase, which directly build ```SDEProblem``` on ```SDEFunction``` by using ```SDEProblem(f, u0, tspan, p)```.

ChrisRackauckas: It's generating an SDEFunction.

ErikQQY: It's strange here, I tested ```f``` here and indeed I got an SDEFunction, but the constructor ```SDEProblem(f::SDEFunction, u0, tspan, p; ...)``` still error here

ChrisRackauckas: yes, that's what I was asking before merge that you were testing. Did you `@which` see that everything is going to the right dispatches? I thought you needed `tspan::Tuple` or something so that it could dispatch correctly.

ErikQQY: Don't sure if this guess is right, I am wondering since this is a ```DiffEqBase.SDEProblem``` dispatch for ```SDESystem``` type, what is the "version" of SDEProblem used here(MTK is still use 1.103.0 version DiffEqBase)?https://github.com/SciML/ModelingToolkit.jl/blob/1aa480673305854e889be2211eb5b0e4487038ae/src/systems/diffeqs/sdesystem.jl#L586-L587

PR #2283, opened 2023-09-26
Support heterogeneous parameters for linearize and remake

YingboMa:
Requires https://github.com/SciML/ModelingToolkitStandardLibrary.jl/pull/225


src/systems/abstractsystem.jl
```diff
@@ -1274,14 +1275,34 @@ See also [`linearize`](@ref) which provides a higher-level interface.
 function linearization_function(sys::AbstractSystem, inputs,
     outputs; simplify = false,
     initialize = true,
+    op = Dict(),
```
baggepinnen, line 1278: This is a very bad interface, now the user has to provide both `p` and `op` even if `op` already contains `p`. The new changes are also lacking documentation, how is the user supposed to know that they now have to provide also `p` for the results to be correct? 

I would prefer to think a bit harder about this design, the current state feels like a quick hack that is going to be very confusing to use

YingboMa: The `op` here doesn't require the correct numerical value. It only cares about types.

baggepinnen: > The op here doesn't require the correct numerical value. It only cares about types.

What does this mean? `op` is the user-provided operating point around which to linearize.

YingboMa: Since we use heterogenous parameters by default now, we need to know the exact type of the parameters to create the linearized function correctly.

PR #2288, opened 2023-09-27
Use SimpleUnPack.jl instead of UnPack.jl

devmotion:
In this PR, I propose to switch from UnPack to SimpleUnPack, for reasons described in its README: https://github.com/devmotion/SimpleUnPack.jl


src/ModelingToolkit.jl
```diff
@@ -70,7 +70,7 @@ using PrecompileTools, Reexport
 
     import Graphs: SimpleDiGraph, add_edge!, incidence_matrix
 
-    @reexport using UnPack
+    @reexport using SimpleUnPack
```
devmotion, line 73: I'm a bit surprised about this re-export, is this actually needed? Or could it be limited to
```suggestion
    @reexport using SimpleUnPack: @unpack
```
?

YingboMa:
We cannot do this because we actually overload the `unpack` function.

devmotion:
> We cannot do this because we actually overload the unpack function.

Yes, I saw that and removed it in the PR - but it seems it's just forwarded to `getproperty` which is what SimpleUnPack uses anyway.

devmotion:
@YingboMa or did you mean something else? Is this `unpack` overload used by some downstream package? In that case, probably the downstream packages could be switched to `getproperty` as well.

YingboMa:
No, `unpack` is deliberately overloaded to have a different behavior, and using `getproperty` is wrong in places where `unpack` is used.

src/systems/abstractsystem.jl
```diff
@@ -1800,10 +1800,6 @@ function compose(syss...; name = nameof(first(syss)))
 end
 Base.:(∘)(sys1::AbstractSystem, sys2::AbstractSystem) = compose(sys1, sys2)
 
-function UnPack.unpack(sys::ModelingToolkit.AbstractSystem, ::Val{p}) where {p}
-    getproperty(sys, p; namespace = false)
```
YingboMa, line 1804: `unpack` changes the default. It doesn't just forwards.

devmotion: Ah, I missed the `namespace` keyword argument - I guess that's the problem here?

devmotion:
Even if https://github.com/SciML/ModelingToolkit.jl/pull/2288#discussion_r1340699824 is problematic and this overload can't be removed, I think the PR should be reconsidered. Many uses of `@unpack` in the package do not deal with `AbstractSystem`s and are just extracting properties (one example is https://github.com/SciML/ModelingToolkit.jl/blob/34d1853f88bdba5de33f988ec197ab2651539de7/src/systems/systems.jl#L25) - and for these SimpleUnPack or the destructuring syntax in Julia >= 1.7 is much better than UnPack.

PR #2300, opened 2023-10-04
Bifurcation kit extension

TorkelE:
This extension creates dispatches for BifurcationKit.jl's `BifurcationProblem` when used on `NonlinearSystem` or `ODESystem`s.

This should work now:
```julia
using ModelingToolkit
using BifurcationKit

# Define ODEFunction
begin
    @variables t x(t) y(t)
    @parameters μ α
    eqs = [0 ~ μ*x - x^3 + α*y,
           0 ~ -y]
    @named nsys = NonlinearSystem(eqs, [x, y], [μ, α])
end

# Define other bifurcation inputs
begin
    p_span = (-4.0, 6.0)
    bif_par = μ
    plot_var = x
    p_start = [μ => -1.0, α => 1.0]
    u0_guess = [x => 1.0, y => 1.0]
end

# Creates a BifurcationProblem.
# This dispatch is what is implemented here, then it is BK stuff all the way.
bprob = BifurcationProblem(nsys, u0_guess, p_start, bif_par; plot_var=plot_var, jac=false)

# Sets bifurcation parameters
opt_newton = NewtonPar(tol = 1e-9, max_iterations = 20)
opts_br = ContinuationPar(dsmin = 0.001, dsmax = 0.05, ds = 0.01,
	max_steps = 100, nev = 2, newton_options = opt_newton,
	p_max = p_span[2], p_min = p_span[1],
	detect_bifurcation = 3, n_inversion = 4, tol_bisection_eigenvalue = 1e-8, dsmin_bisection = 1e-9)

# Creates the bifurcation diagram.
bf = bifurcationdiagram(bprob, PALC(), 2, (args...) -> opts_br; bothside=true)

# You can plot the diagram like
using Plots
plot(bf; putspecialptlegend=false, markersize=2, plotfold=false, title = "#branches = $(size(bf))")
```
image

Tests and docs not added yet, will do if people think this approach looks good.

### DIFF
--- a/ext/MTKBifurcationKitExt.jl
+++ b/ext/MTKBifurcationKitExt.jl
@@ -11,11 +11,11 @@ import BifurcationKit
 ### Creates BifurcationProblem Overloads ###
 
 # When input is a NonlinearSystem.
-function BifurcationKit.BifurcationProblem(nsys::NonlinearSystem, u0_guess, p_start, bif_par, args...; plot_var=nothing, record_from_solution=BifurcationKit.record_sol_default, kwargs...)
+function BifurcationKit.BifurcationProblem(nsys::NonlinearSystem, u0_bif, ps, bif_par, args...; plot_var=nothing, record_from_solution=BifurcationKit.record_sol_default, jac=true, kwargs...)
     # Creates F and J functions.
-    ofun = NonlinearFunction(nsys; jac=true)
+    ofun = NonlinearFunction(nsys; jac=jac)
     F = ofun.f
-    J = ofun.jac
+    J = jac ? ofun.jac : nothing
 
     # Computes bifurcation parameter and plot var indexes.
     bif_idx = findfirst(isequal(bif_par), parameters(nsys))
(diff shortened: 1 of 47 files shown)


ext/MTKBifurcationKitExt.jl
```diff
@@ -0,0 +1,40 @@
...
+# Imports
+using ModelingToolkit, Setfield
+import BifurcationKit
+
+### Creates BifurcationProblem Overloads ###
+
+# When input is a NonlinearSystem.
+function BifurcationKit.BifurcationProblem(nsys::NonlinearSystem, u0_guess, p_start, bif_par, args...; plot_var=nothing, record_from_solution=BifurcationKit.record_sol_default, kwargs...)
+    # Creates F and J functions.
+    ofun = NonlinearFunction(nsys; jac=true)
```
ChrisRackauckas, line 16: is the jac required? The user should request it if it's not.

ChrisRackauckas:
This looks like a great idea. Just needs tests and docs.

ext/MTKBifurcationKitExt.jl
TorkelE: I will make it optional.

TorkelE:
Sounds good, there is a bug in this for for systems with more species/parameters that I need to fix. I might also drop the `p_span` input (will discuss it with Romain). Once I'm sure of this I will make an updated version, and test+docs (and then make an extension in Catalyst that builds on this one).

TorkelE:
Added some docs and a test. If things seems to work out I could add a couple of more tests.

ext/MTKBifurcationKitExt.jl
```diff
@@ -0,0 +1,40 @@
+module MTKBifurcationKitExt
```
ChrisRackauckas, line 1: fix the naming.

ChrisRackauckas:
Tests fail because the Project.toml needs to be updated.

ext/MTKBifurcationKitExt.jl
TorkelE: what should I name it?

ChrisRackauckas:
That rebase did not go well

TorkelE:
Yeah, no idea what happened. Just tried putting it back at master.

ChrisRackauckas:
Well I think it may be ready to go if it's cleaned up.

TorkelE:
I will probably just make a new PR, easier that way.

TorkelE:
Replaced by: https://github.com/SciML/ModelingToolkit.jl/pull/2321

Issue #2296, opened 2023-10-02
Recent bug: parameter tuples no longer preserve parameter types


isaacsas:
It seems like a breaking change was made to how parameter tuples are handled in a recent release? In the last Catalyst release, 13.4.1, we had doc examples that relied on parameter tuples with mixed types as the underlying ODEs involved factorial functions (a very common term within reaction models). The ODEs were:
```julia
 Differential(t)(A(t)) ~ (-k₊*m*(A(t)^m)) / factorial(m) + k₋*B(t)
 Differential(t)(B(t)) ~ (k₊*m*n*(A(t)^m)) / factorial(m) - k₋*B(t)
```
Previously when building an `ODEProblem`
```julia
p  = (k₊ => 1.0, k₋ => 1.0, m => 2, n => 2)
u₀ = [A => 1.0, B => 1.0]
oprob = ODEProblem(osys, u₀, (0.0, 1.0), p)
```
and `p` would contain the same tuple. This seems broken now on master with `p` being promoted to all floating point, causing code that previously worked just fine to fail.

As this is a breaking change, I would have expected MTK to make a major release but this doesn't seem to have occurred?

This seems like a major issue for Catalyst, since we need the ability to have integer parameters as this is intrinsic to many model types.

Can anyone point me to what version of MTK I should cap the project dependency at until this is fixed?

ChrisRackauckas:
That is unintended.

isaacsas:
We unfortunately may not have had a test on this, so I only caught it when the doc examples broke...

isaacsas:
Also, the issue is present with the latest Catalyst release and not just master (since the former declares compatibility with all < v9 versions of MTK).

ChrisRackauckas:
Fixed in v9

PR #2302, opened 2023-10-06
Respect parameter types

YingboMa:
Fix https://github.com/SciML/ModelingToolkit.jl/issues/2296


ChrisRackauckas:
I believe @AayushSabharwal already handled this.

PR #2310, opened 2023-10-09
Simplify `at extend`

YingboMa:


ChrisRackauckas:
The downstream failure in StructuralIdentifiability is expected given the completeness changes.

PR #2319, opened 2023-10-16
Support conditional statements in `@mtkmodel`

ven-k:
This PR adds support for
- Conditional default values
```julia
@mtkmodel A begin
    ...
    @parameters begin
        a = condition ? 1 : 0
    end
end
```
- Conditional equations and components inside the respective blocks
    - [x] Inside `@equations`
    - [x] Inside `@components`
    - [x] Inside `@parameters` and `@variables`
```julia
@mtkmodel A begin
    ...
    @components begin
       if condition1
            b = B()
       end
    end
    @equations begin
       if condition2
            a ~ 0
       end
    end
end
```
- If branch at top level
   - [x] Equations
   - [x] Components
   - [x] Parameters
```julia
@mtkmodel A begin
    ...
    if condition
        @components begin
             b = B()
        end
        @equations begin
                a ~ 0
        end
    end
end
```

- [x] Adds relevant docs.

### DIFF
--- a/docs/src/basics/MTKModel_Connector.md
+++ b/docs/src/basics/MTKModel_Connector.md
@@ -251,3 +251,104 @@ Dict{Symbol, Any} with 7 entries:
   :extend               => Any[[:p2, :p1], Symbol("#mtkmodel__anonymous__ModelB"), :ModelB]
   :equations            => ["model_a.k ~ f(v)"]
 ```
+
+### Using conditional statements
+
+#### Conditional elements of the system
+
+Both `@mtkmodel` and `@connector` support conditionally defining parameters,
+variables, equations, and components.
+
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
Should we also generate events here or lower @YingboMa ?

baggepinnen:
I guess it depends on whether `condition` is a structural parameter or a symbolic expression. It'd be nice to support both, since structural parameters are often used to determine which equations to include.

ven-k:
Bump

YingboMa:
The MTKStdlib failure looks real. Also, could you merge master back to this branch and run formatter?

ven-k:
@YingboMa I've rebased and have fixed issues that threw error in MTKStdlib tests. This can be reviewed.

YingboMa (approved):
This looks really good!

PR #2323, opened 2023-10-20
Keep symbolic expressions as is and minor bugfixes optimization system

Vaibhavdixit02:


ChrisRackauckas:
add tests which catch this?

Vaibhavdixit02:
Failure is unrelated, something's up with RecursiveArrayTools bounds?

test/optimizationsystem.jl
```diff
@@ -283,3 +283,16 @@ end
 
     @test sol1.u ≈ sol2.u
 end
+
+@testset "#2323 keep symbolic exressions and xor condition on constraint bounds" begin
```
ChrisRackauckas, line 287:
```suggestion
@testset "#2323 keep symbolic expressions and xor condition on constraint bounds" begin
```

Vaibhavdixit02:
There is a circular dependency between https://github.com/SciML/Optimization.jl/pull/636 and this PR so the test won't pass. I'll merge this and finish that one up and then create releases such that both CIs start passing with compat bounds.

ChrisRackauckas:
That's fine, I am trying to get BifurcationKit to tag so we can at least see nothing else broke. If

ChrisRackauckas:
Wait, did you run locally? Only seeing 41 seconds of failure isn't a good sign 😅

Vaibhavdixit02:
Yeah have run it locally. The 41 secs is the RecursiveArrayTools thing I mentioned above, the version doesn't get resolved

Issue #2289, opened 2023-09-28
`@mtkmodel` errors when module qualifier is used to create component


baggepinnen:
The following is fine
```julia
using ModelingToolkit
using ModelingToolkitStandardLibrary.Blocks

@mtkmodel TankWithInput begin
    @components begin
        in = Constant(k=308.5)
    end
end
```
but with the namespace `Blocks` qualifying the access to `Constant`, the macro errors
```julia
using ModelingToolkit
using ModelingToolkitStandardLibrary.Blocks

@mtkmodel TankWithInput begin
    @components begin
        in = Blocks.Constant(k=308.5)
    end
end
```

```julia
ERROR: MethodError: Cannot `convert` an object of type Expr to an object of type Symbol

Closest candidates are:
  convert(::Type{T}, ::T) where T
   @ Base Base.jl:64
  Symbol(::Any...)
   @ Base strings/basic.jl:229

Stacktrace:
  [1] setindex!(A::Vector{Symbol}, x::Expr, i1::Int64)
    @ Base ./array.jl:969
  [2] _unsafe_copyto!(dest::Vector{Symbol}, doffs::Int64, src::Vector{Any}, soffs::Int64, n::Int64)
    @ Base ./array.jl:250
  [3] unsafe_copyto!
    @ ./array.jl:304 [inlined]
  [4] _copyto_impl!
    @ ./array.jl:327 [inlined]
  [5] copyto!
    @ ./array.jl:314 [inlined]
  [6] copyto!
    @ ./array.jl:339 [inlined]
  [7] copyto_axcheck!
    @ ./abstractarray.jl:1182 [inlined]
  [8] Vector{Symbol}(x::Vector{Any})
    @ Base ./array.jl:621
  [9] convert
    @ ./array.jl:613 [inlined]
 [10] push!(a::Vector{Vector{Symbol}}, item::Vector{Any})
    @ Base ./array.jl:1060
 [11] parse_components!(exprs::Vector{Any}, cs::Vector{Symbol}, dict::Dict{Symbol, Any}, body::Expr, kwargs::Vector{Any})
    @ ModelingToolkit ~/.julia/packages/ModelingToolkit/BsHty/src/systems/model_parsing.jl:284
[...]

PR #2330, opened 2023-10-26
feat: support components with fully-qualified names with `@mtkmodel`

thazhemadam:
With the `@named` macro, we could use fully-qualified names for components (for instance, while importing components from the MTKStdLib). However, trying to do the same with `@mtkmodel`,
```julia
@mtkmodel Model begin
     @components begin
          resistor = ModelingToolkitStandardLibrary.Electrical.Resistor(R = 1)
          ...
     end
     ...
end
```
throws the following error.
```julia
ERROR: LoadError: MethodError: Cannot `convert` an object of type Expr to an object of type Symbol

Closest candidates are:
  convert(::Type{T}, ::T) where T
   @ Base Base.jl:84
  Symbol(::Any...)
   @ Base strings/basic.jl:229
```

Fix this and support fully qualified names by considering the fully-qualifed name's Expr while parsing components.

### DIFF
--- a/src/systems/model_parsing.jl
+++ b/src/systems/model_parsing.jl
@@ -673,7 +673,7 @@ function _parse_components!(exprs, body, kwargs)
     expr = Expr(:block)
     varexpr = Expr(:block)
     # push!(exprs, varexpr)
-    comps = Vector{Symbol}[]
+    comps = Vector{Union{Symbol, Expr}}[]
     comp_names = []
 
     for arg in body.args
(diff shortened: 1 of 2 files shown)


baggepinnen:
closes #2289

ven-k:
Can you add a test?

thazhemadam:
> Can you add a test?

Refactored the existing test code a bit to test this functionality as well in b1e7a0e.

PR #2336, opened 2023-10-31
handle parameters that only appear as defaults

### LINKED ISSUE #2322: Parameters that appears only as defaults are ignored
From https://discourse.julialang.org/t/mtk-missing-parameters-mwe/96945

In this example, the parameter `a` never appears in the equations, only as part of an expression for the default value of the other parameters. 
```julia
using ModelingToolkit

@parameters a=10 b=a/10 c=a/20

@variables t
Dt = Differential(t)

@variables x(t)=1 z(t)

eqs = [Dt(x) ~ - b*(x-z),
            0 ~ z - c*x]

@named sys = ODESystem(eqs)

sys_simp = structural_simplify(sys)

ModelingToolkit.defaults(sys_simp)
```
Structural simplification removes the parameter `a`
```julia
julia> ModelingToolkit.defaults(sys_simp)
Dict{Any, Any} with 3 entries:
  x(t) => 1
  b    => (1//10)*a
  c    => (1//20)*a
```
which leads to an error:
```julia
tspan = (0.0, 10)
prob = ODEProblem(sys_simp,[],tspan)

julia> prob = ODEProblem(sys_simp,[],tspan)
ERROR: MethodError: no method matching AbstractFloat(::Type{SymbolicUtils.BasicSymbolic{Real}})

Closest candidates are:
  (::Type{T})(::AbstractChar) where T<:Union{AbstractChar, Number}
   @ Base char.jl:50
  (::Type{T})(::SymbolicUtils.Symbolic) where T<:Union{AbstractFloat, Integer, Complex{<:AbstractFloat}, Complex{<:Integer}}
   @ Symbolics ~/.julia/packages/Symbolics/gBKZv/src/Symbolics.jl:148
  (::Type{T})(::Base.TwicePrecision) where T<:Number
   @ Base twiceprecision.jl:266
  ...

Stacktrace:
  [1] float(x::Type)
```

The problem can be worked around by using the 4-arg constructor of `ODESystem` (suggested in the discourse thread), but I believe we don't want people to use this, so it'd be good to fix this bug.

baggepinnen:
closes #2322

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -492,6 +492,10 @@ function collect_var!(states, parameters, var, iv)
     elseif !isconstant(var)
         push!(states, var)
     end
+    # Add also any parameters that appear only as defaults in the var
+    if hasdefault(var)
+        collect_vars!(states, parameters, getdefault(var), iv)
+    end
     return nothing
 end
 
--- a/test/odesystem.jl
+++ b/test/odesystem.jl
@@ -1044,3 +1044,26 @@ new_sys2 = complete(substitute(sys2, Dict(:sub => sub)))
 Set(states(new_sys2)) == Set([new_sys2.x1, new_sys2.sys1.x1,
     new_sys2.sys1.sub.s1, new_sys2.sys1.sub.s2,
     new_sys2.sub.s1, new_sys2.sub.s2])
+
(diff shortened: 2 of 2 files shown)


baggepinnen:
I'm not sure why the formatting fails, it passes locally :shrug:

PR #2345, opened 2023-11-03
feat: implementation of new SymbolicIndexingInterface

AayushSabharwal:


src/systems/abstractsystem.jl
```diff
@@ -174,14 +182,103 @@ function SymbolicIndexingInterface.independent_variables(sys::AbstractSystem)
     end
 end
 
-function SymbolicIndexingInterface.independent_variables(sys::AbstractTimeDependentSystem)
-    [getfield(sys, :iv)]
+#Treat the result as a vector of symbols always
+function SymbolicIndexingInterface.is_variable(sys::AbstractSystem, sym)
+    if unwrap(sym) isa Int    # [x, 1] coerces 1 to a Num
```
oscardssmith, line 187: this seems really ugly.

AayushSabharwal: Yeah it is. It's the only way I could find to cover all the cases

oscardssmith: what does `unwrap` do?

AayushSabharwal: It gets the value inside a `Num`, which is usually either a `SymbolicUtils.BasicSymbolic` for symbolic variables or a number in the case of something like `Num(1)`

src/systems/abstractsystem.jl
```diff
@@ -224,8 +224,12 @@ function SymbolicIndexingInterface.variable_symbols(sys::AbstractSystem)
 end
 
 function SymbolicIndexingInterface.is_parameter(sys::AbstractSystem, sym)
+    usym = unwrap(sym)
     if unwrap(sym) isa Int
-        return unwrap(sym) in 1:length(parameters(sys))
+        return usym in 1:length(parameters(sys))
+    end
+    if istree(usym) && operation(usym) != getindex
```
oscardssmith, line 231: `return !istree(usym) || operation(usym) == getindex`

PR #2349, opened 2023-11-10
[BREAKING Don't merge] Support DynamicQuantities

YingboMa:


ChrisRackauckas (approved):

test/dq_units.jl
```diff
@@ -0,0 +1,177 @@
...
+@named op = OtherPin()
+@named lp = LongPin()
+good_eqs = [connect(p1, p2)]
+bad_eqs = [connect(p1, p2, op)]
+bad_length_eqs = [connect(op, lp)]
+@test MT.validate(good_eqs)
+@test !MT.validate(bad_eqs)
+@test !MT.validate(bad_length_eqs)
+@named sys = ODESystem(good_eqs, t, [], [])
+@test_throws MT.ValidationError ODESystem(bad_eqs, t, [], []; name = :sys)
```
ChrisRackauckas, line 71: This is supposed to pass. It seems  it just converts but I presume the codegen doesn't add a convert?

YingboMa:
I decided to keep `iv` in the unit check. Not checking `iv` will lead to a lot of special cases, and to have canonical t, the user should write
```julia
const t = ModelingToolkit.t_dynamic_quantities
```
or
```julia
const t = ModelingToolkit.t_unitful
```
or
```julia
const t = ModelingToolkit.t
```

ChrisRackauckas:
I see. Given the direction we're going, I'd propose:

```
t
t_nounits
t_unitful

D
D_nounits
D_unitful
```

YingboMa:
To have a nicer syntax, we can have `@import_independent_variables`, `@import_independent_variables Unitful`, etc... Or a shorter variant of the name.

YingboMa:
Because of https://github.com/SymbolicML/DynamicQuantities.jl/issues/105, we cannot use DQ for the standard library.

ChrisRackauckas:
Master is now setup for the v9, no releases till that.

oscardssmith:
for posterity, should the `don't merge` be removed from the title?

MilesCranmer:
Just confirming that precompilation was fixed with https://github.com/SymbolicML/DynamicQuantities.jl/pull/106.

The version you will want is on `v0.11.2`.

PR #2355, opened 2023-11-15
add component-based hybrid system test

baggepinnen:
This PR adds a test for a hybrid system constructed from components. It exposes a few problems:

- There is a delay of exactly 1 sample interval in the output. ~I have not yet located the source of the delay.~ I think the problem is that `d2c_view` needs to be initialized correctly, it currently starts out as 0. #2356 
- All variables associated with the discrete partition are zero in the solution, including observed variables
- The fact that `Hold` is an operator precludes its use as a component with the same name. We could consider renaming the operator `hold` to avoid the conflict. The same goes for the operator `Sample`, but in this case I find the component name `Sampler` okay (whereas `Holder` isn't great :sweat_smile: )

I added some tests that ~are marked as broken~ are fixed, tracked by
- #2356 

Additionally, issue
- #2358 
was discovered


YingboMa:
Tests passed locally for me.

YingboMa (approved):

PR #2384, opened 2023-12-13
Cache hash of ConnectionElement

chriselrod:
Before:
```julia
julia> using ModelingToolkit: ConnectionSet

julia> connectionsets = ConnectionSet[]
ConnectionSet[]

julia> domain_csets = ConnectionSet[]
ConnectionSet[]

julia> ModelingToolkit.generate_connection_set!(connectionsets, domain_csets, sys, nothing, nothing);

julia> length(connectionsets)
40008

julia> @time merge(connectionsets);
 36.909123 seconds (628.06 M allocations: 9.941 GiB, 1.82% gc time, 0.01% compilation time)
```
After:
```julia
julia> @time merge(connectionsets);
  4.970410 seconds (406.83 k allocations: 29.342 MiB, 0.10% compilation time)
```


ChrisRackauckas:
Is there some kind of test we can add for performance regressions on this?

chriselrod:
It'd be great if we could have something like nanosoldier for SciML, some performance benchmark test suite that could run regularly on the same computer to track changes over time.
As is, it's hard to notice when regressions creep in. It's all too easy for a type instability to appear somewhere.

In this case, though, most of `merge`'s time was just calling `hash` so caching the hash was an obvious solution that seemed to really improve things.

However, I fear the validity of things like this. What is `v`? Is it allowed to change?

For now, I added
```julia
    @inbounds begin
        @boundscheck begin
            @assert e.h === _hash_impl(e.sys, e.v, e.isouter)
        end
    end
```
Maybe we should make our own `@dassert` (debug assert) macro that is only on when bounds checking is enforced, and hence is always enabled when running test suites.

ChrisRackauckas:
We have something setup on BoundaryValueDiffEq.jl.

chriselrod:
Blocked by https://github.com/bifurcationkit/BifurcationKit.jl/pull/127

chriselrod:
Can we move forward with this without BifurcationKit? Who uses it/what's our reason for supporting it?

ChrisRackauckas:
Rebase

chriselrod:
> Rebase

The first problem is that we can't even run tests without BifurcationKit having a compatible version.

ChrisRackauckas:
That's what's fixed on master.

chriselrod:
> That's what's fixed on master.

Okay, the changes here should be unrelated to any SymbolicIndexingInterface issues. So if some other changes caused issues, I'll remove them.
Was mostly annoyed by not being able to run tests on CI.

ChrisRackauckas:
Having multiple very low level packages do major bumps in the same month is a nightmare for many reasons. We'll get through this rough patch though.

Issue #2398, opened 2023-12-27
ModelingToolkit does not save array as static arrays (for initial conditions and parameter values)


TorkelE:
I was trying to get the DiffEqGPU tutorial to work for Catalyst. It recommends creating static arrays. However, I noticed this not being the case when I try using MTK approach:

```julia
using Catalyst, StaticArrays

prob_mtk = begin
    @parameters σ ρ β
    @variables t x(t) y(t) z(t)
    D = Differential(t)

    eqs = [D(D(x)) ~ σ * (y - x),
        D(y) ~ x * (ρ - z) - y,
        D(z) ~ x * y - β * z]

    @named sys = ODESystem(eqs)
    sys = structural_simplify(sys)

    u0 = @SVector [D(x) => 2.0,
        x => 1.0,
        y => 0.0,
        z => 0.0]

    p = @SVector [σ => 28.0,
        ρ => 10.0,
        β => 8 / 3]

    tspan = (0.0, 100.0)
    ODEProblem(sys, u0, tspan, p)
end

prob_diffeq = begin
    function lorenz(u, p, t)
        σ = p[1]
        ρ = p[2]
        β = p[3]
        du1 = σ * (u[2] - u[1])
        du2 = u[1] * (ρ - u[3]) - u[2]
        du3 = u[1] * u[2] - β * u[3]
        return SVector{3}(du1, du2, du3)
    end

    u0 = @SVector [1.0; 0.0; 0.0]
    tspan = (0.0, 10.0)
    p = @SVector [10.0, 28.0, 8 / 3.0]
    prob = ODEProblem(lorenz, u0, tspan, p)
end

prob_mtk.u0 # Not static.
prob_mtk.p # Not static.
prob_diffeq.u0 # Static.
prob_diffeq.p # Static.
```

Is this a bug, or is there a different way to specify this for MTK?

PR #2399, opened 2023-12-27
Preserve staticarrays in the problem construction

ChrisRackauckas:
Fixes https://github.com/SciML/ModelingToolkit.jl/issues/2398. This is pretty crucial for giving users a way to target GPUs.

### DIFF
--- a/src/parameters.jl
+++ b/src/parameters.jl
@@ -105,7 +105,6 @@ function split_parameters_by_type(ps)
             parrs = map(x-> SArray{Tuple{size(x)...}}(x), split_ps)
             split_ps = SArray{Tuple{size(parrs)...}}(parrs)
         end
-        @show typeof(split_ps)
         if length(split_ps) == 1  #Tuple not needed, only 1 type
             return split_ps[1], split_idxs
         else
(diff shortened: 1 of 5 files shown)


src/parameters.jl
```diff
@@ -100,6 +100,12 @@ function split_parameters_by_type(ps)
         end
         tighten_types = x -> identity.(x)
         split_ps = tighten_types.(Base.Fix1(getindex, ps).(split_idxs))
+
+        if ps isa StaticArray
+            parrs = map(x-> SArray{Tuple{size(x)...}}(x), split_ps)
+            split_ps = SArray{Tuple{size(parrs)...}}(parrs)
+        end
+        @show typeof(split_ps)
```
ChrisRackauckas, line 108:
```suggestion
```

test/static_arrays.jl
```diff
@@ -0,0 +1,28 @@
+using ModelingToolkit, StaticArrays, Test
```
ChrisRackauckas, line 1:
```suggestion
using ModelingToolkit, SciMLBase, StaticArrays, Test
```

Issue #1939, opened 2022-11-11
Clocks TODO


baggepinnen:
- [x] Hook up codegen when `ODEFunction` is called
- [x] Do something with all the data saved by discrete-time partitions
- [x] Clean up old code generation for discrete-time systems
    - [x] https://github.com/SciML/ModelingToolkit.jl/pull/2413
- [x] Get rid of old, ill-defined discrete operators
    - [x] https://github.com/SciML/ModelingToolkit.jl/pull/2413
- [x] Set initial condition for discrete variables with a sane syntax
- [x] code generation with more than one discrete-time partition
- [x] add tests for multiple clock rates
- [x] Handle negative shifts, #2359 
- [x] Store solution to discrete-time partitions and make it accesible through `idxs` keyword
- [x] handle plotting of discrete-time data with `steppost + marker` appearance
- [x] Specification of data types of state variables and parameters in the MTK model, e.g., `Int, Float64` etc. This type has to be configurable by the user instantiating a block, e.g., the user can select whether an integer variable is `Int64 / Int32` etc., #2491 
- [x] Introduce `Sample` and `Hold` blocks in MTKstdlib, https://github.com/SciML/ModelingToolkitStandardLibrary.jl/pull/242
- [ ] Introduce super and sub sampling #2744 
- [ ] Introduce Clock block that can be connected to discrete-time components to associate them with a clock
- [x] Handle calls to `sampletime` in equations that have `InferredDiscrete`  time domain. This should use a connected clock if available, otherwise set the domain to inferred. #2690
- [x] Test discrete-time partitions that include both positive and negative shifts
- [x] Add documentation
    - [x] Mention implicit `(k)` index for non-indexed variables
    - [x] Nail down the semantics of the shift index w.r.t. initial conditions. The simulation starts at shift index `k`, and the initial condition is all indices less than `k`.
    - [x] https://github.com/SciML/ModelingToolkit.jl/pull/2410
- [x] Add tutorials
- [x] https://github.com/SciML/ModelingToolkit.jl/issues/2356
- [ ] Test inversion of discrete-time model.

PR #2410, opened 2024-01-02
add docs for sampled-data systems

baggepinnen:
Work in progress
- https://github.com/SciML/ModelingToolkit.jl/issues/1939


docs/src/tutorials/SampledData.md
```diff
@@ -0,0 +1,153 @@
...
+- `y` is also automatically inferred to be a discrete-time-time variable, since it appears in an equation with another discrete-time variable `x`. `x,u,y` all belong to the same discrete-time partition, i.e., they are all updated at the same *instantaneous point in time* at which the clock ticks.
+- The equation `y ~ x` does not use any shift index, this is equivalent to `y(k) ~ x(k)`, i.e., discrete-time variables without shift index are assumed to refer to the variable at the current time step.
+- The equation `x(k+1) ~ 0.5x(k) + u(k)` indicates how `x` is updated, i.e., what the value of `x` will be at the *next* time step. The output `y`, however, is given by the value of `x` at the *current* time step, i.e., `y(k) ~ x(k)`. If this logic was implemented in an imperative programming style, the logic would thus be
+
+```julia
+function discrete_step(x, u)
+    y = x # y is assigned the old value of x
+    x = 0.5x + u # x is updated to a new value
+    return x, y # The state x now refers to x at the next time step, while y refers to x at the current time step
+end
```
AayushSabharwal, line 42: Doesn't this make `y` lag `x` by one timestep? This would then imply that `y(k+1) ~ x(k)` which is not what the equation above says.

Basically, if `x, y, u = [1.0, 1.0, 2.0]` would the timeseries be`[[1.0, 1.0, 2.0], [2.5, 1.0, 2.0], [3.75, 2.5, 2.0]]` or `[[1.0, 1.0, 2.0], [2.5, 2.5, 2.0], [3.75, 3.75, 2.0]]`

I'm not very familiar with discrete-time modeling, so this may be some common misunderstanding I'm unaware of 😅

baggepinnen: > Doesn't this make y lag x by one timestep?

Yes, the output `x` of _this function_ is one step ahead of the output `y`, but `y(k)` does not lag `x(k)`. Note how the comment mentions
> The state x now refers to x at the next time step, while y refers to x at the current time step

Where _next_ in `x at the next time step` is key. The returned value of the manually implemented function `discrete_step` is thus to be interpreted as the tuple
```
(state in next time step, current time output)
```

I'll commit a change to the comment to hopefully make this more clear

My goto book is usually "Computer-controlled systems" by KJ Åström, but I'm not sure if it would be particularly useful as a reference here.

baggepinnen: I hope this commit clarifies further? https://github.com/SciML/ModelingToolkit.jl/pull/2410/commits/cbd9f5987021fe4593ff3bd75e004e8eb2ef0f08

ChrisRackauckas:
@baggepinnen can something of this form be merged by the end of the day? It can continue to improve but we need something to go with the v9.

baggepinnen:
This documents functionality that is not yet existing, in particular, the user interface is almost 100% missing, so I'd rather not merge any of it yet.

ChrisRackauckas:
Okay then we'll v9 without this.

docs/pages.jl
```diff
@@ -10,6 +10,7 @@ pages = [
         "tutorials/stochastic_diffeq.md",
         "tutorials/parameter_identifiability.md",
         "tutorials/bifurcation_diagram_computation.md",
+        "tutorials/SampledData.md"
```
ChrisRackauckas, line 13:
```suggestion
        "tutorials/SampledData.md",
```

ChrisRackauckas:
Build failed.

PR #2413, opened 2024-01-03
rm old discrete system

baggepinnen:


ChrisRackauckas:
Rebase

src/utils.jl
```diff
@@ -319,23 +315,6 @@ isoperator(op) = expr -> isoperator(expr, op)
 isdifferential(expr) = isoperator(expr, Differential)
 isdiffeq(eq) = isdifferential(eq.lhs)
 
-isdifference(expr) = isoperator(expr, Difference)
-isdifferenceeq(eq) = isdifference(eq.lhs)
```
AayushSabharwal, line 323: These functions are used in other places. Is there a corresponding implementation for the new discrete system? `hasshift`?

baggepinnen: `isdifference` is no longer the correct question to ask, you can have an equation that is not a difference equation, but it still belongs to the discrete time domain. 

You now have `has_discrete_domain` and other similar functions in the same file, but they require clock inference to have been run on the model before they are fully usable.

AayushSabharwal:
I don't have the permissions to push here, so I opened another PR branched off of this one https://github.com/SciML/ModelingToolkit.jl/pull/2440

PR #2414, opened 2024-01-03
add RealtimeClock

baggepinnen:
This PR adds `RealtimeClock <: AbstractClock` which behaves identically to the standard `Clock`, but forces the simulation to progress in (no faster than) real time, i.e., a simulation with `tspan = (0, 10)` will take 10 seconds of wall-clock time as long as the computations are fast enough to meet the chosen sample rate.

The PR also refactors the generation of the callback associated with the clock so that it is easier to add new clock implementations

### DIFF
--- a/src/clock.jl
+++ b/src/clock.jl
@@ -98,11 +98,14 @@ end
 abstract type AbstractClock <: AbstractDiscrete end
 
 """
-    Clock <: AbstractClock
-    Clock([t]; dt)
+    Clock()
+    Clock(dt)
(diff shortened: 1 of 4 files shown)


ChrisRackauckas:
Rebase?

baggepinnen:
The PR is over a year old now so it's incredibly stale, and some of the features it relies upon are no longer present in MTK. It would probably make more sense to wait until the new clocking support has gotten a bit further and then cut out the parts of this PR that are still relevant into a new PR

PR #2425, opened 2024-01-15
Replace MacroTools with ExprTools & Base.remove_lines!

oxinabox:
## Checklist

## Additional context

Briefly discussed this on twitter with @ChrisRackauckas 

This lets us drop the direct dependency on MacroTools (though we still have like a dozen indirect dependencies on it)
ExprTools is a smaller dependency that is more comprehensively tested than MacroTools, and does actually catch several cases that MacroTools misses.

I will note that `Base.remove_linenum!` is not part of the public API of Base (https://github.com/JuliaLang/julia/issues/52904) however it is already used several times in MacroTools.

I am not super fussed, but if this PR is approved I will try and put a PR in each day or so to remove MacroTools from another package in the SciML ecosystem. Its a nice little exercise to wake brain up in the morning.


ChrisRackauckas (approved):

Project.toml
```diff
@@ -74,6 +74,7 @@ Distributions = "0.23, 0.24, 0.25"
 DocStringExtensions = "0.7, 0.8, 0.9"
 DomainSets = "0.6"
 DynamicQuantities = "0.8, 0.9, 0.10"
+ExprTools = "1"
```
ChrisRackauckas, line 77:
```suggestion
ExprTools = "0.1.10"
```

YingboMa (approved):

PR #2434, opened 2024-01-26
Support for time-dependent parameterizations

cmhyett:
Currently variable/parameter substitution is "eager", in that `build_function` replaces according to a dictionary that looks like `x(t)[1] => _-arg1[1], ..., p[1]=>_-arg2[1], ...`

Which is safe nominal behavior, but if we don't know apriori what index will be used, (e.g., parameters depend on time), and we want to retain a symbolic form like `p[floor(Int, t/10)+1]` `build_function` does not recognize this, and instead retains this as part of the equations, i.e., symbolic result would look like something like `_-arg1[1] = p[floor(Int, t/10)+1]` instead of `_-arg1[1] = _-arg2[floor(Int, t/10)+1]`. The former leads to an undefined variable error, and the latter is the desired behavior.

Currently the proposed change is a hack, but maybe a keyword argument like `eager` which defaults to `true` and maintains the current behavior, while `false` would "weaken" the dictionary to simply find all `x,p` and replace with corresponding args.

### DIFF
--- a/src/systems/diffeqs/abstractodesystem.jl
+++ b/src/systems/diffeqs/abstractodesystem.jl
@@ -178,7 +178,11 @@ function generate_function(sys::AbstractODESystem, dvs = states(sys), ps = param
             if p isa Tuple
                 build_function(rhss, u, p..., t; postprocess_fbody = pre,
                     states = sol_states,
-                    kwargs...)
+                               kwargs...)
+            elseif (length(p) > 1)
+                p = arguments(p[1])[1]
+                build_function(rhss, u, p, t; postprocess_fbody = pre, states = sol_states,
+                               kwargs...)
             else
                 build_function(rhss, u, p, t; postprocess_fbody = pre, states = sol_states,
                     kwargs...)


cmhyett:
@ChrisRackauckas, getting back to this PR, I think the above comment is more descriptive than the discussion we had in slack.

Essentially for "better" symbolic replacement, I propose to tell `build_function` **not** to replace things like `p[1]` with `_-arg3[1]`, but rather `p` with `_-arg3`. Ideally it'd also work for `variables` This allows for indexing into symbolic arrays with values _only known at the time of execution_ (my application is parameterized time-dependent boundary conditions, but one could also use this in molecular dynamics, or time-dependent graph structures, etc)

It'd be a no-change for anything that doesn't use real-time evaluation for indices.

Currently the code is admittedly hacky, because I don't know what the preferred way to strip the variable from the indexing operation. But the change should be quite small, maybe just 2-3 SLOC.

Any feedback on feasibility/preferred ways of performing this change would be appreciated.

cmhyett:
MWE:
using Symbolics, OrdinaryDiffEq, ModelingToolkit;

@parameters p[1:2]
@variables t x(..)[1:3]
Dt = Differential(t);
@register_symbolic Base.floor(T::Type, x)::UInt64

# works
lbdy(_p, t) = _p[1];
rbdy(_p, t) = _p[2];

# LoadError: UndefVarError: `p` not defined
lbdy(_p, t) = _p[floor(Int, t/100)];
rbdy(_p, t) = _p[floor(Int, t/100)];

eqs = [x(t)[1] ~ lbdy(p, t),
       Dt(x(t)[2]) ~ x(t)[1] + x(t)[3],
       x(t)[3] ~ rbdy(p, t)]
@named de = ODESystem(eqs, t, [x(t)...], [p...]);
sys = structural_simplify(de);
u0 = [0.0];
p0 = [1.0; 2.0];
prob = ODEProblem(sys, u0, (0.0, 1.0), Symbolics.scalarize(p .=> p0));
sol = solve(prob, FBDF(), adaptive=false, dt = 0.25)

ChrisRackauckas:
Shouldn't this just work in the v9?

cmhyett:
It does not, but the fix is less hacky because of the v9 changes. I'm slammed right now, but I hope to get back to this soon.

PR #2443, opened 2024-02-02
fix!: require explicit independent variable for ODESystem

AayushSabharwal:
## Checklist

### DIFF
--- a/src/ModelingToolkit.jl
+++ b/src/ModelingToolkit.jl
@@ -192,7 +192,7 @@ const D = Differential(t)
 PrecompileTools.@compile_workload begin
     using ModelingToolkit
     @variables x(ModelingToolkit.t_nounits)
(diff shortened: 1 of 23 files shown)


oscardssmith:
What does this fix? This seems like a really breaking change without a ton of benefit.

ChrisRackauckas:
It fixes a lot since the equation parsing had a few edge cases where it's not generally fixable without requiring that the independent variable is known.

ChrisRackauckas:
If you dig through the closed issues you'll find many cases where ODESystem did it's decomposition incorrectly and that led to an error, with the fix being to use the "right" constructor, i.e. pass in the independent variable too. Some of those cases simply have too little information to solve so the heuristic for determining what is the independent variable is just not generally solvable. Given it's not generally solvable, the best thing to do is to just ask users for it. It's easy and it fixes all of those potential issues.

ChrisRackauckas:
Does @mtkbuild need a change?

PR #2451, opened 2024-02-05
Have CI tests depend on formatter passing

chriselrod:
CI keeps getting merged despite a failing formatter, so this PR's intention is to only have tests run after formatter passes. This should incentivize actually calling `JuliaFormatter.format("path/to/ModelingToolKit")` or using your editor to format the files you edited.


chriselrod:
Confirming that the formatter failed, and tests were skipped.
https://github.com/SciML/ModelingToolkit.jl/actions/runs/7791834752
Now, formatting and committing to confirm that tests do run when the formatter passes.

YingboMa (approved):

PR #2456, opened 2024-02-12
Pin JuliaFormatter

YingboMa:
The latest JuliaFormatter still is not stable https://github.com/domluna/JuliaFormatter.jl/pull/807, so it's better to batch the changes and format later to avoid polluting git blame.


ChrisRackauckas:
Isn't that the last of the changes?

PR #2465, opened 2024-02-14
Fix test groupings and test setup

ChrisRackauckas:


ChrisRackauckas:
@baggepinnen make sure in the future that tests with downstream components properly end up in the downstream environments so they don't block the core tests.

PR #2466, opened 2024-02-14
docs: update NEWS.md with v9 update notes

AayushSabharwal:
## Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -30,6 +30,7 @@
   `@parameters p::Int`. Array-valued parameters must be array symbolics; `@parameters p = [1.0, 2.0]`
   is now invalid and must be changed to `@parameters p[1:2] = [1.0, 2.0]`. The index of a parameter
   in the system is also not guaranteed to be an `Int`, and will instead be a custom undocumented type.
-  - To restore the old behavior, use `ModelingToolkit.@set sys.index_cache = nothing` before creating
-    a problem, and after calling `structural_simplify`.
+  To restore the old behavior:
+  - Pass the `split = false` keyword to `structural_simplify`. E.g. `ss = structural_simplify(sys; split = false)`.
+  - Pass `split = false` to `@mtkbuild`. E.g. `@mtkbuild sys = ODESystem(...) split = false`.
 - Discrete-time system using `Difference` are unsupported. Instead, use the new `Clock`-based syntax.
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,9 +1,35 @@
-# ModelingToolkit v8 Release Notes
+# ModelingToolkit v9 Release Notes
 
 ### Upgrade guide
 
-  - `connect` should not be overloaded by users anymore. `[connect = Flow]`
-    informs ModelingToolkit that particular variable in a connector ought to sum
-    to zero, and by default, variables are equal in a connection. Please check out
-    [acausal components tutorial](https://docs.sciml.ai/ModelingToolkit/stable/tutorials/acausal_components/)
-    for examples.
+- The function `states` is renamed to `unknowns`. In a similar vein:
+  - `unknown_states` is now `solved_unknowns`.
+  - `get_states` is `get_unknowns`.
+  - `get_unknown_states` is now `get_solved_unknowns`.
+- The default backend for using units in models is now `DynamicQuantities.jl` instead of
+  `Unitful.jl`.
+- ModelingToolkit.jl now exports common definitions of `t` (time independent variable)
+  and `D` (the first derivative with respect to `t`). Any models made using ModelingToolkit.jl
+  should leverage these common definitions. There are three variants:
+  - `t` and `D` use DynamicQuantities.jl units. This is the default for standard library
+    components.
+  - `t_unitful` and `D_unitful` use Unitful.jl units.
+  - `t_nounits` and `D_nounits` are unitless.
+- `ODAEProblem` is deprecated in favor of `ODEProblem`.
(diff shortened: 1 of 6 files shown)


NEWS.md
```diff
@@ -1,9 +1,34 @@
...
+- The function `states` is renamed to `unknowns`. In a similar vein:
+  - `unknown_states` is now `solved_unknowns`.
+  - `get_states` is `get_unknowns`.
+  - `get_unknown_states` is now `get_solved_unknowns`.
+- The default backend for using units in models is now `DynamicQuantities.jl` instead of
+  `Unitful.jl`.
+- ModelingToolkit.jl now exports common definitions of `t` (time independent variable)
+  and `D` (the first derivative with respect to `t`). Any models made using ModelingToolkit.jl
+  should leverage these common definitions. There are three variants:
+  - `t` and `D` use DynamicQuantities.jl units.
```
ChrisRackauckas, line 14:
```suggestion
  - `t` and `D` use DynamicQuantities.jl units. This is the default for standard library components.
```

AayushSabharwal:
The discrete-time system docs will be linked to once they exist

NEWS.md
```diff
@@ -1,9 +1,34 @@
...
+  systems will have a custom struct instead of a `Vector` of parameters. The internals of this
+  type are undocumented and subject to change without notice or a breaking release. Parameter values
+  can be queried, updated or manipulated using SciMLStructures.jl or SymbolicIndexingInterface.jl.
+  This also requires that the symbolic type of a parameter match its assigned value. For example,
+  `@parameters p` will always use a `Float64` value for `p`. To use `Int` instead, use
+  `@parameters p::Int`. Array-valued parameters must be array symbolics; `@parameters p = [1.0, 2.0]`
+  is now invalid and must be changed to `@parameters p[1:2] = [1.0, 2.0]`. The index of a parameter
+  in the system is also not guaranteed to be an `Int`, and will instead be a custom undocumented type.
+  - To restore the old behavior, use `ModelingToolkit.@set sys.index_cache = nothing` before creating
+    a problem, and after calling `structural_simplify`.
```
ChrisRackauckas, line 33: Is this really the best way to do this, not a keyword argument in structural simplify?

AayushSabharwal: I'll add the kwarg then. `@mtkbuild` will not support it

ChrisRackauckas: `@mtkbuild` should get a `simplify=true, split=true`.

ChrisRackauckas: The key is that if we don't take control of this, then it won't be deprecatable.

NEWS.md
```diff
@@ -1,9 +1,36 @@
...
+- Specifying the independent variable for an `ODESystem` is now mandatory. The `ODESystem(eqs)`
+  constructor is removed.
+- Systems must be marked as `complete` before creating `*Function`/`*FunctionExpr`/`*Problem`/
+  `*ProblemExpr`. Typically this involved using `@mtkbuild` to create the system or calling
+  `structural_simplify` on an existing system.
+- All systems will perform parameter splitting by default. Problems created using ModelingToolkit.jl
+  systems will have a custom struct instead of a `Vector` of parameters. The internals of this
+  type are undocumented and subject to change without notice or a breaking release. Parameter values
+  can be queried, updated or manipulated using SciMLStructures.jl or SymbolicIndexingInterface.jl.
+  This also requires that the symbolic type of a parameter match its assigned value. For example,
```
isaacsas, line 28:
```suggestion
- All systems will perform parameter splitting by default. Problems created using ModelingToolkit.jl
  systems will have a custom struct instead of a `Vector` of parameters. The internals of this
  type are undocumented and subject to change without notice or a breaking release. Parameter values
  can be queried, updated or manipulated using SciMLStructures.jl or SymbolicIndexingInterface.jl.
  This also requires that the symbolic type of a parameter match its assigned value. For example,
```
Surely this must be documented and standardized for use in the DifferentialEquations solvers? For example, JumpProcesses requires that parameters are stored in a linearly indexable collection in its internal solvers (for MassActionJumps). Mixed parameter types could previously be achieved via MTK by using tuples to store parameters. 

Is it no longer the case the JumpProcesses can assume parameter objects are linearly indexable within solvers?

isaacsas: (Sorry for my confusion, still trying to parse out how this new parameter handling will/does work!)

ChrisRackauckas: > Is it no longer the case that JumpProcesses can assume parameter objects are linearly indexable within solvers?

It should have never assumed that. That's never really been true. We've always supported arbitrary structs in the solvers.

ChrisRackauckas: > this must be documented and standardized for use in the DifferentialEquations solvers?

It got more standardized recently https://docs.sciml.ai/SciMLStructures/dev/interface/.

ChrisRackauckas: We should talk JumpProcesses after this all gets out there. There's things we need to do, but there's so much going on here that I didn't want to overlap the two.

isaacsas: I'm all for improving the interface there. Initializing and updating MassActionJumps for parameter changes right now is clunky... Plus there are the issues of making the library stochastic AD compatible (as much as possible).

ChrisRackauckas: SciMLStructures Tunable canonicalizations are kind of the key to making arbitrary parameters work much better, but there's more to think about with MAJs. Maybe let's do a call the 3 of us on Friday? I want to get this merged independently of that though.

isaacsas: (I'd like to invest a big chunk of time towards the later towards the end of the semester / over the summer as part of our NSF grant.)

PR #2469, opened 2024-02-15
feat!: do not scalarize parameters, fix some tests

AayushSabharwal:
## Checklist


ChrisRackauckas:
format

AayushSabharwal:
JuliaFormatter is getting updated a lot recently

AayushSabharwal:
I have the same version of JuliaFormatter as CI, but locally it doesn't complain and here it does?

ChrisRackauckas (approved):

src/systems/diffeqs/odesystem.jl
```diff
@@ -195,9 +195,8 @@ function ODESystem(deqs::AbstractVector{<:Equation}, iv, dvs, ps;
         gui_metadata = nothing)
     name === nothing &&
         throw(ArgumentError("The `name` keyword must be provided. Please consider using the `@named` macro"))
-    #deqs = scalarize(deqs)
     @assert all(control -> any(isequal.(control, ps)), controls) "All controls must also be parameters."
-
+    deqs = reduce(vcat, scalarize(deqs); init = Equation[])
```
YingboMa: We should never scalarize during construction for MTKv9. We should just follow what I did in https://github.com/SciML/ModelingToolkit.jl/pull/2472

YingboMa: If one wants a scalarized system for simulation, then one needs to call `structural_simplify`.

src/utils.jl
```diff
@@ -342,25 +342,28 @@ v == Set([D(y), u])
...
     (vars!(vars, eq.lhs; op = op); vars!(vars, eq.rhs; op = op); vars)
 end
 function vars!(vars, O; op = Differential)
-    if isvariable(O)
+    if isvariable(O) && !(istree(O) && operation(O) === getindex)
         return push!(vars, O)
     end
+
     !istree(O) && return vars
+    if operation(O) === (getindex)
```
YingboMa, line 358: If we want `structural_simplify` to work correctly, we have to differentiate `x` and `x[1]` alone. https://github.com/SciML/ModelingToolkit.jl/pull/2472/files#diff-47c27891e951c8cd946b850dc2df31082624afdf57446c21cb6992f5f4b74aa2R351-R369 has the correct implementation

src/utils.jl
```diff
@@ -232,7 +232,7 @@ end
 
 function collect_defaults!(defs, vars)
     for v in vars
-        (haskey(defs, v) || !hasdefault(v)) && continue
+        (haskey(defs, v) || !hasdefault(unwrap(v))) && continue
```
YingboMa, line 235: Should `v` ever be wrapped? It will be a pretty hard to catch bug.

Project.toml
```diff
@@ -106,7 +106,7 @@ SpecialFunctions = "0.7, 0.8, 0.9, 0.10, 1.0, 2"
 StaticArrays = "0.10, 0.11, 0.12, 1.0"
 SymbolicIndexingInterface = "0.3.1"
 SymbolicUtils = "1.0"
-Symbolics = "5.7"
+Symbolics = "5.21"
```
ChrisRackauckas, line 109:
```suggestion
Symbolics = "5.20.1"
```

PR #2470, opened 2024-02-15
Enable specification of completeness on ODE/SDE/Nonlinear/Jump system creation (required for conversions from Catalyst's `ReactionSystem`s)

TorkelE:
With MTK v9, systems must be complete when used to create e.g. `ODEProblem`. Catalyst creates `ReactionSystem`, which can be converted to e.g. `ODE`/`SDE`/`Nonlinear`/`Jump` `System`s (which are done internally when `ReactionSystem`s are used as input to e.g. `ODEProblem`s).

Here, if a `ReactionSystem` is marked as `complete`, the `ODESystem`s (and other systems) created from it should also be complete. Currently, it is not possible to set a `ODESystem` as `complete` on creation (using the internal stuff). This PR adds keyword arguments to set `complete = true` (`false` default) for  `ODE`/`SDE`/`Nonlinear`/`Jump` `System`s (which Catalyst can use).

E.g. currently we have
```julia
using Catalyst
rs = @reaction_network begin
    (kp,kd), 0 <--> X
    (k1,k2), X <--> Y
end
rs = complete(rs)
ModelingToolkit.iscomplete(rs)  # true

osys = convert(ODESystem, rs)
ssys = convert(SDESystem, rs)
nsys = convert(NonlinearSystem, rs)
jsys = convert(JumpSystem, rs)

ModelingToolkit.iscomplete(osys) # false
ModelingToolkit.iscomplete(ssys) # false
ModelingToolkit.iscomplete(nsys) # false
ModelingToolkit.iscomplete(jsys) # false
```
This allows me to add an argument:
```julia
function Base.convert(::Type{<:ODESystem}, rs::ReactionSystem; ...)
     ....
    return ODESystem(...,
              complete = ModelingToolkit.iscomplete(rs),
              ...)
end
```
which is used in `convert(ODESystem, rs)`. This turns things to
```julia
ModelingToolkit.iscomplete(osys) # true
ModelingToolkit.iscomplete(ssys) # true
ModelingToolkit.iscomplete(nsys) # true
ModelingToolkit.iscomplete(jsys) # true
end

### DIFF
--- a/src/systems/diffeqs/odesystem.jl
+++ b/src/systems/diffeqs/odesystem.jl
@@ -192,7 +192,8 @@ function ODESystem(deqs::AbstractVector{<:Equation}, iv, dvs, ps;
         discrete_events = nothing,
         checks = true,
         metadata = nothing,
-        gui_metadata = nothing)
+        gui_metadata = nothing,
+        complete = false)
     name === nothing &&
         throw(ArgumentError("The `name` keyword must be provided. Please consider using the `@named` macro"))
     deqs = scalarize(deqs)
@@ -232,7 +233,8 @@ function ODESystem(deqs::AbstractVector{<:Equation}, iv, dvs, ps;
         deqs, iv′, dvs′, ps′, tspan, var_to_name, ctrl′, observed, tgrad, jac,
         ctrl_jac, Wfact, Wfact_t, name, systems, defaults, nothing,
         connector_type, preface, cont_callbacks, disc_callbacks,
-        metadata, gui_metadata, checks = checks)
+        metadata, gui_metadata, 
+        nothing, nothing, complete, checks = checks)
 end
 
 function ODESystem(eqs, iv; kwargs...)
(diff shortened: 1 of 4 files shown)


ChrisRackauckas:
But this isn't correct? Where is `complete` called? This just makes the boolean true even though it should be false?

TorkelE:
You are right, this isn't fully correct. However, I need a function for initiating correctness for MTK systems on creation. I was hoping some of you guys could help with the details (although maybe better for a separate PR).

ChrisRackauckas:
In the conversion call complete?

TorkelE:
Yes, we could put additional stuff in Catalyst'S conversion call to make the output complete. I just figured it would make more sense to have it here, since it is something that all packages that generates internal Mtk systems would want to do.

However, if there is no plans for addressing this in MTK, we can handle it in Catalyst.

ChrisRackauckas:
No, packages shouldn't be doing this. They should be calling complete in order to be complete. Skipping unit validations, correctness tests, etc. is not a good idea. Disabling all errors and doing something that is just wrong is not something that we should figure any package wants to do. Calculating things incorrectly in a silent way is bad. Nobody should want to do that.

ChrisRackauckas:
The correct thing to do is to call `complete` when the system is complete. Catalyst should do that when it thinks its system is complete. It should not try to bypass any and all error handling and validations.

TorkelE:
That was not really what I meant, but we can handle it in Catalyst, no problems.

PR #2478, opened 2024-02-22
Fix and update documentation for v9

ChrisRackauckas:


ChrisRackauckas:
This doesn't get the docs fully building, but handles a lot of it.

Issue #1330, opened 2021-11-08
Algebraic equations are delayed one step too much in DiscreteSystems


baggepinnen:
In the following example, the variables `x` and `y` are equal, but `y` is delayed one step compared to `x` in the solution.
```julia
# test discrete integrator
@variables t x(t)=0 y(t)=0 [output=true]
Dₜ = Difference(t; dt=dt)
eqs = [
    Dₜ(x) ~ x + 1
    y ~ x
]
@named sys = DiscreteSystem(eqs, t)
prob = DiscreteProblem(sys, Pair[], (0.0, 10.0))
sol = solve(prob, FunctionMap())
yind = findfirst(isequal(y),states(sys)) # due to https://github.com/SciML/ModelingToolkit.jl/issues/1329
xind = findfirst(isequal(x),states(sys))
@test Array(sol)[yind,:] == Array(sol)[xind,:] # fails
@test Array(sol)[yind,2:end] == Array(sol)[xind,1:end-1] # passes, but shouldn't
```

baggepinnen:
There are multiple problems here
1. If the system is not simplified, algebraic equations are not moved into `observed` and they are all passed to `DiscreteFunction` which assumes all equations are on the form `x(t+1)  ~ f(x(t))`, causing the delay.
2. Simplification of the discrete system does not do the right thing because `initialize_system_structure` assumes an `ODESystem`, and only looks for `Differential` variables. The call `vars!(vars, eq.rhs)` defaults to `vars!(vars, eq.rhs; op=Differential)`, but adding another `vars!(vars, eq.rhs; op=Difference)` does not solve the problem. 

@YingboMa could there be some modifications to algorithms that should work for both differential and difference equations such that they operate on some common supertype of `Differential` and `Difference`? And similarly calls to functions `isdifferential, isdifference` would call `isdiff(x) = isdifferential(x) || isdifference(x)`?

PR #2507, opened 2024-02-28
feat: initial implementation of new `DiscreteSystem`

AayushSabharwal:
## Checklist


src/systems/discrete_system/discrete_system.jl
```diff
@@ -0,0 +1,367 @@
...
+$(TYPEDEF)
+A system of difference equations.
+# Fields
+$(FIELDS)
+# Example
+```
+using ModelingToolkit
+@parameters σ=28.0 ρ=10.0 β=8/3 δt=0.1
+@variables t x(t)=1.0 y(t)=0.0 z(t)=0.0
+D = Difference(t; dt=δt)
```
baggepinnen, line 11: Needs update?

src/systems/discrete_system/discrete_system.jl
```diff
@@ -0,0 +1,367 @@
...
+    """Dependent (state) variables. Must not contain the independent variable."""
+    unknowns::Vector
+    """Parameter variables. Must not contain the independent variable."""
+    ps::Vector
+    """Time span."""
+    tspan::Union{NTuple{2, Any}, Nothing}
+    """Array variables."""
+    var_to_name::Any
+    """Control parameters (some subset of `ps`)."""
+    ctrls::Vector
```
baggepinnen, line 38: should go away?

src/systems/discrete_system/discrete_system.jl
```diff
@@ -0,0 +1,367 @@
...
+    """
+    Cached data for fast symbolic indexing.
+    """
+    index_cache::Union{Nothing, IndexCache}
+    """
+    The hierarchical parent system before simplification.
+    """
+    parent::Any
+
+    function DiscreteSystem(tag, discreteEqs, iv, dvs, ps, tspan, var_to_name, ctrls,
```
baggepinnen, line 96: ctrls

src/systems/discrete_system/discrete_system.jl
```diff
@@ -0,0 +1,367 @@
...
+            tearing_state, substitutions, complete, index_cache, parent)
+    end
+end
+
+"""
+    $(TYPEDSIGNATURES)
+Constructs a DiscreteSystem.
+"""
+function DiscreteSystem(eqs::AbstractVector{<:Equation}, iv, dvs, ps;
+        controls = Num[],
```
baggepinnen, line 125:
```suggestion
```

baggepinnen:
The old discrete systems had a lot of correctness issues. Some of the issues reporting those have been closed since the old stuff was removed, but it would probably still be good to add the reported MWEs to the tests. Looking through the old issues, many were related to algebraic variables, such as
- https://github.com/SciML/ModelingToolkit.jl/issues/1330

Perhaps adding an error if not all equations are difference equations would be sufficient for now?

src/systems/systems.jl
```diff
@@ -42,6 +42,13 @@ end
 function __structural_simplify(sys::AbstractSystem, io = nothing; simplify = false,
         kwargs...)
     sys = expand_connections(sys)
+    # if sys isa DiscreteSystem
+    #     eqs = difference_order_lowering(sys)
+    #     @set! sys.unknowns = [eq.lhs for eq in eqs]
+    #     @set! sys.eqs = eqs
+    #     @show parameters(sys)
+    #     return sys
+    # end
```
ChrisRackauckas, line 51: what's this commented for?

AayushSabharwal: That should have been deleted

AayushSabharwal: Fixed in https://github.com/SciML/ModelingToolkit.jl/pull/2507/commits/e59639325e9db3bd5105b33851812a737c3629a3

docs/src/tutorials/discrete_system.md
```diff
@@ -0,0 +1,36 @@
...
+u0 = [S => 990.0, I => 10.0, R => 0.0]
+p = [β => 0.05, c => 10.0, γ => 0.25, δt => 0.1]
+tspan = (0.0, 100.0) # value function (from Symbolics) is used to convert a Num to Float64
+prob = DiscreteProblem(sys, u0, tspan, p)
+sol = solve(prob, FunctionMap())
+```
+
+Currently, all updates happen at integer intervals (`t = 1, 2, 3...`). All shifts
+in the equations must be positive. For example, instead of specifying the update equation
+for the Fibonacci series as `x ~ x(k-1) + x(k-2)` use `x(k+2) ~ x(k+1) + x`.
```
ChrisRackauckas, line 36: Let's not touch that right now 😅 . Just remove that from the docs

docs/src/tutorials/discrete_system.md
```diff
@@ -0,0 +1,36 @@
+# Modeling Discrete Systems
```
ChrisRackauckas, line 1: Mark as experimental until the indexing thing is solidified.

src/discretedomain.jl
```diff
@@ -28,7 +28,7 @@ struct Shift <: Operator
     Shift(t, steps = 1) = new(value(t), steps)
 end
 Shift(steps::Int) = new(nothing, steps)
-normalize_to_differential(s::Shift) = Differential(s.t)^s.steps
+normalize_to_differential(s::Shift) = Differential(s.t)^abs(s.steps)
```
YingboMa, line 31: This is not correct. It changes `u(k)~u(k-2)-u(k-1)` to `u(k)~u(k+2)-u(k+1)` which obviously isn't the same.

YingboMa (commented):
Let's support non-positive shift properly before merging.

AayushSabharwal:
https://github.com/SciML/ModelingToolkit.jl/pull/2507/commits/d9891cbbaad21e89766a3843c98c916e7f4b987f fixes a bug introduced by #2537

docs/src/tutorials/discrete_system.md
```diff
@@ -0,0 +1,36 @@
...
+
+# Equations
+eqs = [S(k + 1) ~ S - infection * h,
+    I(k + 1) ~ I + infection - recovery,
+    R(k + 1) ~ R + recovery]
+@mtkbuild sys = DiscreteSystem(eqs, t)
+
+u0 = [S => 990.0, I => 10.0, R => 0.0]
+p = [β => 0.05, c => 10.0, γ => 0.25, δt => 0.1]
+tspan = (0.0, 100.0) # value function (from Symbolics) is used to convert a Num to Float64
```
ChrisRackauckas, line 29: what does this comment mean?

docs/src/tutorials/discrete_system.md
ChrisRackauckas: Is this still correct?

src/discretedomain.jl
ChrisRackauckas: bump

docs/src/tutorials/discrete_system.md
AayushSabharwal: It means I copied the example from tests and didn't notice the comment 😅

docs/src/tutorials/discrete_system.md
AayushSabharwal: Fixed

docs/src/tutorials/discrete_system.md
AayushSabharwal: Fixed

src/discretedomain.jl
AayushSabharwal: Fixed

docs/src/tutorials/discrete_system.md
AayushSabharwal: Fixed

AayushSabharwal:
Codecov is _very_ unhappy for some reason?

ChrisRackauckas:
Something happened with the docs example? https://github.com/SciML/ModelingToolkit.jl/actions/runs/8264109678/job/22607005589?pr=2507#step:6:1747

AayushSabharwal:
Docs don't want to build for some reason.

I also found a minor bug, but it is caused by the changes in this PR so merging this will have to wait a little bit

AayushSabharwal:
Technically this requires https://github.com/JuliaSymbolics/Symbolics.jl/pull/1090

ChrisRackauckas (approved):

YingboMa (changes requested):
We cannot merge this before it's uses ONLY non-positive shifts.

AayushSabharwal:
Non-positive shifts are enforced. Only thing that's left is to make `observed` work with discrete-time variables

src/structural_transformation/utils.jl
```diff
@@ -412,3 +412,41 @@ function numerical_nlsolve(f, u0, p)
...
+end
+
+function isdoubleshift(var)
+    return ModelingToolkit.isoperator(var, ModelingToolkit.Shift) &&
+           ModelingToolkit.isoperator(arguments(var)[1], ModelingToolkit.Shift)
+end
+
+function simplify_shifts(var)
+    ModelingToolkit.hasshift(var) || return var
+    r = @rule ~x::isdoubleshift => begin
```
YingboMa, line 436: I think the rule system drops metadata. Are you sure that this doesn't cause problems later on?

AayushSabharwal: I'll check if it drops metadata. In case it does, I'll rewrite this as a recursive function

src/structural_transformation/symbolics_tearing.jl
```diff
@@ -437,11 +437,12 @@ function tearing_reassemble(state::TearingState, var_eq_matching;
             # We cannot solve the differential variable like D(x)
             if isdervar(iv)
                 order, lv = var_order(iv)
-                dx = D(lower_varname(fullvars[lv], idep, order - 1))
-                eq = dx ~ ModelingToolkit.fixpoint_sub(
+                dx = D(simplify_shifts(lower_varname_withshift(
```
YingboMa, line 440: Why is the `lower_varname_withshift` inside a `isdervar` branch?

YingboMa: Ah, that's the structural info.

src/systems/diffeqs/odesystem.jl
```diff
@@ -378,14 +378,15 @@ function build_explicit_observed_function(sys, ts;
         checkbounds = true,
         drop_expr = drop_expr,
         ps = full_parameters(sys),
+        op = Union{Differential, Shift},
```
YingboMa, line 381: Why not just `Operator`?

src/structural_transformation/utils.jl
AayushSabharwal: Fixed in https://github.com/SciML/ModelingToolkit.jl/pull/2507/commits/33da074d039e47ed7d46e0854dabf42f8a2c2673

YingboMa:
@baggepinnen could you review the updated tests?

src/structural_transformation/utils.jl
```diff
@@ -412,3 +412,39 @@ function numerical_nlsolve(f, u0, p)
...
+end
+
+function isdoubleshift(var)
+    return ModelingToolkit.isoperator(var, ModelingToolkit.Shift) &&
+           ModelingToolkit.isoperator(arguments(var)[1], ModelingToolkit.Shift)
+end
+
+function simplify_shifts(var)
+    ModelingToolkit.hasshift(var) || return var
+    if isdoubleshift(var)
```
YingboMa, line 436: What about triple shifts? Should we recurse on that, too?

docs/src/tutorials/discrete_system.md
```diff
@@ -0,0 +1,45 @@
...
+@mtkbuild sys = DiscreteSystem(eqs, t)
+
+u0 = [S(k - 1) => 990.0, I(k - 1) => 10.0, R(k - 1) => 0.0]
+p = [β => 0.05, c => 10.0, γ => 0.25, δt => 0.1]
+tspan = (0.0, 100.0)
+prob = DiscreteProblem(sys, u0, tspan, p)
+sol = solve(prob, FunctionMap())
+```
+
+All shifts must be negative. If default values are provided, they are treated as the value
```
baggepinnen, line 35:
```suggestion
All shifts must be non-positive, i.e., discrete-time variables may only be indexed at index `k, k-1, k-2, ...`. If default values are provided, they are treated as the value
```

test/clock.jl
```diff
@@ -89,40 +89,31 @@ d = Clock(t, dt)
...
 eqs = [yd ~ Sample(t, dt)(y)
-       ud ~ kp * (r - yd) + z(k)
-       r ~ 1.0
+       ud ~ kp * ud(k - 1) + yd
 
        # plant (time continuous part)
        u ~ Hold(ud)
        D(x) ~ -x + u
-       y ~ x
-       z(k + 2) ~ z(k) + yd
```
baggepinnen, line 105: the purpose of this contrived example was to test multi-step shifts, I don't see any multi-step tests anymore?

AayushSabharwal:
I can't replicate the Optimization bugs, even with the same environment checked out

test/clock.jl
AayushSabharwal: Fixed

docs/src/tutorials/discrete_system.md
AayushSabharwal: Fixed

src/structural_transformation/utils.jl
AayushSabharwal: Fixed

test/clock.jl
```diff
@@ -89,40 +89,31 @@ d = Clock(t, dt)
...
 @named sys = ODESystem(eqs, t)
 ss = structural_simplify(sys);
 
 Tf = 1.0
 prob = ODEProblem(ss, [x => 0.0, y => 0.0], (0.0, Tf),
-    [kp => 1.0; z => 3.0; z(k + 1) => 2.0])
-@test sort(vcat(prob.p...)) == [0, 1.0, 2.0, 3.0, 4.0] # yd, kp, z(k+1), z(k), ud
+    [kp => 1.0; ud(k - 1) => 2.0; ud(k - 2) => 2.0])
+@test sort(vcat(prob.p...)) == [0, 1.0, 2.0, 2.0, 2.0] # yd, Hold(ud), kp, ud(k - 1)
 sol = solve(prob, Tsit5(), kwargshandle = KeywordArgSilent)
```
baggepinnen, line 111: I added this test here
```julia
#      kp  yd(0)  ud(k-1)
ud_0 = 1 * 0.1    + 2; # This computes what ud should be at time t = 0
@test sol(0+eps(), Val{1})[] ≈ -sol(0+eps(), idxs=x) + ud_0
```
it fails. I check that 
`D(x)(t=0)` is correctly calculated (I got errors trying to access `sol(0, idxs=D(x))` by adding `-x(0)` and `ud(0)`

AayushSabharwal: So this discrepancy is because the init function returned by clock inference codegen needs to also perform a discrete update, since the unknowns are now `ud(k-1)` and `ud(k-2)` (past values). Currently, it interprets `ud(k-1)` as the value of `ud` at `t=0`. This sort of discrepancy comes about because the unknowns are named `ud(k-1)` and `ud(k-2)` which implies that they are past values. However, after the `affect!` function is run, the value of `ud(k-1)` is the value of `ud` _at the current time_. Problem initialization assumes that the initial values are the ones after `affect!` has been run, so to speak

AayushSabharwal: With this interpretation, the value of `1.9` returned by `sol(0+eps(), Val{1})` in this example is correct, since `ud` at the current time is considered to be `2.0` and `x` is `0.1`

baggepinnen: > Currently, it interprets ud(k-1) as the value of ud at t=0

Then we should change this. At each tick of the clock, `x(k)` refers to the new value at the tick time, so `ud(k)` refers to the value of `ud` _after_ the clock has ticked at t = 0, while `ud(k-1)` refers to the first past value (before t=0). 

The shift-index notation $x(k)$ is standard in the literature of discrete-time systems and is short for $x(t_0 + k \Delta T)$ where $\Delta T$ is the tick interval of the clock. A discrete-time difference equation describes something that holds true for all $k$, if we insert the start time $t_0 = 0$ and $k=0$ in the equation for `ud`, we get the value of `ud` at the initial time `ud(0) = kp*yd(0) + ud(-1)`

The reason I took the time to clone this PR and add this test is that these details are extremely important. The old discrete-time functionality that we threw away had a lot of small nuances like this left undocumented and untested, and as a result it was completely impossible to rely on it for any serious real-world use case. I want to avoid going back to the same situation again, it's worth spending the time to get this right :)

baggepinnen: > This sort of discrepancy comes about because the unknowns are named ud(k-1) and ud(k-2) 

This is an implementation detail and should not influence how the system behaves from the perspective of the user. We can call the unknowns `Aayush` and `Fredrik` if it makes it less confusing :) 

The choice of state `ud(k-1), ud(k-2) ` is a slightly strange, we are solving for `ud(k)`, so I would probably have chosen this variable to be part of the state rather than the two past values, c.f.
https://juliahub.slack.com/archives/C03UEC0P6G0/p1710482554351029?thread_ts=1710481464.183949&cid=C03UEC0P6G0

AayushSabharwal: > Then we should change this

yeah that's what I'm working on right now

> The choice of state ud(k-1), ud(k-2) is a slightly strange, we are solving for ud(k), so I would probably have chosen this variable to be part of the state rather than the two past values

It's somewhat of an artifact of using nonpositive shifts. Consider the continuous-time case, where `D(D(x)) ~ f(D(x), x)`. We take `x` and `D(x)` (the lowest order terms) as the unknowns, and compute `D(D(x))`. `D(D(x))` isn't actually a symbolic variable in this case. Similarly, for `x(k) ~ f(x(k-1), x(k-2))` we take `x(k-1) and `x(k-2)` as the unknowns, and compute `x(k)` which is now reinterpreted as "the next value of `x(k-1)`".

AayushSabharwal:
The optimization errors seem to be due to bugs upstream in the Optimization stack. The generated functions look good to me, it's probably an error with how they're called

test/clock.jl
AayushSabharwal: Fixed now

baggepinnen: sweet :) 

When we get to the stage where we can access discrete-time variables from the solution, I will be equally picky with exactly what index is obtained. `sol(k*dt, idxs=ud)` should get `ud` as computed for tick `k`, even if this variable happens to be called `u(k-1)` internally. When the solution object is plotted, the plot label should also refer to `u(t0 + k*dt)` as `u(k)`.

AayushSabharwal: I think that shouldn't be too difficult

baggepinnen: > It's somewhat of an artifact of using nonpositive shifts. Consider the continuous-time case

Is this analogy doing us any service? 

>  Similarly, for x(k) ~ f(x(k-1), x(k-2)) we take x(k-1) and x(k-2) as the unknowns, and compute x(k) which is now reinterpreted as "the next value of x(k-1)".

Just a comment on the terminology here. The fundamental _unknown_ is the discrete-time _function_ `x(k)` which is a function of the iteration index `k`. There is only one function `x`, and this function is defined on the domain `k \in {..., -1, 0, 1, 2, ...}`. In order to solve for this unknown function, we choose a _state realization_ comprised out of the _state variables_ `x(k-1), x(k-2)`. Here is where our discussion might become a bit confusing; It's quite common to change the names of the state variables completely into something more like
```
X(k) = [X1(k) X2(k)] = [x(k-1), x(k-2)]
```
so that you can speak about the (singular) _state_ at time index `k` being `X(k)`, and this state is comprised of the state variables `X(k) =  [X1(k) X2(k)] =[x(k-1), x(k-2)]`. In this discussion so far, we have conflated the unknown we are solving for with the state realization we have chosen in order to do so. The reason that these two are typically separated is partly to avoid this confusion.

Somewhat confusingly, MTK introduces new "unknowns" to represent the same function`x` but indexed at different points in time, like `x(k-1)` and `x(k-2)`, but I consider these unknowns computational tool rather than a property of the system being modeled. 

The terms "state" and "unknowns" are not interchangeable, which was the fundamental reason we made the switch from "states" (the plural form was wrong on so many levels) to "unknowns". After structural simplify, the term "state" is more descriptive, since structural simplification can be seen as the selection of a state realizaition among all the unknowns. 

> I think that shouldn't be too difficult

Probably not, but will it forever haunt us in the sense that every person looking into this in the future will be confused by the presence of a one-step shift in many places?


Sorry for all the long text here, I just want to make sure we fully understand each other very precisely and end up with a nice and maintainable solution that will be easy to communicate precisely to users so that we avoid the long list of bugs and confusions that plagued the old implementation.

test/discrete_system.jl
```diff
@@ -0,0 +1,219 @@
...
+@test_throws ErrorException @mtkbuild sys = DiscreteSystem([x(k + 1) ~ x + x(k - 1)], t)
+
+@inline function rate_to_proportion(r, t)
+    1 - exp(-r * t)
+end;
+
+# Independent and dependent variables and parameters
+@parameters c nsteps δt β γ
+@constants h = 1
+@variables S(t) I(t) R(t)
```
baggepinnen, line 22: If I modify this problem to add an observable
```julia
@variables S(t) I(t) R(t) R2(t)

# Equations
eqs = [S ~ S(k - 1) - infection * h,
    I ~ I(k - 1) + infection - recovery,
    R ~ R(k - 1) + recovery,
    R2 ~ R]
```
and then try to plot the observed variable it does not work
```julia
julia> plot(sol_map, idxs=[R, R2])
ERROR: ArgumentError: R2(t) is neither an observed nor an unknown variable.
```

Will this somehow be handled?

AayushSabharwal: If you check `observed(sys)` you'll notice that it'll be something like `R2(k+1) ~ R + recovery` which is an artifact of shifting the entire system forward one timestep. Plotting `R2(k+1)` should work. While it's not great user API, it's the only way I found of making this correct, since in general we only have the capability of calculating the next state given the current state, even for observed variables.

Semantically, everything for `DiscreteSystem` should be correct. The saved values correspond to the states (i.e. `S(k)` is chosen to be a state, and the saved timeseries is for `S(k)`) and the observed equations are correct (albeit for the next timestep). Plotting `R2(k+1)` should plot a shifted `R(k)`. I haven't tested this, which I'll do and add to the testcases.

ChrisRackauckas: Can you open an issue about the user interface piece here and we merge?

baggepinnen (approved):
:tada:

docs/src/tutorials/discrete_system.md
```diff
@@ -0,0 +1,51 @@
+# (Experimental) Modeling Discrete Systems
```
ChrisRackauckas, line 1: Is it still experimental?

baggepinnen: kind of, unless observed variables are working as intended now?

AayushSabharwal: Yeah, since shifting the entire system by one step is basically a workaround and eventually we'll need proper compiler support for this. I've discussed with @baggepinnen and that will end up changing things such as the state realization and observed equations in the simplified system.

ChrisRackauckas:
Test failures

AayushSabharwal:
Fixed and rebased

AayushSabharwal:
CI failures are due to codecov

AayushSabharwal:
Finally 🥳

Issue #2508, opened 2024-02-28
Non-DAE ODEs Bypass Initialization


bradcarman:
I'm attempting to achieve the same idea as the Modelica `initial equation` feature.  If I have the following:

```
Real dx, ddx;

equation
der(dx) = ddx;
0 = ddx + dx + 1;
```

I can use either 

```
initial equation
dx=1;
```

Or

```
initial equation
ddx=-2;
```

And Modelica can sort out the correct initial value for `dx` or `ddx`.  With the new initialization system in MTK, I was assuming I could do this like...

```julia
using ModelingToolkit, OrdinaryDiffEq, Test
using ModelingToolkit: t_nounits as t, D_nounits as D

function System(;name)
    vars = @variables begin
        dx(t), [guess=0]
        ddx(t), [guess=0]
    end
    eqs = [
        D(dx) ~ ddx
        0 ~ ddx + dx + 1
    ]
    return ODESystem(eqs, t, vars, []; name)
end

@mtkbuild sys = System()
prob = ODEProblem(sys, [sys.dx => 1], (0,1)) # OK
prob = ODEProblem(sys, [sys.ddx => -2], (0,1)) # ERROR: ArgumentError: SymbolicUtils.BasicSymbolic{Real}[dx(t)] are missing from the variable map.
```
My assumption was the `dx(t)` had a guess value and would be solved with the initialization system.  Am I mistaken?

bradcarman:
All I can sort out so far is that I can do the following to get this functionality.  Trying to sort out how to do this in a cleaner way, but at least we can see the initialization feature I'm looking for here is possible...

```julia
using ModelingToolkit, OrdinaryDiffEq, Test
using ModelingToolkit: t_nounits as t, D_nounits as D
using DifferentialEquations

function System(;name, dx_=nothing, ddx_=nothing, f=1, m=1, d=1)
    pars = @parameters begin
        m = m
        d = d
        f = f
    end
    vars = @variables begin
        dx(t)
        ddx(t)
    end
    defaults = Dict()
    isnothing(dx_) || push!(defaults, dx => dx_)
    isnothing(ddx_) || push!(defaults, ddx => ddx_)
    eqs = [
        D(dx) ~ ddx
        0 ~ m*ddx + d*dx + f
    ]
    return ODESystem(eqs, t, vars, pars; name, defaults)
end

@named odesys1 = System(; dx_=1)
@named odesys2 = System(; ddx_=-2)

[...]

PR #2512, opened 2024-02-29
Initialization on non-DAE models

ChrisRackauckas:
https://github.com/SciML/ModelingToolkit.jl/issues/2508 pointed out that the heuristic of `implicit_dae || calculate_massmatrix(sys) !== I`, i.e. "only do initialization on DAEs", while sensible for the DAE model, doesn't quite fit in the MTK context. Instead, what we need to do is check whether the varmap is correct and consistent for the chosen ODE. If it's not consistent for the selected ODE, then we still need to run an initialization step for the resulting ODE solve.

This needs a few changes in the ODE solver as though, since currently only DAE solvers will run the initialization step. Thus the test case:

```julia
using ModelingToolkit, OrdinaryDiffEq, Test

function System(;name)
    vars = @variables begin
        dx(t), [guess=0]
        ddx(t), [guess=0]
    end
    eqs = [
        D(dx) ~ ddx
        0 ~ ddx + dx + 1
    ]
    return ODESystem(eqs, t, vars, []; name)
end

@mtkbuild sys = System()
prob = ODEProblem(sys, [sys.dx => 1], (0,1)) # OK
prob = ODEProblem(sys, [sys.ddx => -2], (0,1), guesses = [sys.dx => 1])
sol = solve(prob, Rodas5P())
sol = solve(prob, Tsit5())
```

gives an erroneous success as it skips initialization, using `prob.u0 == [0.0]`, the sentinel value since initialization is only run on DAEs. We will need to override that so that initialization is always run on any system that provides an initializeprob.

Fixes https://github.com/SciML/ModelingToolkit.jl/issues/2508


ChrisRackauckas:
This also coincidentally adds steady state initializations as almost a free piece, and late binding initialization_eqs

ChrisRackauckas:
Ran locally since CI is taken over by codecov bumps. Passes!

PR #2531, opened 2024-03-10
Add `getindex`/`setindex!` methods for `MTKParameters` with `ParameterIndex`

MasonProtter:
Without this method, we hit the generic methods for `MTKParameters` which assumes `ParameterIndex` is an integer

## Checklist

## Additional context

### DIFF
--- a/src/systems/parameter_buffer.jl
+++ b/src/systems/parameter_buffer.jl
@@ -362,12 +362,12 @@ function Base.setindex!(p::MTKParameters, val, i)
 end
 
 function Base.getindex(p::MTKParameters, pind::ParameterIndex)
-    (;portion, idx) = pind
+    (; portion, idx) = pind
     i, j, k... = idx
     if isempty(k)
         indexer = (v) -> v[i][j]
     else
-        indexer = (v) -> v[i][j][k...] 
+        indexer = (v) -> v[i][j][k...]
     end
     if portion isa SciMLStructures.Tunable
         indexer(p.tunable)
--- a/src/systems/parameter_buffer.jl
+++ b/src/systems/parameter_buffer.jl
@@ -361,6 +361,41 @@ function Base.setindex!(p::MTKParameters, val, i)
     end
 end
 
+function Base.getindex(p::MTKParameters, pind::ParameterIndex)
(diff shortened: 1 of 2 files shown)


AayushSabharwal (changes requested):
`pind.idx` is not necessarily a 2-tuple. It can also have extra indices in case `p.portion[i][j]` is an array. For example, for `@parameters p[1:3]` you might have a `ParameterIndex(Tunable(), (2, 1, 2))` to refer to `p[2]`.

MasonProtter:
I see, thank you. does this change I just pushed make more sense?

src/systems/parameter_buffer.jl
```diff
@@ -361,6 +361,52 @@ function Base.setindex!(p::MTKParameters, val, i)
...
+        setindexer = (v) -> v[i][j][k...] = val 
+    end
+    if portion isa SciMLStructures.Tunable
+        setindexer(p.tunable)
+    elseif portion isa SciMLStructures.Discrete
+        setindexer(p.discrete)
+    elseif portion isa SciMLStructures.Constants
+        setindexer(p.constant)
+    elseif portion === DEPENDENT_PORTION
+        setindexer(p.dependent)
```
AayushSabharwal, line 402:
```suggestion
```
The dependent portion shouldn't be set directly

AayushSabharwal (changes requested):
I forgot this last time around 😅 but I think `setindex!` should just call `set_parameter!`. It handles the extra indexes, doesn't allow setting the dependent portion (as intended) and updates the dependent values.

src/systems/parameter_buffer.jl
MasonProtter: Okay, I'd switched it to just fall back to `set_parameter`

AayushSabharwal (approved):
LGTM, but I don't have permissions to make CI run.

@ChrisRackauckas could you please approve CI

ChrisRackauckas:
Needs formatting

MasonProtter:
I added a basic testset to try and cover all the branches. Let me know if that's appropriate.

ChrisRackauckas:
Format's last PR

PR #2540, opened 2024-03-12
Add guesses to `linearize`

ven-k:
Adds the guesses of the sys to x0 (Preference = guess < defaults < op.)

While here, fixes the metadata in mtkmodel `@defaults` for parameters of subcomponents.

## Checklist

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -92,5 +92,3 @@ end
         @safetestset "BifurcationKit Extension Test" include("extensions/bifurcationkit.jl")
     end
 end
-
-@safetestset "Model Parsing Test" include("model_parsing.jl")


src/systems/abstractsystem.jl
```diff
@@ -1997,7 +1997,7 @@ lsys_sym, _ = ModelingToolkit.linearize_symbolic(cl, [f.u], [p.x])
 """
 function linearize(sys, lin_fun; t = 0.0, op = Dict(), allow_input_derivatives = false,
         p = DiffEqBase.NullParameters())
-    x0 = merge(defaults(sys), op)
+    x0 = merge(defaults(sys), Dict(missing_variable_defaults(sys)), op)
```
ven-k, line 2000: I've matched with `linearization_function`, which doesn't account for guesses either.

Should both be instead
```julia
x0 = merge(defaults(sys), get_guesses(sys), Dict(missing_variable_defaults(sys)), op)
```

test/downstream/linearize.jl
```diff
@@ -19,6 +19,24 @@ lsys, ssys = linearize(sys, [r], [y])
...
+@parameters kp = 1
+D = Differential(t)
+
+eqs = [u ~ kp * (r - y)
+       D(x) ~ -x + u
+       y ~ x]
+
+@named sys = ODESystem(eqs, t)
+
+lsys, ssys = linearize(sys, [r], [y])
```
ChrisRackauckas, line 33: This should error as most of the values are not defined. It's not clear what to linearize at.

test/downstream/linearize.jl
```diff
@@ -19,6 +19,24 @@ lsys, ssys = linearize(sys, [r], [y])
 @test lsys.C[] == 1
 @test lsys.D[] == 0
 
+# With guesses
+@variables t x(t) [guess = 0] y(t) [guess = 0] u(t) [guess = 0] r(t)=0.0 [input = true]
```
ChrisRackauckas, line 23:
```suggestion
@variables t x(t) [guess = 0] y(t) [guess = 1] u(t) [guess = 0] r(t)=0.0 [input = true]
```

This is a much better test case which then highlights the issue with your proposed approach.

ven-k: Yes; will add non zero guesses.

PR #2544, opened 2024-03-13
refactor: implement new additions to SII interface

AayushSabharwal:
Also some bug fixes and test improvements

## Checklist


src/systems/systemstructure.jl
```diff
@@ -630,7 +630,7 @@ function _structural_simplify!(state::TearingState, io; simplify = false,
     if has_io
         ModelingToolkit.markio!(state, orig_inputs, io...)
     end
-    if io !== nothing
+    if io !== nothing || any(isinput, state.fullvars)
```
baggepinnen, line 633: This might (almost certain) have unintended consequences. `input = true` does not have the same meaning as being part of `io`.

baggepinnen: for example, all `RealInput` in MTKstdlib are marked as `input = true`, but they should **not** be considered `io` unless explicitly requested by the user

baggepinnen: hmm, I seem to have been a bit confused about how it works internally here, which test was it that failed?

AayushSabharwal: https://github.com/SciML/ModelingToolkit.jl/blob/539fe81a800a8fc559234ba3cfa50720ab5256f1/test/input_output_handling.jl#L53 errors

baggepinnen: A few days ago, this code was actually being run unconditionally on `io == nothing`, so I don't think it will break anything. Also for MTKstdlib, it's probably more correct not run the `inputs_to_parameters` unless io is provided, in case we want to make more use of the `input= true` metadata in the future. When the test was written, I think `input = true` was the only way to indicate an input, but since we added the io argument we changed approach to let the user-provided io indicate the externally visible inputs and outputs, while the `input = true` metadata lost most of its meaning. the reason is that whether something is an externally visible input is a property of how a component is used, rather than an inherent property of the component. Hence, specifying this information inside component does not make sense, and we instead pass io when we have assembled the whole system

AayushSabharwal: https://github.com/SciML/ModelingToolkit.jl/pull/2547

PR #2545, opened 2024-03-13
Fix `_varmap_to_vars`, re-enable dde.jl test

MasonProtter:
Closes https://github.com/SciML/ModelingToolkit.jl/issues/2543

cc @AayushSabharwal 

It turns out the problem was actually just that we weren't putting symbolic constraints into the `values` dict, it wasn't about `t` being missing from the initial `u0map` (adding that gives the wrong answer actually).  I thought I had checked this before but I must have done something else that was breaking alongside it, so this took me way longer to track down than it should have.

This has made it so I can now re-enable the `dde.jl` test file which was previously excluded from the test suite. 

## Checklist

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -115,6 +115,7 @@ AmplNLWriter = "7c4d4715-977e-5154-bfe0-e096adeac482"
 BenchmarkTools = "6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf"
 ControlSystemsBase = "aaaaaaaa-a6ca-5380-bf3e-84a91bcd477e"
 DeepDiffs = "ab62b9b5-e342-54a8-a765-a90f495de1a6"
+DelayDiffEq = "bcd4f6db-9728-5f36-b5f7-82caef46ccdb"
 ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
 Ipopt = "b6b21f68-93f8-5de0-b562-5493be1d77c9"
 Ipopt_jll = "9cc047cb-c261-5740-88fc-0cf96f7bdcc7"
@@ -136,4 +137,4 @@ Sundials = "c3572dad-4567-51f8-b174-8c6c989267f4"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["AmplNLWriter", "BenchmarkTools", "ControlSystemsBase", "NonlinearSolve", "ForwardDiff", "Ipopt", "Ipopt_jll", "ModelingToolkitStandardLibrary", "Optimization", "OptimizationOptimJL", "OptimizationMOI", "Random", "ReferenceTests", "SafeTestsets", "StableRNGs", "Statistics", "SteadyStateDiffEq", "Test", "StochasticDiffEq", "Sundials", "StochasticDelayDiffEq", "Pkg"]
+test = ["AmplNLWriter", "BenchmarkTools", "ControlSystemsBase", "DelayDiffEq", "NonlinearSolve", "ForwardDiff", "Ipopt", "Ipopt_jll", "ModelingToolkitStandardLibrary", "Optimization", "OptimizationOptimJL", "OptimizationMOI", "Random", "ReferenceTests", "SafeTestsets", "StableRNGs", "Statistics", "SteadyStateDiffEq", "Test", "StochasticDiffEq", "Sundials", "StochasticDelayDiffEq", "Pkg"]
--- a/src/systems/diffeqs/abstractodesystem.jl
+++ b/src/systems/diffeqs/abstractodesystem.jl
@@ -957,7 +957,6 @@ function process_DEProblem(constructor, sys::AbstractODESystem, u0map, parammap;
         ddvs = nothing
     end
     check_eqs_u0(eqs, dvs, u0; kwargs...)
-
     f = constructor(sys, dvs, ps, u0; ddvs = ddvs, tgrad = tgrad, jac = jac,
         checkbounds = checkbounds, p = p,
         linenumbers = linenumbers, parallel = parallel, simplify = simplify,
@@ -1243,6 +1242,8 @@ function DiffEqBase.SDDEProblem{iip}(sys::AbstractODESystem, u0map = [],
     h_oop, h_iip = generate_history(sys, u0)
     h(out, p, t) = h_iip(out, p, t)
     h(p, t) = h_oop(p, t)
+    h(p::MTKParameters, t) = h_oop(p..., t)
+    h(out, p::MTKParameters, t) = h_iip(out, p..., t)
     u0 = h(p, tspan[1])
     cbs = process_events(sys; callback, kwargs...)
     if has_discrete_subsystems(sys) && (dss = get_discrete_subsystems(sys)) !== nothing
--- a/src/variables.jl
+++ b/src/variables.jl
@@ -197,7 +197,7 @@ function _varmap_to_vars(varmap::Dict, varlist; defaults = Dict(), check = false
     for var in varlist
         var = unwrap(var)
         val = unwrap(fixpoint_sub(var, varmap; operator = Symbolics.Operator))
-        if symbolic_type(val) === NotSymbolic()
+        if !isequal(val, var)
             values[var] = val
         end
     end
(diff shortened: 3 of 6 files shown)


src/variables.jl
```diff
@@ -193,9 +193,7 @@ function _varmap_to_vars(varmap::Dict, varlist; defaults = Dict(), check = false
     for var in varlist
         var = unwrap(var)
         val = unwrap(fixpoint_sub(fixpoint_sub(var, varmap), defaults))
-        if symbolic_type(val) === NotSymbolic()
-            values[var] = val
-        end
+        values[var] = val
```
AayushSabharwal, line 196: Why is this necessary? This could fail in cases where a dependent initialization is not fully defined. For example, if `x => 2y` but `y` is not specified, `val` will be `2y` when we want it to not be a symbolic quantity.

AayushSabharwal (changes requested):

src/variables.jl
MasonProtter: This just restores the v8 behaviour. In v8 we had
```julia
function _varmap_to_vars(varmap::Dict, varlist; defaults = Dict(), check = false,
        toterm = Symbolics.diff2term)
    varmap = merge(defaults, varmap) # prefers the `varmap`
    varmap = Dict(toterm(value(k)) => value(varmap[k]) for k in keys(varmap))
    # resolve symbolic parameter expressions
    for (p, v) in pairs(varmap)
        varmap[p] = fixpoint_sub(v, varmap)
    end

    missingvars = setdiff(varlist, collect(keys(varmap)))
    check && (isempty(missingvars) || throw_missingvars(missingvars))

    out = [varmap[var] for var in varlist]
end
```
which was changed in v9 to only allow `values` to have non-symbolic entries, which caused this bug. 

e.g. in v8, for the test case linked above, `_varmap_to_vars` would take in arguments
```
varmap = Dict{Num, Num}(x₂(t) => 1.0 - 10q1*t, x₀(t) => 1.0 - 10q1*t, x₁(t) => 1.0 - 10q1*t)
varlist = SymbolicUtils.BasicSymbolic{Real}[x₀(t), x₁(t), x₂(t)]
defaults = Dict{Any, Any}(x₂(t) => 1.0 - 10q1*t, d2 => 1, beta0 => 1, v0 => 1, p0 => 0.2, d0 => 5, beta1 => 1, p1 => 0.2, q0 => 0.3, x₁(t) => 1.0 - 10q1*t, q1 => 0.3, d1 => 1, v1 => 1, x₀(t) => 1.0 - 10q1*t)
```
and compute
```julia
out = SymbolicUtils.BasicSymbolic{Real}[1.0 - 3.0t, 1.0 - 3.0t, 1.0 - 3.0t]
```

whereas in v9 we also have it taking in the identical arguments:
```julia
varmap = Dict{Num, Num}(x₂(t) => 1.0 - 10q1*t, x₀(t) => 1.0 - 10q1*t, x₁(t) => 1.0 - 10q1*t)
varlist = SymbolicUtils.BasicSymbolic{Real}[x₀(t), x₁(t), x₂(t)]
defaults = Dict{Any, Any}(x₂(t) => 1.0 - 10q1*t, d2 => 1, beta0 => 1, v0 => 1, p0 => 0.2, d0 => 5, beta1 => 1, p1 => 0.2, q0 => 0.3, x₁(t) => 1.0 - 10q1*t, q1 => 0.3, d1 => 1, v1 => 1, x₀(t) => 1.0 - 10q1*t)
```
and erroring. 

It's totally possible there's another way to fix this, but I *think* what I did is just the minimal change to un-break the change in v9.

src/systems/diffeqs/abstractodesystem.jl
```diff
@@ -1205,7 +1205,7 @@ function DiffEqBase.SDDEProblem{iip}(sys::AbstractODESystem, u0map = [],
     h_oop, h_iip = generate_history(sys, u0)
     h(out, p, t) = h_iip(out, p, t)
     h(p, t) = h_oop(p, t)
-    u0 = h(p, tspan[1])
+    u0 = h(p..., tspan[1])
```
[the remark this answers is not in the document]
MasonProtter: Right, sorry about that, I'll fix that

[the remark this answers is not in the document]
MasonProtter: should be fixed now, let me know if you see anything else potentially problematic

AayushSabharwal:
Please rebase and format so CI runs

AayushSabharwal:
Those commits shouldn't have appeared here after the rebase 😅

MasonProtter:
hmm, not sure what went wrong, I just configured `rebase = true` and pulled from master..

AayushSabharwal:
Try resetting to before the rebase using `git reflog`, then
```git
git checkout master
git pull sciml master # assuming sciml is the remote pointing to this repo
git checkout fix-dde
git rebase master
# fix conflicts
git push -f origin fix-dde
```

usually works for me

MasonProtter:
Aha, yeah I think I just did a bad order of operations or something, thank you for spelling it out.

MasonProtter:
So, comparing with https://github.com/SciML/ModelingToolkit.jl/pull/2548, it seems that there's an overall improvement here with a couple of previously failing suites now failing, and I'm not seeing any new regressions.

AayushSabharwal:
The failures in #2548 that don't fail here are due to codecov being flaky. This PR does have test regressions in `InterfaceI`. See: https://github.com/SciML/ModelingToolkit.jl/actions/runs/8279178750/job/22653091634?pr=2545#step:6:726

ChrisRackauckas:
What is the state of this?

MasonProtter:
Okay, this is finally working and ready to go. The failing downstream test is unrelated, and also present on other current PRs.

Project.toml
```diff
@@ -115,6 +115,7 @@ AmplNLWriter = "7c4d4715-977e-5154-bfe0-e096adeac482"
 BenchmarkTools = "6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf"
 ControlSystemsBase = "aaaaaaaa-a6ca-5380-bf3e-84a91bcd477e"
 DeepDiffs = "ab62b9b5-e342-54a8-a765-a90f495de1a6"
+DelayDiffEq = "bcd4f6db-9728-5f36-b5f7-82caef46ccdb"
```
AayushSabharwal, line 118: This needs a compat bound

test/dde.jl
```diff
@@ -29,21 +31,21 @@ prob2 = DDEProblem(bc_model, u0, h2, tspan, constant_lags = lags)
...
 tau = 1
-D = Differential(t)
 eqs = [D(x₀) ~ (v0 / (1 + beta0 * (x₂(t - tau)^2))) * (p0 - q0) * x₀ - d0 * x₀
        D(x₁) ~ (v0 / (1 + beta0 * (x₂(t - tau)^2))) * (1 - p0 + q0) * x₀ +
                (v1 / (1 + beta1 * (x₂(t - tau)^2))) * (p1 - q1) * x₁ - d1 * x₁
        D(x₂(t)) ~ (v1 / (1 + beta1 * (x₂(t - tau)^2))) * (1 - p1 + q1) * x₁ - d2 * x₂(t)]
-@named sys = System(eqs)
-prob = DDEProblem(sys,
+@named sys = System(eqs, t)
+final_sys = structural_simplify(sys)
```
AayushSabharwal, line 41: Can this just be an `@mtkbuild` instead of `@named`? Or is there a reason we need `sys` and `final_sys` separately?

AayushSabharwal (changes requested):
Just a couple minor comments. Looks good otherwise. Please also rebase this to remove commits such as `retrigger CI`.

Project.toml
MasonProtter: It's only in the testing `[extras]`, it doesn't seem that any of the other extras have compat bounds.

test/dde.jl
MasonProtter: nope, I've switched it to `@mtkbuild`

MasonProtter:
> Please also rebase this to remove commits such as `retrigger CI`.

I've rebased it, but you can also just use the `squash and merge` button.

AayushSabharwal (approved):

PR #2548, opened 2024-03-14
[DO NOT MERGE] trigger CI

MasonProtter:
Just want to cross-compare failure modes with https://github.com/SciML/ModelingToolkit.jl/pull/2545


ChrisRackauckas:
It should just be Downstream 1, which is the linearization thing (@ven-k you were looking to finish that?)

The other X's if you dive in you'll see it's codecov's server that's a bit wonky, but other than that those should pass.

PR #2557, opened 2024-03-19
Fix iteration bounds in `tearing_with_dummy_derivatives`

YingboMa:


YingboMa:
This change is probably unnecessary.

YingboMa:
Turns out that it's necessary.

PR #2567, opened 2024-03-23
Document and export `continuous_events` and `discrete_events`

MasonProtter:
## Checklist

## Additional context

I'd like a blessed API for fetching the callbacks of an ODESystem without reaching into internals, so I figure marking these functions as exported and giving them docstrings would be a good way to go about this. If there's a better, existing way to do that please let me know.

### DIFF
--- a/docs/src/basics/Events.md
+++ b/docs/src/basics/Events.md
@@ -58,6 +58,10 @@ be handled via more [general functional affects](@ref func_affects).
 
 Finally, multiple events can be encoded via a `Vector{Pair{Vector{Equation}, Vector{Equation}}}`.
 
+Given an `AbstractSystem`, one can fetch its continuous events, and the continuous events of any
+subsystem inside of it using the `continuous_events(::AbstractSystem)` method, returning a vector
+of `ModelingToolkit.SymbolicContinuousCallback` objects.
+
 ### Example: Friction
 
 The system below illustrates how continuous events can be used to model Coulomb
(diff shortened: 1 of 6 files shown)


ChrisRackauckas:
Seems reasonable to me. Add it to the docs?

docs/src/basics/Events.md
```diff
@@ -221,6 +225,10 @@ equations can either all change unknowns (i.e. variables) or all change
 parameters, but one cannot currently mix unknown variable and parameter changes within one
 individual event.
 
+Given an `AbstractSystem`, one can fetch its discrete events, and the discrete events of any
```
ChrisRackauckas, line 228: Add it to the system docs for the docstrings of the available functions?

MasonProtter:
Done, let me know if that's a sufficient level of detail. I'm not sure how much we should explain the structure of `SymbolicContinuousCallback` and `SymbolicDiscreteCallback`. Maybe just say something like "these encode the first element of the input `Pair` in the `.eqs` field and the second element of the input `Pair` in the `.affect` field"?

docs/src/basics/Events.md
MasonProtter: Ahh, I see, I missed that section.

MasonProtter: Should I remove the mention of it from `Events.md` and just have it in the system docs?

ChrisRackauckas: Yes

ChrisRackauckas (approved):

ChrisRackauckas:
Odd docs build failure, I'm going to assume that was a freak event, but also do a pass over this repo soon to fix a few red checkmarks.

PR #2570, opened 2024-03-25
feat: implement SII.remake_buffer for MTKParameters, fix bugs, add tests

### LINKED ISSUE #2571: Error with MTKParameters and Dual Numbers
It seems that the new MTKParameters system does not work with Dual Numbers from things like Automatic Differentiation. Trying to follow the advice given in the FAQ for using ModelingToolkit with Optimization/AutomaticDifferentiation results in something akin to `MethodError: no method matching Float64(::ForwardDiff.Dual{ForwardDiff.Tag{OptimizationForwardDiffExt.var"#37#55"{…}, Float64}, Float64, 4})`

```julia
using ModelingToolkit
using DifferentialEquations
using Plots
using Optimization
using OptimizationOptimJL
using SymbolicIndexingInterface
using SciMLStructures: replace!, Tunable, canonicalize

function lotka_volterra(;name)
    @parameters t
    pars = @parameters α β γ δ
    D = Differential(t)
    vars = @variables x(t) y(t)
    eqs = [
        D(x) ~ (α - β * y) * x, # prey
        D(y) ~ (δ * x - γ) * y # predator
    ]
    tspan = (0.0, 10.0)
    sys = ODESystem(eqs, t, vars, pars; tspan = tspan, name = name)
    return sys
end

@mtkbuild lvsys = lotka_volterra();
prob = ODEProblem(lvsys, [lvsys.x => 1.0, lvsys.y => 1.0], lvsys.tspan, [lvsys.α => 1.5, lvsys.β => 1.0, lvsys.γ => 3.0, lvsys.δ => 1.0]);

# Plot simulation.
plot(solve(prob, Tsit5()))

sol = solve(prob, Tsit5(); saveat=0.1);
data = Array(sol) + 0.01 * randn(size(Array(sol)));

function loss(x, p)
    prob = p[1]
    prob = remake(prob, p = [:α => x[1], :β => x[2], :γ => x[3], :δ => x[4]])
    sol = solve(prob, Tsit5(), saveat=0.1);
    return sum((sol .- data).^2)
end

f = OptimizationFunction(loss, Optimization.AutoForwardDiff())
oprob = OptimizationProblem(f, rand(4), [prob]);
sol = solve(oprob, BFGS())
```

```julia
ERROR: MethodError: no method matching Float64(::ForwardDiff.Dual{ForwardDiff.Tag{OptimizationForwardDiffExt.var"#37#55"{…}, Float64}, Float64, 4})

Closest candidates are:
  (::Type{T})(::Real, ::RoundingMode) where T<:AbstractFloat
   @ Base rounding.jl:207
  (::Type{T})(::T) where T<:Number
   @ Core boot.jl:792
  Float64(::IrrationalConstants.Invsqrt2π)
   @ IrrationalConstants ~/.julia/packages/IrrationalConstants/vp5v4/src/macro.jl:112
  ...
```

The other methods in the FAQ (i.e `setter!` and using `SciMLStructures`) results in the same issue. As far as I can tell, it seems that passing a Dual Number into `MTKParameters` causes the problem.

```julia
Pkg.status()
  [0c46a032] DifferentialEquations v7.13.0
  [961ee093] ModelingToolkit v9.7.0
  [7f7a1694] Optimization v3.24.3
  [36348300] OptimizationOptimJL v0.2.3
  [91a5bcdd] Plots v1.40.2
  [1ed8b502] SciMLSensitivity v7.56.2
  [53ae85a6] SciMLStructures v1.1.0
  [2efcf032] SymbolicIndexingInterface v0.3.12
```

AayushSabharwal:
Close #2571 

## Checklist


test/mtkparameters.jl
```diff
@@ -90,3 +91,13 @@ end
...
+
+ps = MTKParameters(sys, ivs)
+function loss(value, sys, ps)
+    @test value isa ForwardDiff.Dual
+    vals = merge(Dict(parameters(sys) .=> getp(sys, parameters(sys))(ps)), Dict(a => value))
+    ps = remake_buffer(sys, ps, vals)
+    getp(sys, a)(ps) + getp(sys, b)(ps)
+end
+
+@test ForwardDiff.derivative(x -> loss(x, sys, ps), 1.5) == 3.0
```
ChrisRackauckas, line 103: what about a test that actually uses remake?

AayushSabharwal: `remake` doesn't use `remake_buffer` yet so that would just fail. I've planned refactoring `remake` for after this since without it `remake` would fail for all MTK problems

PR #2575, opened 2024-03-26
fix: fix initialization and observed bugs

### LINKED ISSUE #2565: `ODEProblem` get stuck (when creating `MTKProblem`) for certain ODE inputs (where parameter have defaults depending on initial conditions)
This is a simple example (originating from Catalyst when we create systems where we eliminate conservation laws, but implemented fully in MTK).
```julia
# Imports.
using ModelingToolkit
import ModelingToolkit: t_nounits as t, D_nounits as D

# Create ODESystem.
@variables X1(t) X2(t)
@parameters k1 k2 Γ[1:1] = X1 + X2
eq = D(X1) ~ -k1*X1 + k2*(-X1 + Γ[1])
obs = X2 ~ Γ[1] - X1
@mtkbuild osys_m = ODESystem([eq], t, [X1], [k1, k2, Γ[1]]; observed = [X2 ~ Γ[1] - X1])

# Creates ODEProblem.
u0 = [X1 => 1.0, X2 => 2.0]
tspan = (0.0, 1.0)
ps = [k1 => 1.0, k2 => 5.0]
oprob = ODEProblem(osys_m, u0, tspan, ps)
```
The system gets stuck on the last statement, and nothing happens. I have tracked it to this line of code in the `MTKParameters` constructor function:
```julia
p = Dict(k => fixpoint_sub(v, p)
for (k, v) in p if k in all_ps || default_toterm(k) in all_ps)
for (sym, _) in p
    if istree(sym) && operation(sym) === getindex &&
       first(arguments(sym)) in all_ps
        error("Scalarized parameter values ($sym) are not supported. Instead of `[p[1] => 1.0, p[2] => 2.0]` use `[p => [1.0, 2.0]]`")
    end
end
```

Setting both `X1` and `X2` as states in the `ODESystem`, i.e. `@mtkbuild osys_m = ODESystem([eq], t, [X1, X2], [k1, k2, Γ[1]]; observed = [X2 ~ Γ[1] - X1])` does not seem to solve it.

### LINKED ISSUE #2566: `ModelingToolkit.varmap_to_vars` throws error when input have default value depending on paraemters or initial conditions
Another thing that comes up when we use `ModelingToolkit.varmap_to_vars`. Minimal example:
```julia
using ModelingToolkit
@variables t X(t)
@parameters p1 p2 p3

p_vals = [p1 => 1.0, p2 => 2.0]
u_vals = [X => 3.0]

var_vals = [p1 => 1.0, p2 => 2.0, X => 3.0]
desired_values = [p1, p2, p3]
defaults = Dict([p3 => X])
ModelingToolkit.varmap_to_vars(var_vals, desired_values; defaults = defaults)
```
This gives a
```
ERROR: Initial condition underdefined. Some are missing from the variable map.
for the following variables:

Any[p3]

Stacktrace:
 [1] _varmap_to_vars(varmap::Dict{…}, varlist::Vector{…}; defaults::Dict{…}, check::Bool, toterm::Function, initialization_phase::Bool)
   @ ModelingToolkit ~/.julia/dev/ModelingToolkit/src/variables.jl:205
 [2] varmap_to_vars(varmap::Vector{…}, varlist::Vector{…}; defaults::Dict{…}, check::Bool, toterm::Function, promotetoconcrete::Nothing, tofloat::Bool, use_union::Bool)
   @ ModelingToolkit ~/.julia/dev/ModelingToolkit/src/variables.jl:155
 [3] top-level scope
   @ ~/Desktop/Julia Playground/Environment - Catalyst test/test_playground.jl:11

```
We also get the problem if `p3`'s default value depends on another parameter:
```julia
defaults = Dict([p3 => p2])
ModelingToolkit.varmap_to_vars(var_vals, desired_values; defaults = defaults) # Errors
```
However, if it depends on a normal number, it is fine.
If `p3` has a default value that does not depend on the variable, we are fine:
```
defaults = Dict([p3 => 1.0])
ModelingToolkit.varmap_to_vars(var_vals, desired_values; defaults = defaults) # Works.
```

AayushSabharwal:
Close #2566 
Close #2565 

## Checklist

### DIFF
--- a/src/systems/diffeqs/odesystem.jl
+++ b/src/systems/diffeqs/odesystem.jl
@@ -460,9 +460,6 @@ function build_explicit_observed_function(sys, ts;
     else
         ps = (DestructuredArgs(ps, inbounds = !checkbounds),)
     end
-    if isempty(ps)
-        ps = (DestructuredArgs([]),)
-    end
     dvs = DestructuredArgs(unknowns(sys), inbounds = !checkbounds)
     if inputs === nothing
         args = [dvs, ps..., ivs...]
--- a/src/systems/parameter_buffer.jl
+++ b/src/systems/parameter_buffer.jl
@@ -39,7 +39,7 @@ function MTKParameters(
         u0 = Dict()
     end
     defs = merge(defs, u0)
-    defs = merge(defs, Dict(eq.lhs => eq.rhs for eq in observed(sys)))
+    defs = merge(Dict(eq.lhs => eq.rhs for eq in observed(sys)), defs)
     p = merge(defs, p)
     p = merge(Dict(unwrap(k) => v for (k, v) in p),
         Dict(default_toterm(unwrap(k)) => v for (k, v) in p))
--- a/src/variables.jl
+++ b/src/variables.jl
@@ -192,11 +192,11 @@ function _varmap_to_vars(varmap::Dict, varlist; defaults = Dict(), check = false
         toterm = Symbolics.diff2term, initialization_phase = false)
     varmap = canonicalize_varmap(varmap; toterm)
     defaults = canonicalize_varmap(defaults; toterm)
+    varmap = merge(defaults, varmap)
     values = Dict()
     for var in varlist
         var = unwrap(var)
-        val = unwrap(fixpoint_sub(fixpoint_sub(var, varmap; operator = Symbolics.Operator),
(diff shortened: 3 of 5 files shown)


AayushSabharwal:
Tests are actually passing here, codecov is failing

PR #2576, opened 2024-03-26
Don't bold model name etc in show

oxinabox:
The idea i propose is that nothing that is derived from the model should be printed bold.
That then makes it stand out more because of contrast.
This is however very bike-shed-able

**before**
Screenshot from 2024-03-26 23-58-32

**after**
image

## Checklist

## Additional context

I didn't even realize that the model name was printed til today when I was like "why isn't the model name printed?"


oxinabox:
bump

ChrisRackauckas:
Do we want this? It's not clear to me.

oxinabox:
my argument is that we should consistently format anything coming from the model (currently nonbold in this PR)
vs the "template" text we use to describe it (currently bold) in this PR.

PR #2577, opened 2024-03-27
bareiss: Fix corner case assertion failure

Keno:
This failure was reported by @oxinabox on an equation-reordered of one of the MSL examples [1].
Essentially the bug was the we were doing a consistency check on the post-update coefficient,
rather than the pre-update one. See commit message for details. There's two commits here.
One to fix the actual bug and add a test and one to just improve the code slightly to remove the
code pattern that led to the bug and also speed up this code slightly.

[1] https://github.com/SciML/ModelingToolkitStandardLibrary.jl/blob/main/test/Mechanical/rotational.jl#L134


Keno:
Bump @YingboMa.

YingboMa:
The formatter CI fails.

Keno:
Fixed

PR #2582, opened 2024-03-27
Empty events should be empty

YingboMa:
`ModelingToolkit.get_continuous_events(sys)` should be empty when there are no continuous events.


YingboMa:
Test failures are unrelated.

PR #2583, opened 2024-03-28
fix: pass `VariableUnit` of diff var while converting it to a `Term`

ven-k:
Units are defined only in MTK, so pass this piece of info to `Symbolics.diff2term` to add this to the returning term-var.

With this PR: `k_t`'s unit will be `1.0 s⁻¹` 
```julia
using ModelingToolkit
using ModelingToolkit: t, D, get_unit, value, VariableUnit, default_toterm
@variables k(t)
get_unit(default_toterm(D(k))) # returns 1.0 s⁻¹
```

Needs https://github.com/JuliaSymbolics/Symbolics.jl/pull/1104.

## Checklist

### DIFF
--- a/src/variables.jl
+++ b/src/variables.jl
@@ -109,7 +109,7 @@ function default_toterm(x)
             end
             x = normalize_to_differential(op)(arguments(x)...)
         end
-        Symbolics.diff2term(x)
+        Symbolics.diff2term(x, Dict(VariableUnit => get_unit(x)))
     else
         x
     end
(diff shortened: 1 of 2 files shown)


ven-k:
The newly added tests in "Variable Utils Tests" have passed [here](https://github.com/SciML/ModelingToolkit.jl/actions/runs/8911493073/job/24472967508?pr=2583#step:6:1390)

As `Clock` calls `default_toterm` and the behavior shouldn't change "Parameter Dependency Test" are related too. It has passed [here](https://github.com/SciML/ModelingToolkit.jl/actions/runs/8911493073/job/24472967508?pr=2583#step:6:1398)


This PR is ready for review.

PR #2591, opened 2024-03-31
Add initialization_equations flattening constructor

ChrisRackauckas:


ChrisRackauckas:
Whenever I get a second to code freely.

PR #2613, opened 2024-04-04
Speed up `replace` and `canonicalize` by improving type stability

### LINKED ISSUE #2608: Make replace of  MTKParameters type-stable
**Is your feature request related to a problem? Please describe.**

Currently, `replace(MTKParameters, Tunable(), parameter_values(odeprob), newvals)` is not type-stable, i.e
result type is not inferred.
This is a big performance disadvantage for solving ODEProblems of updated parameters, which is used in
a differentiable cost-function.

**Describe the solution you’d like**

Replacing tunable parameters in a MTKParameters returns a type-stable result and is differentiable.

**Describe alternatives you’ve considered**

One could type-annotate the new MTKParameters object inside the cost-function. However, this would involve internal details
and knowledge on how to construct the updated type when supplying e.g. Dual numbers in newval.

**MWE**

```
Pkg.add(["OrdinaryDiffEq","ComponentArrays","ForwardDiff","SciMLStructures"])
Pkg.develop("SymbolicIndexingInterface")
Pkg.develop("ModelingToolkit")

using ModelingToolkit
using OrdinaryDiffEq
using SymbolicIndexingInterface: parameter_values, state_values
using SciMLStructures: Tunable, replace, replace!, canonicalize, isscimlstructure

@parameters α β γ δ
@variables x(t) y(t)
eqs = [D(x) ~ (α - β * y) * x
       D(y) ~ (δ * x - γ) * y]
@mtkbuild odesys = ODESystem(eqs, t)


odeprob = ODEProblem(
    odesys, [x => 1.0, y => 1.0], (0.0, 10.0), [α => 1.5, β => 1.0, γ => 3.0, δ => 1.0])

@code_warntype replace(Tunable(), parameter_values(odeprob), collect(1.0:4.0))
```

SebastianM-C:
## Checklist

## Additional context

This PR attempts to fix #2608 by avoiding boxing and runtime dispatch. So far this works for the case of homogeneous parameters, but for heterogeneous ones, we end up iterating over a tuple of  different types, which is unstable.

### DIFF
--- a/src/systems/parameter_buffer.jl
+++ b/src/systems/parameter_buffer.jl
@@ -143,18 +143,24 @@ end
 
 _split_helper(buf_v::T, recurse, raw, idx) where {T} = _split_helper(eltype(T), buf_v, recurse, raw, idx)
 
-function _split_helper(::Type{<:AbstractArray}, buf_v, recurse, raw, idx)
-    recurse ? map(b -> _split_helper(b, recurse, raw, idx), buf_v) : _split_helper(Any, b, recurse, raw, idx)
+function _split_helper(::Type{<:AbstractArray}, buf_v, ::Val{true}, raw, idx)
+    map(b -> _split_helper(eltype(b), b, Val(false), raw, idx), buf_v)
 end
 
-function _split_helper(::Any, buf_v, recurse, raw, idx)
+function _split_helper(::Type{<:AbstractArray}, buf_v, ::Val{false}, raw, idx)
+    _split_helper((), buf_v, (), raw, idx)
+end
+
+function _split_helper(_, buf_v, _, raw, idx)
     res = reshape(raw[idx[]:(idx[] + length(buf_v) - 1)], size(buf_v))
     idx[] += length(buf_v)
     return res
 end
 
-function split_into_buffers(raw::AbstractArray, buf; recurse = true)
-    ntuple(i->_split_helper(buf[i], recurse, raw, Ref(1)), Val(length(buf)))
+function split_into_buffers(raw::AbstractArray, buf, recurse = Val(true))
+    idx = Ref(1)
+    ntuple(i->_split_helper(buf[i], recurse, raw, idx), Val(length(buf)))
+end
 end
 
 function update_tuple_of_buffers(raw::AbstractArray, buf)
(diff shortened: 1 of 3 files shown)


src/systems/parameter_buffer.jl
```diff
@@ -171,7 +187,7 @@ for (Portion, field) in [(SciMLStructures.Tunable, :tunable)
         repack = let as_vector = as_vector, p = p
             function (new_val)
                 if new_val !== as_vector
-                    update_tuple_of_buffers(new_val, p.$field)
+                    @set! p.$field = update_tuple_of_buffers(new_val, p.$field)
```
AayushSabharwal, line 190: `update_tuple_of_buffers` was written to be an in-place update. If we want to do an out-of-place update here, we should use `split_into_buffers` and delete `update_tuple_of_buffers`

SebastianM-C: Ah, that might explain the test failures 😅

SebastianM-C: I think one question is whether `repack` is supposed to be in-place or out-of-place. I'm not very clear on that.

AayushSabharwal: How I understand it:
Since `alias = true`, if `repack` receives the same array object that `canonicalize` returned, it doesn't need to do anything. Since `MTKParameters` is a mutable SciMLStructure, `repack` should ideally be in-place.

SebastianM-C: But if I want to promote the element type of a `$field`, how would I use `replace`? For `replace` I'd have to use the `newvals`, but that would need the canonicalized version of the parameters, which is why I was thinking to use `repack`.

AayushSabharwal: You don't _have_ to use `repack` if you called `canonicalize`, could just use `replace`.

SebastianM-C:
I fixed a few more cases and added tests with JET and `@inferred` for `replace`/`replace!`. For `canonicalize` there are some cases which can't be inferred due to storing things in vectors, but that could potentially be addressed in the future with a  vector type that stores its size in the type domain. I was a bit surprised that `replace` avoids the issue, but I think that the main factor is that we're effectively unrolling the iterations at compile time.

I think that this should not increase compile times as the tuples are used for the number of types, so we don't expect a ridiculous number of types in the portions.

AayushSabharwal (approved):
LGTM!

Issue #2616, opened 2024-04-05
Support scaled SI units


TorkelE:
E.g. this fails:
```
using ModelingToolkit, DynamicQuantities
@parameters p [unit = u"L/s"] d [unit = u"s^(-1)"]
@parameters t [unit = u"s"] 
@variables X [unit = u"L"]
D = Differential(t)
eqs =[D(X) ~ p - d*X]
ModelingToolkit.validate(eqs) # Does not work.
```
Instead, one have to do:
```julia
using ModelingToolkit, DynamicQuantities
@parameters p [unit = u"m^3/s"] d [unit = u"s^(-1)"]
@parameters t [unit = u"s"] 
@variables X [unit = u"m^3"]
D = Differential(t)
eqs =[D(X) ~ p - d*X]
ModelingToolkit.validate(eqs) # Does not work.
```

These are identical. For highly composite units, rather cumbersome expressions will be generated. Allowing composite units should be very desirable (and was how things used to work).

ChrisRackauckas:
@YingboMa do you know why this limitation exists?

YingboMa:
First, `L` is not a SI unit; `m^3` is a SI unit.

In principle, unit checking can pass as long as they have the same dimension, but what happens when there's `x ~ y` where `x` has the unit of `L` but `y` has the unit of `m^3`? The reason that we error for non-SI units is because that requires additional transformation to equations to be correct in general.

isaacsas:
Mixing `L` and `m^3` resulting in an error seems reasonable, but why can't non-SI units be supported anymore? They were supported previously with Unitful, and are incredibly common in chemistry / biology contexts (i.e. Molar concentrations for example which I guess are not SI but the most common concentration measure in most bio papers). @TorkelE's example that uses units of just `L` and `s` seems like it shouldn't require extra transformations to validate, just support for non-SI units.

YingboMa:
There's no way of tell if someone uses constants like `u"cm"` or `1e-2u"m"` though.
```julia
julia> using DynamicQuantities

julia> u"cm"
0.01 m

julia> 1e-2u"m"
0.01 m

julia> 1e-2u"m" === u"cm"
true
```

YingboMa:
To support non-SI units, we need to use `SymbolicDimensions` from `DynamicQuantities`.

PR #2621, opened 2024-04-05
Add support for symbolic units

YingboMa:
Close https://github.com/SciML/ModelingToolkit.jl/issues/2616


ChrisRackauckas:
Are we planning to finish this?

ctessum:
+1 for finishing this!

ChrisRackauckas:
Well it looks like tests were passing when I re-ran it, so let me re-run it again with the merge fix and I'll merge if it's passing.

ctessum:
Thanks!

ctessum:
Would it be possible to create a release that includes this change? Thanks!

ChrisRackauckas:
done

PR #2629, opened 2024-04-10
fix: propagate parameter dependencies in `extend`

SebastianM-C:
## Checklist

## Additional context

It looks like `parameter_dependencies` are not propagated when extending the system.

### DIFF
--- a/src/systems/abstractsystem.jl
+++ b/src/systems/abstractsystem.jl
@@ -2225,6 +2225,10 @@ function extend(sys::AbstractSystem, basesys::AbstractSystem; name::Symbol = nam
     eqs = union(get_eqs(basesys), get_eqs(sys))
     sts = union(get_unknowns(basesys), get_unknowns(sys))
     ps = union(get_ps(basesys), get_ps(sys))
+    base_deps = get_parameter_dependencies(basesys)
+    deps = get_parameter_dependencies(sys)
+    dep_ps = isnothing(base_deps) ? deps :
+             isnothing(deps) ? base_deps : union(base_deps, deps)
     obs = union(get_observed(basesys), get_observed(sys))
     cevs = union(get_continuous_events(basesys), get_continuous_events(sys))
     devs = union(get_discrete_events(basesys), get_discrete_events(sys))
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
Needs a test

src/systems/abstractsystem.jl
```diff
@@ -2395,7 +2400,7 @@ end
 """
     is_diff_equation(eq)
 
-Returns `true` if the input is a differential equation, i.e. is an equatation that contain some 
+Returns `true` if the input is a differential equation, i.e. is an equatation that contain some
 form of differential.
```
AayushSabharwal, line 2404:
```suggestion
Returns `true` if the input is a differential equation, i.e. an equation that contains a
differential term.
```

src/systems/abstractsystem.jl
```diff
@@ -2421,7 +2426,7 @@ end
 """
     is_alg_equation(eq)
 
-Returns `true` if the input is an algebraic equation, i.e. is an equatation that does not contain 
+Returns `true` if the input is an algebraic equation, i.e. is an equatation that does not contain
 any differentials.
```
AayushSabharwal, line 2430:
```suggestion
Returns `true` if the input is an algebraic equation, i.e. an equation that does not contain
any differentials.
```

src/systems/abstractsystem.jl
```diff
@@ -2603,7 +2608,7 @@ has_alg_eqs(sys::AbstractSystem) = any(is_alg_equation, get_eqs(sys))
 """
     has_diff_eqs(sys::AbstractSystem)
 
-For a system, returns true if it contain at least one differential equation (i.e. that contain a 
+For a system, returns true if it contain at least one differential equation (i.e. that contain a
 differential) in its *top-level system*.
```
AayushSabharwal, line 2612:
```suggestion
Return `true` if a system contains at least one differential equation (i.e. an equation with a
differential term). Note that this does not consider subsystems, and only takes into account
equations in the top-level system.
```

AayushSabharwal (commented):
Not directly related to the PR, but might as well fix it

SebastianM-C:
Tests fail because nonlinear systems don't have parameter dependencies. Are they expected to have that?

AayushSabharwal:
I didn't add parameter dependencies to nonlinear systems because it didn't seem necessary. Can parameters change while solving a `NonlinearProblem`? Or do we still want them so that the dependent parameter is always calculated as a function of the other parameters every time the problem is created?

ChrisRackauckas:
The latter

AayushSabharwal:
#2632 needs to be merged for CI to pass here then

ChrisRackauckas:
Rebase

SebastianM-C:
I'm not sure why the integration tests fail, it looks unrelated to this PR.

PR #2630, opened 2024-04-11
fix: fix intialization of array parameters with unknown size

### LINKED ISSUE #2615: Weird behaviour when declaring a paraemter as a vector
When designating a parameter as a vector, using it as such, and giving it vector values, the ode simulation fails quietly. Not sure how standard the workflow is, but intuitively it makes sense. If nothing else it should yield an error.
```julia
@parameters p::Vector{Float64}
@variables X(t)
eq = D(X) ~ p[1] - p[2]*X
@mtkbuild osys = ODESystem([eq], t)

u0 = [X => 1.0]
ps = [p => [2.0, 0.1]]
oprob = ODEProblem(osys, u0, (0.0, 100.0), ps)
sol = solve(oprob, Tsit5())
sol[X][end] # Gives `1.0` (the original value, should be ~20.0)
```

AayushSabharwal:
Close #2615 

## Checklist

### DIFF
--- a/src/systems/parameter_buffer.jl
+++ b/src/systems/parameter_buffer.jl
@@ -43,8 +43,7 @@ function MTKParameters(
     p = merge(defs, p)
     p = merge(Dict(unwrap(k) => v for (k, v) in p),
         Dict(default_toterm(unwrap(k)) => v for (k, v) in p))
-    p = Dict(k => fixpoint_sub(v, p)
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
Test failures?

AayushSabharwal:
Hmm that's an annoying one, looking into it

PR #2655, opened 2024-04-17
Allow Array unknowns in OptimizationSystem

Vaibhavdixit02:
## Checklist

### DIFF
--- a/src/systems/optimization/optimizationsystem.jl
+++ b/src/systems/optimization/optimizationsystem.jl
@@ -26,7 +26,7 @@ struct OptimizationSystem <: AbstractOptimizationSystem
     """Objective function of the system."""
     op::Any
     """Unknown variables."""
-    unknowns::Vector
+    unknowns::Array
     """Parameters."""
     ps::Vector
     """Array variables."""


src/systems/optimization/optimizationsystem.jl
```diff
@@ -26,7 +26,7 @@ struct OptimizationSystem <: AbstractOptimizationSystem
     """Objective function of the system."""
     op::Any
     """Unknown variables."""
-    unknowns::Vector
```
ChrisRackauckas, line 29: I'm not sure how this becomes an array? I guess we just vec it. But how do we get to this?

Vaibhavdixit02: I am not sure I understand. This will allow us to handle matrix variables that's all

ChrisRackauckas: How did we get a non-vector array?

Vaibhavdixit02: You pass it in as the initial guess for the variables

PR #2668, opened 2024-04-24
Fixes for dependent parameters

SebastianM-C:
## Checklist

## Additional context

When `build_explicit_observed_function` [was updated](https://github.com/SebastianM-C/ModelingToolkit.jl/blob/9184ebe8d9450646553082b2cd9f0e141cb5ec9f/src/systems/diffeqs/odesystem.jl#L381) for parameter dependencies, only the `ps` argument was updated, but internally the function still ckecked for the `parameters` instead of the `full_parameters`.

### DIFF
--- a/src/systems/abstractsystem.jl
+++ b/src/systems/abstractsystem.jl
@@ -878,6 +878,11 @@ function namespace_guesses(sys)
     Dict(unknowns(sys, k) => namespace_expr(v, sys) for (k, v) in guess)
 end
 
+function namespace_parameter_dependencies(sys)
+    pdeps = get_parameter_dependencies(sys)
+    Dict(dependent_parameters(sys, k) => namespace_expr(v, sys) for (k, v) in pdeps)
+end
+
 function namespace_equations(sys::AbstractSystem, ivs = independent_variables(sys))
     eqs = equations(sys)
     isempty(eqs) && return Equation[]
@@ -976,11 +981,27 @@ end
 
 function dependent_parameters(sys::AbstractSystem)
     if has_parameter_dependencies(sys) &&
-       (pdeps = get_parameter_dependencies(sys)) !== nothing
-        collect(keys(pdeps))
+       !isempty(parameter_dependencies(sys))
+        collect(keys(parameter_dependencies(sys)))
     else
         []
     end
+end
+
+function parameter_dependencies(sys::AbstractSystem)
+    pdeps = get_parameter_dependencies(sys)
+    if isnothing(pdeps)
+        pdeps = Dict()
+    end
+    systems = get_systems(sys)
+    isempty(systems) && return pdeps
+    for subsys in systems
+        isnothing(get_parameter_dependencies(subsys)) && continue
+
+        pdeps = merge(pdeps, namespace_parameter_dependencies(subsys))
+    end
+    # @info pdeps
+    return pdeps
 end
 
 function full_parameters(sys::AbstractSystem)
@@ -1044,6 +1065,8 @@ for f in [:unknowns, :parameters]
         map(v -> $f(sys, v), vs)
     end
 end
+
+dependent_parameters(sys::Union{AbstractSystem, Nothing}, v) = renamespace(sys, v)
 
 flatten(sys::AbstractSystem, args...) = sys
 
--- a/src/systems/abstractsystem.jl
+++ b/src/systems/abstractsystem.jl
@@ -880,7 +880,7 @@ end
 
 function namespace_parameter_dependencies(sys)
     pdeps = get_parameter_dependencies(sys)
-    Dict(dependent_parameters(sys, k) => namespace_expr(v, sys) for (k, v) in pdeps)
+    Dict(parameters(sys, k) => namespace_expr(v, sys) for (k, v) in pdeps)
 end
 
 function namespace_equations(sys::AbstractSystem, ivs = independent_variables(sys))
@@ -1065,8 +1065,6 @@ for f in [:unknowns, :parameters]
         map(v -> $f(sys, v), vs)
     end
 end
-
-dependent_parameters(sys::Union{AbstractSystem, Nothing}, v) = renamespace(sys, v)
 
 flatten(sys::AbstractSystem, args...) = sys
 
--- a/src/systems/abstractsystem.jl
+++ b/src/systems/abstractsystem.jl
@@ -996,8 +996,6 @@ function parameter_dependencies(sys::AbstractSystem)
     systems = get_systems(sys)
     isempty(systems) && return pdeps
     for subsys in systems
-        isnothing(get_parameter_dependencies(subsys)) && continue
-
         pdeps = merge(pdeps, namespace_parameter_dependencies(subsys))
     end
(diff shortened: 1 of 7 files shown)


ChrisRackauckas:
Format or the tests won't run.

SebastianM-C:
I realized that this is not enough, as the parameter dependencies are dropped in several other places. The above 2 are commits are what I got so far, I'm not sure what other places need this. I'll try to add more tests too.

src/systems/diffeqs/odesystem.jl
```diff
@@ -354,6 +354,7 @@ function flatten(sys::ODESystem, noeqs = false)
             get_iv(sys),
             unknowns(sys),
             parameters(sys),
+            parameter_dependencies = get_parameter_dependencies(sys),
```
AayushSabharwal, line 357: We could have parameter dependencies in hierarchical systems, which means this needs the `guesses` treatment where dependencies from subsystems are namespaced and included in the flattened system

AayushSabharwal (changes requested):
This would also benefit from a test where a subsystem has a parameter dependency alongside the parent system it is included in. The subsystem's dependency should be appropriately namespaced

src/systems/diffeqs/odesystem.jl
SebastianM-C: Isn't the system structurally simplified by the time we get here? [95c8567](https://github.com/SciML/ModelingToolkit.jl/pull/2668/commits/95c85674555414b40f6649312c8ff26e4d94796b) would take care of that.

test/parameter_dependencies.jl
```diff
@@ -67,6 +67,20 @@ end
...
+    @named sys = ODESystem(
+        [D(x) ~ p1 * t + p2],
+        t;
+        parameter_dependencies = [p2 => 2p1]
+    )
+    prob = ODEProblem(complete(sys))
+    get_dep = getu(prob, 2p2)
+    @test get_dep(prob) == 4
+end
+
```
AayushSabharwal, line 83: Could you also add a testcase for hierarchical systems where the subsystems have parameter dependencies? I have a feeling that will fail

AayushSabharwal (changes requested):

test/parameter_dependencies.jl
SebastianM-C: Yeah, it failed. I added a test in [db89f68](https://github.com/SciML/ModelingToolkit.jl/pull/2668/commits/db89f6892a671c0bc80a807f7b3662833483385f) and I tried to do the renamespacing, but something is off and the dependent parameter does not seem to update correctly.

src/systems/diffeqs/odesystem.jl
SebastianM-C: I changed this to use `parameter_dependencies(sys)` instead, so now it gets the renamespaced dependencies of the subsystems.

src/systems/abstractsystem.jl
```diff
@@ -870,6 +870,11 @@ function namespace_guesses(sys)
     Dict(unknowns(sys, k) => namespace_expr(v, sys) for (k, v) in guess)
 end
 
+function namespace_parameter_dependencies(sys)
+    pdeps = get_parameter_dependencies(sys)
+    Dict(dependent_parameters(sys, k) => namespace_expr(v, sys) for (k, v) in pdeps)
```
AayushSabharwal, line 875:
```suggestion
    Dict(parameters(sys, k) => namespace_expr(v, sys) for (k, v) in pdeps)
```

No need to create a new method that does the same thing as an old one

src/systems/abstractsystem.jl
```diff
@@ -870,6 +870,11 @@ function namespace_guesses(sys)
     Dict(unknowns(sys, k) => namespace_expr(v, sys) for (k, v) in guess)
 end
 
+function namespace_parameter_dependencies(sys)
+    pdeps = get_parameter_dependencies(sys)
```
AayushSabharwal, line 874:
```suggestion
    pdeps = parameter_dependencies(sys)
```

This needs to be recursive, since we can have multiple levels of nesting

src/systems/abstractsystem.jl
```diff
@@ -967,13 +972,29 @@ end
...
 
+function parameter_dependencies(sys::AbstractSystem)
+    pdeps = get_parameter_dependencies(sys)
+    if isnothing(pdeps)
+        pdeps = Dict()
+    end
+    systems = get_systems(sys)
+    isempty(systems) && return pdeps
+    for subsys in systems
+        isnothing(get_parameter_dependencies(subsys)) && continue
```
AayushSabharwal, line 990:
```suggestion
```

Don't need this check for two reasons
- `subsys` might not have parameter dependencies, but subsystems of `subsys` might
- `namespace_parameter_dependencies` will handle checking subsystems of `subsys` by recursively calling this function

src/systems/abstractsystem.jl
```diff
@@ -1028,6 +1049,8 @@ for f in [:unknowns, :parameters]
     end
 end
 
+dependent_parameters(sys::Union{AbstractSystem, Nothing}, v) = renamespace(sys, v)
+
```
AayushSabharwal, line 1053:
```suggestion
```

src/systems/index_cache.jl
```diff
@@ -104,8 +104,8 @@ function IndexCache(sys::AbstractSystem)
         end
     end
 
-    if has_parameter_dependencies(sys) &&
-       (pdeps = get_parameter_dependencies(sys)) !== nothing
+    if has_parameter_dependencies(sys)
+        pdeps = parameter_dependencies(sys)
```
AayushSabharwal, line 108: This is not technically necessary, although it will work regardless. If the system is simplified, all parameter dependencies will be available at the top level. If it isn't simplified, we can't simulate it anyway so an index cache is immaterial.

src/systems/nonlinear/initializesystem.jl
```diff
@@ -95,6 +95,7 @@ function generate_initializesystem(sys::ODESystem;
         full_states,
         pars;
         defaults = merge(ModelingToolkit.defaults(sys), todict(u0), dd_guess),
+        parameter_dependencies = get_parameter_dependencies(sys),
```
AayushSabharwal, line 98:
```suggestion
        parameter_dependencies = parameter_dependencies(sys),
```

AayushSabharwal (changes requested):

AayushSabharwal:
Also needs a rebase, since MTK CI passes now

AayushSabharwal:
I can't suggest changes on unchanged files, but in `src/systems/parameter_buffer.jl:139` `get_parameter_dependencies` needs to be `parameter_dependencies`. If you run a global search for `get_parameter_dependencies`, all of them except the ones in `parameter_dependencies` itself need to be replaced this way.

SebastianM-C:
Thanks! That seems to have solved the issue.

AayushSabharwal (approved):

ChrisRackauckas (approved):

Issue #2670, opened 2024-04-25
Structural simplification fails on non-1-indexed variable array 


hersle:
The example
```julia
using ModelingToolkit
using ModelingToolkit: t_nounits as t, D_nounits as D

@testset "Non-1-indexed variable array" begin
    @variables x(t)[0:1] # 0-indexed variable array
    sys = ODESystem([
        x[0] ~ 0.0
        D(x[1]) ~ x[0]
    ], t, [x], []; name=:sys)
    @test_nowarn sys = structural_simplify(sys)
    @test equations(sys) == [D(x[1]) ~ 0.0]
end
```
fails because [this hack](https://github.com/SciML/ModelingToolkit.jl/blob/b4f14a4fae26a11436b36bc3a968b5cb6b5eb7c7/src/structural_transformation/symbolics_tearing.jl#L576) tries to access the 1-indexed vector `[x[0], x[1]]` at index `[0]`:
```julia
  BoundsError: attempt to access 2-element Vector{SymbolicUtils.BasicSymbolic{Real}} at index [0]
  Stacktrace:
    [1] getindex(A::Vector{SymbolicUtils.BasicSymbolic{Real}}, i1::Int64)
      @ Base .\essentials.jl:13
    [2] fast_substitute(expr::SymbolicUtils.BasicSymbolic{Real},
 subs::Dict{Any, Any}; operator::Type)
      @ Symbolics C:\Users\herma\.julia\packages\Symbolics\Eas9m\src\variable.jl:489
    [3] fast_substitute(eq::Equation, subs::Dict{Any, Any}; operator::Type)
      @ Symbolics C:\Users\herma\.julia\packages\Symbolics\Eas9m\src\variable.jl:455
    [4] tearing_reassemble(state::TearingState{ODESystem}, var_eq_matching::ModelingToolkit.BipartiteGraphs.Matching{Union{ModelingToolkit.BipartiteGraphs.Unassigned, ModelingToolkit.StructuralTransformations.SelectedState}, Vector{Union{ModelingToolkit.BipartiteGraphs.Unassigned, ModelingToolkit.StructuralTransformations.SelectedState, Int64}}}, full_var_eq_matching::Nothing; simplify::Bool, mm::ModelingToolkit.SparseMatrixCLIL{Int64, Int64}) 
[...]

PR #2671, opened 2024-04-25
Fix structural simplification of non-1-indexed variable arrays

hersle:
## Checklist

## Additional context

This is my attempt at fixing #2670.
It is my first PR. Please let me know if I should make any changes, or if you have suggestions on other ways to fix the problem! 😃


hersle:
I am checking if this still runs. It's not a perfect solution, but neither is the hack it hacks.

ChrisRackauckas (approved):

ChrisRackauckas:
It's not unreasonable. Agree that maybe we could simplify it in the future, but we should go with this for now so that way we're a bit safer every day.

PR #2673, opened 2024-04-25
Add scalarization for systems

YingboMa:


ChrisRackauckas:
Does this effect the SII and remake on the state?

YingboMa:
I don't think so. It just makes structural simplify to at least work.

PR #2676, opened 2024-04-29
fix: create and solve initialization system in linearization_function

AayushSabharwal:
## Checklist


src/systems/nonlinear/initializesystem.jl
```diff
@@ -100,3 +100,81 @@ function generate_initializesystem(sys::ODESystem;
 
     return sys_nl
 end
+
+function generate_algebraic_initializesystem(sys::ODESystem;
+        u0map = Dict(),
+        name = nameof(sys),
+        guesses = Dict(), check_defguess = false,
+        default_dd_value = 0.0,
+        kwargs...)
```
ChrisRackauckas, line 109: It seems like this is a lot of a copy, maybe we can do this instead with some undocumented kwargs in the other?

src/systems/nonlinear/initializesystem.jl
```diff
@@ -68,28 +69,34 @@ function generate_initializesystem(sys::ODESystem;
...
-            push!(u0, st => guesses[st])
-        elseif check_defguess
-            error("Invalid setup: unknown $(st) has no default value or initial guess")
         end
     end
 
     pars = [parameters(sys); get_iv(sys)]
-    nleqs = [eqs_ics; get_initialization_eqs(sys); observed(sys)]
+    nleqs = if algebraic_only
+        eqs_ics
```
ChrisRackauckas, line 96: you always need the observed equations?

ChrisRackauckas (approved):

src/systems/abstractsystem.jl
```diff
@@ -1782,6 +1780,8 @@ function linearization_function(sys::AbstractSystem, inputs,
                 length(sts) == length(u) ||
                     error("Number of unknown variables ($(length(sts))) does not match the number of input unknowns ($(length(u)))")
                 if initialize && !isempty(alge_idxs) # This is expensive and can be omitted if the user knows that the system is already initialized
+                    initprob = NonlinearLeastSquaresProblem(initfn, u, p)
+                    @set! fun.initializeprob = initprob
                     residual = fun(u, p, t)
                     if norm(residual[alge_idxs]) > √(eps(eltype(residual)))
```
baggepinnen, line 1786: this is intended to be a fast path such that no expensive operation has to be performed in case the operating point provided by the user is consistent. If you know of a better way to do this nowadays we could adopt that instead. This function is very performance critical, it may be called 10k times on the same system but for different operating points.

AayushSabharwal: Fixed in https://github.com/SciML/ModelingToolkit.jl/pull/2676/commits/d23ce3ae1a379fe7405c0b7a645b425b3271f582

test/downstream/linearize.jl
```diff
@@ -120,10 +120,14 @@ lsys = ModelingToolkit.reorder_unknowns(lsys0, unknowns(ssys), desired_order)
 lsyss, _ = ModelingToolkit.linearize_symbolic(pid, [reference.u, measurement.u],
     [ctr_output.u])
 
-@test substitute(lsyss.A, ModelingToolkit.defaults(pid)) == lsys.A
-@test substitute(lsyss.B, ModelingToolkit.defaults(pid)) == lsys.B
-@test substitute(lsyss.C, ModelingToolkit.defaults(pid)) == lsys.C
-@test substitute(lsyss.D, ModelingToolkit.defaults(pid)) == lsys.D
+@test substitute(
+    lsyss.A, merge(ModelingToolkit.defaults(pid), ModelingToolkit.guesses(pid))) == lsys.A
```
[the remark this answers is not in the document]
AayushSabharwal: Fixed in https://github.com/SciML/ModelingToolkit.jl/pull/2676/commits/4745a9be8dc024f0cd80eb25ac0f48e2ec6e0d29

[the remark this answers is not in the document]
baggepinnen: I think the order of the arguments to `merge` should be reversed, the last collection takes precedence if a key is present in both, and I think `defaults` should be considered more important if present.

baggepinnen (approved):

ChrisRackauckas:
Looks like there's still linearization tests failing?

AayushSabharwal:
Yeah that's the one I'm having trouble reproducing. It errors in tests, but not when I run the code block manually in the REPL.

ChrisRackauckas:
Avoid any redefinition. If I had to guess it could be https://github.com/JuliaLang/julia/issues/52635 cropping back up. If you split the safetestset into different parts does it go away?

AayushSabharwal:
Nope, still errors even in it's own file and `@safetestset`, but runs fine copy-pasted into the REPL

ChrisRackauckas:
Start Julia with forced bounds checks.

AayushSabharwal:
Thanks! Found the bug.

ChrisRackauckas:
What's the state here?

AayushSabharwal:
`InterfaceI` doesn't pass because of the OrdinaryDiffEq stuff, but everything else is good (except the downstream tests marked as broken)

ChrisRackauckas:
```
InitializationSystem Test: Test Failed at /home/runner/work/ModelingToolkit.jl/ModelingToolkit.jl/test/initializationsystem.jl:238
  Expression: sol.retcode == SciMLBase.ReturnCode.Unstable
   Evaluated: SciMLBase.ReturnCode.MaxIters == SciMLBase.ReturnCode.Unstable
```

@oscardssmith I thought this was handled?

oscardssmith:
so did I. Was this using the most recent version of SciMLBase?

ChrisRackauckas:
Make a PR that bumps it and let's see if master fails.

PR #2702, opened 2024-05-06
introduce cache array to make discrete-time state update atomic

### LINKED ISSUE #2701: Miss-compilation when `Hold` is applied to delayed discrete variable
The following models a discrete-time delay $y(k) = u(k-d)$. At $t = 2$, a step in the input occurs from 0 to 1, it should take $d$ seconds for the step to appear in the output, instead, it takes $d-1$ seconds, and the behavior for $d=0$ and $d=1$ are identical.

The problem is exposed by the following analysis
1. We use `d` discrete-time state variables to store the delayed signal. Let's assume `d = 1` for simplicity
2. At time `t`, the input to the discrete partition changes (in the example `t = 2` and the change is from 0 to 1)
3. At time `t`, the c2d variable is thus changed by `c2d_obs(integrator.u, p, t)`
4. At time `t`, the `disc` update stores the changed signal value in the discrete-time state variable which is used as memory
5. At time `t`, **the `d2c` variable is set to the value of the discrete-time state variable**. This is the incorrect step, the discrete update correctly represents `x(k) = f(x(k-1), c2d(k)`, but the `d2c` update in this case has to output the old discrete-time state, not the new, since the `Hold` operator was applied to the delayed input signal `input(k-d)`. 

There are two ways to handle this issue
1. The compiler can realize that an additional state variable is required in order to keep the old value in memory until the `d2c` pass is run. Right now, the compiler correctly handles cases when the `d2c` involves state variables at time `k`, but not at `k-d` for some positive `d`.
2. Split the `d2c` update in two, one that is run before the discrete affect, and one after. This would reduce the number of state variables, but adds complexity instead.

MWE

```julia
using ModelingToolkit, Plots, OrdinaryDiffEq
using ModelingToolkit: D_nounits as D
using ModelingToolkit: t_nounits as t
k = ShiftIndex(Clock(t, 1))
@mtkmodel DelayModel begin
    @variables begin
        input(t) = 0
        delay(t) = 0
        x(t) = 0
    end
    @structural_parameters begin
        d
    end
    @equations begin
        input ~ (t >= 2)
        delay(k) ~ input(k-d)
        D(x) ~ (-x + Hold(delay))/1e-3 # Simple fast lowpass filter to be able to show a continuous-time output variable
    end
end

plot()
for d = 0:3
    @mtkbuild m = DelayModel(; d)
    prob = ODEProblem(m, [m.delay(k-3)=>0, m.delay(k-2)=>0, m.delay(k-1)=>0], (0.0, 10.0))
    sol = solve(prob, Tsit5(), kwargshandle = KeywordArgSilent)
    plot!(sol, lab="d = $d")
end
display(current())
```
image

the plot shows the result for a few different values of $d$

baggepinnen:
closes #2701

system in issue above now produces the correct answer

image


src/systems/clock_inference.jl
```diff
@@ -321,6 +321,8 @@ function generate_discrete_affect(
             end)
         end
 
+        cache = copy(p[$disc_range]) # Cache needed for atomic state update
```
AayushSabharwal, line 324: Is the `copy` necessary? Won't indexing already allocate a new array

AayushSabharwal (approved):

baggepinnen:
@AayushSabharwal the test failure indicates that the indices used to index into `p` are of some magical type? Tests pass locally so I assume this is some recent change upstream?

src/systems/clock_inference.jl
```diff
@@ -326,6 +326,7 @@ function generate_discrete_affect(
         # @show disc_range
         affect! = :(function (integrator, saved_values)
             @unpack u, p, t = integrator
+            cache = p[$disc_range] # Cache needed for atomic state update
```
AayushSabharwal, line 329: Nope this is all in MTK. `ParameterIndex` is an index into `MTKParameters`. `disc_range isa Vector{<:ParameterIndex}`. I didn't notice the error, but we don't have `getindex(::MTKParameters, ::Vector{<:ParameterIndex})` defined, so this call is invalid. I'm not sure how it passes locally.

I guess we either need a method for the above, which just broadcasts the `getindex` or do it manually here

AayushSabharwal: That's interesting. I'll look into why `view` works, but I suspect it is because view internally just relies on `getindex` working for scalar indices.

baggepinnen: ah, I see the difference now, I'll push a fix

AayushSabharwal:
So `view` won't work either. There are a bunch of branches based on whether `has_index_cache(sys) && get_index_cache(sys) !== nothing` (which is basically a check for if the parameter object is an array or `MTKParameters`). The `view` version is never used for `MTKParameters`.

baggepinnen:
There are likely additional compiler-related problems, outlined in https://github.com/SciML/ModelingToolkit.jl/issues/2701

PR #2726, opened 2024-05-18
Backport #2719 to MTK v8

devmotion:
The PR backports #2719 to a MTK v8 on a new `backport_8` branch. I can't use MTK v9 yet due to the incompatibility with Catalyst, so #2719 is not sufficient for my project.


ChrisRackauckas:
Format so tests run

devmotion:
I reformatted the code (which blows up the diff). It seems only a subset of CI is run on PRs that do not target the master branch - one problem with the existing version bounds for MTK v8 are the breaking changes in SymbolicUtils 1.6.0 (see e.g. https://github.com/SciML/ModelingToolkit.jl/actions/runs/9150353436/job/25155026031#step:6:1243) and the ADTypes changes (and some incorrect version bounds in NonlinearSolve it seems: https://github.com/SciML/ModelingToolkit.jl/actions/runs/9150353436/job/25155025675?pr=2726#step:6:1903). In my package that's not too problematic since I pin these packages anyway but generally that's annoying for users.

ChrisRackauckas:
@thazhemadam do you know how to make this run?

ChrisRackauckas:
Can we just bound SymbolicUtils?

ChrisRackauckas:
Limit SymbolicIndexingInterface as well?

devmotion:
Why?

gdalle:
I think the problem here is the same as https://github.com/SciML/SciMLBase.jl/pull/689#issuecomment-2124463847

It's not an ADTypes bug per se, it's just that an environment with ADTypes v1 installed leaks into an environment where only ADTypes v0.2 should be allowed, due to Pkg manipulations during testing. With fresh environments this probably wouldn't happen

Vaibhavdixit02:
Yeah we can do that, I am not sure how it can be changed in the registry though. Can it just be bounded to v1.0.1 here?

gdalle:
One option is to yank those versions, aka get rid of them completely. That's what I did in https://github.com/JuliaRegistries/General/pull/107391

https://github.com/JuliaRegistries/General?tab=readme-ov-file#when-is-yanking-a-release-appropriate

gdalle:
Let's discuss on the General PR https://github.com/JuliaRegistries/General/pull/107391

devmotion:
@gdalle Tests still fail due to ADTypes it seems.

gdalle:
It's installing OptimizationBase v0.0.7, which I thought I had yanked?

ChrisRackauckas:
The package servers can take awhile to get the updated registries.

ChrisRackauckas:
r

PR #2744, opened 2024-05-27
add ClockChange operator

baggepinnen:
This PR adds a discrete-time operator `ClockChange` that changes the clock of a discrete-time variable. This allows for sub and super sampling.

**Status:** In the simple test model added to the tests, I get an error from the compiler 
```julia
julia> @mtkbuild cc = CC()
ERROR: Some specified inputs were not found in system. The following variables were not found Any[ModelingToolkit.ClockChange(Clock(t, 2.0), Clock(t, 1.0))(y(t))]
Stacktrace:
  [1] error(::String, ::Base.KeySet{Any, Dict{Any, Bool}})
    @ Base ./error.jl:44
  [2] markio!(state::TearingState{ODESystem}, orig_inputs::Set{Any}, inputs::Vector{Any}, outputs::Vector{Any}; check::Bool)
    @ ModelingToolkit ~/.julia/dev/ModelingToolkit/src/systems/abstractsystem.jl:2038
  [3] markio!
    @ ~/.julia/dev/ModelingToolkit/src/systems/abstractsystem.jl:2009 [inlined]
  [4] _structural_simplify!(state::TearingState{…}, io::Tuple{…}; simplify::Bool, check_consistency::Bool, fully_determined::Bool, warn_initialize_determined::Bool, dummy_derivative::Bool, kwargs::@Kwargs{})
    @ ModelingToolkit ~/.julia/dev/ModelingToolkit/src/systems/systemstructure.jl:680
```

The problem appears realted to the hacky forward shift employed by the compiler, since
```julia
fullvars[end]
ModelingToolkit.ClockChange(Clock(t, 2.0), Clock(t, 1.0))(Shift(t, 2)(y(t)))
```
while the operator variable is not shifted
```julia
collect(keys(inputset))[]
ModelingToolkit.ClockChange(Clock(t, 2.0), Clock(t, 1.0))(y(t))
```

---

EDIT: I pushed a commit that applies the shift to the new operator as well. The remaining problem is that `x` is updated before `y`, while it should be the opposite, the equation for `x` depends on `y` and has to come after.

### DIFF
--- a/src/ModelingToolkit.jl
+++ b/src/ModelingToolkit.jl
@@ -268,7 +268,7 @@ export debug_system
 #export Continuous, Discrete, sampletime, input_timedomain, output_timedomain
 #export has_discrete_domain, has_continuous_domain
 #export is_discrete_domain, is_continuous_domain, is_hybrid_domain
-export Sample, Hold, Shift, ShiftIndex, sampletime, SampleTime
+export Sample, Hold, Shift, ShiftIndex, sampletime, SampleTime, ClockChange
 export Clock #, InferredDiscrete,
 
 end # module
--- a/src/clock.jl
+++ b/src/clock.jl
@@ -55,7 +55,7 @@ See also [`is_discrete_domain`](@ref)
 """
 function has_discrete_domain(x)
     issym(x) && return is_discrete_domain(x)
-    hasshift(x) || hassample(x) || hashold(x)
+    hasshift(x) || hassample(x) || hashold(x) || hasclockchange(x)
 end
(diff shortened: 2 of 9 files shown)


baggepinnen:
@Aayush Sabharwal @Yingbo Ma 
It looks like we need to take some time to clean up the current implementation that forward-shifts discrete time variables, it leads to headaches like above. 

I'd be in favor of getting rid of the forward-shift hack completely, are there any fundamental reasons left that prevents us from doing that?

Issue #1586, opened 2022-05-19
Solving differential equations symbolicly


mahdiarsadeghi:
Is it possible to solve differential equations symbolically, something like the `dsolve()` function in [Matlab](https://www.mathworks.com/help/symbolic/dsolve.html)?

PR #2747, opened 2024-05-28
feat: allow parameters in ODESystem to be unknowns in initialization system

### LINKED ISSUE #2665: Cannot solve for dependent parameters during initialization
Consider this (trivial) ODE for `x` and `y`:

```julia
using Test
using ModelingToolkit
using DifferentialEquations

@variables x(t) y(t) s(t)
sys1 = structural_simplify(ODESystem([D(x) ~ 0, D(y) ~ 0, s ~ x + y], t, [x, y, s], []; name=:sys))
prob1 = ODEProblem(sys1, [s => 1.0, x => 0.3], (0.0, 1.0))
sol1 = solve(prob1)
@test all(sol1[x] .≈ 0.3) && all(sol1[y] .≈ 0.7) # passes
```

This works. The constraint `s(0) == x(0)+y(0) == 1` correctly initializes `y(0) == 0.7`. I do not have to specify a guess for `y(0)`.

---

But consider an implementation of its (trivial) analytical solution, using parameters for the initial values `x(0) == x0` and `y(0) == y0`:

```julia
@parameters x0 y0
sys2 = structural_simplify(ODESystem([x ~ x0, y ~ y0, s ~ x + y], t, [x, y, s], [x0, y0]; name=:sys))
prob2 = ODEProblem(sys2, [s => 1.0, x0 => 0.3], (0.0, 1.0))
sol2 = solve(prob2)
```

This does not work. The initialization gives an error that asks for a guess:
```julia
ERROR: LoadError: Initial condition underdefined. Some are missing from the variable map.
for the following variables:

Any[y0]
```
But adding e.g. `; guesses = [y0 => 0.0]` to `ODEProblem(sys2, ...` does not change anything.

Is it possible to make this work?

AayushSabharwal:
Close #2665 

Requires:
- https://github.com/SciML/OrdinaryDiffEq.jl/pull/2228
- https://github.com/SciML/DiffEqBase.jl/pull/1051
- https://github.com/SciML/SciMLBase.jl/pull/698

## Checklist

### DIFF
--- a/src/systems/abstractsystem.jl
+++ b/src/systems/abstractsystem.jl
@@ -1849,6 +1849,8 @@ function linearization_function(sys::AbstractSystem, inputs,
     end
     initfn = NonlinearFunction(initsys)
     initprobmap = getu(initsys, unknowns(sys))
+    initprob_init! = generate_initializeprob_init(sys, initsys)
+    initprob_update! = generate_initializeprob_update(sys, initsys)
     ps = full_parameters(sys)
     lin_fun = let diff_idxs = diff_idxs,
         alge_idxs = alge_idxs,
@@ -1856,7 +1858,8 @@ function linearization_function(sys::AbstractSystem, inputs,
         sts = unknowns(sys),
         get_initprob_u_p = get_initprob_u_p,
         fun = ODEFunction{true, SciMLBase.FullSpecialize}(
-            sys, unknowns(sys), ps; initializeprobmap = initprobmap),
+            sys, unknowns(sys), ps; initializeprob_init! = initprob_init!,
+            initializeprob_update! = initprob_update!),
         initfn = initfn,
         h = build_explicit_observed_function(sys, outputs),
         chunk = ForwardDiff.Chunk(input_idxs),
--- a/src/systems/abstractsystem.jl
+++ b/src/systems/abstractsystem.jl
@@ -495,10 +495,18 @@ function SymbolicIndexingInterface.is_observed(sys::AbstractSystem, sym)
 end
 
 function SymbolicIndexingInterface.observed(sys::AbstractSystem, sym)
-    return let _fn = build_explicit_observed_function(sys, sym)
-        fn(u, p, t) = _fn(u, p, t)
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
Why would parameters be unknowns?

hersle:
This is the way I see it. Consider a system of ODEs where some can be solved analytically. Suppose also that their initial values are constrained. For example (with "alternative 1"):
```julia
using ModelingToolkit
using DifferentialEquations

@parameters y0 # represents y(0) in alternative 2 below
@variables x(t) y(t)
@mtkbuild sys = ODESystem([
    D(x) ~ sin(y^2)  # hard to solve analytically for arbitrary y
    D(y) ~ t         # alternative 1: use ODE (works)
    # y ~ t^2/2 + y0 # alternative 2: use analytical solution of ODE (fails)
], t; initialization_eqs = [y * tan(y) ~ 1.0]) # non-trivial initialization constraint
prob = ODEProblem(sys, [x => 0.0], (0.0, 1.0), []; guesses = [y, y0] .=> 1.0)
sol = solve(prob, Tsit5()) # expect sol[y][begin] ≈ 0.8603335890193797
```
This works as expected: `y` is correctly initialized from the constraint.

But what if I want to implement the same system, but with the analytical solution for `y` instead? This would reduce the number of ODEs and improve performance. Then I need a parameter `y0` to represent the initial value of `y(0)`, and this would have to be treated as unknown during initialization. I expect the same code to work, only with `D(y) ~ ...` commented and `# y ~ ...` uncommented. But this fails precisely because `y0` is not treated as unknown.

Note that if the initialization equation was something simpler, like `y ~ 1.0 - x`, this could be handled in both alternative 1 and 2 with defaults for both `[y, y0] .=> 1.0 - x`. In this case, one can interpret `y0` as an "unknown" parameter. This issue can thus be seen as a matter of generalizing unknown parameters to work with arbitrary initialization equations (harder than substitution).

Looking (very) far ahead, I also think overcoming this is a prerequisite to solve e.g. #1586.

AayushSabharwal:
This PR is too old and will be reimplemented

PR #2780, opened 2024-06-09
Test master

ChrisRackauckas:


src/ModelingToolkit.jl
```diff
@@ -102,6 +102,8 @@ for fun in [:toexpr]
     end
 end
 
+1+1
```
ChrisRackauckas, line 105:
```suggestion
1 + 1
```

ChrisRackauckas:
Master is fixed by https://github.com/JuliaSymbolics/Symbolics.jl/pull/1166

PR #2783, opened 2024-06-09
Only use the observed for simplification in MTKParameters construction

ChrisRackauckas:
You don't need it in the p itself, just wanted to use it for simplification, so don't loop over the whole thing. This fixes https://github.com/SciML/MethodOfLines.jl/pull/397

### DIFF
--- a/src/systems/parameter_buffer.jl
+++ b/src/systems/parameter_buffer.jl
@@ -43,10 +43,10 @@ function MTKParameters(
     end
     defs = merge(defs, u0)
     defs = merge(Dict(eq.lhs => eq.rhs for eq in observed(sys)), defs)
-    p = merge(defs, p)
+    bigdefs = merge(defs, p)
     p = merge(Dict(unwrap(k) => v for (k, v) in p),
         Dict(default_toterm(unwrap(k)) => v for (k, v) in p))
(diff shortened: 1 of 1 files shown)


xtalax:
Massive thanks for finding and fixing this!

src/systems/parameter_buffer.jl
```diff
@@ -43,10 +43,10 @@ function MTKParameters(
     end
     defs = merge(defs, u0)
     defs = merge(Dict(eq.lhs => eq.rhs for eq in observed(sys)), defs)
-    p = merge(defs, p)
+    bigdefs = merge(defs, p)
     p = merge(Dict(unwrap(k) => v for (k, v) in p),
         Dict(default_toterm(unwrap(k)) => v for (k, v) in p))
-    p = Dict(unwrap(k) => fixpoint_sub(v, p) for (k, v) in p)
+    p = Dict(unwrap(k) => fixpoint_sub(v, bigdefs) for (k, v) in p)
```
AayushSabharwal, line 49: This only includes parameters in `p` and ignores the ones with default values. This causes tests to fail everywhere, and is why it was implemented this way.

AayushSabharwal:
Superseded by #2785

PR #2795, opened 2024-06-12
fix: fix observed function generation for systems with inputs

### LINKED ISSUE #2794: `build_explicit_observed_function` no longer handles inputs correctly
When `sys` has been processed by `io_preprocessing` and `inputs` are passed to `build_explicit_observed_function`, these inputs are sorted out of the parameter vector by the code
```julia
    if inputs !== nothing
        ps = setdiff(ps, inputs) # Inputs have been converted to parameters by io_preprocessing, remove those from the parameter list
    end
```
However, the following code does not account for this
```julia
reorder_parameters(get_index_cache(sys), ps)
```
so the generated code still expects a parameter vector that is the length of `length(ps) + length(inputs)` rather than just `length(ps)`. This causes the generated code to be incorrect, accessing parameter values at the wrong indices. 

I tried
```julia
reorder_parameters(get_index_cache(sys), ps, drop_missing=true)
```
which makes the parameter array in the generated code of the correct length, but the order is still wrong. @AayushSabharwal I don't understand what `reorder_parameters` is supposed to do, can you have a look?

It appears possible for us to work around the issue by using `split=false` to `structural_simplify`

CC @shobhitvoleti

AayushSabharwal:
Close #2794 

## Checklist

### DIFF
--- a/src/systems/diffeqs/odesystem.jl
+++ b/src/systems/diffeqs/odesystem.jl
@@ -473,6 +473,9 @@ function build_explicit_observed_function(sys, ts;
         ps = DestructuredArgs.(ps, inbounds = !checkbounds)
     elseif has_index_cache(sys) && get_index_cache(sys) !== nothing
         ps = DestructuredArgs.(reorder_parameters(get_index_cache(sys), ps))
+        if isempty(ps) && inputs !== nothing
+            ps = (:EMPTY,)
+        end
     else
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
Test failure looks real.

AayushSabharwal:
Downstream failure is an SII bug fixed in https://github.com/SciML/SymbolicIndexingInterface.jl/pull/86

PR #2809, opened 2024-06-17
fix: fix unknowns not present in initialization system in linearization

AayushSabharwal:
## Checklist

### DIFF
--- a/src/systems/abstractsystem.jl
+++ b/src/systems/abstractsystem.jl
@@ -1896,6 +1896,7 @@ function linearization_function(sys::AbstractSystem, inputs,
         zero_dummy_der = false,
         initialization_solver_alg = TrustRegion(),
         eval_expression = false, eval_module = @__MODULE__,
+        warn_initialize_determined = true,
         kwargs...)
     inputs isa AbstractVector || (inputs = [inputs])
     outputs isa AbstractVector || (outputs = [outputs])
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
Test the case I described, `x => 2, x ~ y`, and `y => 2, x ~ y`.

AayushSabharwal:
Does this test work?

ChrisRackauckas:
No it needs to be structurally simplified.

src/systems/abstractsystem.jl
```diff
@@ -2002,7 +2002,9 @@ function linearization_function(sys::AbstractSystem, inputs,
                 p = todict(p)
                 newps = deepcopy(sys_ps)
                 for (k, v) in p
-                    setp(sys, k)(newps, v)
+                    if is_parameter(sys, p)
```
baggepinnen, line 2005: what happens if there is something for which `is_parameter(sys, p)`?

AayushSabharwal: That's a bug, thanks 😅

baggepinnen: Haha, you fixed something other than I had in mind actually, let me rephrase the question after the fix

what happens if there is something for which !is_parameter(sys, k)?

not `p` -> `k`

AayushSabharwal: Ah. Well that piece of code handles parameters specified as a `Dict`. Its sole purpose is to update the `MTKParameters` object. If `k` is not a parameter, it's not in the parameter object, so it's ignored. If it's an unknown, it's already taken care of in `u`.

PR #2826, opened 2024-06-28
feat: complete `eval_expression` and `eval_module` support

AayushSabharwal:
## Checklist

### DIFF
--- a/src/systems/abstractsystem.jl
+++ b/src/systems/abstractsystem.jl
@@ -161,7 +161,8 @@ time-independent systems. If `split=true` (the default) was passed to [`complete
 object.
 """
 function generate_custom_function(sys::AbstractSystem, exprs, dvs = unknowns(sys),
-        ps = parameters(sys); wrap_code = nothing, postprocess_fbody = nothing, states = nothing, kwargs...)
+        ps = parameters(sys); wrap_code = nothing, postprocess_fbody = nothing, states = nothing,
+        expression = Val{true}, eval_expression = false, eval_module = @__MODULE__, kwargs...)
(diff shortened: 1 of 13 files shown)


ChrisRackauckas (approved):

ChrisRackauckas:
https://github.com/SciML/ModelingToolkit.jl/actions/runs/9714593463/job/26814397836?pr=2826#step:6:1871 the failures look real

AayushSabharwal:
The inference failures are weird, even Cthulhu output doesn't make sense.

src/systems/abstractsystem.jl
```diff
@@ -1210,19 +1223,30 @@ end
...
+
+# This is hit because ensemble problems do a deepcopy
+function Base.deepcopy_internal(ofc::ObservedFunctionCache, stackdict::IdDict)
+    sys = deepcopy(ofc.sys)
+    dict = deepcopy(ofc.dict)
+    eval_expression = ofc.eval_expression
+    eval_module = ofc.eval_module
+    newofc = ObservedFunctionCache(sys, dict, eval_expression, eval_module)
+    stackdict[ofc] = newofc
+    return newofc
```
ChrisRackauckas, line 1242: This is not a good idea, it's not a public API. What's the real problem here?

AayushSabharwal: It's in `deepcopy`'s docstring to overload this for custom `deepcopy`. The problem here is that ensemble simulations deepcopy the problem, which tries to deepcopy the module inside `ObservedFunctionCache`

PR #2834, opened 2024-07-01
fix: infer oop form for SDEProblem/SDEFunction with StaticArrays

### LINKED ISSUE #2814: Unable to simualte SDEs with static array u0/ps
This is important since this is required for GPU simulations.

MWE:
```julia
using ModelingToolkit, StaticArrays, StochasticDiffEq

# Create SDE system.
@parameters p d 
@variables t X(t)
D = Differential(t)
eqs = [D(X) ~ p - d*X]
noise_eqs = [sqrt(p), sqrt(p - d*X)]
@named ssys = SDESystem(eqs, noise_eqs, t, [X], [p, d])
ssys = complete(ssys)

# Create SDEProblem
u0 = @SVector [X => 10.0f0]
tspan = (0.0f0, 10.0f0)
ps = @SVector [p => 2.0f0, d => 0.5f0]
sprob = SDEProblem(ssys, u0, tspan, ps)
solve(sprob, ImplicitEM())
```
Gives an `ERROR: Initial condition incompatible with functional form.` error.

For ODEs, this work:
```julia
using OrdinaryDiffEq
@named osys = ODESystem(eqs, t, [X], [p, d])
osys = complete(osys)

# Create SDEProblem
u0 = @SVector [X => 10.0f0]
tspan = (0.0f0, 10.0f0)
ps = @SVector [p => 2.0f0, d => 0.5f0]
oprob = ODEProblem(osys, u0, tspan, ps)
solve(oprob, Tsit5())
```

AayushSabharwal:
Close #2814 

## Checklist

### DIFF
--- a/src/systems/diffeqs/sdesystem.jl
+++ b/src/systems/diffeqs/sdesystem.jl
@@ -403,14 +403,14 @@ function Girsanov_transform(sys::SDESystem, u; θ0 = 1.0)
         checks = false)
 end
 
-function DiffEqBase.SDEFunction{iip}(sys::SDESystem, dvs = unknowns(sys),
+function DiffEqBase.SDEFunction{iip, specialize}(sys::SDESystem, dvs = unknowns(sys),
         ps = parameters(sys),
         u0 = nothing;
         version = nothing, tgrad = false, sparse = false,
         jac = false, Wfact = false, eval_expression = false,
         eval_module = @__MODULE__,
         checkbounds = false,
-        kwargs...) where {iip}
+        kwargs...) where {iip, specialize}
     if !iscomplete(sys)
         error("A completed `SDESystem` is required. Call `complete` or `structural_simplify` on the system before creating an `SDEFunction`")
     end
@@ -480,7 +480,7 @@ function DiffEqBase.SDEFunction{iip}(sys::SDESystem, dvs = unknowns(sys),
 
     observedfun = ObservedFunctionCache(sys; eval_expression, eval_module)
 
-    SDEFunction{iip}(f, g,
+    SDEFunction{iip, specialize}(f, g,
         sys = sys,
         jac = _jac === nothing ? nothing : _jac,
         tgrad = _tgrad === nothing ? nothing : _tgrad,
@@ -503,6 +503,16 @@ respectively.
 """
 function DiffEqBase.SDEFunction(sys::SDESystem, args...; kwargs...)
     SDEFunction{true}(sys, args...; kwargs...)
+end
+
+function DiffEqBase.SDEFunction{true}(sys::SDESystem, args...;
+        kwargs...)
+    SDEFunction{true, SciMLBase.AutoSpecialize}(sys, args...; kwargs...)
+end
+
+function DiffEqBase.SDEFunction{false}(sys::SDESystem, args...;
+        kwargs...)
+    SDEFunction{false, SciMLBase.FullSpecialize}(sys, args...; kwargs...)
 end
 
 """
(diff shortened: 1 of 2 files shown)


AayushSabharwal:
The problem here is that [this](https://github.com/SciML/StochasticDiffEq.jl/blob/4af4ab8382c257bf2b802aafca4f15baa20cf08b/src/perform_step/sdirk.jl#L30) line doesn't work with static arrays. `K` is an `SVector{1}` whereas `L` is an `SMatrix{1, 1}` (since noise is a matrix). I guess that line should have an `@..`?

```julia
julia> [1;;] + [1]
1×1 Matrix{Int64}:
 2
julia> @SArray[1;;] + @SArray[1]
ERROR: DimensionMismatch: Sizes (Size(1, 1), Size(1,)) of input arrays do not match
```

ChrisRackauckas:
That's a method only for diagonal noise, so that method won't work on array

AayushSabharwal:
1x1 noise is diagonal, right? Which I guess is why I guess it hits that

ChrisRackauckas:
Yeah, what we should do is detect that `if isdiagonal` then turn it it into a vector so it represents diagonal noise.

ChrisRackauckas (approved):

TorkelE:
This update causes Catalyst CI to fail. Given that Catalyst has a very extensive test set, and often has catched things that the MTK tests have missed, would it make sense to not merge things which cause Catalyst downstream errors?

TorkelE:
(I have modified Catalyst so that it works with this new version. However, generally, I do think we should try and not ignore the CI that we got)

ChrisRackauckas:
I don't think the Catalyst tests are right here?

TorkelE:
The Catalyst tests are fine. Basically, this PR goes from requiring an internal matrix to be `Matrix{Num}` from `Matrix{Any}` (I know Sam has argued for the latter representation being used, personally I have no strong opinion). Anyway, this breaks a couple of things (which I Catalyst I switched around so that it should be fine).

ChrisRackauckas:
Oh I thought the Catalyst issue was the sizing thing we were talking about? Requiring Matrix{Num} isn't intentional, though it is somewhat bettter.

TorkelE:
No, the sizing got fixed (thanks for that!). Practically, I'm not sure of changing to `Matrix{Num}` is a big change, but next time it would be useful to at least have a note that there has been an update which breaks the Catalyst tests 👍

isaacsas:
Shouldn’t one be able to generate SDESystems without using Nums? If not, wouldn’t this require all compositional modeling tools to rewrap internal unwrapped expressions when generating SDESystems from other systems?

src/systems/diffeqs/sdesystem.jl
```diff
@@ -231,6 +231,9 @@ function generate_diffusion_function(sys::SDESystem, dvs = unknowns(sys),
     if isdde
         eqs = delay_to_function(sys, eqs)
     end
+    if eqs isa AbstractMatrix && isdiag(eqs)
+        eqs = diag(eqs)
+    end
```
isaacsas, line 236: @ChrisRackauckas @AayushSabharwal I don't think this is correct code for symbolics. The problem is that `isdiag` calls `iszero` which for a non-Num will create a symbolic boolean expression, i.e. 
```julia
julia> @variables A
e1-element Vector{Num}:
 A

julia> eq = 5*A^2 + 2*A + 65
65 + 2A + 5(A^2)

julia> iszero(eq)
false

julia> iszero(ModelingToolkit.unwrap(eq))
(65 + 2A + 5(A^2)) == 0
```
So this now explicitly breaks creating an `SDESystem` with unwrapped expressions.

isaacsas: Also, as @TorkelE mentioned Catalyst master tests should now generally pass, so if Catalyst tests fail when running downstream tests on an MTK PR that is meaningful again (unlike before we updated Catalyst to MTK 9).

ChrisRackauckas: The Catalyst v14 release hasn't been made yet, so it's still not testing the updated code?

ChrisRackauckas: > The problem is that isdiag calls iszero which for a non-Num will create a symbolic boolean expression, i.e.

Thanks, we need to have a different `isdiag` that uses `isequal`. @AayushSabharwal can you prioritize this?

ChrisRackauckas: @isaacsas is there an issue open on this? If so, let's assign @AayushSabharwal to it.

isaacsas: I opened https://github.com/SciML/ModelingToolkit.jl/issues/2868

isaacsas: > The Catalyst v14 release hasn't been made yet, so it's still not testing the updated code?

Also, doesn't this:

https://github.com/SciML/ModelingToolkit.jl/actions/runs/9772401615/job/26976712250#step:6:579

indicate it was tested against Catalyst master?

Issue #2619, opened 2024-04-05
Inconsistency for how to give intiial conditions to DAEs with higher order derivatives.


TorkelE:
Normally, one can/should give initial values/guesses for variables occurring only in the algebraic equations parts of a DAE. However, when one does this for a DAE with higher-order derivatives there are weird behaviours:
```julia
using OrdinaryDiffEq, ModelingToolkit
using ModelingToolkit: t_nounits as t, D_nounits as D

# Creates systems.
@parameters k1 k2 ω
@variables X(t) Y(t)
eqs_1st_order = [D(Y) + Y - ω ~ 0,
       X + k1 ~ Y + k2]
eqs_2nd_order = [D(D(Y)) + 2ω*D(Y) + (ω^2)*Y ~ 0,
        X + k1 ~ Y + k2]
@mtkbuild sys_1st_order = ODESystem(eqs_1st_order, t)
@mtkbuild sys_2nd_order = ODESystem(eqs_2nd_order, t)

# Creates initial conditions and parameters.
u0_1st_order_1 = [X => 1.0, Y => 2.0]
u0_1st_order_2 = [Y => 2.0]
u0_2nd_order_1 = [X => 1.0, Y => 2.0, D(Y) => 0.5]
u0_2nd_order_2 = [Y => 2.0, D(Y) => 0.5]
tspan = (0.0, 10.)
ps = [ω => 0.5, k1 => 2.0, k2 => 3.0]

# Creates ODEProblems.
oprob_1st_order_1 = ODEProblem(sys_1st_order, u0_1st_order_1, tspan, ps)
oprob_1st_order_2 = ODEProblem(sys_1st_order, u0_1st_order_2, tspan, ps)
oprob_2nd_order_1 = ODEProblem(sys_2nd_order, u0_2nd_order_1, tspan, ps) # gives sys_2nd_order
oprob_2nd_order_2 = ODEProblem(sys_2nd_order, u0_2nd_order_2, tspan, ps)

# Solves ODEProblems.
solve(oprob_1st_order_1, Rosenbrock23()) # retcode: Success
solve(oprob_1st_order_2, Rosenbrock23()) # retcode: Success
solve(oprob_2nd_order_1, Rosenbrock23()) # retcode: InitialFailure
solve(oprob_2nd_order_2, Rosenbrock23()) # retcode: Success
```

ChrisRackauckas:
What is the weird behavior?

TorkelE:
[...]

PR #2850, opened 2024-07-06
Run initialization on ODEs with superset initial conditions

ChrisRackauckas:
InitializationProblem is not always built, however it missed a case. If you have an ODE and you define all of the initial conditions, then it does not need an initialization problem. But, if you simplify down to an ODE and give initial conditions for all of the original variables, including some simplified out, then you have defined initial conditions on some observed quantities and that needs to be accounted for by an initialization process.

This fixes https://github.com/SciML/ModelingToolkit.jl/issues/2619

### DIFF
--- a/src/systems/diffeqs/abstractodesystem.jl
+++ b/src/systems/diffeqs/abstractodesystem.jl
@@ -811,7 +811,7 @@ function process_DEProblem(constructor, sys::AbstractODESystem, u0map, parammap;
 
     # ModelingToolkit.get_tearing_state(sys) !== nothing => Requires structural_simplify first
     if sys isa ODESystem && build_initializeprob &&
-       (((implicit_dae || !isempty(missingvars) || !isempty(setboserved)) &&
+       (((implicit_dae || !isempty(missingvars) || !isempty(setobserved)) &&
          all(isequal(Continuous()), ci.var_domain) &&
          ModelingToolkit.get_tearing_state(sys) !== nothing) ||
         !isempty(initialization_equations(sys))) && t !== nothing
--- a/src/systems/diffeqs/abstractodesystem.jl
+++ b/src/systems/diffeqs/abstractodesystem.jl
@@ -797,7 +797,7 @@ function process_DEProblem(constructor, sys::AbstractODESystem, u0map, parammap;
     varmap = canonicalize_varmap(varmap)
     varlist = collect(map(unwrap, dvs))
     missingvars = setdiff(varlist, collect(keys(varmap)))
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
@AayushSabharwal did some of the cases of initialization with array variables simply require it wasn't using the initialization?

src/systems/diffeqs/abstractodesystem.jl
```diff
@@ -777,6 +777,7 @@ function process_DEProblem(constructor, sys::AbstractODESystem, u0map, parammap;
     varmap = canonicalize_varmap(varmap)
     varlist = collect(map(unwrap, dvs))
     missingvars = setdiff(varlist, collect(keys(varmap)))
+    setboserved = setdiff(collect(keys(varmap)), varlist)
```
AayushSabharwal, line 780:
```suggestion
    setobserved = setdiff(collect(keys(varmap)), varlist)
```

src/systems/diffeqs/abstractodesystem.jl
```diff
@@ -824,7 +825,7 @@ function process_DEProblem(constructor, sys::AbstractODESystem, u0map, parammap;
     # TODO: make it work with clocks
     # ModelingToolkit.get_tearing_state(sys) !== nothing => Requires structural_simplify first
     if sys isa ODESystem && build_initializeprob &&
-       (((implicit_dae || !isempty(missingvars)) &&
+       (((implicit_dae || !isempty(missingvars) || !isempty(setboserved)) &&
```
AayushSabharwal, line 828:
```suggestion
       (((implicit_dae || !isempty(missingvars) || !isempty(setobserved)) &&
```

AayushSabharwal (approved):
Just a typo

src/systems/diffeqs/abstractodesystem.jl
```diff
@@ -810,7 +811,8 @@ function process_DEProblem(constructor, sys::AbstractODESystem, u0map, parammap;
 
     # ModelingToolkit.get_tearing_state(sys) !== nothing => Requires structural_simplify first
     if sys isa ODESystem && build_initializeprob &&
-       (((implicit_dae || !isempty(missingvars)) &&
+       (((implicit_dae || !isempty(missingvars) || !isempty(setobserved)) &&
+         all(isequal(Continuous()), ci.var_domain) &&
```
ChrisRackauckas, line 815:
```suggestion
       (((implicit_dae || !isempty(missingvars) || !isempty(setobserved)) &&
```

PR #2862, opened 2024-07-16
Create independent variables with @independent_variables

### LINKED ISSUE #2818: Custom independent variable leads to underdetermined initialization system
This trivial example works as it should, where the independent parameter `t` is imported as `ModelingToolkit.t_nounits`:
```julia
using ModelingToolkit
using DifferentialEquations
using ModelingToolkit: t_nounits as t, D_nounits as d
@variables x(t)
@named sys1 = ODESystem([d(x) ~ 0], t; initialization_eqs = [x ~ t], guesses = [x => 0.0])
prob1 = ODEProblem(structural_simplify(sys1), [], (1.0, 2.0), [])
sol1 = solve(prob1)
```
Now run the same example, but using our own independent parameter (`T` instead of `t`) and differential (`D`instead of `d`):
```julia
using ModelingToolkit
using DifferentialEquations
@variables T; D = Differential(T) # only difference from first example
@variables x(T)
@named sys2 = ODESystem([D(x) ~ 0], T; initialization_eqs = [x ~ T], guesses = [x => 0.0])
prob2 = ODEProblem(structural_simplify(sys2), [], (1.0, 2.0), [])
sol2 = solve(prob2)
```
In the second example (and not the first), the initialization system becomes underdetermined, as reported by the warning
```
┌ Warning: Initialization system is underdetermined. 1 equations for 2 unknowns. Initialization will default to using least squares. To suppress this warning pass warn_initialize_determined = false.
└ @ ModelingToolkit ~/.julia/packages/ModelingToolkit/353ne/src/systems/diffeqs/abstractodesystem.jl:1564
```
By inspecting `structural_simplify(generate_initializesystem(sys2); fully_determined = false)`, I see that the bug is caused by the independent parameter appearing twice, as both a parameter and unknown:
```
Model sys2 with 1 equations
Unknowns (2):
  T
  x(T) [defaults to 0.0]
Parameters (1):
  T
```
I expected the second example to behave like the first. This is with `ModelingToolkit v9.19.0`.

hersle:
I suggest this to close #2818 and prevent others from making the same mistake.

This checks that independent variables are `@parameters`, to be consistent with `t_nounits`.

## Checklist

### DIFF
--- a/docs/src/basics/FAQ.md
+++ b/docs/src/basics/FAQ.md
@@ -228,7 +228,7 @@ using ModelingToolkit: t_nounits as t, D_nounits as D
 sts = @variables x1(t) = 0.0
 eqs = [D(x1) ~ 1.1 * x1]
 @mtkbuild sys = ODESystem(eqs, t)
-prob = ODEProblem{false}(sys, [], (0, 1); u0_constructor = x->SVector(x...))
+prob = ODEProblem{false}(sys, [], (0, 1); u0_constructor = x -> SVector(x...))
 ```
 
 ## Using a custom independent variable
--- a/docs/src/basics/FAQ.md
+++ b/docs/src/basics/FAQ.md
@@ -186,7 +186,7 @@ p, replace, alias = SciMLStructures.canonicalize(Tunable(), prob.p)
 
 This error can come up after running `structural_simplify` on a system that generates dummy derivatives (i.e. variables with `ˍt`).  For example, here even though all the variables are defined with initial values, the `ODEProblem` generation will throw an error that defaults are missing from the variable map.
 
-```
+```julia
 using ModelingToolkit
 using ModelingToolkit: t_nounits as t, D_nounits as D
 
@@ -202,7 +202,7 @@ prob = ODEProblem(sys, [], (0,1))
 
 We can solve this problem by using the `missing_variable_defaults()` function
 
-```
+```julia
 prob = ODEProblem(sys, ModelingToolkit.missing_variable_defaults(sys), (0,1))
 ```
 
@@ -221,7 +221,7 @@ julia> ModelingToolkit.missing_variable_defaults(sys, [1,2,3])
 Use the `u0_constructor` keyword argument to map an array to the desired
 container type. For example:
 
-```
+```julia
 using ModelingToolkit, StaticArrays
 using ModelingToolkit: t_nounits as t, D_nounits as D
 
@@ -229,4 +229,18 @@ sts = @variables x1(t)=0.0
 eqs = [D(x1) ~ 1.1 * x1]
 @mtkbuild sys = ODESystem(eqs, t)
 prob = ODEProblem{false}(sys, [], (0,1); u0_constructor = x->SVector(x...))
+```
+
+## Using a custom independent variable
+
+When possible, we recommend `using ModelingToolkit: t_nounits as t, D_nounits as D` as the independent variable and its derivative.
+However, if you want to use your own, you can do so:
+
+```julia
+using ModelingToolkit
+
+@parameters x # independent variables must be created as parameters
+D = Differential(x)
+@variables y(x)
+@named sys = ODESystem([D(y) ~ x], x)
 ```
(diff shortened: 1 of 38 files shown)


isaacsas:
This would be a breaking change though? For years MTK had made `t` a variable and such warnings were not included in MTK9 so users would’ve assumed it was still possible.

hersle:
I suppose, I think some tests will fail without modification.

The alternative, of course, is to require the independent variable to be defined with `@variables` instead of `@parameters`.
Or both definitions should be supported throughout.

isaacsas:
I don’t have an opinion on it either way; MTK has bounced between using parameters vs variables for ivs multiples times now, but if this is merged it should only be released within an MTK10 release since it seems like a breaking change.

ChrisRackauckas:
We already made the breaking change with v9 that everything documented says to use the global time anyways.

```julia
```

it's in every tutorial and it's in the upgrade guide.

ChrisRackauckas:
You need to run the formatter for tests to run.

```julia
using JuliaFormatter
format(raw"C:\Users\accou\.julia\dev\ModelingToolkit/") # or whatever the path is
````

**Comment by isaacsas** [comment]
The `ODESystem` docstring itself still uses `@variables`…

**Comment by isaacsas** [comment]
And I thought the migration guide just said that the defaults should be used, not that the old style was deprecated?

**Comment by ChrisRackauckas** [comment]
Well we should definitely update any docs that were missed to import the `t` from the package. We probably missed a few since it was a huge change, that's not intended.

But ultimately, if structural simplify needs it in one form or the other, that determines it. It really didn't seem to matter before, but maybe @YingboMa can comment on why it matters here.

**Comment by ChrisRackauckas** [comment]
oh actually I see what's going on. This is a byproduct of initialization, since when initialization is running it has to be treated as a parameter and during that process the structural simplification cares. This means this bug only effects DAEs with non-trivial initialization problems where the non-trivial subset is non-autonomous, i.e. the initialization system is required and an initialization equation is directly a function of `t`. It seems fairly expensive to support it not being a parameter then, so that settles the debate once and for all.

But I do want to check in on something. A seemingly unrelated error https://github.com/SciML/SciMLBenchmarks.jl/pull/1007 in dummy derivative sorting in initialization seems suspiciously related.

**Comment by hersle** [comment]
> This means this bug only effects DAEs with non-trivial initialization problems where the non-trivial subset is non-autonomous, i.e. the initialization system is required and an initialization equation is directly a function of t.

Exactly. #2818 is indeed a very simple such example (although with a trivial linear initialization equation).

**Comment by isaacsas** [comment]
If such a change is now being considered, would it perhaps make sense to add metadata to tag an `iv` as an actual `iv`? This seems useful anyways...

**Comment by isaacsas** [comment]
So an `iv` could now be a parameter with specific metadata indicating it is an `iv`?

**Comment by isaacsas** [comment]
(It just seems like this back and forth between releases switching `iv`s from parameters to variables and now back to parameters is indicative that they aren't really either, and so need some more specific way to be flagged internally.)

**Comment by hersle** [comment]
It is also confusing that the independent *variable* is a *parameter* (e.g. #564).

**Comment by hersle** [comment]
I just "pushed through" on this to see that everything can work as expected with the change.

**Comment by hersle** [comment]
We could also export a macro like `@independent_variables t` that (for now) just returns `@parameters t`, but can be changed internally later with smaller chances of breaking user code.

**Comment by isaacsas** [comment]
Yes, I think something like that would be a good approach. Though adding metadata at some point would be useful because then a function like `isiv(sym)` would be trivial to implement (and useful to have).

**Comment by ChrisRackauckas** [comment]
>We could also export a macro like @independent_variables t that (for now) just returns @parameters t, but can be changed internally later with smaller chances of breaking user code.

I think that would be good. And to address Sam's concern of being a breaking change, we can change this to a deprecation warning that the independent variables should be defined with `@independent_variables` and this will error in future versions.

**Comment by hersle** [comment]
I have added an `@independent_variable t` macro. Now it just wraps `@parameters t`, though, so they are both allowed and fully equivalent.

I guess it is possible to treat independent variables as a separate "variable type", but it requires even more work. With a separate treatment, I suppose `check_independent_variables()` could issue a warning if the independent variable is created with `@parameters` or `@variables`.

What do you think?

**Comment by hersle** [comment]
Should
```julia
@variables x y(x)
@named sys = ODESystem([y ~ 0], x)
```
just issue a warning, instead of failing hard?

ChrisRackauckas (approved):

ChrisRackauckas:
Wait, not fully approved because error to warn.

Issue #2858, opened 2024-07-14
Broken Modeling Nonlinear Systems tutorial when using analytical jac


ArnoStrouwen:
https://docs.sciml.ai/ModelingToolkit/stable/tutorials/nonlinear/

At first glance, the generated jac looks overly simplistic?
```jl
julia> prob.f.jac.jac_iip
RuntimeGeneratedFunction(#=in ModelingToolkit=#, #=using ModelingToolkit=#, :((ˍ₋out, ˍ₋arg1, ˍ₋arg2)->begin
          #= C:\Users\arno\.julia\packages\SymbolicUtils\dtCid\src\code.jl:373 =#
          #= C:\Users\arno\.julia\packages\SymbolicUtils\dtCid\src\code.jl:374 =#
          #= C:\Users\arno\.julia\packages\SymbolicUtils\dtCid\src\code.jl:375 =#
          begin
              begin
                  y = (*)(ˍ₋arg1[1], (+)((*)(-1, ˍ₋arg1[2]), ˍ₋arg2[1]))
                  begin
                      #= C:\Users\arno\.julia\packages\Symbolics\Nk48t\src\build_function.jl:546 =#
                      #= C:\Users\arno\.julia\packages\SymbolicUtils\dtCid\src\code.jl:422 =# @inbounds begin
                              #= C:\Users\arno\.julia\packages\SymbolicUtils\dtCid\src\code.jl:418 =#
                              ˍ₋out[1] = (*)(-1, ˍ₋arg2[3])
                              ˍ₋out[2] = y
                              ˍ₋out[3] = 0
                              ˍ₋out[4] = (*)(-1, ˍ₋arg2[2])
                              #= C:\Users\arno\.julia\packages\SymbolicUtils\dtCid\src\code.jl:420 =#
                              nothing
                          end
                  end
              end
          end
      end))
```

hersle:
I don't think it's too simplistic. The simplified equations are
```
julia> equations(ns)
2-element Vector{Equation}:
 0 ~ (-x + y)*σ
 0 ~ x*y - z*β
```
with Jacobian
```
julia> calculate_jacobian(ns)
2×2 Matrix{Num}:
 -σ   0
  y  -β
```
which seems to match the generated expressions.

hersle:
But it's still strange that the `solve` that uses the Jacobian fails.

ArnoStrouwen:
But is `y` not a function of `z` and `x` here?

[...]

PR #2863, opened 2024-07-16
Capture observed dependence on unknowns in simplified nonlinear systems

hersle:
This fixes the incorrect Jacobian in #2858.

## Checklist

### DIFF
--- a/src/systems/nonlinear/nonlinearsystem.jl
+++ b/src/systems/nonlinear/nonlinearsystem.jl
@@ -194,6 +194,8 @@ function calculate_jacobian(sys::NonlinearSystem; sparse = false, simplify = fal
     end
 
     # observed equations may depend on unknowns, so substitute them in first
+    # TODO: must do the same fix in e.g. calculate_hessian?
+    # TODO: rather keep observed derivatives unexpanded, like "Differential(obs)(expr)"?
     obs = map(eq -> eq.lhs => eq.rhs, observed(sys))
     rhs = map(eq -> substitute(eq.rhs, obs), equations(sys))
     vals = [dv for dv in unknowns(sys)]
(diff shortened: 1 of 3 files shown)


hersle:
Would it be better if, for example,
```julia
using ModelingToolkit
@variables x y z
@parameters σ ρ β
@mtkbuild ns = NonlinearSystem([
    0 ~ σ * (y - x)
    0 ~ x * (ρ - z) - y
    0 ~ x * y - β * z
])
J = calculate_jacobian(ns)
```
did not expand observed derivatives like
```
2×2 Matrix{Num}:
 (-1 - z + ρ)*σ        -x*σ
    2x*(-z + ρ)  -β - (x^2)
```
but rather kept them in a form like
```
2×2 Matrix{Num}:
 (-1 + Differential(x)(y))*σ     Differential(z)(y)*σ
    y + x*Differential(x)(y)  x*Differential(z)(y) -β
```
?

src/systems/nonlinear/nonlinearsystem.jl
```diff
@@ -193,8 +193,13 @@ function calculate_jacobian(sys::NonlinearSystem; sparse = false, simplify = fal
         return cache[1]
     end
 
-    rhs = [eq.rhs for eq in equations(sys)]
+    # observed equations may depend on unknowns, so substitute them in first
+    # TODO: must do the same fix in e.g. calculate_hessian?
+    # TODO: rather keep observed derivatives unexpanded, like "Differential(obs)(expr)"?
+    obs = map(eq -> eq.lhs => eq.rhs, observed(sys))
+    rhs = map(eq -> substitute(eq.rhs, obs), equations(sys))
```
ChrisRackauckas, line 200: This won't catch a chain of observed, it would need a fixed point sub I think

hersle: Like in the most recent commit? Can you think of a simple example on to test?

ChrisRackauckas: `z ~ y, `y ~ x`, `x^2 - 2z` should hit it, since if you just plug in z you get y.

hersle: Not quite after structural simplification, it seems.
But I added `0 ~ x^2 + 2*z + y`, `z ~ y`, `y ~ x`, which seems to do the trick.

ChrisRackauckas (approved):

PR #2874, opened 2024-07-18
Throw helpful error when unknowns in the initialization system don't have guesses

### LINKED ISSUE #2873: Helpful error for missing guesses has disappeared
Solving an ODE that needs a guess in initialization, like
```julia
using ModelingToolkit

@variables x(t) y(t)
@named sys = ODESystem([x^2 + y^2 ~ 25, D(x) ~ 1], t)
ssys = structural_simplify(sys)
prob = ODEProblem(ssys, [x => 3], (0, 1), [] #=; guesses = [y => 5]=#)
```
gives the bad error
```
ERROR: LoadError: MethodError: no method matching AbstractFloat(::Type{SymbolicUtils.BasicSymbolic{Real}})

Closest candidates are:
  AbstractFloat(::Bool)
   @ Base float.jl:264
  AbstractFloat(::UInt32)
   @ Base float.jl:272
  AbstractFloat(::UInt8)
   @ Base float.jl:270
  ...

Stacktrace:
  [1] float(x::Type)
    @ Base .\float.jl:294
  [2] promote_to_concrete(vs::Vector{SymbolicUtils.BasicSymbolic{Real}}; tofloat::Bool, use_union::Bool)
    @ ModelingToolkit c:\Users\herma\.julia\dev\Symboltz\dev\ModelingToolkit\src\utils.jl:657
  [3] varmap_to_vars(varmap::Dict{…}, varlist::Vector{…}; defaults::Dict{…}, check::Bool, toterm::Function, promotetoconcrete::Nothing, tofloat::Bool, use_union::Bool)
    @ ModelingToolkit c:\Users\herma\.julia\dev\Symboltz\dev\ModelingToolkit\src\variables.jl:165
  [4] get_u0(sys::NonlinearSystem, u0map::Dict{Any, Any}, parammap::Vector{Pair{…}}; symbolic_u0::Bool, toterm::Function)
    @ ModelingToolkit c:\Users\herma\.julia\dev\Symboltz\dev\ModelingToolkit\src\systems\diffeqs\abstractodesystem.jl:742
  [5] get_u0
    @ c:\Users\herma\.julia\dev\Symboltz\dev\ModelingToolkit\src\systems\diffeqs\abstractodesystem.jl:719 [inlined]
  [6] process_NonlinearProblem(constructor::Type, sys::NonlinearSystem, u0map::Dict{…}, parammap::Vector{…}; version::Nothing, jac::Bool, checkbounds::Bool, sparse::Bool, simplify::Bool, linenumbers::Bool, parallel::Symbolics.SerialForm, eval_expression::Bool, eval_module::Module, use_union::Bool, tofloat::Bool, kwargs::@Kwargs{…})
    @ ModelingToolkit c:\Users\herma\.julia\dev\Symboltz\dev\ModelingToolkit\src\systems\nonlinear\nonlinearsystem.jl:391
```
I believe it used to give something much nicer saying which variables have missing guesses. What happened to that?

hersle:
Attempt to fix #2873.

First draft is very much a WIP. Want to see which tests fail.

## Checklist

### DIFF
--- a/src/systems/diffeqs/abstractodesystem.jl
+++ b/src/systems/diffeqs/abstractodesystem.jl
@@ -1567,11 +1567,9 @@ function InitializationProblem{iip, specialize}(sys::AbstractODESystem,
     defs = defaults(isys)
 
     # Check that all unknowns have guesses
-    # TODO: which, if any, unknowns should be excluded from this guess?
-    for x in unknowns(isys)
-        x in keys(guesses) || x in keys(defs) ||
-            throw(ArgumentError("Missing guess for $x."))
-    end
+    unknowns_missing = setdiff(unknowns(isys), [keys(guesses), keys(defs)])
+    !isempty(unknowns_missing) &&
+        throw(ArgumentError("Missing guesses for unknowns $(join(unknowns_missing, ", "))."))
 
     u0map = merge(guesses, u0map)
     if neqs == nunknown
--- a/src/systems/diffeqs/abstractodesystem.jl
+++ b/src/systems/diffeqs/abstractodesystem.jl
@@ -1564,11 +1564,12 @@ function InitializationProblem{iip, specialize}(sys::AbstractODESystem,
                merge(todict(parammap), Dict(get_iv(sys) => t))
     u0map = isempty(u0map) ? Dict() : todict(u0map)
     guesses = isempty(guesses) ? Dict() : todict(guesses)
+    defaults = defaults(isys)
 
     # Check that all unknowns have guesses
     # TODO: which, if any, unknowns should be excluded from this guess?
     for x in unknowns(isys)
-        x in keys(guesses) || throw(ArgumentError("Missing guess for $x."))
+        x in keys(guesses) || x in keys(defaults) || throw(ArgumentError("Missing guess for $x."))
     end
 
     u0map = merge(guesses, u0map)
--- a/src/systems/diffeqs/abstractodesystem.jl
+++ b/src/systems/diffeqs/abstractodesystem.jl
@@ -1567,7 +1567,7 @@ function InitializationProblem{iip, specialize}(sys::AbstractODESystem,
     defs = defaults(isys)
 
     # Check that all unknowns have guesses
-    unknowns_missing = setdiff(unknowns(isys), [keys(guesses), keys(defs)])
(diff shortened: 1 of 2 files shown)


ChrisRackauckas (approved):

ChrisRackauckas:
Thanks, I did notice in a few cases we were missing an error message somewhere but didn't track down the best spot to put it. This looks correct.

hersle:
Yeah I'm also not 100% sure where to put it. I guess this applies more generally to any `NonlinearSystem`, so maybe it should go there?

ChrisRackauckas:
But an error at that level might be confusing because people might not understand the lack of contextualization, i.e. it would say "u0 is missing a value", when it means "u0 of the initialization" which means "guess". So I think this has to be specialized.

hersle:
Maybe. The stacktrace should be very indicative of the context in such a case, though.

src/systems/diffeqs/abstractodesystem.jl
```diff
@@ -1562,14 +1562,16 @@ function InitializationProblem{iip, specialize}(sys::AbstractODESystem,
...
-    end
-    if isempty(guesses)
-        guesses = Dict()
+    u0map = isempty(u0map) ? Dict() : todict(u0map)
+    guesses = isempty(guesses) ? Dict() : todict(guesses)
+
+    # Check that all unknowns have guesses
+    # TODO: which, if any, unknowns should be excluded from this guess?
+    for x in unknowns(isys)
+        x in keys(guesses) || throw(ArgumentError("Missing guess for $x."))
```
ChrisRackauckas, line 1571:
```suggestion
    guesses = isempty(guesses) ? Dict() : todict(guesses)
    defaults = defaults(isys)

    # Check that all unknowns have guesses
    # TODO: which, if any, unknowns should be excluded from this guess?
    for x in unknowns(isys)
        x in keys(guesses) || x in keys(defaults ) || throw(ArgumentError("Missing guess for $x."))
```

ChrisRackauckas (approved):

hersle:
I am doubtful that this is a good way to fix the issue, at all. It would be better to hit the old error again, instead of throwing a new error a new place. Also, `get_u0()` does some processing further down the pipe that extends the list of guesses from that in `InitializationProblem()`, so it is not really correct to throw an error at such a high level. See comment in #2873.

Issue #2845, opened 2024-07-05
extend() does not pass on initialization equations


hersle:
This does not work:
```julia
using Test
using ModelingToolkit
using ModelingToolkit: t_nounits as t, D_nounits as D

@variables x(t) y(t)
@named sysx = ODESystem([D(x) ~ 0], t; initialization_eqs = [x ~ 1])
@named sysy = ODESystem([D(y) ~ 0], t; initialization_eqs = [y ~ 2])
sys = extend(sysx, sysy)
@test length(equations(generate_initializesystem(sys))) == 2
```

Issue #2502, opened 2024-02-27
Some `ODESystem` fields such as `metadata` are lost when using the `extend` function


avinashresearch1:
**Minimal Reproducible Example 👇**
```
julia> @named sys1 = ODESystem(Equation[],t, [],[]; metadata = Dict(:a => 2))
Model sys1 with 0 equations
States (0):
Parameters (0):

julia> @named sys2 = ODESystem(Equation[],t, [],[])
Model sys2 with 0 equations
States (0):
Parameters (0):

julia> sys_extend = extend(sys1, sys2)
Model sys1 with 0 equations
States (0):
Parameters (0):

julia> sys_extend.metadata

julia> sys1.metadata
Dict{Symbol, Int64} with 1 entry:
  :a => 2
```

**Expected behavior**
One would expect to retain the metadata in the extended system

**Environment**

```julia
[49dc2e85] Calculus v0.5.1
  [0c46a032] DifferentialEquations v7.12.0
⌅ [961ee093] ModelingToolkit v8.75.0
⌃ [16a59e39] ModelingToolkitStandardLibrary v2.3.4
  [2774e3e8] NLsolve v4.5.1
⌃ [1dea7af3] OrdinaryDiffEq v6.70.1
  [91a5bcdd] Plots v1.40.1
  [189a3867] Reexport v1.2.2
  [1bc83da4] SafeTestsets v0.1.0
⌃ [0bca4576] SciMLBase v2.24.0
  [efcf1570] Setfield v1.1.1
⌃ [2efcf032] SymbolicIndexingInterface v0.3.5
  [d1185830] SymbolicUtils v1.5.0
⌃ [0c5d862f] Symbolics v5.17.0
  [3a884ed6] UnPack v1.0.2
  [37e2e46d] LinearAlgebra
  [2f01184e] SparseArrays v1.10.0
```

Issue #1752, opened 2022-08-15
Unsafe to call constructor to create new systems in system transformation methods


bowenszhu:
Some system transformation methods, such as [`convert_system`](https://github.com/SciML/ModelingToolkit.jl/blob/e180e14bf3ea8468d4bcbe4c343009770b8c72bf/src/systems/diffeqs/odesystem.jl#L377) and [`liouville_transform`](https://github.com/SciML/ModelingToolkit.jl/blob/e180e14bf3ea8468d4bcbe4c343009770b8c72bf/src/systems/diffeqs/basic_transformations.jl#L48), call the constructor to create a new system. Usually this works fine, except dropping some fields as mentioned in issue #1710. 

The method [`tearing_reassemble`](https://github.com/SciML/ModelingToolkit.jl/blob/e180e14bf3ea8468d4bcbe4c343009770b8c72bf/src/structural_transformation/symbolics_tearing.jl#L175), which is often called by [`dummy_derivative`](https://github.com/SciML/ModelingToolkit.jl/blob/e180e14bf3ea8468d4bcbe4c343009770b8c72bf/src/structural_transformation/symbolics_tearing.jl#L627) and [`structural_simplify`](https://github.com/SciML/ModelingToolkit.jl/blob/e180e14bf3ea8468d4bcbe4c343009770b8c72bf/src/systems/abstractsystem.jl#L959), sets the field [`substitutions`](https://github.com/SciML/ModelingToolkit.jl/blob/e180e14bf3ea8468d4bcbe4c343009770b8c72bf/src/systems/diffeqs/odesystem.jl#L117) of [`ODESystem`](https://github.com/SciML/ModelingToolkit.jl/blob/e180e14bf3ea8468d4bcbe4c343009770b8c72bf/src/systems/diffeqs/odesystem.jl#L25). In most cases, after calling `structural_simplify`, the field `substitutions` of the resulting `ODESystem` is empty. However, when `substitutions` is not empty and the resulting system is transformed using a method which calls the constructor, an error occurs when one tries to `solve` it.

The following is a test case:
Version: julia v1.7.3, DifferentialEquations v7.2.0, MethodOfLines v0.4.0, ModelingToolkit v8.18.9
```julia
using ModelingToolkit, MethodOfLines, DifferentialEquations
@variables x t v(..) w(..)
Dx = Differential(x)
Dxx = Dx * Dx
Dt = Differential(t)
const L = 1.0
const ε = 0.015
[...]

Issue #1710, opened 2022-07-30
System transformations and alterations sometimes dropping fields


isaacsas:
This is something that came up as I was looking at adding events into other system types. Maybe this isn't an easy fix, or desired, but right now when new fields get added to systems they don't always get accounted for in all places systems are transformed. For example, 

https://github.com/SciML/ModelingToolkit.jl/blob/9c0d0437e300dab6d50ea1f7369e554e66f41b45/src/systems/diffeqs/odesystem.jl#L377

drops fields like observed, as does

https://github.com/SciML/ModelingToolkit.jl/blob/9c0d0437e300dab6d50ea1f7369e554e66f41b45/src/systems/diffeqs/sdesystem.jl#L164

and 

https://github.com/SciML/ModelingToolkit.jl/blob/9c0d0437e300dab6d50ea1f7369e554e66f41b45/src/systems/diffeqs/sdesystem.jl#L261

It seems like there is a need for a transformation/copying mechanism where one can just specify values for sub-fields that have changed, but the remaining fields get automatically propagated. 

I guess this could potentially introduce silent bugs by incorrectly propagating fields that actually require transformations, as opposed to now where there are silent bugs with fields being dropped. Maybe the copying mechanism needs to require that all fields be explicitly passed in unless one passes a flag that default values can be propagated for non-selected fields (forcing one to make a choice to propagate the remaining values unchanged). 

It does seem though like it would be helpful to have such a functionality, since system transformations with only modifying a few fields crop up all over the place.

ChrisRackauckas:
Agreed on this. Some kind of remake functionality could improve the robustness of pass writing.

bowenszhu:
The transformation methods mentioned in the PR description all create a new system by calling the corresponding constructor, while some other methods, for example `alias_elimination` and `ode_order_lowering`, make use of [`@set!`](https://jw3126.github.io/Setfield.jl/stable/#Setfield.@set!-Tuple{Any}) from [Setfield.jl](https://github.com/jw3126/Setfield.jl), which safely keep the original fileds including `observed`, `defaults`, `continuous_events`.

https://github.com/SciML/ModelingToolkit.jl/blob/a09b9140eff8ebd87dfbbc78a1a92037fc2ab7a8/src/systems/diffeqs/first_order_transform.jl#L7-L13

[...]

PR #2875, opened 2024-07-18
Transform systems robustly by remaking systems

hersle:
This is an attempt to strengthen the robustness of system transformations (i.e. processes that turn one system into another).

Internally, as pointed out in #1710 and #1752, there is inconsistency between

1. transformations that re-calls the system constructor with the modified fields (dropping unspecified fields, but redoing checks),
2. transformations that uses `@set! sys.field = value` from Setfield.jl (keeping unspecified fields, but skipping checks).

Field dropping has spawned bugs like #2845 and #2502, which usually gets "duct tape patched" by adding that single field for one particular system type. Transforming systems in a uniform way could prevent such bugs in the future, and maybe simplify some internal machinery. Maybe it would also be more efficient to "batch remake" several modified fields with a single new constructor call, rather than `@set! sys.field1 = value1; @set! sys.field2 = value2; ...` which I think calls the constructor several times.

Externally, I "envision" that a simple and robust interface could facilitate users making their own custom transformations. Maybe another package using ModelingToolkit would like to build models in a way that requires some simple application-dependent transformation of the system. Every user shouldn't have to reinvent a machinery for doing this themselves.

This is a draft for now. I would really appreciate thoughts and suggestions, or to know if you think this is not a good idea.

## Checklist

### DIFF
--- a/src/systems/abstractsystem.jl
+++ b/src/systems/abstractsystem.jl
@@ -741,7 +741,9 @@ WARNING: intended for internal use; does not perform any sanity checks.
 """
 # TODO: optionally re-call constructor to sanity check?
 # TODO: use SciMLBase's generic remake()? doesn't work out of the box, though
-function remake(sys::AbstractSystem; skip_nonexisting = false, kwargs...)
+# TODO: should register new tag? move Threads.atomic_add!(SYSTEM_COUNT, ...) to a separate function?
+# TODO: recreate the struct once with all new fields, instead of once for every field
+function remake(sys::AbstractSystem; kwargs...)
     for (field, value) in kwargs
         # like `Setfield.@set! sys.field = value`, but with `field` replaced by an arbitrarily named symbol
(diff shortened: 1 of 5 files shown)


hersle:
For now I've tried the `remake(sys)` functionality out in a few select places. In particular, I "showcase" how it can be used to implement an `extend()` function that handles all fields of arbitrary system types (hopefully).

I would try to take it further, but first I think it is more important to decide whether this is a good idea at all, and if so to settle the base mechanism 😅

ChrisRackauckas:
I think this is a great idea. Indeed mutation is not a good idea on systems, and a safe `remake` would help make things more robust for people to do things.

isaacsas:
FWIW in Catalyst I came to the conclusion this would be very helpful to have and perhaps reduce errors when transforming systems a little while ago: https://github.com/SciML/Catalyst.jl/issues/935

We would definitely hook into this for `ReactionSystem`s.

PR #2877, opened 2024-07-19
Restore error when unknowns in the initialization system don't have guesses 

hersle:
Another attempt to fix https://github.com/SciML/ModelingToolkit.jl/issues/2873 in a simpler way; supersedes #2874.

## Checklist

### DIFF
--- a/docs/src/tutorials/initialization.md
+++ b/docs/src/tutorials/initialization.md
@@ -195,7 +195,7 @@ long enough you will see that `λ = 0` is required for this equation, but since 
 `λ = 1` we end up with a set of equations that are impossible to satisfy.
 
 !!! note
-
+    
     If you would prefer to have an error instead of a warning in the context of non-fully
     determined systems, pass the keyword argument `fully_determined = true` into the
     problem constructor. Additionally, any warning about not being fully determined can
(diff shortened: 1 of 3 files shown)


hersle:
Seems weird that just disabling one early-return optimization causes one test to fail. Looks like there is just a `t` that is not substituted. Have DDEs just not gotten the non-autonomous treatment?

ChrisRackauckas:
>  Looks like there is just a t that is not substituted. Is there something DDE-specific going on here, or have DDEs just not gotten the non-autonomous treatment?

DDEs naturally do a bit more with `t` since it's common to have say `x(t - tau)`, or `x(t - y(t))`. I wonder if those cases trip it up.

ChrisRackauckas (approved):

ChrisRackauckas:
This looks good to go though? Just the latexify regression test needs an update, but I'd be fine handling that after merge.

hersle:
Yes, that's what I realized. It seems that DDEs relied on the early return, which is why they pass `tofloat = false` in the first place.

I think it's good to merge, yes.

ChrisRackauckas:
Thanks for seeing this one through. I hadn't found the time to do it since it was pretty non-trivial, but in the end I think you figured out a good solution.

PR #2898, opened 2024-07-26
Make default value units consistent

contradict:
When using units in a component, it would be nice if those units could be checked on the default and initial values. With old ModelingToolkit, it was possible to do `@parameters p = ustrip(expected_unit, provided_value), [unit=expected_unit]`, but that is not longer possible. This PR is a draft of a way to restore that functionality. It only works with Unitful and only with scalar variables, but these could both be fixed if this is desirable.

The result is a component defined like this
```
julia> @mtkmodel ATest begin
       @parameters begin
       some=1.0, [description="Some", unit=u"kg"]
       more, [description="more", unit=u"L"]
       end
       end
ModelingToolkit.Model{typeof(__ATest__), Dict{Symbol, Any}}(__ATest__, Dict{Symbol, Any}(:kwargs => Dict{Symbol, Dict}(:some => Dict{Symbol, Any}(:value => 1.0, :type => Real), :more => Dict{Symbol, Union{Nothing, DataType}}(:value => nothing, :type => Real)), :independent_variable => t, :parameters => Dict{Symbol, Dict{Symbol, Any}}(:some => Dict(:default => 1.0, :unit => :(u"kg"), :type => Real, :description => "Some"), :more => Dict(:unit => :(u"L"), :type => Real, :description => "more"))), false)
```

can be instantiated like this

```
julia> ATest(;some=3.0u"g", name=:t)
Model t with 0 equations
Unknowns (0):
Parameters (2):
  some [defaults to 0.003]: Some
  more: more
```

And the units are converted correctly. Errors are given for incompatible or missing units.

Is this a desirable feature? If this PR were fleshed out to handle DynamicQuantities and array variables as well, is there a chance it could be merged?

### DIFF
--- a/src/systems/model_parsing.jl
+++ b/src/systems/model_parsing.jl
@@ -628,7 +628,8 @@ function convert_units(varunits::DynamicQuantities.Quantity, value)
         DynamicQuantities.SymbolicUnits.as_quantity(varunits), value))
 end
 
-function convert_units(varunits::DynamicQuantities.Quantity, value::AbstractArray{T}) where T
+function convert_units(
+        varunits::DynamicQuantities.Quantity, value::AbstractArray{T}) where {T}
     DynamicQuantities.ustrip.(DynamicQuantities.uconvert.(
         DynamicQuantities.SymbolicUnits.as_quantity(varunits), value))
 end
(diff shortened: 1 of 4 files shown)


ChrisRackauckas:
I think it's desirable. Handling DynamicQuantities is essential.

ChrisRackauckas (approved):

PR #2903, opened 2024-07-28
Getting started doc rewrite

ArnoStrouwen:
The getting started documentation is currently a confusing combination of sometimes using the `mtkmodel` macro, and sometimes not. This PR puts as much functionality as possible into said macro, and removes all sidetracking explanations of the more manual definitions of `ODESystems` and `ODEProblem`.

I'll also take the opportunity to note the sentences in the tutorial I did not fully understand:

> And the independent variable t is automatically added by @mtkmodel.

Added where exactly?

> Vector-valued parameters and variables are possible. A cleaner, more
> consistent treatment of these is still a work in progress, however. Once finished,
> this introductory tutorial will also cover this feature.

Is this ready by now?

> Observed equations are variables which can be computed on-demand but are not necessary for the solution of the system.

Calculating RHS is definitely needed for solving the system (you can see this by inspecting `prob.f.f.f_oop`, it is just not a state of the dynamic system? I don't know how to succinctly rewrite this.

### DIFF
--- a/docs/src/tutorials/ode_modeling.md
+++ b/docs/src/tutorials/ode_modeling.md
@@ -16,7 +16,7 @@ Pkg.add("ModelingToolkit")
 ## Copy-Pastable Simplified Example
 
 A much deeper tutorial with forcing functions and sparse Jacobians is below.
-But if you want to just see some code and run, here's an example:
+But if you want to just see some code and run it, here's an example:
 
 ```@example first-mtkmodel
 using ModelingToolkit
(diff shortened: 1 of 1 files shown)


docs/src/tutorials/ode_modeling.md
```diff
@@ -36,7 +36,7 @@ end
 
 using DifferentialEquations: solve
 @mtkbuild fol = FOL()
-prob = ODEProblem(fol, [fol.x => 0.0], (0.0, 10.0), [fol.τ => 3.0])
+prob = ODEProblem(fol, [], (0.0, 10.0), [])
```
ChrisRackauckas, line 39: It would be good to add a section with "creating the model with new initial conditions and parameters", since with these changes the later overrides are unused.

ChrisRackauckas (approved):

PR #2909, opened 2024-07-30
Default to using fully_determined when possible in initialization

ChrisRackauckas:
This should give better error messages


ChrisRackauckas:
I'm not sure this is legal in general. Counting the number of equations does not actually give if fully determined since there might be trivial equations, and the tests suggest that's not too uncommon for initializations? @YingboMa opinion?

AayushSabharwal:
What if we make it so that passing `fully_determined = missing` will `try-catch` the consistency check and take the appropriate code path?

YingboMa (commented):
This is not correct in general. You should check if the DAE system has finite index.

ChrisRackauckas:
The DAE system would have to already be structurally simplified at this point, so it should be index 1 or have errored already?

PR #2911, opened 2024-07-31
Support more of the SciMLBase events API

BenChung:
## Checklist

## Additional context

This extends `SymbolicContinuousCallback` to support the `affect_neg` and `rootfind` mechanisms provided by the underlying SciMLBase callback API. In particular, the semantics of `affect_neg` are mirrored into the symbolic domain (same as affect by default; disabled if `nothing`, and usable as either the equation or `FunctionalAffect` interface). The implementation of `rootfind` is somewhat more cursed; we cannot generate a single `VectorContinuousCallback` for different `rootfind` options, so instead generate several using the `rootfind` value as an equivalence class and then sort the resulting callbacks in order of `rootfind`. The alternative approach would be to generate a `CallbackSet` if the order violated the order of `VectorContinousCallbacks` (e.g. if we had a left then right then left rootfind selection), but this comes with a performance penalty. As a result, we have a choice between impacting performance and odd semantics; I'd appreciate feedback on what the right choice is here.

### DIFF
--- a/src/systems/callbacks.jl
+++ b/src/systems/callbacks.jl
@@ -84,11 +84,11 @@ as well as the positive-edge `affect` and negative-edge `affect_neg` that apply 
 By default `affect_neg = affect`; to only get rising edges specify `affect_neg = nothing`.
 
 Assume without loss of generality that the equation is of the form `c(u,p,t) ~ 0`; we denote the integrator state as `i.u`. 
-For simplicty, we define `prev_sign = sign(c(u[t-1], p[t-1], t-1))` and `cur_sign = sign(c(u[t], p[t], t))`.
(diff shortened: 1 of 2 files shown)


BenChung:
Stuff that I want to do for this commit:

* [ ] Fully check compliance with the style guide. 
* [ ] Extend the narrative documentation that exploits the affect_neg and rootfinding side functionality
* [ ] Write more extensive docstrings for the relevant methods & expose this into the public facing documentation.

BenChung:
I'm going to put of writing more extensive docstrings until after the API reference gets generated.

ChrisRackauckas (approved):

PR #2914, opened 2024-07-31
Detect cases where there's fewer brownians than equations, but noise still 'diagonal'

MasonProtter:
## Checklist

## Additional context

Followup to #2882 and #2886

### DIFF
--- a/test/sdesystem.jl
+++ b/test/sdesystem.jl
@@ -725,12 +725,12 @@ end
 
 @testset "Non-diagonal noise check" begin
     @parameters σ ρ β
-    @variables x(t) y(t) z(t)
+    @variables x(tt) y(tt) z(tt)
     @brownian a b c
     eqs = [D(x) ~ σ * (y - x) + 0.1a * x + 0.1b * y,
         D(y) ~ x * (ρ - z) - y + 0.1b * y,
         D(z) ~ x * y - β * z + 0.1c * z]
-    @mtkbuild de = System(eqs, t)
+    @mtkbuild de = System(eqs, tt)
 
     u0map = [
         x => 1.0,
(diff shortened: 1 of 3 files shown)


test/sdesystem.jl
```diff
@@ -746,5 +746,32 @@ end
...
+
+    parammap = [
+        σ => 10.0,
+        β => 26.0,
+        ρ => 2.33
+    ]
+
+    prob = SDEProblem(de, u0map, (0.0, 100.0), parammap)
+    @test solve(prob, SOSRI()).retcode == ReturnCode.Success
+end
```
MasonProtter, line 777: Tested this example and it failed before this PR.

ChrisRackauckas (approved):

MasonProtter:
The downstream failure in Catalyst is caused by this, not entirely sure why it failed though it 's a bit weird. I'll try a more explicit version of the check for whether the noise is diagonal.

MasonProtter:
Okay, I think I figured out the problem, I was iterating over the axes of the noise equations in the wrong order, effectively looking at the transpose of the noise equations.

This bug was masked in the test suite, but it reared its head in the Catalyst tests because it creates some reaction systems where the noise matrix is of the form
```julia
[a 0 0 0 e 0 0 0
 0 b 0 0 0 f 0 0
 0 0 c 0 0 0 g 0
 0 0 0 d 0 0 0 h]
```
and since I was iterating over the transpose of that, I was mistakenly detecting it as diagonal noise.

Should be fixed now, and I'd updated the non-diagonal noise test to watch for this case.

MasonProtter:
Hm, so now the failing Catalyst test is happening because Catalyst does this:
```julia
    p_matrix = zeros(length(get_unknowns(sde_sys)), numreactions(rs))
    return SDEProblem(sde_sys, u0map, tspan, pmap, args...; check_length,
        noise_rate_prototype = p_matrix, kwargs...)
```
in https://github.com/SciML/Catalyst.jl/blob/v14.1.0/src/reactionsystem_conversions.jl#L767. What this ends up doing is hard-coding into the SDEProblem that the noise is going to be a matrix, rather than diagonal, but that interacts badly with  the change added in https://github.com/SciML/ModelingToolkit.jl/pull/2886, which will unconditionally make `g` assume the noise matrix is a vector, regardless of what the user is doing with the noise equations.

I've removed that change from #2886 and things seems to work now, we'll see how this test-run goes.

isaacsas:
If calling SDEProblem on an SDESystem now infers the prototype optimally we could drop that in Catalyst.

isaacsas:
By drop I just mean we won’t pass a prototype.

MasonProtter:
Yeah, dropping that would be be good because some cases in Catalyst would then get a performance boost, but it's also important that we not break code that *does* assume a specific form of the prototype like that, so I think the change I made is still necessary (though perhaps I misunderstood something, and @AayushSabharwal can chime in).

MasonProtter:
@ChrisRackauckas can you merge this?

PR #2916, opened 2024-08-01
rewrite optimization tutorial

ArnoStrouwen:
Turns the optimization tutorial into an actual Documenter example.

However, I'm stuck at the guess metadata.
`OptimizationProblem` complains that no guess is present for `x.`

I also think the comma present after `x` and `y` looks a bit weird:

```jl
@variables begin
    x, [bounds = (-2.0, 2.0), guess = 1.0]
    y, [bounds = (-1.0, 3.0), guess = 3.0]
end
```

No such comma is present in the examples on:
https://docs.sciml.ai/ModelingToolkit/stable/basics/Variable_metadata/#Additional-functions
Yet deleting them here gives a whitespace error.

### DIFF
--- a/docs/src/tutorials/optimization.md
+++ b/docs/src/tutorials/optimization.md
@@ -19,38 +19,27 @@ Let's optimize the classical _Rosenbrock function_ in two dimensions.
 ```@example optimization
 using ModelingToolkit, Optimization, OptimizationOptimJL
 @variables begin
-    x, [bounds = (-2.0, 2.0), guess = 1.0]
-    y, [bounds = (-1.0, 3.0), guess = 3.0]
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
> I also think the comma present after x and y looks a bit weird:

It's a detail that comes from Julia's parser.

ArnoStrouwen:
Apparently, the guess metadata is only compatible with initialization problems for ODE.
I mistakenly thought this would have also transferred to nonlinear problem and optimization problem.
Instead, I had to use the default value.

When this passes tests, it should be good to go.

ChrisRackauckas:
Yes, guess semantics are for embedded initialization problems. The starting value for an optimization or nonlinear problem is not considered a guess in these semantics, but the initial state.

PR #2919, opened 2024-08-02
Error hint on Num in boolean context

baggepinnen:
Use of a symbolic expression in a boolean context is a rather common user mistake, this PR adds a hint with possible fixes
image


ChrisRackauckas:
This is the wrong repository. It's effectively piracy here. It should be in Symbolics.jl

PR #2922, opened 2024-08-03
Improved FunctionalAffect alternative

BenChung:
## Checklist

## Additional context

This introduces `MutatingFunctionalAffect` (better name suggestions much appreciated) which presents a higher level interface when compared to FunctionalAffect. In particular, it decouples the inputs and outputs to the affect from the precise problem structure post-simplification.

This isn't quite complete yet; I need to implement it marking unknowns mutating by affects irreducible so that they're always there come runtime. Close, however.


BenChung:
I'm increasingly happy with this implementation; the main caveat to that statement is that the use of ComponentArrays for MutatingFunctionalAffects means that the values passed in and out are homogenously typed (that is they're all float64s, which is not great).

src/systems/callbacks.jl
```diff
@@ -749,7 +977,144 @@ function compile_user_affect(affect::FunctionalAffect, cb, sys, dvs, ps; kwargs.
...
+        function (integ)
+            # update the to-be-mutated values; this ensures that if you do a no-op then nothing happens
+            modvals = mod_og_val_fun(integ.u, integ.p..., integ.t)
+            upd_component_array = NamedTuple{mod_names}(modvals)
+
+            # update the observed values
+            obs_component_array = NamedTuple{obs_sym_tuple}(obs_fun(integ.u, integ.p..., integ.t))
+
+            # let the user do their thing
+            modvals = if applicable(user_affect, upd_component_array, obs_component_array, ctx, integ)
```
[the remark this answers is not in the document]
BenChung: I think that applicable and hasmethod are essentially identical modulo hasmethod makes you write out the type. I don't think that (in inferrable code, for which I haven't checked this yet) there should be a lot of difference performance-wise; hasmethod is more general but somewhat more annoying to use.

src/systems/callbacks.jl
```diff
@@ -73,6 +71,112 @@ function namespace_affect(affect::FunctionalAffect, s)
...
+    obs_vals = join(map((ob,nm) -> "$ob => $nm", mfa.obs, mfa.obs_syms), ", ")
+    mod_vals = join(map((md,nm) -> "$md => $nm", mfa.modified, mfa.mod_syms), ", ")
+    affect = mfa.f
+    print(io, "MutatingFunctionalAffect(observed: [$obs_vals], modified: [$mod_vals], affect:$affect)")
+end
+func(f::MutatingFunctionalAffect) = f.f
+context(a::MutatingFunctionalAffect) = a.ctx
+observed(a::MutatingFunctionalAffect) = a.obs
+observed_syms(a::MutatingFunctionalAffect) = a.obs_syms
+discretes(a::MutatingFunctionalAffect) = filter(ModelingToolkit.isparameter, a.modified)
```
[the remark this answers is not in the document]
BenChung: It is not at the present (we would need to have the writeback happen into the initialization system), and my detection of that is sub-par. I'll see about fixing that.

Project.toml
```diff
@@ -9,6 +9,7 @@ ArrayInterface = "4fba245c-0d91-5ea0-9b3e-6abc04ee57a9"
 BlockArrays = "8e7c35d0-a365-5155-bbbb-fb81a777f24e"
 Combinatorics = "861a8166-3701-5b0c-9a16-15d98fcdc6aa"
 Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
+ComponentArrays = "b0b7db55-cfe3-40fc-9ded-d10e2dbeff66"
```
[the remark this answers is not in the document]
BenChung: Can be removed now that I've switched to tuples, sorry about that.

AayushSabharwal:
This looks like a rebase gone wrong 😅

BenChung:
Yep, alas. I'll fix it with a force push

docs/src/basics/Events.md
```diff
@@ -378,3 +378,201 @@ sol.ps[c] # sol[c] will error, since `c` is not a timeseries value
...
+            @set! x.hA = x.qA
+            @set! x.hB = o.qB
+            @set! x.qA = 0
+            @set! x.cnt += decoder(x.hA, x.hB, x.qA, o.qB)
+            x
+        end)
+```
+
+The other way we can implement a sensor is by changing the root find.
+Normally, we use left root finding; the affect will be invoked instantaneously before
```
baggepinnen, line 564:
```suggestion
Normally, we use left root finding; the affect will be invoked instantaneously _before_
```

docs/src/basics/Events.md
```diff
@@ -378,3 +378,201 @@ sol.ps[c] # sol[c] will error, since `c` is not a timeseries value
...
 
 It can be seen that the timeseries for `c` is not saved.
+
+
+## [(Experimental) Imperative affects](@id imp_affects)
+The `ImperativeAffect` can be used as an alternative to the aforementioned functional affect form. Note
+that `ImperativeAffect` is still experimental; to emphasize this, we do not export it and it should be
+included as `ModelingToolkit.ImperativeAffect`. It abstracts over how values are written back to the 
+system, simplifying the definitions and (in the future) allowing assignments back to observed values
+by solving the nonlinear reinitialization problem afterwards.
```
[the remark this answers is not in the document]
ChrisRackauckas: Indeed it's not always possible to invert such functions.

docs/src/basics/Events.md
```diff
@@ -378,3 +378,201 @@ sol.ps[c] # sol[c] will error, since `c` is not a timeseries value
...
+Bang-bang control of a heater connected to a leaky plant requires hysteresis in order to prevent control oscillation.
+
+```@example events 
+@variables temp(t)
+params = @parameters furnace_on_threshold=0.5 furnace_off_threshold=0.7 furnace_power=1.0 leakage=0.1 furnace_on(t)::Bool=false
+eqs = [
+    D(temp) ~ furnace_on * furnace_power - temp^2 * leakage
+]
+```
+Our plant is simple. We have a heater that's turned on and off by the clocked parameter `furnace_on`
```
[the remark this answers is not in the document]
ChrisRackauckas: yes

docs/src/basics/Events.md
```diff
@@ -378,3 +378,220 @@ sol.ps[c] # sol[c] will error, since `c` is not a timeseries value
...
+
+We need a controller with hysteresis to control the plant. We wish the furnace to turn on when the temperature
+is below `furnace_on_threshold` and off when above `furnace_off_threshold`, while maintaining its current state
+in between. To do this, we create two continuous callbacks:
+
+```@example events
+using Setfield
+furnace_disable = ModelingToolkit.SymbolicContinuousCallback(
+    [temp ~ furnace_off_threshold],
+    ModelingToolkit.ImperativeAffect(modified = (; furnace_on)) do x, o, i, c
```
ChrisRackauckas, line 419: can we use more explicit canonical naming of the args?

docs/src/basics/Events.md
```diff
@@ -378,3 +378,220 @@ sol.ps[c] # sol[c] will error, since `c` is not a timeseries value
...
+evaluate this before calling our function to fill out all of the numerical values, then apply them back to the system
+once our affect function returns. Furthermore, it will check that it is possible to do this assignment.
+
+The function given to `ImperativeAffect` needs to have one of four signatures, checked in this order:
+
+  - `f(modified::NamedTuple, observed::NamedTuple, ctx, integrator)::NamedTuple` if the function needs the low-level integrator,
+  - `f(modified::NamedTuple, observed::NamedTuple, ctx)::NamedTuple` if the function needs the user-defined context,
+  - `f(modified::NamedTuple, observed::NamedTuple)::NamedTuple` if the function also reads observed values from the system,
+  - `f(modified::NamedTuple)::NamedTuple` if the function only writes values (unknowns or parameters) to the system.
+    The `do` block in the example implicitly constructs said function inline. For exposition, we use the full version (e.g. `x, o, i, c`) but this could be simplified to merely `x`.
```
ChrisRackauckas, line 463: Just simplify it down to one? This seems overly complicated.

src/systems/diffeqs/odesystem.jl
```diff
@@ -419,7 +443,8 @@ function build_explicit_observed_function(sys, ts;
         return_inplace = false,
         param_only = false,
         op = Operator,
-        throw = true)
+        throw = true,
+        array_type = :array)
```
ChrisRackauckas, line 447: it should be a type not a symbol?

src/systems/callbacks.jl
```diff
@@ -73,6 +71,116 @@ function namespace_affect(affect::FunctionalAffect, s)
         context(affect))
 end
 
+"""
```
ChrisRackauckas, line 74: move this to its own file?

src/systems/callbacks.jl
```diff
@@ -123,17 +256,79 @@ make_affect(affect::NamedTuple) = FunctionalAffect(; affect...)
 
 function Base.:(==)(e1::SymbolicContinuousCallback, e2::SymbolicContinuousCallback)
     isequal(e1.eqs, e2.eqs) && isequal(e1.affect, e2.affect) &&
+        isequal(e1.initialize, e2.initialize) && isequal(e1.finalize, e2.finalize) &&
         isequal(e1.affect_neg, e2.affect_neg) && isequal(e1.rootfind, e2.rootfind)
 end
 Base.isempty(cb::SymbolicContinuousCallback) = isempty(cb.eqs)
 function Base.hash(cb::SymbolicContinuousCallback, s::UInt)
```
ChrisRackauckas, line 263: this is an unrelated change? Can we split this to a separate PR and merge it? It's much easier to review things that are one idea.

src/systems/callbacks.jl
```diff
@@ -749,7 +1002,165 @@ function compile_user_affect(affect::FunctionalAffect, cb, sys, dvs, ps; kwargs.
...
+            elseif applicable(user_affect, upd_component_array, obs_component_array, ctx)
+                user_affect(upd_component_array, obs_component_array, ctx)
+            elseif applicable(user_affect, upd_component_array, obs_component_array)
+                user_affect(upd_component_array, obs_component_array)
+            elseif applicable(user_affect, upd_component_array)
+                user_affect(upd_component_array)
+            else
+                @error "User affect function $user_affect needs to implement one of the supported ImperativeAffect callback forms; see the ImperativeAffect docstring for more details"
+                user_affect(upd_component_array, obs_component_array, integ, ctx) # this WILL error but it'll give a more sensible message
+            end
```
ChrisRackauckas, line 1150: Yeah very over complicated. Just simplify the interface.

BenChung: Would you suggest all four, all the time?

baggepinnen: > Would you suggest all four, all the time?

I see no downside with having them all. it makes it easier to read the docs since there is only one way to do it

ChrisRackauckas:
Rebase.

src/systems/diffeqs/odesystem.jl
```diff
@@ -564,12 +589,12 @@ function build_explicit_observed_function(sys, ts;
         oop_mtkp_wrapper = mtkparams_wrapper
     end
 
+    output_expr = isscalar ? ts[1] :
+                  (array_type <: Vector ? MakeArray(ts, output_type) : MakeTuple(ts))
```
baggepinnen, line 593: If I pass `StaticVector{Float64, 3}` here, I still get a `Vector` right? Seems slightly confusing, although the docstring is rather clear on what options are available. 

There is an option to get static arrays out
https://github.com/JuliaSymbolics/Symbolics.jl/blob/10a61ee505038f4faaf9e1edbcda1caeed8d4cc7/src/build_function.jl#L285
which I've often used for nice performance improvements in MPC and state estimation

BenChung:
@ChrisRackauckas In terms of simplification, what'd be your thoughts on how to trim this down in that case?

src/systems/diffeqs/odesystem.jl
ChrisRackauckas: Yes it would make sense to do the same thing here.

ChrisRackauckas:
It's just hard to review because there's multiple things going on here. There's an extension to SymbolicContinuousCallback, there's changing options in `build_explicit_observed_function`, and there's ImperativeAffect callback. So it's effectively 3 PRs at once. I think the current merge blocker would be the arguments in `build_explicit_observed_function` vs `build_function`.

Also, since the ImperativeAffect is marked as experimental, it should all be isolated to a file so that if it's removed it can be done cleanly. It seems like there are utils for ImperativeAffect that are mixed in with the other callbacks?

src/systems/callbacks.jl
```diff
@@ -927,8 +915,41 @@ function generate_vector_rootfinding_callback(
             end
         end
     end
-    initialize = handle_optional_setup_fn(
-        map(fn -> fn.initialize, affect_functions), SciMLBase.INITIALIZE_DEFAULT)
+    initialize = nothing
+    if has_index_cache(sys) && (ic = get_index_cache(sys)) !== nothing
+        initialize = handle_optional_setup_fn(
+            map(
+                (cb, fn) -> begin
```
baggepinnen, line 922:
```suggestion
                function (cb, fn) 
```
anonymous functions can be written this way as well, which maybe looks slightly

src/systems/callbacks.jl
```diff
@@ -88,17 +90,26 @@ By default `affect_neg = affect`; to only get rising edges specify `affect_neg =
...
 guaranteed to be triggered.
 
 Once detected the integrator will "wind back" through a root-finding process to identify the point when the condition became active; the method used 
-is specified by `rootfind` from [`SciMLBase.RootfindOpt`](@ref). Multiple callbacks in the same system with different `rootfind` operations will be resolved 
-into separate VectorContinuousCallbacks in the enumeration order of `SciMLBase.RootfindOpt`, which may cause some callbacks to not fire if several become
-active at the same instant. See the `SciMLBase` documentation for more information on the semantic rules.
-
-The positive edge `affect` will be triggered iff an edge is detected and if `prev_sign < 0`; similarly, `affect_neg` will be
-triggered iff an edge is detected `prev_sign > 0`. 
+is specified by `rootfind` from [`SciMLBase.RootfindOpt`](@ref). If we denote the time when the condition becomes active as `tc``,
```
AayushSabharwal, line 101:
```suggestion
is specified by `rootfind` from [`SciMLBase.RootfindOpt`](@ref). If we denote the time when the condition becomes active as `tc`,
```

AayushSabharwal (approved):
Just a minor doc thing I found

ChrisRackauckas (approved):

PR #2923, opened 2024-08-04
Reorganizing tutorials

ArnoStrouwen:
The tutorial list has become very unwieldy.

This PR gives a starting point to discuss further splitting them into groups.

I also do not understand the difference with some pages in the examples section.
Why is initializing an ODE a tutorial, but lowering the order of an ODE an example?


ChrisRackauckas:
> Why is initializing an ODE a tutorial

It describes the initialization features

> lowering the order of an ODE an example

It's just a quick example that we support higher order ODEs. It doesn't explain much about the features.

cstjean:
As a beginner who went through the documentation in the past few weeks... Consolidating the basics into a few tutorials would be very helpful.

Issue #2905, opened 2024-07-29
Optimizing a control signal


cstjean:
(from https://discourse.julialang.org/t/sciml-optimizing-a-control-signal/117504 and https://discourse.julialang.org/t/solving-ode-parameters-using-experimental-data-with-control-inputs/66614/20; apologies if this isn't using quite the right MTK jargon, I'm new at this)

Suppose I’ve got a differential equation that takes some forcing / control signal as input defined as a `DataInterpolation.LinearInterpolation(u, t)` (similar to the [tutorial example](https://docs.sciml.ai/ModelingToolkit/dev/tutorials/ode_modeling/#Specifying-a-time-variable-forcing-function)). I’d like to find the `u` vector which minimizes some loss function. How should I do that with ModelingToolkit?

- I can pass the individual `u1`, `u2`, `u3` as parameters, then call `LinearInterpolation([u1,u2,u3], t)`. That doesn’t look very practical, but maybe I can create many parameters in a comprehension?
- I believe that I can pass non-numerical values (like Vector, or LinearInterpolation objects) as parameters, by specifying their type somehow. I can’t find the documentation anymore, but is that how it’s done?
- [DiffEqFlux.jl](https://juliahub.com/ui/Packages/General/DiffEqFlux)?

On top of that, we would like to be able to specify a control signal that depends on the value of the dependent variables. I.e. as temperature rises above 80°C, reduce the heat source. I believe this kind of thing can be done with optimal control / model-predictive-control but for our problem we are interested in a reinforcement learning solution. 

In theory we could run it as 

1. Solve ODEProblem from initial state S0, from `t=0` to `t=1` with parameter HEAT=H0, yielding final temperature T1
2. Decide new heat level H1 based on temperature T1
3. Solve ODEProblem from initial state S1, from `t=1` to `t=2` with parameter HEAT=H1, yielding final temperature T2
4. ...

But I was wondering if there was a nicer interface for this kind of setup. I'd really just like to be able to pass an arbitrary function / functor `f`  in the `ODEProblem` constructor.

Related: https://github.com/SciML/ModelingToolkitStandardLibrary.jl/issues/123

ChrisRackauckas:
@baggepinnen has some things to point to. You wouldn't want to just fit the interpolation since you have a lot of sparsity in time to exploit.

baggepinnen:
We don't have much that is open source unfortunately.

[...]

PR #2926, opened 2024-08-05
add an example to IO docs

baggepinnen:


cstjean:
Thank you for the example! I'll confess that I'm still confused. I came here looking for solutions to #2905 (i.e. general black box `u = some_arbitrary_f(current_state)` control). From my understanding, https://docs.sciml.ai/ModelingToolkit/dev/basics/Events/#Discrete-events-support (and https://docs.sciml.ai/DiffEqDocs/stable/features/callback_functions/#Example-2:-A-Control-Problem) looks like the best solution out there. I was trying my best to see if `generate_control_function` would be useful, and my rough impression from reading the docs it's that it's kinda meant for "continuous control" (classical control theory problems, I guess?) This isn't what I want, I think, but it's still interesting. However, while I see that you can pass an `u` to `f` above, and it returns "the dynamics of the system", how would you use that to achieve control?

cstjean:
> Like here
https://help.juliahub.com/juliasimcontrol/dev/examples/mtk_control/#State-feedback-using-ModelingToolkit

My background is reinforcement learning, I struggle to read through the documentation on control topics. I think I'm just out of my depth here, so I'll refrain from further comments. Thank you for the link.

> What exactly is it that you want?

In plain Julia, it'd be something like:

```julia
control_setpoint(x) = maximum(u->Q_value(x, u), [10,15,20,25])  # choose the setpoint which has highest Q value

integral = x = 0.0
for t in 0:10
	u = control_setpoint(x)   # assumed to be constant for that timestep
	x = g(x, u, t)       # arbitrary non-linear state dynamic
	integral += some_f(x)     # Riemann sum
end

# learn a Q_value function to maximize `integral`
```
I've been looking for how to express the `u = control_setpoint(x)` line in ModelingToolkit. Discrete events look like a valid way to do it.

I don't want to derail this further, thank you again for providing an example in this PR.

baggepinnen:
Your example is completely in discrete time, including the dynamics, while `generate_control_function` is for continuous time systems. 

Hybrid continuous-discrete systems are not yet fully supported in MTK (as described [here](https://discourse.julialang.org/t/how-to-implement-similar-functionality-as-in-simulink-simscape-by-modelingtoolkitstandardlibrary-jl/106597/15?u=baggepinnen)), so you cannot implement a discrete-time controller in any convenient way yet. 

Furthermore, MTK will not give any guarantees of what `x` actually is, it can change between versions of MTK. You thus always have to implement feedback (`u = control_setpoint(x)`) in terms of variables you specify yourself rather than the state.

cstjean:
In my real example, the state is continuous and changes quickly (minutes), while my setpoints (my control signal `u`) ought to stay put for hours. If I use discrete events to change the setpoints, and I specify those setpoints as [`d_discontinuities`](https://docs.sciml.ai/DiffEqDocs/stable/basics/common_solver_opts/), won't it work?

ChrisRackauckas (approved):

ChrisRackauckas:
@thazhemadam how could the formatter fail after taking its suggestions? Different versions? 😅

PR #2929, opened 2024-08-06
fix: several MTKParameters fixes

AayushSabharwal:
These are mainly centered around enabling AD.

Do not merge without approval from @DhairyaLGandhi


ChrisRackauckas:
What's the state of this one?

ChrisRackauckas (approved):

PR #2931, opened 2024-08-06
update JumpSystem for auto-alg support

isaacsas:
This enables `JumpProblem(::JumpSystem, ::DiscreteProblem)`, i.e. eliminating the requirement to pass an aggregator, and using the new auto-alg in JumpProcesses.


isaacsas:
@ChrisRackauckas I think this is good to go. (And it would be great for finishing this off in Catalyst if there can be a MTK release with it once merged.) Thanks!

isaacsas:
Strike that, this needs a tweak to ensure it generates dependency graphs when appropriate.

isaacsas:
Ok, now this should be good to go.

PR #2944, opened 2024-08-09
Experimental variable filtering in dummy derivative

YingboMa:


baggepinnen:
Interesting, is this useful for end users?

YingboMa:
Nope. It just saves some computations.

PR #2946, opened 2024-08-09
support JumpProblems over ODEProblems

isaacsas:
To enable more automatic VariableRateJump handling.

~~@AayushSabharwal this is supposed to allow JumpProblems over ODEProblems, but symbolic indexing of solutions and plotting / plot legends are not working. I just copied how we did JumpProblems over DiscreteProblems, but that doesn't seem to work? Can you let me know how to update the dummy ODEProblem(::JumpSystem,...) here to get it working? Thanks!~~ 

TODO:
- ~~Fix dependency graph generation to ignore `VariableRateJump`s and add tests.~~
- ~~Get symbolic indexing working for the resulting solutions~~

Basic `plot(sol)` is currently broken but that is more of a SciMLBase / JumpProcesses issue from the ExtendedJumpArrays. Plotting works with explicit specification of indexes to plot.

Edit: Needs [JumpProcesses 438](https://github.com/SciML/JumpProcesses.jl/pull/438) and a JumpProcesses release.

### DIFF
--- a/src/systems/dependency_graphs.jl
+++ b/src/systems/dependency_graphs.jl
@@ -143,7 +143,6 @@ variable_dependencies(jumpsys)
 """
 function variable_dependencies(sys::AbstractSystem; variables = unknowns(sys),
         variablestoids = nothing, eqs = equations(sys))
-
     vtois = isnothing(variablestoids) ? Dict(v => i for (i, v) in enumerate(variables)) :
             variablestoids
 
--- a/src/systems/dependency_graphs.jl
+++ b/src/systems/dependency_graphs.jl
@@ -36,8 +36,8 @@ equation_dependencies(jumpsys)
 equation_dependencies(jumpsys, variables = parameters(jumpsys))
 ```
 """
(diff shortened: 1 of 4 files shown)


isaacsas:
Also, just to add I tried alternatively creating an ODEProblem via an ODESystem with the unknowns, parameters, defaults, and observed from the JumpSystem (but no equations), however, that also did not result in solutions that could be symbolically indexed or plots that had proper symbolic naming/indexing.

isaacsas:
@ChrisRackauckas @AayushSabharwal so this won't work cleanly until symbolic indexing / plotting for solutions that are represented as ExtendedJumpArrays is supported. I don't know what getting such support to work would entail, but it would be nice to have (and should I guess also be added for JumpProblems over SDEProblems where one gets ExtendedJumpArrays for `u`).

ChrisRackauckas:
What's the current state here?

isaacsas:
@ChrisRackauckas This is good I think.

ChrisRackauckas:
okay cool, that's what I thought.

Issue #2969, opened 2024-08-15
`ODEProblem` from `ODESystem` does not sub nested constants


aml5600:
**Describe the bug 🐞**

When building an `ODEProblem` from an `ODESystem` that contains nested systems, the function generation step does not fully substitute all nested constants in the nested systems.

**Expected behavior**

We can evaluate `prob.f` without error.

**Minimal Reproducible Example 👇**

Without MRE, we would only be able to help you to a limited extent, and attention to the issue would be limited. to know more about MRE refer to [wikipedia](https://en.wikipedia.org/wiki/Minimal_reproducible_example) and [stackoverflow](https://stackoverflow.com/help/minimal-reproducible-example).

```julia
using ModelingToolkit
using ModelingToolkit: t_nounits as t, D_nounits as D

@mtkmodel ModelA begin
    @constants begin
        a = 1.234
    end
    @variables begin
        x(t)
    end
    @equations begin
        D(x) ~ -a * x
    end
end

@mtkmodel ModelB begin
    @constants begin
        b = 5.678
    end
    @variables begin
        y(t)
    end
    @components begin
        modela = ModelA()
    end
    @equations begin
        D(y) ~ -b * y
    end
end

##
@named sys = ModelB()
sys = complete(sys)

u0 = [sys.y => 1.0, sys.modela.x => -1.0]
p = defaults(sys)

prob = ODEProblem(sys, u0, (0.0, 1.0), p)

# evaluate
u0_v, p_v, _ = ModelingToolkit.get_u0_p(sys, u0, p)
prob.f(u0_v, p_v, 0.0)
```

**Error & Stacktrace ⚠️**

```julia
ERROR: UndefVarError: `modela₊a` not defined
Stacktrace:
 [1] macro expansion
   @ ~/.julia/packages/SymbolicUtils/ZwKM7/src/code.jl:480 [inlined]
 [2] macro expansion
   @ ~/.julia/packages/RuntimeGeneratedFunctions/M9ZX8/src/RuntimeGeneratedFunctions.jl:163 [inlined]
 [3] macro expansion
   @ ./none:0 [inlined]
 [4] generated_callfunc
   @ ./none:0 [inlined]
 [5] (::RuntimeGeneratedFunctions.RuntimeGeneratedFunction{…})(::Vector{…}, ::SciMLBase.NullParameters, ::Float64)
[...]

PR #2970, opened 2024-08-15
sub constants in `generate_function`

aml5600:
In the case of nested systems, `@constants` in the nested systems were not being substituted. See [2969](https://github.com/SciML/ModelingToolkit.jl/issues/2969)

Similar to [2904](https://github.com/SciML/ModelingToolkit.jl/pull/2904), we simply call `subs_constants` on the equations before using them.

Included test fails before fix, passes after.

Thanks!

## Checklist


aml5600:
the same tests fail locally, at least locally

PR #2975, opened 2024-08-19
Support SymbolicUtils 1.7.0 and 1.7.1 in MTK v8

devmotion:
This PR adds support for SymbolicUtils 1.7.1 for MTK v8 in a backport to make the bugfix for the ConstructionBase/SymbolicUtils issue available for people stuck on MTK v8. Compatibility with SymbolicUtils was previously limited to <= 1.5.1 due to an incompatibility in 1.6.0 which was later reverted in 1.7.0: https://github.com/SciML/ModelingToolkit.jl/pull/2726

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "ModelingToolkit"
 uuid = "961ee093-0014-501f-94e3-6117800e7a78"
 authors = ["Yingbo Ma <mayingbo5@gmail.com>", "Chris Rackauckas <accounts@chrisrackauckas.com> and contributors"]
-version = "8.76.0"
+version = "8.77.0"
 
 [deps]
 AbstractTrees = "1520ce14-60c1-5f80-bbc7-55ef81b5835c"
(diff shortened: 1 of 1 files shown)


devmotion:
Is it clear why tests are failing? MTK v8.76.0 + SymbolicUtils v1.5.1 works fine for us, so maybe it's some other dependency that's causing tests to fail. v1.7.0 should be almost the same as v1.5.1, with the problematic changes reverted: https://github.com/JuliaSymbolics/SymbolicUtils.jl/compare/v1.5.1...v1.7.0

ChrisRackauckas:
It's not obvious to me.

devmotion:
I started a CI run on the latest v8 release in https://github.com/SciML/ModelingToolkit.jl/pull/2976, to check if the test errors show up there as well.

devmotion:
In #2976 (latest release of MTK v8 + SymbolicUtils 1.5.1) the same errors are thrown, so it seems the update of some other dependency might have broken something. I guess this means the PR is fine and it's OK to make another release of MTK v8 with the bugfix for ConstructionBase 1.5.7 since it will be strictly superior to the existing release?

ChrisRackauckas:
Yes I'd accept it. That's a pretty bad looking error on v8 though, I'd try restricting SymbolicUtils.jl or Symbolics.jl and see if you can get it to pass.

devmotion:
This is not needed anymore.

PR #3019, opened 2024-09-05
Respect positional arguments in `@mtkmodel`

baggepinnen:


baggepinnen:
@SebastianM-C see this

@ven-k do you think you could take ´over this one? I could not easily handle all corner cases

ven-k:
'Will take a look.

PR #3026, opened 2024-09-06
Make has_time_domain and get_time_domain more generic

YingboMa:


YingboMa:
The test failures are unrelated.

Issue #3012, opened 2024-09-04
repeated application of `structural_simplify` leads to error


hexaeder:
**Describe the bug 🐞**

Repeated aplication of `structural_simplify` can lead to errors. I've observed at least two:
- unbalanced system error and
- (far less important) out of bounds error on equations in a case where structural simplify reduced the system to 0 equations

**Expected behavior**

Either error because it should not be called on a `completed` system or be a no-op.

**Minimal Reproducible Example 👇**

```julia
using Pkg
pkg"activate --temp"
pkg"add ModelingToolkit, ModelingToolkitStandardLibrary"

using ModelingToolkit
using ModelingToolkitStandardLibrary.Electrical
using ModelingToolkitStandardLibrary.Blocks: Constant
using ModelingToolkit: t_nounits as t

R = 1.0
C = 1.0
V = 1.0
systems = @named begin
    resistor = Resistor(R = R)
    capacitor = Capacitor(C = C, v = 0.0)
    source = Voltage()
    constant = Constant(k = V)
    ground = Ground()
end

rc_eqs = [connect(constant.output, source.V)
          connect(source.p, resistor.p)
          connect(resistor.n, capacitor.p)
          connect(capacitor.n, source.n, ground.g)]

@named rc_model = ODESystem(rc_eqs, t; systems)
sys = structural_simplify(rc_model)
structural_simplify(sys)
```

**Error & Stacktrace ⚠️**

```julia
julia> structural_simplify(sys)
ERROR: ExtraVariablesSystemException: The system is unbalanced. There are 2 highest order derivative variables and 1 equations.
More variables than equations, here are the potential extra variable(s):
 capacitor₊v(t)
 capacitor₊i(t)
Note that the process of determining extra variables is a best-effort heuristic. The true extra variables are dependent on the model and may not be in this list.
Stacktrace:
  [1] error_reporting(state::TearingState{…}, bad_idxs::Vector{…}, n_highest_vars::Int64, iseqs::Bool, orig_inputs::Set{…})
    @ ModelingToolkit.StructuralTransformations ~/.julia/packages/ModelingToolkit/yfT8s/src/structural_transformation/utils.jl:50
  [2] check_consistency(state::TearingState{ODESystem}, orig_inputs::Set{Any})
    @ ModelingToolkit.StructuralTransformations ~/.julia/packages/ModelingToolkit/yfT8s/src/structural_transformation/utils.jl:85
[...]

PR #3027, opened 2024-09-08
Throw an error if structural simplification is applied twice

ChrisRackauckas:
Fixes https://github.com/SciML/ModelingToolkit.jl/issues/3012


ChrisRackauckas:
@YingboMa that failing test... doesn't make much sense?

ChrisRackauckas:
Since downstreams all seem fine other than test bugs, this is good to go. I'm sure someone will complain about this one, but this is pretty much required in order to make sure no weird hidden bugs show up. Calling a system complete means you have completed the system: if you try to change it after you say you will never change it (i.e. that it's complete), bad things will happen. We have to assume users don't lie about when their system is completed or else we could never optimize things like schedules. So... this error is pretty necessary, and if you were violating it there's issues you could have hit before which is why it was made explicit.

ChrisRackauckas:
I'll follow up with SBMLToolkit which is over zealous at calling `complete` in the system construction which is simply invalid: a file reader cannot read the future on what a user wants to do.

baggepinnen:
>  I'm sure someone will complain about this one

yes I will, this was quite breaking. Calling `complete` is required to access variables with the correct name space, and one has to access variables before `structural_simplify` for things like linearization and generation of dynamics functions with inputs. Why is `iscomplete` used for this check and not a check that actually determines if `structural_simplify` has been called?

ChrisRackauckas:
`iscomplete` is a flag for whether no more changes will happen to a system. You're literally saying that you are breaking that promise, and there are many things that rely on that promise. Of course you want an error when that happens!

src/systems/systems.jl
```diff
@@ -21,6 +29,7 @@ function structural_simplify(
         sys::AbstractSystem, io = nothing; simplify = false, split = true,
         allow_symbolic = false, allow_parameter = true, conservative = false, fully_determined = true,
         kwargs...)
+    iscomplete(sys) && throw(RepeatedStructuralSimplificationError())
```
YingboMa, line 32: This is the wrong check and breaking. You should just check the existence of schedule.

PR #3034, opened 2024-09-09
Add unit aware diff2term

YingboMa:
Fix https://github.com/SciML/ModelingToolkit.jl/issues/2959

### DIFF
--- a/test/dq_units.jl
+++ b/test/dq_units.jl
@@ -270,5 +270,6 @@ using DynamicQuantities
 end
 
 @mtkbuild pend = UnitsExample()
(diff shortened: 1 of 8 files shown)


test/dq_units.jl
```diff
@@ -246,3 +246,30 @@ let
     @test MT.get_unit(x_vec) == u"1"
     @test MT.get_unit(x_mat) == u"1"
 end
+
+module UnitTD
```
YingboMa, line 250:
```suggestion
module UnitTD
using Test
```

PR #3038, opened 2024-09-09
Fix test setup

ChrisRackauckas:
https://github.com/SciML/ModelingToolkit.jl/pull/3035 accidentally reverted fixes to the test setup and checks that the error is thrown


test/structural_transformation/errors.jl
```diff
@@ -0,0 +1,19 @@
...
+        x(t) = 0.0 # dependent variables
+    end
+    @equations begin
+        D(x) ~ (1 - x) / τ
+    end
+end
+
+@named rc_model = FOL()
+sys = structural_simplify(rc_model)
+@test_throws ModelingToolkit.RepeatedStructuralSimplificationError structural_simplify(sys)
```
YingboMa, line 19: This is already added as a part of components.jl https://github.com/SciML/ModelingToolkit.jl/blob/690a659bed6956b9e37cb72f6036a68ba7816444/test/components.jl#L47

Issue #2178, opened 2023-06-01
Incidence matrix structural_simplify


ArnoStrouwen:
Would it be possible to add some extra information to the docstring of structural_simplify to explain how the incidence matrix should be interpreted: on the left original dependence of equations on variables, on the right simplified...

ChrisRackauckas:
@YingboMa can you take a second to add this?

YingboMa:
The incidence matrix is actually not useful for humans. I am just showing it because it looks cool. We can just disable it if it's confusing.

ChrisRackauckas:
Disable it then from the default printing and surface a function in the docs for showing it.

PR #3039, opened 2024-09-10
Remove incidence matrix from system show

ChrisRackauckas:
Fixes https://github.com/SciML/ModelingToolkit.jl/issues/2178

>
The incidence matrix is actually not useful for humans. I am just showing it because it looks cool. We can just disable it if it's confusing.

I'm not even sure it's worth documenting, but who it's useful for is for the compiler writers since if structural simplification works, then you're good. Maybe it's useful for diagnosing singularities, but that should get a debugging page because someone would need to get introduced to what it is for it to be useful. So I keep it around as a `ModelingToolkit.incidence_matrix` function that can be easily queried, but it's removed from the default show.

### DIFF
--- a/src/systems/abstractsystem.jl
+++ b/src/systems/abstractsystem.jl
@@ -876,8 +876,7 @@ end
 """
 $(TYPEDSIGNATURES)
 
-Mark a system as completed. A completed system is assumed to be a 
-If a system is complete, the system will no longer
+Mark a system as completed. If a system is complete, the system will no longer
 namespace its subsystems or variables, i.e. `isequal(complete(sys).v.i, v.i)`.
 """
 function complete(sys::AbstractSystem; split = true)
--- a/src/systems/abstractsystem.jl
+++ b/src/systems/abstractsystem.jl
@@ -876,7 +876,8 @@ end
 """
 $(TYPEDSIGNATURES)
 
-Mark a system as completed. If a system is complete, the system will no longer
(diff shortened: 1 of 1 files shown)


src/systems/abstractsystem.jl
```diff
@@ -876,7 +876,8 @@ end
 """
 $(TYPEDSIGNATURES)
 
-Mark a system as completed. If a system is complete, the system will no longer
+Mark a system as completed. A completed system is assumed to be a
```
YingboMa, line 879: Is this changed by mistake?

src/systems/abstractsystem.jl
```diff
@@ -876,7 +876,8 @@ end
 """
 $(TYPEDSIGNATURES)
 
-Mark a system as completed. If a system is complete, the system will no longer
+Mark a system as completed. A completed system is assumed to be a 
+If a system is complete, the system will no longer
```
ChrisRackauckas, line 880:
```suggestion
Mark a system as completed. If a system is complete, the system will no longer
```

src/systems/abstractsystem.jl
ChrisRackauckas: fixed.

ChrisRackauckas:
I assume since you caught that you reviewed this and didn't have an issue with the removal, so I'll merge.

YingboMa:
Yeah, I am fine with this change.

PR #3041, opened 2024-09-10
Tolerate empty equations and uneven matching

YingboMa:
Alias elimination might introduce empty equations. We need to handle that properly.


ChrisRackauckas:
The indexing one looks kind of funny though @AayushSabharwal

AayushSabharwal:
I think the ODESystem in that PR fails to initialize with the superset initialization PR, and so the solution doesn't have as many elements as expected.

PR #3042, opened 2024-09-11
feat: add adjoint for `remake_buffer`

AayushSabharwal:
## Checklist

### DIFF
--- a/ext/MTKChainRulesCoreExt.jl
+++ b/ext/MTKChainRulesCoreExt.jl
@@ -2,14 +2,102 @@ module MTKChainRulesCoreExt
 
 import ModelingToolkit as MTK
 import ChainRulesCore
-import ChainRulesCore: NoTangent
+import ChainRulesCore: Tangent, ZeroTangent, NoTangent, zero_tangent, unthunk
 
 function ChainRulesCore.rrule(::Type{MTK.MTKParameters}, tunables, args...)
     function mtp_pullback(dt)
+        dt = unthunk(dt)
         (NoTangent(), dt.tunable[1:length(tunables)],
             ntuple(_ -> NoTangent(), length(args))...)
     end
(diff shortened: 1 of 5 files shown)


ext/MTKChainRulesCoreExt.jl
```diff
@@ -2,14 +2,102 @@ module MTKChainRulesCoreExt
...
+
+    function remake_buffer_pullback(buf′)
+        buf′ = unthunk(buf′)
+        f′ = NoTangent()
+        indp′ = NoTangent()
+
+        tunable = selected_tangents(buf′.tunable, tunable_idxs)
+        discrete = selected_tangents(buf′.discrete, disc_idxs)
+        constant = selected_tangents(buf′.constant, const_idxs)
+        nonnumeric = selected_tangents(buf′.nonnumeric, nn_idxs)
```
ChrisRackauckas, line 94: is this type stable?

ChrisRackauckas (approved):

ext/MTKChainRulesCoreExt.jl
AayushSabharwal: I haven't checked, though the type-stability of the adjoint will largely depend on the type stability of `idxs` and `vals`

AayushSabharwal: `selected_tangents` is type-stable, the pullback is almost type stable with a couple changes if `idxs` and `vals` are type-stable. The only source of instability is the one `selected_tangents` method which uses `similar(tangents, Union{T, NoTangent})` (which in fairness is a concretely typed union) and `vals′`

AayushSabharwal: However, if we restrict to the case of `idxs` only referring to values in the tunables, it's type-stable.

PR #3079, opened 2024-09-26
Relax type constraints to allow callable parameters in pdeps

SebastianM-C:
## Checklist

## Additional context

Currently parameter dependencies assume that we only have `BasicSymbolic`. I tried to relax that constraint a bit to allow callable parameters to be stored too. If one has both callable parameters and normal ones in the parameter dependencies, I think that there might be a performance issue, but otherwise it should not affect nortmal usage.

@AayushSabharwal is this the right way of handling this issue? I think that for normal parametrs we store things in the non-numeric portion and that's splitted by types (?), so that might be better for the heterogenuous case.

### DIFF
--- a/src/systems/index_cache.jl
+++ b/src/systems/index_cache.jl
@@ -39,7 +39,7 @@ const UnknownIndexMap = Dict{
 const TunableIndexMap = Dict{BasicSymbolic,
     Union{Int, UnitRange{Int}, Base.ReshapedArray{Int, N, UnitRange{Int}} where {N}}}
 
-struct IndexCache
+struct IndexCache{D}
     unknown_idx::UnknownIndexMap
     # sym => (bufferidx, idx_in_buffer)
     discrete_idx::Dict{BasicSymbolic, DiscreteIndex}
@@ -49,7 +49,7 @@ struct IndexCache
     constant_idx::ParamIndexMap
     nonnumeric_idx::NonnumericMap
     observed_syms::Set{BasicSymbolic}
-    dependent_pars::Set{BasicSymbolic}
+    dependent_pars::Set{D}
     discrete_buffer_sizes::Vector{Vector{BufferTemplate}}
     tunable_buffer_size::BufferTemplate
     constant_buffer_sizes::Vector{BufferTemplate}
@@ -275,7 +275,15 @@ function IndexCache(sys::AbstractSystem)
         end
     end
 
-    dependent_pars = Set{BasicSymbolic}()
+    pdeps = parameter_dependencies(sys)
+
+    D = if isempty(pdeps)
+        BasicSymbolic
+    else
+        mapreduce(typeof, promote_type, getproperty.(pdeps, :lhs))
+    end
+    dependent_pars = Set{D}()
+
     for eq in parameter_dependencies(sys)
         sym = eq.lhs
         ttsym = default_toterm(sym)
@@ -289,7 +297,7 @@ function IndexCache(sys::AbstractSystem)
         end
     end
 
-    return IndexCache(
+    return IndexCache{D}(
         unk_idxs,
         disc_idxs,
         callback_to_clocks,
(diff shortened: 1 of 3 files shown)


src/systems/index_cache.jl
```diff
@@ -39,7 +39,7 @@ const UnknownIndexMap = Dict{
 const TunableIndexMap = Dict{BasicSymbolic,
     Union{Int, UnitRange{Int}, Base.ReshapedArray{Int, N, UnitRange{Int}} where {N}}}
 
-struct IndexCache
+struct IndexCache{D}
```
AayushSabharwal, line 42: I would prefer if this struct were not generically typed. `dependent_pars` can be a `Set{Union{BasicSymbolic, CallWithMetadata}}`. This shouldn't have much of a performance impact since it's a small union and the field is only used in a few places for `in` checks.

SebastianM-C: But `CallWithMetadata` will have additional type parameters for the function args & return type, so if we do that, we might get inference failures.

AayushSabharwal: Currently, no. That field isn't actually used anywhere on `master`, and the only place it is used is one function call in the unmerged parameter initialization. At runtime, this will never be accessed.

SebastianM-C: Ah, good to know.

SebastianM-C:
I'm having some trouble with the test. The codegen doesn't seem to do the replacement of the parameter in `prob.f`. In the function it does `ˍ₋arg2[1]`, but in the pdep call it's still `CallableFoo(p)`.

```
julia> prob.f.f.f_iip
RuntimeGeneratedFunction(#=in ModelingToolkit=#, #=using ModelingToolkit=#, :((ˍ₋out, ˍ₋arg1, ˍ₋arg2, t)->begin    
          #= C:\Users\sebastian\.julia\packages\SymbolicUtils\ij6YM\src\code.jl:385 =#
          #= C:\Users\sebastian\.julia\packages\SymbolicUtils\ij6YM\src\code.jl:386 =#
          #= C:\Users\sebastian\.julia\packages\SymbolicUtils\ij6YM\src\code.jl:387 =#
          begin
              begin
                  begin
                      i = CallableFoo(p)
                      begin
                          #= C:\Users\sebastian\.julia\packages\Symbolics\b4I7P\src\build_function.jl:546 =#       
                          #= C:\Users\sebastian\.julia\packages\SymbolicUtils\ij6YM\src\code.jl:434 =# @inbounds begin
                                  #= C:\Users\sebastian\.julia\packages\SymbolicUtils\ij6YM\src\code.jl:430 =#     
                                  ˍ₋out[1] = (+)(ˍ₋arg2[1], i(t))
                                  #= C:\Users\sebastian\.julia\packages\SymbolicUtils\ij6YM\src\code.jl:432 =#     
                                  nothing
                              end
                      end
                  end
              end
          end
      end))
```

src/systems/index_cache.jl
SebastianM-C: Fixed in [0871fd4](https://github.com/SciML/ModelingToolkit.jl/pull/3079/commits/0871fd42d4f0766ead9a614f2bdf29a93d468d90). What is the advantage of avoiding the type parameters for this struct?

AayushSabharwal:
> but in the pdep call it's still CallableFoo(p).

You probably need to `@register_symbolic CallableFoo(p)`

SebastianM-C:
@ChrisRackauckas test failures seem the same as master. Does this look okay?

ChrisRackauckas:
I don't like that they are just relaxed, but if @AayushSabharwal thinks it's required then let's go with it.

SebastianM-C:
What would be the alternative? I understand that we don't want to introduce type parameters (to avoid increasing inference time?).

AayushSabharwal:
> I don't like that they are just relaxed

What's not to like? 😅 Looks good to me, and a fairly simple change allows new features.

ChrisRackauckas:
```
    lhss = BasicSymbolic[]
    lhss = []
```

lost a validation step, but if it's required, sure.

PR #3081, opened 2024-09-27
feat: improve error message when system contains array equations

### LINKED ISSUE #2597: Symbolic-array system bug in counting equations and variables
MWE
```julia
@mtkmodel Foo begin
    @variables begin
        x(t)[1:2]
    end
    @equations begin
        D.(x) ~ .-x
    end
end
@named foo = Foo()
foo = complete(foo)
prob = ODEProblem(foo, Dict(foo.x => [1.0, 1]), (0.0, 1.0))
sol = solve(prob, Rodas4())
```
```
julia> prob = ODEProblem(foo, Dict(foo.x => [1.0, 1]), (0.0, 1.0))
ERROR: ArgumentError: Equations (1), unknowns (2), and initial conditions (2) are of different lengths. To allow a different number of equations than unknowns use kwarg check_length=false.
```

AayushSabharwal:
Close #2597 

## Checklist

### DIFF
--- a/src/systems/abstractsystem.jl
+++ b/src/systems/abstractsystem.jl
@@ -2878,6 +2878,15 @@ function Base.eltype(::Type{<:TreeIterator{ModelingToolkit.AbstractSystem}})
     ModelingToolkit.AbstractSystem
 end
 
+function check_array_equations_unknowns(eqs, dvs)
+    if any(eq -> Symbolics.isarraysymbolic(eq.lhs), eqs)
+        throw(ArgumentError("The system has array equations. Call `structural_simplify` to handle such equations or scalarize them manually."))
+    end
+    if any(x -> Symbolics.isarraysymbolic(x), dvs)
+        throw(ArgumentError("The system has array unknowns. Call `structural_simplify` to handle this or scalarize them manually."))
+    end
+end
+
 function check_eqs_u0(eqs, dvs, u0; check_length = true, kwargs...)
     if u0 !== nothing
         if check_length
--- a/src/systems/diffeqs/abstractodesystem.jl
+++ b/src/systems/diffeqs/abstractodesystem.jl
@@ -815,6 +815,7 @@ function process_DEProblem(constructor, sys::AbstractODESystem, u0map, parammap;
     ps = parameters(sys)
     iv = get_iv(sys)
 
+    check_array_equations_unknowns(eqs, dvs)
     # TODO: Pass already computed information to varmap_to_vars call
     # in process_u0? That would just be a small optimization
     varmap = u0map === nothing || isempty(u0map) || eltype(u0map) <: Number ?
--- a/test/odesystem.jl
+++ b/test/odesystem.jl
@@ -1404,3 +1404,34 @@ end
     @test !ModelingToolkit.is_dde(sys)
     @test is_markovian(sys)
 end
+
+@testset "Issue #2597" begin
+    @variables x(t)[1:2]=ones(2) y(t)=1.0
+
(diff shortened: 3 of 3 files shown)


ChrisRackauckas:
Why would this be required? The codegen doesn't require it if it's a true ODE.

AayushSabharwal:
It doesn't really work very reliably.

```julia
@variables x(t)[1:2]
@named sys = ODESystem(D(x) ~ x, t)
sys = complete(sys)
prob = ODEProblem(sys, ...)
```
This errors in `calculate_massmatrix` because that can't handle array equations and needs them scalarized

```julia
@named sys = ODESystem(D(x) ~ x, t, [x...], [])
```

If I comment out `check_eqs_u0` this errors in `generate_function` because `check_lhs` finds that the variable inside the derivative `D(x)` isn't in the list of unknowns.

I _could_ make these cases work, but all of the fixes boil down to scalarizing the equations and/or unknowns.

AayushSabharwal:
And if there are multiple array unknowns or a mix of array/scalars then codegen also won't work

ChrisRackauckas:
`calculate_massmatrix` needs to semantically scalarize and show it's constant. 

> And if there are multiple array unknowns or a mix of array/scalars then codegen also won't work without scalarizing the equations.

Is that really the case? Symbolics.jl handled this awhile back.

ChrisRackauckas:
I see no issue with a temporary error message that told people as a temporary workaround this is required, though I think we should look at some of the codegen here since Symbolics should already handle these cases.

AayushSabharwal:
The problem with the codegen is that this system:
```julia
julia> @named sys = ODESystem([D(x) ~ x, D(y) ~ t], t, [x..., y])
```
Generates this code:
```julia
RuntimeGeneratedFunction(#=in ModelingToolkit=#, #=using ModelingToolkit=#, :((ˍ₋out, ˍ₋arg1, t)->begin
          #= /Users/aayush/.julia/packages/SymbolicUtils/ij6YM/src/code.jl:385 =#
          #= /Users/aayush/.julia/packages/SymbolicUtils/ij6YM/src/code.jl:386 =#
          #= /Users/aayush/.julia/packages/SymbolicUtils/ij6YM/src/code.jl:387 =#
          begin
              begin
                  var"x(t)" = view(ˍ₋arg1, 1:2)
                  begin
                      begin
                          #= /Users/aayush/.julia/packages/Symbolics/XnDVB/src/build_function.jl:546 =#
                          #= /Users/aayush/.julia/packages/SymbolicUtils/ij6YM/src/code.jl:434 =# @inbounds begin
                                  #= /Users/aayush/.julia/packages/SymbolicUtils/ij6YM/src/code.jl:430 =#
                                  ˍ₋out[1] = var"x(t)"
                                  ˍ₋out[2] = t
                                  #= /Users/aayush/.julia/packages/SymbolicUtils/ij6YM/src/code.jl:432 =#
                                  nothing
                              end
                      end
                  end
              end
          end
      end))
```

Which again points to the fact that the system needs to be scalarized

AayushSabharwal:
Also without `structural_simplify` it's up to the user to ensure the equations and unknowns are in the correct order.

ChrisRackauckas:
Okay we should fix that, it's fixable. For now, mention that a temporary workaround is to scalarize or structural simplify.

ChrisRackauckas (approved):

ChrisRackauckas:
@AayushSabharwal some test greening https://github.com/SciML/ModelingToolkit.jl/actions/runs/11185499006/job/31098502328#step:6:2267 ?

PR #3084, opened 2024-09-29
Rewrite of SDE tutorial

ArnoStrouwen:
Rewrite of MTK documentation continues with SDE modelling.

What this tutorial is still lacking, is an exact mapping of the MTK `@brownian` concept to the LaTeX equation.

I'm not at all an expert in SDEs, but `@brownian` does not seem to map directly to what we call Brownian motion but more something on the level of dB/dt.

### DIFF
--- a/docs/src/tutorials/stochastic_diffeq.md
+++ b/docs/src/tutorials/stochastic_diffeq.md
@@ -1,38 +1,79 @@
 # Modeling with Stochasticity
 
-All models with `ODESystem` are deterministic. `SDESystem` adds another element
-to the model: randomness. This is a
+All previous differential equations tutorials deal with deterministic `ODESystem`s.
+In this tutorial, we add randomness.
+In particular, we show how to represent a
 [stochastic differential equation](https://en.wikipedia.org/wiki/Stochastic_differential_equation)
-which has a deterministic (drift) component and a stochastic (diffusion)
-component. Let's take the Lorenz equation from the first tutorial and extend
-it to have multiplicative noise by creating `@brownian` variables in the equations.
+as a `SDESystem`.
+
+!!! note
+    
+    The high level `@mtkmodel` macro used in the
+    [getting started tutorial](@ref getting_started)
+    is not yet compatible with `SDESystem`.
+    We thus have to use a lower level interface to define stochastic differential equations.
+    For an introduction to this interface, read the
+    [programmatically generating ODESystems tutorial](@ref programmatically).
+
+Let's take the Lorenz equation and add noise to each of the states.
+To show the flexibility of ModelingToolkit,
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
> I'm not at all an expert in SDEs, but @brownian does not seem to map directly to what we call Brownian motion but more something on the level of dB/dt.

They are Brownian increments, yes, so variables which when integrated are Brownian motions. Not referred to as dB/dt, instead just dB since it's only defined in the integral sense.

docs/src/tutorials/stochastic_diffeq.md
```diff
@@ -1,38 +1,79 @@
...
+plot(sol, idxs = [(1, 2, 3)])
+```
+
+The noise present in all 3 equations is correlated, as can be seen on the below figure.
+If you want uncorrelated noise for each equation,
+multiple `@brownian` variables have to be declared.
+
+```@example SDE
+@brownian Bx By Bz
+```
```
ChrisRackauckas, line 71: Maybe expand this out and show the whole multiplicative case?

docs/src/tutorials/stochastic_diffeq.md
```diff
@@ -1,38 +1,79 @@
...
-    σ => 10.0,
-    β => 26.0,
-    ρ => 2.33
-]
+We continue by solving and plotting the SDE.
 
-prob = SDEProblem(de, u0map, (0.0, 100.0), parammap)
+```@example SDE
+prob = SDEProblem(de, [], (0.0, 100.0), [])
 sol = solve(prob, LambaEulerHeun())
```
ChrisRackauckas, line 61: This probably isn't a good method choice for this case.

ArnoStrouwen:
> They are Brownian increments, yes, so variables which when integrated are Brownian motions. Not referred to as dB/dt, instead just dB since it's only defined in the integral sense.

I understand that this is only mathematically rigorously defined in the integral sense.
However, in the code, we do have: `D = Differential(t); D(x)`, which in the Symbolics docs is defined as differential w.r.t time.

I need to figure out a way to succinctly express that even though we work in the integral sense, the `dt` present in the deterministic terms is omitted (when no `@brownian` is present), while the `dW` is implied from `@brownian`

ChrisRackauckas:
The formulation here is effectively as a Langevin equation https://en.wikipedia.org/wiki/Langevin_equation

ArnoStrouwen:
Can be merged when tests pass.

ArnoStrouwen:
Errors seem to be in another tutorial:
https://github.com/SciML/ModelingToolkit.jl/actions/runs/11306893338/job/31448000464?pr=3084#step:6:2450

PR #3085, opened 2024-10-01
Start rewriting programmatically generating systems docs

ArnoStrouwen:
This slightly expands the first few subsections of the docs on how to programmatically create mtk systems.

This page also contains a section on "Understanding the Difference Between the Julia Variable and the Symbolic Variable".
I think that most of this section, except for the explanation of the `@named` macro should be moved to the Symbolics.jl docs, where this information is missing.

### DIFF
--- a/docs/src/tutorials/programmatically_generating.md
+++ b/docs/src/tutorials/programmatically_generating.md
@@ -1,42 +1,44 @@
 # [Programmatically Generating and Scripting ODESystems](@id programmatically)
 
-In the following tutorial we will discuss how to programmatically generate `ODESystem`s.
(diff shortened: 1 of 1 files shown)


ArnoStrouwen:
This example should also just be merged with this tutorial IMO:
https://docs.sciml.ai/ModelingToolkit/stable/examples/parsing/

ChrisRackauckas:
I don't see why. Working on Expr is quite an odd case and we really shouldn't promote it much since there's lots of downsides to it.

ArnoStrouwen:
Then I would just move it to Symbolics. ModelingToolkit docs are far too bloated.

ChrisRackauckas:
Sure, that's worth doing.

PR #3090, opened 2024-10-03
fix: promote `resid_prototype` using tunables

AayushSabharwal:
## Checklist

### DIFF
--- a/src/systems/nonlinear/nonlinearsystem.jl
+++ b/src/systems/nonlinear/nonlinearsystem.jl
@@ -295,7 +295,7 @@ function SciMLBase.NonlinearFunction(sys::NonlinearSystem, args...; kwargs...)
 end
 
 function SciMLBase.NonlinearFunction{iip}(sys::NonlinearSystem, dvs = unknowns(sys),
-        ps = parameters(sys), u0 = nothing;
+        ps = parameters(sys), u0 = nothing, p = nothing;
         version = nothing,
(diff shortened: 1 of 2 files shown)


ChrisRackauckas (approved):

src/systems/nonlinear/nonlinearsystem.jl
```diff
@@ -295,7 +295,7 @@ function SciMLBase.NonlinearFunction(sys::NonlinearSystem, args...; kwargs...)
 end
 
 function SciMLBase.NonlinearFunction{iip}(sys::NonlinearSystem, dvs = unknowns(sys),
-        ps = parameters(sys), u0 = nothing;
+        ps = parameters(sys), u0 = nothing, p = nothing;
```
ChrisRackauckas, line 298: Why ps and p here?

AayushSabharwal: ps is the list of parameter symbols. p is the parameter object, which is required for type promotion in resid_prototype.

ChrisRackauckas: How is p already generated by here?

AayushSabharwal: It's generated in `process_NonlinearProblem`, which calls this and is responsible for providing `u0` and now `p` as well.

PR #3091, opened 2024-10-03
flesh  out next steps after getting started docs

ArnoStrouwen:


docs/src/tutorials/ode_modeling.md
```diff
@@ -366,24 +366,55 @@ exploiting the structural information. For more information, see the
...
+
+If you want to further investigate the properties of your ODE model:
+
+  - [Check if the parameters of your model are identifiable](@ref identifiability)
+  - [Check if your system exhibits any bifurcation](@ref bifurcation_diagrams)
+  - [Perform a perturbation analysis to see your system change as parameter values change](@ref perturb_diff)
+  - [Linearize your system](@ref linearization) and [turn it into an input-output system](@ref inputoutput) (useful for feedback control)
+  - [Use your MTK ODE model in an `OptimizationProblem`](@ref remake)
+
+Still want to know more about MTK?
```
ChrisRackauckas, line 415: This is too comprehensive and no longer is a simplified way into MTK. Things which are non-essential should be cut from this.

ArnoStrouwen: Sure, some of the more low level ones, like tearing, could be removed.
However, some of the higher level ones should then link to the tearing one to explain how MTK is able to simulate only 1 ODE even though you wrote down 100 equations.

Also, everything removed from this list is a prime target for being consolidated into another doc page.
The most obvious example is modelingtoolkitize, we don't need both a Rober and a Brusselator example, particularly not spread over two pages.

ChrisRackauckas: Yeah, all good points.

PR #3098, opened 2024-10-07
fix first ODE perturbation example

### LINKED ISSUE #1994: Perturbation theory example exposes issue in `structural_simplify`?
Noted in https://github.com/SciML/ModelingToolkit.jl/pull/1993#issuecomment-1345576052. Would be good to get that fixed and doctested.

karlwessel:
## Checklist

- [x] Appropriate tests were added (see very last bullet point below)
  
## Additional context
This PR makes the first part from the Perturbation theory for ODEs example work. This fixes #1994.
The following things needed to be changed:
- `collect powers` included zero order terms in each order term,
- link to algebraic perturbation example from `Symbolics.jl` was wrong,
- indexing in definition and usage of `def_taylor` did not match,
- use `enumerate` instead of `eachindex` for symbolic arrays to get a linear index instead of a CartesianIndex (`eachindex(IndexLinear(), ...)` would also work).

The following things were not changed but would probably improve the example:
- Use `Symbolics.coeff` in `collect_powers`, but `coeff` currently can't be used to get the zeroth order term correctly (see second point in https://github.com/JuliaSymbolics/Symbolics.jl/issues/910).
- Start indexing of `y` and its derivative at zero, to match the order they represent, but non-one-based indexing isn't well supported yet (see #2670).
- It would probably be good to use the `@example ...` feature from `Documenter.jl` to add the output of commands to the documentation. This way it would be easier to detect when some change breaks the example.

### DIFF
--- a/docs/src/examples/perturbation.md
+++ b/docs/src/examples/perturbation.md
@@ -4,8 +4,8 @@
 
 In the previous tutorial, [Mixed Symbolic-Numeric Perturbation Theory](https://symbolics.juliasymbolics.org/stable/examples/perturbation/), we discussed how to solve algebraic equations using **Symbolics.jl**. Here, our goal is to extend the method to differential equations. First, we import the following helper functions that were introduced in [Mixed Symbolic/Numerical Methods for Perturbation Theory - Algebraic Equations](https://symbolics.juliasymbolics.org/stable/examples/perturbation/):
 
-```julia
-using Symbolics, SymbolicUtils
+```@example perturbation
+using Symbolics
 
 def_taylor(x, ps) = sum([a * x^(i - 1) for (i, a) in enumerate(ps)])
 
@@ -17,10 +17,10 @@ function collect_powers(eq, x, ns; max_power = 100)
         powers = Dict(x^j => (i == j ? 1 : 0) for j in 1:last(ns))
         e = substitute(eq, powers)
 
-		# manually remove zeroth order from higher orders
-		if 0 in ns && i != 0
-			e = e - eqs[1]
-		end
+        # manually remove zeroth order from higher orders
+        if 0 in ns && i != 0
+            e = e - eqs[1]
+        end
 
         push!(eqs, e)
     end
@@ -50,7 +50,7 @@ As the first ODE example, we have chosen a simple and well-behaved problem, whic
 
 with the initial conditions $x(0) = 0$, and $\dot{x}(0) = 1$. Note that for $\epsilon = 0$, this equation transforms back to the standard one. Let's start with defining the variables
 
-```julia
+```@example perturbation
 using ModelingToolkit: t_nounits as t, D_nounits as D
 order = 2
 n = order + 1
@@ -59,69 +59,69 @@ n = order + 1
 
 Next, we define $x$.
 
-```julia
+```@example perturbation
 x = def_taylor(ϵ, y)
 ```
 
 We need the second derivative of `x`. It may seem that we can do this using `Differential(t)`; however, this operation needs to wait for a few steps because we need to manipulate the differentials as separate variables. Instead, we define dummy variables `∂∂y` as the placeholder for the second derivatives and define
 
-```julia
+```@example perturbation
 ∂∂x = def_taylor(ϵ, ∂∂y)
 ```
 
 as the second derivative of `x`. After rearrangement, our governing equation is $\ddot{x}(t)(1 + \epsilon x(t))^{-2} + 1 = 0$, or
 
-```julia
+```@example perturbation
 eq = ∂∂x * (1 + ϵ * x)^2 + 1
 ```
 
 The next two steps are the same as the ones for algebraic equations (note that we pass `1:n` to `collect_powers` because the zeroth order term is needed here)
 
-```julia
+```@example perturbation
 eqs = collect_powers(eq, ϵ, 0:order)
 ```
 
 and,
 
-```julia
+```@example perturbation
 vals = solve_coef(eqs, ∂∂y)
 ```
 
 Our system of ODEs is forming. Now is the time to convert `∂∂`s to the correct **Symbolics.jl** form by substitution:
 
-```julia
+```@example perturbation
 subs = Dict(∂∂y[i] => D(D(y[i])) for i in eachindex(y))
 eqs = [substitute(first(v), subs) ~ substitute(last(v), subs) for v in vals]
 ```
 
 We are nearly there! From this point on, the rest is standard ODE solving procedures. Potentially, we can use a symbolic ODE solver to find a closed form solution to this problem. However, **Symbolics.jl** currently does not support this functionality. Instead, we solve the problem numerically. We form an `ODESystem`, lower the order (convert second derivatives to first), generate an `ODEProblem` (after passing the correct initial conditions), and, finally, solve it.
 
-```julia
+```@example perturbation
 using ModelingToolkit, DifferentialEquations
 
 @mtkbuild sys = ODESystem(eqs, t)
 unknowns(sys)
 ```
 
-```julia
+```@example perturbation
 # the initial conditions
 # everything is zero except the initial velocity
 u0 = zeros(2order + 2)
 u0[2] = 1.0   # yˍt₁
 
 prob = ODEProblem(sys, u0, (0, 3.0))
-sol = solve(prob; dtmax = 0.01)
+sol = solve(prob; dtmax = 0.01);
 ```
 
 Finally, we calculate the solution to the problem as a function of `ϵ` by substituting the solution to the ODE system back into the defining equation for `x`. Note that `𝜀` is a number, compared to `ϵ`, which is a symbolic variable.
 
-```julia
+```@example perturbation
 X = 𝜀 -> sum([𝜀^(i - 1) * sol[yi] for (i, yi) in enumerate(y)])
 ```
 
 Using `X`, we can plot the trajectory for a range of $𝜀$s.
 
-```julia
+```@example perturbation
 using Plots
 
 plot(sol.t, hcat([X(𝜀) for 𝜀 in 0.0:0.1:0.5]...))
@@ -135,17 +135,18 @@ For the next example, we have chosen a simple example from a very important clas
 
 The goal is to solve $\ddot{x} + 2\epsilon\dot{x} + x = 0$, where the dot signifies time-derivatives and the initial conditions are $x(0) = 0$ and $\dot{x}(0) = 1$. If $\epsilon = 0$, the problem reduces to the simple linear harmonic oscillator with the exact solution $x(t) = \sin(t)$. We follow the same steps as the previous example.
 
-```julia
-n = 3
-@variables ϵ t y[1:n](t) ∂y[1:n] ∂∂y[1:n]
-x = def_taylor(ϵ, y[3:end], y[2])
-∂x = def_taylor(ϵ, ∂y[3:end], ∂y[2])
-∂∂x = def_taylor(ϵ, ∂∂y[3:end], ∂∂y[2])
+```@example perturbation
+order = 2
+n = order + 1
+@variables ϵ (y(t))[1:n] (∂y)[1:n] (∂∂y)[1:n]
+x = def_taylor(ϵ, y)
+∂x = def_taylor(ϵ, ∂y)
+∂∂x = def_taylor(ϵ, ∂∂y)
 ```
 
 This time we also need the first derivative terms. Continuing,
 
-```julia
+```@example perturbation
 eq = ∂∂x + 2 * ϵ * ∂x + x
 eqs = collect_powers(eq, ϵ, 0:n)
 vals = solve_coef(eqs, ∂∂y)
@@ -153,7 +154,7 @@ vals = solve_coef(eqs, ∂∂y)
 
 Next, we need to replace `∂`s and `∂∂`s with their **Symbolics.jl** counterparts:
 
-```julia
+```@example perturbation
 subs1 = Dict(∂y[i] => D(y[i]) for i in eachindex(y))
 subs2 = Dict(∂∂y[i] => D(D(y[i])) for i in eachindex(y))
 subs = subs1 ∪ subs2
@@ -162,19 +163,19 @@ eqs = [substitute(first(v), subs) ~ substitute(last(v), subs) for v in vals]
 
 We continue with converting 'eqs' to an `ODEProblem`, solving it, and finally plot the results against the exact solution to the original problem, which is $x(t, \epsilon) = (1 - \epsilon)^{-1/2} e^{-\epsilon t} \sin((1- \epsilon^2)^{1/2}t)$,
 
-```julia
+```@example perturbation
 @mtkbuild sys = ODESystem(eqs, t)
 ```
 
-```julia
+```@example perturbation
 # the initial conditions
-u0 = zeros(2n + 2)
-u0[3] = 1.0   # y₀ˍt
+u0 = zeros(2order + 2)
+u0[1] = 1.0   # yˍt₁
 
 prob = ODEProblem(sys, u0, (0, 50.0))
 sol = solve(prob; dtmax = 0.01)
 
-X = 𝜀 -> sum([𝜀^(i - 1) * sol[y[i]] for i in eachindex(y)])
+X = 𝜀 -> sum([𝜀^(i - 1) * sol[yi] for (i, yi) in enumerate(y)])
 T = sol.t
 Y = 𝜀 -> exp.(-𝜀 * T) .* sin.(sqrt(1 - 𝜀^2) * T) / sqrt(1 - 𝜀^2)    # exact solution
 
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
> It would probably be good to use the @example ... feature from Documenter.jl to add the output of commands to the documentation. This way it would be easier to detect when some change breaks the example.

We are doing that over time. Most examples are already updated. We shouldn't merge this one until it's using `@example`.

hersle:
Nice! Since I am working on something somewhat similar to this example, just a few suggestions:

1. Agreed, definitely use `@example`.
2. Can you get away with `using Symbolics` instead of `using Symbolics, SymbolicUtils`?
3. To simplify `collect_powers()`, I have used this solution:
    ```julia
    def_power(x, ϵ) = sum([xi * ϵ^(i - 1) for (i, xi) in enumerate(x)]) # define power series expansion of x
    get_power(x, ϵ, p) = p == 0 ? substitute(x, ϵ => 0) : get_power(expand_derivatives(Differential(ϵ)(x)) / p, ϵ, p - 1) # get O(ϵ^p) term of x
    ```
    `get_power()` is basically a recursion implementation of that the $O(\epsilon^p)$ term of $x$ is $(1/p!) \cdot \partial^p x / \partial^p \epsilon |_{\epsilon = 0}$ (i.e. get the factor in front of $\epsilon^p$ by differentiating away $p$ factors of $\epsilon$ and removing the $p!$ factor that "comes down with it"). I think it should be equivalent to `collect_powers()` and (IMO) maybe a little more symmetrical to the `def_power` definition, and you avoid the annoying `max_power` variable. You can then do `eqs = get_power.(eq, ϵ, 0:order)`. Maybe this is less obvious, though.
4. I agree that 0-based indices would make this much nicer. I reported #2670 and tried to make a fix. I will see if I can "revive" that.

ChrisRackauckas:
If you've just got this working, mark the whole thing as `@example` so it won't regress.

karlwessel:
> 3\. To simplify `collect_powers()`, I have used this solution:


Very nice! But why the recursive implementation and not directly implement the mathematical equation you gave:
```julia
function power(poly, sym, p)
    x = sym
    ∂ₓ = Differential(x)
    p! = factorial(p)
    substitute(expand_derivatives((∂ₓ^p)(poly)) / p!, x => 0)
end
```

If there aren't any performance concerns, this probably could/should be used as implementation for an zeroth order respecting `Symbolics.coeff`.

hersle:
Beautiful, that's even better! I didn't think about the `Differential(ϵ)^p` syntax.

> If there aren't any performance concerns, this probably could/should be used as implementation for an zeroth order respecting Symbolics.coeff.

I think it primarily depends on what the function should do. Suppose you have a non-polynomial expression, like `x = sin(ϵ*t)`. If I want its `O(ϵ)` perturbation, I probably want `t` (i.e. the derivative trick, effectively treating `sin` with its Taylor series expansion). But if I want the coefficient in front of `ϵ` of the same expression, I probably want `0` (i.e. what `Symbolics.coeff` currently gives).

Also, https://github.com/SciML/ModelingToolkit.jl/pull/2671 was merged, so 0-based arrays should work now (on the master branch), if you would like to try it.

karlwessel:
> Suppose you have a non-polynomial expression, like `x = sin(ϵ*t)`. If I want its `O(ϵ)` perturbation, I probably want `t` (i.e. the derivative trick, effectively treating `sin` with its Taylor series expansion). But if I want the coefficient in front of `ϵ` of the same expression, I probably want `0` (i.e. what `Symbolics.coeff` currently gives).

You are right. But this also means `get_power` could be used in [the algebraic perturbation example](https://symbolics.juliasymbolics.org/stable/examples/perturbation/#Solving-the-Kepler's-Equation) to automatically linearize `sin(x)` without having to expand it manually, right?

karlwessel:
> Also, #2671 was merged, so 0-based arrays should work now (on the master branch), if you would like to try it.

This still fails:
```julia
julia> using Symbolics

julia> def_taylor(x, ps) = sum([a * x^(i - 1) for (i, a) in enumerate(ps)])
def_taylor (generic function with 1 method)

julia> function collect_powers(eq, x, ns; max_power = 100)
           eq = substitute(expand(eq), Dict(x^j => 0 for j in (last(ns) + 1):max_power))

           eqs = []
           for i in ns
               powers = Dict(x^j => (i == j ? 1 : 0) for j in 1:last(ns))
               e = substitute(eq, powers)

               # manually remove zeroth order from higher orders
               if 0 in ns && i != 0
                   e = e - eqs[1]
               end

               push!(eqs, e)
           end
           eqs
       end
collect_powers (generic function with 1 method)

julia> function solve_coef(eqs, ps)
           vals = Dict()

           for (i, p) in enumerate(ps)
               eq = substitute(eqs[i], vals)
               vals[p] = Symbolics.symbolic_linear_solve(eq ~ 0, p)
           end
           vals
       end
solve_coef (generic function with 1 method)

julia> using ModelingToolkit: t_nounits as t, D_nounits as D

julia> order = 2
2

julia> @variables ϵ (y(t))[0:order] (∂∂y(t))[0:order]
3-element Vector{Any}:
 ϵ
  (y(t))[0:2]
  (∂∂y(t))[0:2]

julia> x = def_taylor(ϵ, y)
(y(t))[0] + (y(t))[1]*ϵ + (y(t))[2]*(ϵ^2)

julia> ∂∂x = def_taylor(ϵ, ∂∂y)
(∂∂y(t))[0] + ϵ*(∂∂y(t))[1] + (ϵ^2)*(∂∂y(t))[2]

julia> eq = ∂∂x * (1 + ϵ * x)^2 + 1
1 + ((1 + ((y(t))[0] + (y(t))[1]*ϵ + (y(t))[2]*(ϵ^2))*ϵ)^2)*((∂∂y(t))[0] + ϵ*(∂∂y(t))[1] + (ϵ^2)*(∂∂y(t))[2])

julia> eqs = collect_powers(eq, ϵ, 0:order)
3-element Vector{Any}:
                                                                           1 + (∂∂y(t))[0]
                                                      (∂∂y(t))[1] + 2(y(t))[0]*(∂∂y(t))[0]
 (∂∂y(t))[2] + 2(y(t))[0]*(∂∂y(t))[1] + 2(y(t))[1]*(∂∂y(t))[0] + ((y(t))[0]^2)*(∂∂y(t))[0]

julia> vals = solve_coef(eqs, ∂∂y)
Dict{Any, Any} with 3 entries:
  (∂∂y(t))[2] => 2.0(y(t))[1] - 3.0((y(t))[0]^2)
  (∂∂y(t))[1] => 2.0(y(t))[0]
  (∂∂y(t))[0] => -1.0

julia> subs = Dict(∂∂y[i] => D(D(y[i])) for i in eachindex(y))
Dict{Num, Num} with 3 entries:
  (∂∂y(t))[2] => Differential(t)(Differential(t)((y(t))[2]))
  (∂∂y(t))[1] => Differential(t)(Differential(t)((y(t))[1]))
  (∂∂y(t))[0] => Differential(t)(Differential(t)((y(t))[0]))

julia> eqs = [substitute(first(v), subs) ~ substitute(last(v), subs) for v in vals]
3-element Vector{Equation}:
 Differential(t)(Differential(t)((y(t))[2])) ~ 2.0(y(t))[1] - 3.0((y(t))[0]^2)
 Differential(t)(Differential(t)((y(t))[1])) ~ 2.0(y(t))[0]
 Differential(t)(Differential(t)((y(t))[0])) ~ -1.0

julia> using ModelingToolkit, DifferentialEquations

julia> @mtkbuild sys = ODESystem(eqs, t)
Model sys with 6 equations
Unknowns (6):
  var"y(t)ˍt"[2]
  var"y(t)ˍt"[1]
  var"y(t)ˍt"[0]
  (y(t))[2]
  (y(t))[1]
  (y(t))[0]
Parameters (0):

julia> unknowns(sys)
6-element Vector{SymbolicUtils.BasicSymbolic{Real}}:
 var"y(t)ˍt"[2]
 var"y(t)ˍt"[1]
 var"y(t)ˍt"[0]
 (y(t))[2]
 (y(t))[1]
 (y(t))[0]

julia> u0 = zeros(2order + 2)
6-element Vector{Float64}:
 0.0
 0.0
 0.0
 0.0
 0.0
 0.0

julia> u0[3] = 1.0
1.0

julia> prob = ODEProblem(sys, u0, (0, 3.0))
ERROR: BoundsError: attempt to access 3-element Vector{SymbolicUtils.BasicSymbolic{Real}} at index [0]
Stacktrace:
  [1] getindex(A::Vector{SymbolicUtils.BasicSymbolic{Real}}, i1::Int64)
    @ Base ./essentials.jl:13
  [2] fast_substitute(expr::SymbolicUtils.BasicSymbolic{Real}, subs::Dict{Any, Any}; operator::Type)
    @ Symbolics ~/.julia/packages/Symbolics/rtkf9/src/variable.jl:577
  [3] fast_substitute(eq::Equation, subs::Dict{Any, Any}; operator::Type)
    @ Symbolics ~/.julia/packages/Symbolics/rtkf9/src/variable.jl:543
  [4] tearing_reassemble(state::TearingState{…}, var_eq_matching::ModelingToolkit.BipartiteGraphs.Matching{…}, full_var_eq_matching::ModelingToolkit.BipartiteGraphs.Matching{…}; simplify::Bool, mm::ModelingToolkit.SparseMatrixCLIL{…})
    @ ModelingToolkit.StructuralTransformations ~/.julia/packages/ModelingToolkit/5OzIt/src/structural_transformation/symbolics_tearing.jl:594
  [5] tearing_reassemble
    @ ~/.julia/packages/ModelingToolkit/5OzIt/src/structural_transformation/symbolics_tearing.jl:230 [inlined]
  [6] tearing(sys::NonlinearSystem, state::TearingState{…}; mm::ModelingToolkit.SparseMatrixCLIL{…}, simplify::Bool, kwargs::@Kwargs{…})
    @ ModelingToolkit.StructuralTransformations ~/.julia/packages/ModelingToolkit/5OzIt/src/structural_transformation/symbolics_tearing.jl:641
  [7] _structural_simplify!(state::TearingState{…}, io::Nothing; simplify::Bool, check_consistency::Bool, fully_determined::Bool, warn_initialize_determined::Bool, dummy_derivative::Bool, kwargs::@Kwargs{…})
    @ ModelingToolkit ~/.julia/packages/ModelingToolkit/5OzIt/src/systems/systemstructure.jl:706
  [8] _structural_simplify!
    @ ~/.julia/packages/ModelingToolkit/5OzIt/src/systems/systemstructure.jl:676 [inlined]
  [9] #structural_simplify!#1221
    @ ~/.julia/packages/ModelingToolkit/5OzIt/src/systems/systemstructure.jl:669 [inlined]
 [10] __structural_simplify(sys::NonlinearSystem, io::Nothing; simplify::Bool, kwargs::@Kwargs{…})
    @ ModelingToolkit ~/.julia/packages/ModelingToolkit/5OzIt/src/systems/systems.jl:94
 [11] __structural_simplify
    @ ~/.julia/packages/ModelingToolkit/5OzIt/src/systems/systems.jl:75 [inlined]
 [12] structural_simplify(sys::NonlinearSystem, io::Nothing; simplify::Bool, split::Bool, allow_symbolic::Bool, allow_parameter::Bool, conservative::Bool, fully_determined::Bool, kwargs::@Kwargs{})
    @ ModelingToolkit ~/.julia/packages/ModelingToolkit/5OzIt/src/systems/systems.jl:33
 [13] structural_simplify (repeats 2 times)
    @ ~/.julia/packages/ModelingToolkit/5OzIt/src/systems/systems.jl:28 [inlined]
 [14] ModelingToolkit.InitializationProblem{…}(sys::ODESystem, t::Int64, u0map::Vector{…}, parammap::SciMLBase.NullParameters; guesses::Dict{…}, check_length::Bool, warn_initialize_determined::Bool, initialization_eqs::Vector{…}, fully_determined::Bool, check_units::Bool, kwargs::@Kwargs{…})
    @ ModelingToolkit ~/.julia/packages/ModelingToolkit/5OzIt/src/systems/diffeqs/abstractodesystem.jl:1475
 [15] (ModelingToolkit.InitializationProblem{…})(::ODESystem, ::Int64, ::Vararg{…}; kwargs::@Kwargs{…})
    @ ModelingToolkit ~/.julia/packages/ModelingToolkit/5OzIt/src/systems/diffeqs/abstractodesystem.jl:1434
 [16] ModelingToolkit.InitializationProblem(::ODESystem, ::Int64, ::Vararg{…}; kwargs::@Kwargs{…})
    @ ModelingToolkit ~/.julia/packages/ModelingToolkit/5OzIt/src/systems/diffeqs/abstractodesystem.jl:1422
 [17] process_DEProblem(constructor::Type, sys::ODESystem, u0map::Vector{…}, parammap::SciMLBase.NullParameters; implicit_dae::Bool, du0map::Nothing, version::Nothing, tgrad::Bool, jac::Bool, checkbounds::Bool, sparse::Bool, simplify::Bool, linenumbers::Bool, parallel::Symbolics.SerialForm, eval_expression::Bool, eval_module::Module, use_union::Bool, tofloat::Bool, symbolic_u0::Bool, u0_constructor::typeof(identity), guesses::Dict{…}, t::Int64, warn_initialize_determined::Bool, build_initializeprob::Bool, initialization_eqs::Vector{…}, fully_determined::Bool, check_units::Bool, kwargs::@Kwargs{…})
    @ ModelingToolkit ~/.julia/packages/ModelingToolkit/5OzIt/src/systems/diffeqs/abstractodesystem.jl:852
 [18] process_DEProblem
    @ ~/.julia/packages/ModelingToolkit/5OzIt/src/systems/diffeqs/abstractodesystem.jl:792 [inlined]
 [19] (ODEProblem{…})(sys::ODESystem, u0map::Vector{…}, tspan::Tuple{…}, parammap::SciMLBase.NullParameters; callback::Nothing, check_length::Bool, warn_initialize_determined::Bool, eval_expression::Bool, eval_module::Module, kwargs::@Kwargs{})
    @ ModelingToolkit ~/.julia/packages/ModelingToolkit/5OzIt/src/systems/diffeqs/abstractodesystem.jl:1022
 [20] ODEProblem
    @ ~/.julia/packages/ModelingToolkit/5OzIt/src/systems/diffeqs/abstractodesystem.jl:1010 [inlined]
 [21] (ODEProblem{true, SciMLBase.AutoSpecialize})(sys::ODESystem, u0map::Vector{Float64}, tspan::Tuple{Int64, Float64})
    @ ModelingToolkit ~/.julia/packages/ModelingToolkit/5OzIt/src/systems/diffeqs/abstractodesystem.jl:1010
 [22] (ODEProblem{true})(::ODESystem, ::Vector{Float64}, ::Vararg{Any}; kwargs::@Kwargs{})
    @ ModelingToolkit ~/.julia/packages/ModelingToolkit/5OzIt/src/systems/diffeqs/abstractodesystem.jl:997
 [23] (ODEProblem{true})(::ODESystem, ::Vector{Float64}, ::Vararg{Any})
    @ ModelingToolkit ~/.julia/packages/ModelingToolkit/5OzIt/src/systems/diffeqs/abstractodesystem.jl:996
 [24] ODEProblem(::ODESystem, ::Vector{Float64}, ::Vararg{Any}; kwargs::@Kwargs{})
    @ ModelingToolkit ~/.julia/packages/ModelingToolkit/5OzIt/src/systems/diffeqs/abstractodesystem.jl:986
 [25] ODEProblem(::ODESystem, ::Vector{Float64}, ::Vararg{Any})
    @ ModelingToolkit ~/.julia/packages/ModelingToolkit/5OzIt/src/systems/diffeqs/abstractodesystem.jl:985
 [26] top-level scope
    @ REPL[22]:1
```

karlwessel:
Both examples work now (with one-based arrays only) and I added the outputs using the `@example` feature.

I am not completely happy with the definition of the initial condition with
```julia
u0 = zeros(2order + 2)
u0[3] = 1.0
```
The order of the equations defined by `@mtkbuild` might change and the correct index might change with it. But I don't know how to define it in a more robust way.

I think making the indices zero based and using a different implementation for `collect_powers` can be done later with a separate PR.

hersle:
> You are right. But this also means get_power could be used in [the algebraic perturbation example](https://symbolics.juliasymbolics.org/stable/examples/perturbation/#Solving-the-Kepler's-Equation) to automatically linearize sin(x) without having to expand it manually, right?

I think so, yes! I will try to add it in https://github.com/JuliaSymbolics/Symbolics.jl/pull/1298.

> This still fails:

Ah shit. If I get time I will have a look.

> I think making the indices zero based and using a different implementation for collect_powers can be done later with a separate PR.

Agreed, making the example correct is most important! :)

hersle:
> I am not completely happy with the definition of the initial condition with

I think you can do something like (with 1-based `y[1:3]`)
```julia
u0 = Dict(unknowns(sys) .=> 0.0) # set all initial conditions
u0[D(y[1])] => 1.0 # ... except this one
```

karlwessel:
> > I am not completely happy with the definition of the initial condition with
> 
> I think you can do something like (with 1-based `y[1:3]`)
> 
> ```julia
> u0 = Dict(unknowns(sys) .=> 0.0) # set all initial conditions
> u0[D(y[1])] => 1.0 # ... except this one
> ```

Thank you, that worked.

Now everything is good to merge from my side.

PR #3108, opened 2024-10-09
fix: fix unnecessary `oneunit` call in `get_unit`

AayushSabharwal:
## Checklist


isaacsas:
Note that this causes this simple unit test to fail now: 

https://github.com/SciML/ModelingToolkit.jl/actions/runs/11250561770/job/31279691694?pr=3108#step:6:925

which is indicative that units in registered functions still have an issue that this was masking.

ChrisRackauckas:
I think we need a way to register unit arithmetic on a registered function.

PR #3117, opened 2024-10-13
Backwards compatibility of scope detection for Catalyst

ChrisRackauckas:
Attempt to fix https://github.com/SciML/Catalyst.jl/issues/1075


src/utils.jl
```diff
@@ -492,6 +492,7 @@ recursively searches through all subsystems of `sys`, increasing the depth if it
 function collect_scoped_vars!(unknowns, parameters, sys, iv; depth = 1, op = Differential)
     if has_eqs(sys)
         for eq in get_eqs(sys)
+            hasfield(eq, :lhs) || continue
```
isaacsas, line 495:
```suggestion
            eq isa Equation || continue
```
This should work as long as `eq` is an equation right?

PR #3122, opened 2024-10-14
fix: only use `OffsetArray` if starting index is not 1

AayushSabharwal:
## Checklist


ChrisRackauckas:
what does this fix?

AayushSabharwal:
It seemed to fix something on a slightly older version using MTKNeuralNets, but that error is gone now. I was having trouble making an MWE anyway.

PR #3123, opened 2024-10-15
add trait and dispatch for collect_vars!

isaacsas:
This should allow us to add a dispatch for `Reaction`s in Catalyst, along with refactoring to simplify the code in MTK a bit.

Note that I left a number of manual dispatches between `Pair`s and `Equation`s that are present in the current code. It would be cleaner to leave them to a dynamic dispatch, but I wasn't sure if this would have performance implications where it is currently being used.


ChrisRackauckas (approved):

ChrisRackauckas:
Yes this is a really nice refactor.

PR #3126, opened 2024-10-16
fix: some fixes related to usage of array symbolics

AayushSabharwal:
Requires https://github.com/JuliaSymbolics/Symbolics.jl/pull/1307

Regarding the improvement of the "hack", the issue with the old implementation was that it put `OffsetArray`s inside equations, which we can't codegen (`SymbolicUtils.Code.create_array(::Type{<:OffsetArray}, ...)` doesn't exist). This fixes that so we generate code equivalent to `y = OffsetArrays.Origin(0)([1, 2, 3])`.

## Checklist

### DIFF
--- a/src/structural_transformation/symbolics_tearing.jl
+++ b/src/structural_transformation/symbolics_tearing.jl
@@ -584,7 +584,7 @@ function tearing_reassemble(state::TearingState, var_eq_matching,
     end
     @set! sys.unknowns = unknowns
 
-    obs, subeqs = cse_and_array_hacks(
+    obs, subeqs, deps = cse_and_array_hacks(
         obs, subeqs, unknowns, neweqs; cse = cse_hack, array = array_hack)
 
     @set! sys.eqs = neweqs
@@ -637,10 +637,7 @@ function cse_and_array_hacks(obs, subeqs, unknowns, neweqs; cse = true, array = 
         vars!(all_vars, rhs)
 
         # HACK 1
-        if cse &&
-           (!ModelingToolkit.isvariable(rhs) || ModelingToolkit.iscalledparameter(rhs)) &&
-           iscall(rhs) && operation(rhs) === getindex &&
-           Symbolics.shape(rhs) != Symbolics.Unknown()
+        if cse && is_getindexed_array(rhs)
             rhs_arr = arguments(rhs)[1]
             if !haskey(rhs_to_tempvar, rhs_arr)
                 tempvar = gensym(Symbol(lhs))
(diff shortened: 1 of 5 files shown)


AayushSabharwal:
Requires https://github.com/JuliaSymbolics/Symbolics.jl/pull/1309

AayushSabharwal:
This should be good to go now

ChrisRackauckas:
It looks like this causes a lot of issues to MethodOfLines?

AayushSabharwal:
Yeah, it seems the CSE hack is invalid in some cases. I'll work on fixing it.

src/systems/nonlinear/initializesystem.jl
```diff
@@ -12,7 +12,9 @@ function generate_initializesystem(sys::ODESystem;
         algebraic_only = false,
         check_units = true, check_defguess = false,
         name = nameof(sys), kwargs...)
-    vars = unique([unknowns(sys); getfield.((observed(sys)), :lhs)])
+    trueobs = unhack_observed(observed(sys))
+    @show trueobs
```
ChrisRackauckas, line 16:
```suggestion
```

src/systems/nonlinear/initializesystem.jl
```diff
@@ -12,7 +12,9 @@ function generate_initializesystem(sys::ODESystem;
         algebraic_only = false,
         check_units = true, check_defguess = false,
         name = nameof(sys), kwargs...)
-    vars = unique([unknowns(sys); getfield.((observed(sys)), :lhs)])
+    trueobs = unhack_observed(observed(sys))
```
ChrisRackauckas, line 15: is this function defined?

src/structural_transformation/symbolics_tearing.jl
```diff
@@ -574,48 +574,122 @@ function tearing_reassemble(state::TearingState, var_eq_matching,
     # TODO: compute the dependency correctly so that we don't have to do this
     obs = [fast_substitute(observed(sys), obs_sub); subeqs]
 
-    # HACK: Substitute non-scalarized symbolic arrays of observed variables
-    # E.g. if `p[1] ~ (...)` and `p[2] ~ (...)` then substitute `p => [p[1], p[2]]` in all equations
-    # ideally, we want to support equations such as `p ~ [p[1], p[2]]` which will then be handled
-    # by the topological sorting and dependency identification pieces
-    obs_arr_subs = Dict()
+    unknowns = Any[v
```
ChrisRackauckas, line 577: It might be good to refactor this into a function call that is able to be turned on/off.

AayushSabharwal: Sure. I'll pull it out into a function and propagate a keyword argument to toggle it. Should it be on by default or off? I feel like HACK2 should be on by default, and CSE off since that's the more bug-prone of the two

ChrisRackauckas: The CSE one seems pretty essential to a lot of applications? I would try to see if we can get them on by default. But yes making an option would then make it much easier to isolate any potential bugs to it.

ChrisRackauckas (approved):

AayushSabharwal:
There's a bug

PR #3137, opened 2024-10-22
feat: use observed equations for guesses of observed variables

AayushSabharwal:
## Checklist

### DIFF
--- a/test/initializationsystem.jl
+++ b/test/initializationsystem.jl
@@ -836,3 +836,11 @@ end
     integ = init(prob, Rosenbrock23())
     @test integ[y] ≈ -0.5
 end
+
+@testset "Use observed equations for guesses of observed variables" begin
+    @variables x(t) y(t) [state_priority = 100]
+    @mtkbuild sys = ODESystem(
+        [D(x) ~ x + t, y ~ 2x + 1], t; initialization_eqs = [x^3 + y^3 ~ 1])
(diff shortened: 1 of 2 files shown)


test/initializationsystem.jl
```diff
@@ -836,3 +836,11 @@ end
     integ = init(prob, Rosenbrock23())
     @test integ[y] ≈ -0.5
 end
+
+@testset "Use observed equations for guesses of observed variables" begin
+    @variables x(t) y(t) [state_priority = 100]
+    @mtkbuild sys = ODESystem(
+        [D(x) ~ x + t, y ~ 2x + 1], t; initialization_eqs = [x^3 + y^3 ~ 1])
+    isys = ModelingToolkit.generate_initializesystem(sys)
+    @test isequal(defaults(isys)[y], 2x + 1)
```
ChrisRackauckas, line 845: but in this case `x` doesn't have a guess, so how would it work?

AayushSabharwal: The system wasn't meant to be solved, just to ensure that `y` has a guess using the observed equation.

ChrisRackauckas: I see but it's still not solvable?

AayushSabharwal: I can add a guess for `x` but I don't see how that helps the test 😅 the initializationsystem won't be solved during the test anyway

PR #3151, opened 2024-10-25
feat: support rational functions in `HomotopyContinuationProblem`

AayushSabharwal:
Requires SciML/SciMLBase.jl#829

## Checklist


ext/MTKHomotopyContinuationExt.jl
```diff
@@ -139,51 +188,74 @@ function MTK.HomotopyContinuationProblem(
...
+        return 0 ~ num
     end
 
-    nlfn, u0, p = MTK.process_SciMLProblem(NonlinearFunction{true}, sys, u0map, parammap;
+    sys2 = MTK.@set sys.eqs = eqs2
+
+    nlfn, u0, p = MTK.process_SciMLProblem(NonlinearFunction{true}, sys2, u0map, parammap;
         jac = true, eval_expression, eval_module)
 
+    denominator = MTK.build_explicit_observed_function(sys, denoms)
```
ChrisRackauckas, line 206: don't you just want the numerator?

AayushSabharwal: The denominator is used to rule out roots that cause a `0/0`. A trivial example would be `0 ~ (x - y) / (x - 3), 0 ~ y - 3`. If we take the numerators and solve, we get `x = y = 3`, but that causes 0 / 0 in the first polynomial and hence the system is actually infeasible.

ChrisRackauckas (approved):

PR #3155, opened 2024-10-28
fix: fix flattening in `complete`

AayushSabharwal:
The old implementation broke Catalyst, but it wasn't detected due to differing BifurcationKit compats preventing tests from running.

## Checklist


ChrisRackauckas:
Extension test failing? Seems unrelated though.

AayushSabharwal:
Huh, weird, it only fails on lts.

ChrisRackauckas:
Must be an alternative solution or something, just choosing the wrong root?

AayushSabharwal:
Nope, it's a bug. `x = 2` is an invalid root because it occurs in the denominator, but I forgot to add an `abs` in the check so it accepts `2 - 2e-8` but rejects `2 + 2e-8`.

PR #3161, opened 2024-10-28
fixing record from solution for observables in the BifurcationKit extension

vyudu:
## Additional context

Making the ObservedRecordFromSolution object work with the new definition of record_from_solution in BifurcationKit.


ChrisRackauckas (approved):

ChrisRackauckas:
@AayushSabharwal remember to bump this after the test fixes.

PR #3168, opened 2024-10-30
fix: fix `complete` not properly expanding systems

AayushSabharwal:
## Checklist

### DIFF
--- a/src/systems/abstractsystem.jl
+++ b/src/systems/abstractsystem.jl
@@ -933,8 +933,8 @@ function complete(sys::AbstractSystem; split = true, flatten = true)
         @set! sys.ps = unique!(vcat(get_ps(sys), collect(newparams)))
     end
     if flatten
-        if (eqs = equations(sys)) isa Vector &&
-           any(eq -> eq isa Equation && isconnection(eq.lhs), eqs)
+        eqs = equations(sys)
+        if eqs isa AbstractArray && eltype(eqs) <: Equation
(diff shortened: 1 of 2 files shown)


src/systems/abstractsystem.jl
```diff
@@ -933,8 +933,8 @@ function complete(sys::AbstractSystem; split = true, flatten = true)
         @set! sys.ps = unique!(vcat(get_ps(sys), collect(newparams)))
     end
     if flatten
-        if (eqs = equations(sys)) isa Vector &&
-           any(eq -> eq isa Equation && isconnection(eq.lhs), eqs)
+        eqs = equations(sys)
+        if eqs isa AbstractArray && eltype(eqs) <: Equation
```
ChrisRackauckas, line 937: What's a case where it's not a vector?

AayushSabharwal: `OptimizationSystem`. `equations(sys)` returns the objective.

ChrisRackauckas: but is there any case that is an abstractarray that's not a vector?

AayushSabharwal: `JumpSystem` stores it in an `ArrayPartition`, which is an `AbstractVector` but we don't want to `expand_connections` on that, and it would fail `eltype <: Equation` anyway.

ChrisRackauckas (approved):

PR #3180, opened 2024-11-04
JumpSystem cleanup

isaacsas:
As a first step before working on making JumpSystems support coupled ODEs/SDEs, I wanted to clean up the code a bit. This 

1. Adds support for `collect_vars!` for `JumpSystems`.
2. Cleans up the constructor to be more in line with how the code works for other system types. If at some point refactoring of common code in constructors is done it should now be easier to see what is different in JumpSystems.

Note that JumpSystems would still need some more work to support hierarchical modeling (i.e. sub-systems and scoping).

### DIFF
--- a/src/systems/jumps/jumpsystem.jl
+++ b/src/systems/jumps/jumpsystem.jl
@@ -219,7 +219,8 @@ end
 ##### MTK dispatches for JumpSystems #####
 function collect_vars!(unknowns, parameters, j::MassActionJump, iv; depth = 0, 
         op = Differential)    
-    for field in (j.scaled_rates, j.reactant_stoch, j.net_stoch)
+    collect_vars!(unknowns, parameters, j.scaled_rates, iv; depth, op)
+    for field in (j.reactant_stoch, j.net_stoch)
         for el in field
             collect_vars!(unknowns, parameters, el, iv; depth, op)
         end
--- a/src/systems/jumps/jumpsystem.jl
+++ b/src/systems/jumps/jumpsystem.jl
@@ -96,6 +96,11 @@ struct JumpSystem{U <: ArrayPartition} <: AbstractTimeDependentSystem
     """
     discrete_events::Vector{SymbolicDiscreteCallback}
     """
+    A `Vector{SymbolicContinuousCallback}` that model events.
+    The integrator will use root finding to guarantee that it steps at each zero crossing.
+    """
+    continuous_events::Vector{SymbolicContinuousCallback}
+    """
     Topologically sorted parameter dependency equations, where all symbols are parameters and
     the LHS is a single parameter.
     """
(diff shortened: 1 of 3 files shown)


isaacsas:
My plan is to next make the `ArrayPartition` that stores the equations also have a component for equations, which will be used to generate an `ODESystem` during `ODEProblem(::JumpSystem, ...)` to be used in making a coupled ODE-jump system.

AayushSabharwal (commented):
Looks good. Could you check if removing the `if !(sys isa JumpSystem)` [here](https://github.com/SciML/ModelingToolkit.jl/blob/76805bea9036d56a17d667c0325e444ceda707e2/src/systems/abstractsystem.jl#L926-L934) now works? It was added because `collect_vars!` didn't work

isaacsas:
Will update later for that. I think we need a dispatch on the trait I added in `collect_scoped_vars!`.

isaacsas:
Should be updated now.

isaacsas:
Lots of failures here, but as far as I can tell none are related to this PR. I formatted with JuliaFormatter v1.0.62 so not sure why that is failing either. Locally it reports there is nothing to format when run on the repo.

src/systems/jumps/jumpsystem.jl
```diff
@@ -179,30 +206,42 @@ function JumpSystem(eqs, iv, unknowns, ps;
...
+
     JumpSystem{typeof(ap)}(Threads.atomic_add!(SYSTEM_COUNT, UInt(1)),
-        ap, value(iv), unknowns, ps, var_to_name, observed, name, description, systems,
+        ap, iv′, us′, ps′, var_to_name, observed, name, description, systems,
         defaults, connector_type, disc_callbacks, parameter_dependencies,
         metadata, gui_metadata, checks = checks)
 end
 
+##### MTK dispatches for JumpSystems #####
+eqtype_supports_collect_vars(j::MassActionJump) = true
```
AayushSabharwal, line 221: What is the point of this trait? It doesn't seem to be used anywhere.

isaacsas: https://github.com/SciML/ModelingToolkit.jl/blob/76805bea9036d56a17d667c0325e444ceda707e2/src/utils.jl#L503

isaacsas: It is already used to make collect_scoped_vars work with non-Equation equations that have the trait.

isaacsas: More generally, by using a trait like this that by default opts systems/features out of supporting new functionality, it gives an easy way for downstream packages like Catalyst (or even other system types like JumpSystem) to subsequently turn the feature on, and helps avoid hard coded checks like `isa JumpSystem` or `isa Equation`.

isaacsas:
@ChrisRackauckas I think this is good to merge now -- the test failures all seem unrelated.

ChrisRackauckas (approved):

ChrisRackauckas:
This looks good for now. Though I think the path forward really is that a symbolic JumpSystem should have its variables as discretes and use the hybrid system stuff when possible. The reason is the purely discrete variables would then not be in the ODE when the two are mixed, that could be a big performance win. We probably need to sync about this.

isaacsas:
Happy to chat about that at some point. Maybe we can talk at the next office hours. 

That would require changes in JumpProcesses too presumably? I thought what we are setting up here — really in the other WIP PR — would be pretty similar to what one would normally write by hand.

PR #3181, opened 2024-11-04
Support ODEs as equations in JumpSystem

isaacsas:
- [x] Needs https://github.com/SciML/ModelingToolkit.jl/pull/3180 merged and to then rebase on master.
- [x] Needs tests.

Note this assumes a user has properly classified their jumps (i.e. they haven't made something a constant rate jump that has a condition that depends on a variable rate jump variable and/or an ODE modified variable).


src/systems/jumps/jumpsystem.jl
```diff
@@ -436,19 +482,31 @@ function DiffEqBase.ODEProblem(sys::JumpSystem, u0map, tspan::Union{Tuple, Nothi
...
+    # forward everything to be an ODESystem but the jumps
+    if has_equations(sys)
+        osys = ODESystem(equations(sys).x[4], get_iv(sys), unknowns(sys), parameters(sys);
+            observed = observed(sys), name = nameof(sys), description = description(sys),
+            systems = get_systems(sys), defaults = defaults(sys), 
+            discrete_events = discrete_events(sys), 
+            parameter_dependencies = parameter_dependencies(sys),
+            metadata = get_metadata(sys), gui_metadata = get_gui_metadata(sys))
+        osys = complete(osys)
+        return ODEProblem(osys, u0map, tspan, parammap; check_length, kwargs...)
```
isaacsas, line 500: This seemed like the best way to handle generating the underlying ODEProblem, but happy to consider alternative approaches if anyone has feedback.

isaacsas:
@ChrisRackauckas where should callbacks be handled here, the ODEProblem or the JumpProblem? (And actually, I'm not even sure in the lower-level interface what is supposed to be done within JumpProcesses -- are passing some callbacks to both problems and/or just the wrapped ODEProblem allowed and supposed to work?)

isaacsas:
For now I've made it that continuous events are passed to the generated ODEProblem but discrete events go via the JumpProblem. This will need some tests to check that works consistently.

src/systems/callbacks.jl
```diff
@@ -1,7 +1,9 @@
 #################################### system operations #####################################
-get_continuous_events(sys::AbstractSystem) = SymbolicContinuousCallback[]
-get_continuous_events(sys::AbstractODESystem) = getfield(sys, :continuous_events)
 has_continuous_events(sys::AbstractSystem) = isdefined(sys, :continuous_events)
+function get_continuous_events(sys::AbstractSystem)
+    has_continuous_events(sys) || return SymbolicContinuousCallback[]
+    getfield(sys, :continuous_events)
+end
```
isaacsas, line 6: This and the changes below are to allow non-AbstractODESystems to also work use continuous events (JumpSystems in this case).

src/systems/jumps/jumpsystem.jl
```diff
@@ -478,14 +490,24 @@ function DiffEqBase.ODEProblem(sys::JumpSystem, u0map, tspan::Union{Tuple, Nothi
...
-    ODEProblem(df, u0, tspan, p; kwargs...)
+    # forward everything to be an ODESystem but the jumps and discrete events
+    if has_equations(sys)
+        osys = ODESystem(equations(sys).x[4], get_iv(sys), unknowns(sys), parameters(sys);
+            observed = observed(sys), name = nameof(sys), description = description(sys),
+            systems = get_systems(sys), defaults = defaults(sys), 
+            parameter_dependencies = parameter_dependencies(sys),
+            metadata = get_metadata(sys), gui_metadata = get_gui_metadata(sys))
+        osys = complete(osys)
+        return ODEProblem(osys, u0map, tspan, parammap; check_length = false, kwargs...)
```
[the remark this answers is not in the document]
isaacsas: By isn't working I mean the solution indexing issue I reported in https://github.com/SciML/ModelingToolkit.jl/issues/3188 comes up now.

test/jumpsystem.jl
```diff
@@ -422,3 +422,39 @@ let
...
+
+    times = range(0.0, 20.0, length = 100)
+    Nsims = 2000
+    X = zeros(length(times))
+    Y = similar(X)
+    for n in 1:Nsims
+        sol = solve(jprob, Tsit5())
+        X .+= Array(sol(times; idxs = X))
+        Y .+= Array(sol(times; idxs = Y))
+    end
```
isaacsas, line 451: @AayushSabharwal this is the MWE for the indexing issue I mentioned.

src/systems/jumps/jumpsystem.jl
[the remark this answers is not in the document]
AayushSabharwal: My guess would be it uses the ODESystem for indexing instead of the `JumpSystem`, but that seems like it should still work bar maybe some idiosyncrasies with parameters. I still need something I can run to help debug this 😅

test/jumpsystem.jl
AayushSabharwal: Ah, GitHub didn't reload. I just noticed this comment. Will investigate

AayushSabharwal: Indexing works fine. You shadowed the symbolic variable `X` with `X = zeros(length(times))`, which is why it won't work

isaacsas: > Indexing works fine. You shadowed the symbolic variable `X` with `X = zeros(length(times))`, which is why it won't work

Doh. Thanks for catching that!

isaacsas:
@ChrisRackauckas assuming tests pass I think this is gtg.

ChrisRackauckas (approved):

isaacsas:
@ChrisRackauckas I think the doc/extension/Catalyst failures here are all unrelated (they seem to be BifurcationKit related).

rveltz:
can you link me to the issue please?

ChrisRackauckas:
```

ChrisRackauckas:
I assume that will show in your CI tests as of about 2 days ago?

rveltz:
Yes that's a mistake from KrylovKit. I notified them

PR #3186, opened 2024-11-05
Rewrite perturbation example with Symbolics' new Taylor series functions

hersle:
Follow-up to https://github.com/JuliaSymbolics/Symbolics.jl/pull/1298/ to use the new Taylor series functions.

I think this makes the perturbation example flow really well. It completely avoids any workarounds that were present before.


ChrisRackauckas (approved):
Yeah this looks great, major simplification.

PR #3192, opened 2024-11-08
feat: cache start system and solver in HomotopyContinuation interface

AayushSabharwal:
## Checklist


AayushSabharwal:
I realized polynomial checking doesn't take into account observed variables, so we don't report a nice error.

test/extensions/homotopy_continuation.jl
```diff
@@ -61,20 +61,11 @@ end
...
-    prob = HomotopyContinuationProblem(sys, [])
-    sol = solve(prob; threading = false)
-    @test SciMLBase.successful_retcode(sol)
-end
-
 @testset "Polynomial check and warnings" begin
     @variables x = 1.0
-    @parameters n = 4
+    @parameters n::Integer = 4
     @mtkbuild sys = NonlinearSystem([x^n + x^2 - 1 ~ 0])
```
isaacsas, line 67: No supporting integer exponents is a pretty big limitation. It means that very common expressions in bio/pharma modeling, Hill expressions, would not be usable here (i.e. terms like ` x^n / (K + x^n)` where `n` is an integer parameter). Is there no way to handle them? They should, ultimately, reduce down to polynomials once the denominators are removed and the integer exponent values are substituted in.

AayushSabharwal: Interesting. We can't cache the start system with parametric exponents. @ChrisRackauckas I guess we just have to do the caching optionally?

ChrisRackauckas: Why can't the hill functions in this case use `@constant`?

AayushSabharwal: It's also worth noting that pure HomotopyContinuation.jl also doesn't allow parametric exponents

AayushSabharwal: Parametric exponents are re-allowed with a warning

AayushSabharwal:
Extension test failures are BifurcationKit failing to precompile

ChrisRackauckas:
But this would be in extension tests. Just flip the order so we can see it pass.

test/extensions/homotopy_continuation.jl
isaacsas: Great, thanks.

PR #3201, opened 2024-11-13
Add a mechanism to add additional passes to the end of structural simplification before completion.

BenChung:
## Checklist

## Additional context

This is a preliminary idea for how to do it; we might want to have a trait-like system for this that's more powerful?

### DIFF
--- a/src/systems/systems.jl
+++ b/src/systems/systems.jl
@@ -26,7 +26,7 @@ topological sort of the observed equations in `sys`.
 + `fully_determined=true` controls whether or not an error will be thrown if the number of equations don't match the number of inputs, outputs, and equations.
 """
 function structural_simplify(
-        sys::AbstractSystem, io = nothing; simplify = false, split = true,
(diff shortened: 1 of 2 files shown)


src/systems/systems.jl
```diff
@@ -49,6 +49,9 @@ function structural_simplify(
     if newsys isa ODESystem || has_parent(newsys)
         @set! newsys.parent = complete(sys; split, flatten = false)
     end
+    for pass in additional_passes 
+        newsys = pass(newsys)
+    end
```
AayushSabharwal, line 54:
```suggestion
    for pass in additional_passes 
        newsys = pass(newsys)
    end
    if newsys isa ODESystem || has_parent(newsys)
        @set! newsys.parent = complete(sys; split, flatten = false)
    end
```

This way regardless of the passes, the system has the appropriate parent. Otherwise if the pass doesn't forward the parent to the transformed system, it'll break user workflows.

ChrisRackauckas (approved):

PR #3224, opened 2024-11-19
feat: add function to parse variable from string

AayushSabharwal:
## Checklist

### DIFF
--- a/src/systems/abstractsystem.jl
+++ b/src/systems/abstractsystem.jl
@@ -3285,18 +3285,25 @@ function dump_unknowns(sys::AbstractSystem)
     end
 end
 
-# syntax:
-# varname                  = "D(" varname ")" | arrvar | maybe_dummy_var
-# arrvar                   = maybe_dummy_var "[idxs...]"
-# idxs                     = int | int "," idxs
-# maybe_dummy_var          = namespacedvar | namespacedvar "(t)" |
-#                            namespacedvar "(t)" "ˍ" ts | namespacedvar "ˍ" ts |
-#                            namespacedvar "ˍ" ts "(t)"
-# ts                       = "t" | "t" ts
-# namespacedvar            = ident "₊" namespacedvar | ident "." namespacedvar | ident
-# 
-# I'd write a regex to validate this, but https://xkcd.com/1171/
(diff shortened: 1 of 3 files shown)


src/systems/abstractsystem.jl
```diff
@@ -3285,6 +3285,86 @@ function dump_unknowns(sys::AbstractSystem)
...
+Roughly supports the following CFG:
+
+```
+varname                  = "D(" varname ")" | arrvar | maybe_dummy_var
+arrvar                   = maybe_dummy_var "[idxs...]"
+idxs                     = int | int "," idxs
+maybe_dummy_var          = namespacedvar | namespacedvar "(t)" |
+                           namespacedvar "(t)" "ˍ" ts | namespacedvar "ˍ" ts |
+                           namespacedvar "ˍ" ts "(t)"
+ts                       = "t" | "t" ts
```
ChrisRackauckas, line 3301: `nameof(independent_variable(sys))`?

src/systems/abstractsystem.jl
```diff
@@ -3285,6 +3285,87 @@ function dump_unknowns(sys::AbstractSystem)
...
+
+Where `iv` is the independent variable, `int` is an integer and `ident` is an identifier.
+"""
+function parse_variable(sys::AbstractSystem, str::AbstractString)
+    iv = has_iv(sys) ? string(getname(get_iv(sys))) : nothing
+
+    # I'd write a regex to validate `str`, but https://xkcd.com/1171/
+    str = strip(str)
+    derivative_level = 0
+    while startswith(str, "D(") && endswith(str, ")")
```
ChrisRackauckas, line 3313: We may need to validate that `D` is also the differential?

AayushSabharwal: `D` never shows up in expressions, it always prints as `Differential(t)(x(t))`. In the interest of concision, I assumed that `D` refers to the derivative with respect to the independent variable. I could also add a case that handles `Differential(t)(x(t))` and validates that the independent variable matches.

ChrisRackauckas: Okay

PR #3273, opened 2024-12-16
add option to include disturbance args in `generate_control_function`

baggepinnen:
This PR adds the option to include the disturbance inputs as input arguments for the generated function. This is useful when designing state estimators where the noise model is non-additive.

@AayushSabharwal the PR includes a test, which is currently failing, likely because there is a modification to `wrap_array_vars` required. `wrap_array_vars` appears long and complicated without any comments or documentation, what does this function do and can it be made easier to extend?

### DIFF
--- a/src/inputoutput.jl
+++ b/src/inputoutput.jl
@@ -252,10 +252,11 @@ function generate_control_function(sys::AbstractODESystem, inputs = unbound_inpu
     process = get_postprocess_fbody(sys)
     wrapped_arrays_vars = disturbance_argument ?
                           wrap_array_vars(
-        sys, rhss; dvs, ps, inputs, cachesyms = (disturbance_inputs,)) :
+        sys, rhss; dvs, ps, inputs, extra_args = (disturbance_inputs,)) :
                           wrap_array_vars(sys, rhss; dvs, ps, inputs)
     f = build_function(rhss, args...; postprocess_fbody = process,
-        expression = Val{true}, wrap_code = wrap_mtkparameters(sys, false, 3) .∘
+        expression = Val{true}, wrap_code = wrap_mtkparameters(
+            sys, false, 3, Int(disturbance_argument) + 1) .∘
                                             wrapped_arrays_vars .∘
                                             wrap_parameter_dependencies(sys, false),
         kwargs...)
--- a/src/inputoutput.jl
+++ b/src/inputoutput.jl
@@ -219,7 +219,7 @@ function generate_control_function(sys::AbstractODESystem, inputs = unbound_inpu
         # ps = [ps; disturbance_inputs]
     end
     inputs = map(x -> time_varying_as_func(value(x), sys), inputs)
-    disturbance_inputs = map(x -> time_varying_as_func(value(x), sys), disturbance_inputs)
+    disturbance_inputs = unwrap.(disturbance_inputs)
 
     eqs = [eq for eq in full_equations(sys)]
     eqs = map(subs_constants, eqs)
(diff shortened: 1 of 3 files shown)


src/inputoutput.jl
```diff
@@ -219,10 +219,11 @@ function generate_control_function(sys::AbstractODESystem, inputs = unbound_inpu
         # ps = [ps; disturbance_inputs]
     end
     inputs = map(x -> time_varying_as_func(value(x), sys), inputs)
+    disturbance_inputs = map(x -> time_varying_as_func(value(x), sys), disturbance_inputs)
```
AayushSabharwal, line 222:
```suggestion
    disturbance_inputs = unwrap.(disturbance_inputs)
```

The `time_varying_as_func` stuff is hacky handling of DDEs.

src/inputoutput.jl
```diff
@@ -239,16 +240,22 @@ function generate_control_function(sys::AbstractODESystem, inputs = unbound_inpu
...
+        args = (u, inputs, p..., t)
+    end
     if implicit_dae
         ddvs = map(Differential(get_iv(sys)), dvs)
         args = (ddvs, args...)
     end
     process = get_postprocess_fbody(sys)
+    wrapped_arrays_vars = disturbance_argument ?
+                          wrap_array_vars(sys, rhss; dvs, ps, inputs, disturbance_inputs) :
+                          wrap_array_vars(sys, rhss; dvs, ps, inputs)
```
AayushSabharwal, line 255:
```suggestion
    wrapped_arrays_vars = disturbance_argument ?
                          wrap_array_vars(sys, rhss; dvs, ps, inputs, cachesyms = (disturbance_inputs,)) :
                          wrap_array_vars(sys, rhss; dvs, ps, inputs)
```

`wrap_array_vars` is basically a workaround in our codegen to handle cases where an array variable is scalarized in unknowns/parameters but used in the unscalarized form in equations. The best solution is to pass `disturbance_inputs` to `cachesyms` as a tuple. This only works because `w` is the last argument to the generated function. I'll add documentation to `wrap_array_vars` in another PR, but it's _very_ dependent on what the arguments to the generated function are. As a result, it carries some bad API baggage to avoid having to refactor the million places it's called. Once I find the time, I plan to overhaul our codegen so it isn't implemented in such a scattered manner and has some new optimizations. That will also get rid of this API baggage and hopefully make `wrap_array_vars` more concise.

baggepinnen:
The updates did not make the test pass. Looking at the generated function
```julia
julia> f[1]
RuntimeGeneratedFunction(#=in ModelingToolkit=#, #=using ModelingToolkit=#, :((ˍ₋arg1, ˍ₋arg2, ___mtkparameters___, ˍ₋arg4)->begin
          #= /home/fredrikb/.julia/packages/SymbolicUtils/jf8aQ/src/code.jl:385 =#
          #= /home/fredrikb/.julia/packages/SymbolicUtils/jf8aQ/src/code.jl:386 =#
          #= /home/fredrikb/.julia/packages/SymbolicUtils/jf8aQ/src/code.jl:387 =#
          begin
              begin
                  begin
                      begin
                          begin
                              #= /home/fredrikb/.julia/packages/SymbolicUtils/jf8aQ/src/code.jl:480 =#
                              (SymbolicUtils.Code.create_array)(typeof(ˍ₋arg1), nothing, Val{1}(), Val{(1,)}(), (+)((+)(ˍ₋arg2[1], (*)(-1, ˍ₋arg1[1])), (^)(ˍ₋arg4[1], 2)))
                          end
                      end
                  end
              end
          end
      end))
```
it looks like `_arg4` is the `disturbance_inputs`, but time should have been `_arg4` and the disturbance inputs should have been `_arg5`. Right now, time does not appear to be included among the inputs arguments

baggepinnen:
It turns out that it works correctly with `split = false`, but test failing with `split=true`

baggepinnen:
looks good now, the test failures appear to be old

ChrisRackauckas (approved):

PR #3279, opened 2024-12-19
refactor: use Accessors instead of Setfield

AayushSabharwal:
I need Accessors for another PR, so it seemed like a good excuse to perform this refactor. Setfield is retained for compatibility.


ChrisRackauckas:
Looks like there's a lot of test failures?

AayushSabharwal:
Yeah I'll look into those. While fixing bugs in the other PR I managed to not need Accessors :D so this PR might not end up being necessary.

PR #3282, opened 2024-12-20
feat: add `FMUComponent`

AayushSabharwal:
This is only here temporarily as it enables easier development. Eventually we can move this to either an Stdlib extension or its own package.


AayushSabharwal:
Requires https://github.com/SciML/ModelingToolkit.jl/pull/3284

ChrisRackauckas:
Resolve conflicts?

AayushSabharwal:
Draft since there might be a way to not have to do https://github.com/SciML/ModelingToolkit.jl/pull/3282/commits/6cf8d9837f01e592550f55a870d3ed3cae6894d3

ChrisRackauckas:
Still needs docs, but let's merge and follow up with docs so this doesn't go stale.

PR #3285, opened 2024-12-29
feat: add analysis points

AayushSabharwal:
## Checklist


test/analysis_points.jl
```diff
@@ -0,0 +1,259 @@
...
+        (nameof(inputap), outputap),
+        (inputap, nameof(outputap)),
+        (nameof(inputap), nameof(outputap)),
+        (inputap, [outputap]),
+        (nameof(inputap), [outputap]),
+        (inputap, [nameof(outputap)]),
+        (nameof(inputap), [nameof(outputap)])
+    ]
+        if input isa Symbol
+            # broken because MTKStdlib defines the method for
```
baggepinnen, line 168: Can this method not be moved to MTK?

src/systems/analysis_points.jl
```diff
@@ -0,0 +1,645 @@
+"""
+    $(TYPEDEF)
+    $(TYPEDSIGNATURES)
+
+Create an AnalysisPoint for linear analysis. Analysis points can be created by calling
```
baggepinnen, line 5:
```suggestion
Create an AnalysisPoint for input-output and linear analysis. Analysis points can be created by calling
```

I'd like to widen the scope to handle any IO analysis

src/systems/analysis_points.jl
```diff
@@ -0,0 +1,645 @@
...
+    fn, root, namespace_hierarchy(target))
+
+"""
+    $(TYPEDSIGNATURES)
+
+Apply the modification to the nested subsystem of `root` where the name of the subsystem at
+each level in the hierarchy is given by elements of `hierarchy`. For example, if
+`hierarchy = [:a, :b, :c]`, the system being searched for will be `root.a.b.c`. Note that
+the hierarchy may include the name of the root system, in which the first element will be
+ignored. For example, `hierarchy = [:root, :a, :b, :c]` also searches for `root.a.b.c`.
```
baggepinnen, line 181: will this be problematic if the system contains `root.root.a.b.c`? How do we know whether one `root` should be removed or not?

baggepinnen (commented):
This PR only includes the `get_sensitivity_function`, but not `get_sensitivity`, is this a deliberate choice?

src/systems/analysis_points.jl
```diff
@@ -0,0 +1,645 @@
...
+
+The added variables will have a default of zero, of the appropriate type and size.
+"""
+struct ComplementarySensitivityTransform <: AnalysisPointTransformation
+    ap::AnalysisPoint
+end
+
+function apply_transformation(cst::ComplementarySensitivityTransform, sys::AbstractSystem)
+    sys, (u,) = apply_transformation(GetInput(cst.ap), sys)
+    sys, (du,) = apply_transformation(AddVariable(cst.ap, Symbol(:comp_sens_du)), sys)
```
baggepinnen, line 518:
```suggestion
    sys, (du,) = apply_transformation(AddVariable(cst.ap, :comp_sens_du), sys)
```
?

test/analysis_points.jl
AayushSabharwal: Will look into disabling the method in Stdlib if this one is defined in MTK

src/systems/analysis_points.jl
AayushSabharwal: Yeah that is a bit of a problem. The main issue here is that with `getproperty` syntax to access analysis points, the returned analysis point may or may not have the namespace of the outermost system depending on whether the outermost system is marked with `complete` or not. We have to handle both cases, but it leads to ambiguities. In your example, the best way to do it would be to ensure the outermost `root` is not marked as `complete`. I'm open to better suggestions.

src/systems/analysis_points.jl
AayushSabharwal: This is fixed now. Each of these added variables has a name derived from the analysis point to avoid issues where `get_comp_sensitivity` is called with multiple analysis points in the same (sub-)system.

AayushSabharwal:
I'd really appreciate feedback on the docstrings, and suggestions for the docstring for the new `linearization_function` method.

src/systems/analysis_points.jl
baggepinnen: I think it has to follow the same convention as for variable access, otherwise it'll be confusing. This is why I was using complete everywhere, and why I was sad to see `complete` change behavior since it broke my code

baggepinnen:
There are lots of docstrings from the old implementation missing, [all these](https://github.com/SciML/ModelingToolkitStandardLibrary.jl/blob/main/src/Blocks/analysis_points.jl#L531C2-L588C17) + the ones for `AnalysisPoint` and `connect(a, :name, b)` have lost a lot of detail

AayushSabharwal:
Should we make this new interface experimental as well?

baggepinnen:
> Should we make this new interface experimental as well?

we could, but ideally not for too long

AayushSabharwal:
I'm not advocating for it being experimental, so it would be great to have this be stable if you're happy with it.

baggepinnen:
> it would be great to have this be stable if you're happy with it.

If you feel like it should, I'm fine with that as well

baggepinnen:
Should we add some version of this warning?
https://github.com/SciML/ModelingToolkitStandardLibrary.jl/blob/main/src/Blocks/analysis_points.jl#L136C2-L138C4

AayushSabharwal:
While that check can be implemented in MTK, the concept of a `RealInput` and `RealOutput` doesn't exist here. I can check that the connection variables in the input/output systems have the correct `input` and `output` metadata, but that would restrict `AnalysisPoint` to be used only with such variables

baggepinnen:
Yeah this would act on the IO metadata, and it only works in setting that have those properties anyways (regardless of the metadata), so that's likely fine. If the user commits the mistake the warnings warn against, the result will be silently wrong, so we have to either warn or error.

src/systems/analysis_points.jl
AayushSabharwal: This will mean that in the following example:
```julia
@named P = FirstOrder(k = 1, T = 1)
@named C = Gain(; k = -1)

ap = AnalysisPoint(:plant_input)
eqs = [connect(P.output, C.input), connect(C.output, ap, P.input)]
sys = ODESystem(eqs, t, systems = [P, C], name = :hej)
```
The analysis point must be referred to as `sys.plant_input` or `Symbol(:hej, Symbolics.NAMESPACE_SEPARATOR, :plant_input)` or `Symbol(:hej, ".", :plant_input)` (or the `Symbol`s wrapped in an `AnalysisPoint`). Simply passing `ap` above won't work.

test/analysis_points.jl
AayushSabharwal: Resolved as of MTKStdlib@2.19

src/systems/analysis_points.jl
AayushSabharwal: The discussed solution:
> A more consistent way to do this is to make it so `AnalysisPoint`s passed to `get_sensitivity` have to be fully namespaced, but `Symbol` names omit the top-level namespace and we always assume it exists.

AayushSabharwal:
Once CI passes and the PR is approved, I'll rebase and make SpellCheck happy

baggepinnen (approved):

AayushSabharwal:
Apparently CI won't deploy docs for preview because the PR is from a fork

baggepinnen:
> Apparently CI won't deploy docs for preview because the PR is from a fork

I guess we can tackle any docs issues after merge?

AayushSabharwal:
Yeah. It's not a blocker, it just would've been nice to see.

ChrisRackauckas (approved):

PR #3286, opened 2025-01-02
build: bump OrdinaryDiffEqDefault compat

AayushSabharwal:
This should fix InterfaceI test failures


ChrisRackauckas (approved):

AayushSabharwal:
`SciMLBase/Downstream` is fixed in https://github.com/SciML/SciMLBase.jl/pull/899. ODEDefault needs a tag.

PR #3296, opened 2025-01-08
Generic mechanism for debugging/logging functions

hersle:
I find `debug_system()` very helpful for seeing where code generated by symbolics crashes. But it currently only handles a limited set of error types. It also duplicates domain-checking logic that must be added for every function that is to be debugged, so I think it scales poorly.

I propose this generic mechanism instead, which handles any error type, relies on the original functions' error/domain handling, and lets the user pass which functions they want debugged.

## Checklist

### DIFF
--- a/src/systems/abstractsystem.jl
+++ b/src/systems/abstractsystem.jl
@@ -2266,18 +2266,14 @@ Replace functions with singularities with a function that errors with symbolic
 information. E.g.
 
 ```julia-repl
-julia> sys = debug_system(sys)
+julia> sys = debug_system(complete(sys))
 
-julia> sys = complete(sys)
-
-julia> prob = ODEProblem(sys, [1.0, 0.0], (0.0, 1.0))
+julia> prob = ODEProblem(sys, [0.0, 2.0], (0.0, 1.0))
 
 julia> prob.f(prob.u0, prob.p, 0.0)
-ERROR: Function log(-cos(Q(t))) errors with input
-  -cos(Q(t)) => -1.0
-Stacktrace:
-  [1] (::ModelingToolkit.LoggedFun{typeof(log)})(num_args::Float64)
-  ...
+ERROR: Function /(1, sin(P(t))) output non-finite value Inf with input
+  1 => 1
+  sin(P(t)) => 0.0
 ```
 """
 function debug_system(sys::AbstractSystem)
--- a/src/systems/abstractsystem.jl
+++ b/src/systems/abstractsystem.jl
@@ -2266,19 +2266,17 @@ Replace functions with singularities with a function that errors with symbolic
 information. E.g.
 
 ```julia-repl
-julia> sys = debug_system(sys);
+julia> sys = debug_system(sys)
 
-julia> sys = complete(sys);
+julia> sys = complete(sys)
 
-julia> prob = ODEProblem(sys, [], (0, 1.0));
+julia> prob = ODEProblem(sys, [1.0, 0.0], (0.0, 1.0))
 
-julia> du = zero(prob.u0);
-
-julia> prob.f(du, prob.u0, prob.p, 0.0)
-ERROR: DomainError with (-1.0,):
-log errors with input(s): -cos(Q(t)) => -1.0
+julia> prob.f(prob.u0, prob.p, 0.0)
+ERROR: Function log(-cos(Q(t))) errors with input
+  -cos(Q(t)) => -1.0
 Stacktrace:
-  [1] (::ModelingToolkit.LoggedFun{typeof(log)})(args::Float64)
+  [1] (::ModelingToolkit.LoggedFun{typeof(log)})(num_args::Float64)
   ...
 ```
 """
--- a/src/systems/abstractsystem.jl
+++ b/src/systems/abstractsystem.jl
@@ -2282,7 +2282,7 @@ Stacktrace:
 """
 function debug_system(sys::AbstractSystem)
     if has_systems(sys) && !isempty(get_systems(sys))
-        error("debug_system only works on systems with no sub-systems!")
+        error("debug_system(sys) only works on systems with no sub-systems! Consider flattening it with flatten(sys) or structural_simplify(sys) first.")
     end
     if has_eqs(sys)
         @set! sys.eqs = debug_sub.(equations(sys))
--- a/src/systems/abstractsystem.jl
+++ b/src/systems/abstractsystem.jl
@@ -2278,7 +2278,8 @@ ERROR: Function /(1, sin(P(t))) output non-finite value Inf with input
   sin(P(t)) => 0.0
 ```
 """
-function debug_system(sys::AbstractSystem; functions = [log, sqrt, (^), /, inv, asin, acos], kw...)
+function debug_system(
+        sys::AbstractSystem; functions = [log, sqrt, (^), /, inv, asin, acos], kw...)
     if !(functions isa Set)
         functions = Set(functions) # more efficient "in" lookup
     end
--- a/src/systems/abstractsystem.jl
+++ b/src/systems/abstractsystem.jl
@@ -2260,11 +2260,11 @@ macro mtkbuild(exprs...)
 end
 
 """
-    debug_system(sys::AbstractSystem; error_nonfinite = true)
+    debug_system(sys::AbstractSystem; functions = [log, sqrt, (^), /, inv, asin, acos], error_nonfinite = true)
 
-Replace functions with singularities with a function that errors with symbolic
-information. If `error_nonfinite`, debugged functions that output nonfinite values
-(like `Inf` or `NaN`) also display errors, even though the raw function itself
+Wrap `functions` in `sys` so any error thrown in them shows helpful symbolic-numeric
+information about its input. If `error_nonfinite`, functions that output nonfinite
+values (like `Inf` or `NaN`) also display errors, even though the raw function itself
 does not throw an exception (like `1/0`). For example:
 
 ```julia-repl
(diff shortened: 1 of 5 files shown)


ChrisRackauckas:
Try catch can be quite expensive though, how much of an overhead does this have? It can be okay for a debug mode though

hersle:
It actually seems free for an example that runs without errors, and better than the old behavior:

1) Save this [spring-mass example](https://docs.sciml.ai/ModelingToolkit/dev/examples/spring_mass/) to `bench.jl` (it has some "debuggable" division operations):
```julia
using ModelingToolkit, Plots, OrdinaryDiffEq, LinearAlgebra, BenchmarkTools
using Symbolics: scalarize

function Mass(; name, m = 1.0, xy = [0.0, 0.0], u = [0.0, 0.0])
    ps = @parameters m = m
    sts = @variables pos(t)[1:2]=xy v(t)[1:2]=u
    eqs = scalarize(D.(pos) .~ v)
    ODESystem(eqs, t, [pos..., v...], ps; name)
end

function Spring(; name, k = 1e4, l = 1.0)
    ps = @parameters k=k l=l
    @variables x(t), dir(t)[1:2]
    ODESystem(Equation[], t, [x, dir...], ps; name)
end

function connect_spring(spring, a, b)
    [spring.x ~ norm(scalarize(a .- b))
     scalarize(spring.dir .~ scalarize(a .- b))]
end

function spring_force(spring)
    -spring.k .* scalarize(spring.dir) .* (spring.x - spring.l) ./ spring.x
end

m = 1.0
xy = [1.0, -1.0]
k = 1e4
l = 1.0
center = [0.0, 0.0]
g = [0.0, -9.81]
@named mass = Mass(m = m, xy = xy)
@named spring = Spring(k = k, l = l)

eqs = [connect_spring(spring, mass.pos, center)
       scalarize(D.(mass.v) .~ spring_force(spring) / mass.m .+ g)]

@named _model = ODESystem(eqs, t, [spring.x; spring.dir; mass.pos], [])
@named model = compose(_model, mass, spring)

println("\nDebugging OFF, branch ", read(`git branch --show-current`, String))
prob = ODEProblem(sys, [], (0.0, 3.0))
bench = @benchmark solve(prob, Rosenbrock23())
display(bench)

println("\nDebugging ON, branch ", read(`git branch --show-current`, String))
dsys = debug_system(sys)
dprob = ODEProblem(dsys, [], (0.0, 3.0))
dbench = @benchmark solve(dprob, Rosenbrock23())
display(dbench)
```
2) Compare from the shell:
```sh
git checkout --quiet master && julia bench.jl
git checkout --quiet generic_logged_functions && julia bench.jl
```
3) Output:
```
Debugging OFF, branch master

BenchmarkTools.Trial: 2066 samples with 1 evaluation per sample.
 Range (min … max):  2.257 ms … 30.838 ms  ┊ GC (min … max): 0.00% … 91.47%
 Time  (median):     2.294 ms              ┊ GC (median):    0.00%
 Time  (mean ± σ):   2.411 ms ±  1.146 ms  ┊ GC (mean ± σ):  3.51% ±  6.86%

   ▂▆██▇▇▅▅▃▃▂▂▁           ▁▁  ▁            ▁    ▁           ▁
  ▅███████████████▇▇▇▆▆█▆▆▇███▆██▆▇▇█▆▄▄▅▅████▆█▇█▇▇▅▄▄▁▄▁▄▅ █
  2.26 ms      Histogram: log(frequency) by time     2.65 ms <

 Memory estimate: 546.52 KiB, allocs estimate: 10677.

Debugging ON, branch master

BenchmarkTools.Trial: 632 samples with 1 evaluation per sample.
 Range (min … max):  7.026 ms … 42.955 ms  ┊ GC (min … max): 0.00% … 79.87%
 Time  (median):     7.148 ms              ┊ GC (median):    0.00%
 Time  (mean ± σ):   7.913 ms ±  3.002 ms  ┊ GC (mean ± σ):  6.25% ± 11.68%

  █▂     ▃▂                                                   
  ██▄▁▁▄▄██▆▄▁▄▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▄▁▁▄▄▅▅▄▁▁▁▄ ▆
  7.03 ms      Histogram: log(frequency) by time       22 ms <

 Memory estimate: 3.62 MiB, allocs estimate: 129585.
Precompiling ModelingToolkit...
  2 dependencies successfully precompiled in 90 seconds. 299 already precompiled.

Debugging OFF, branch generic_logged_functions

BenchmarkTools.Trial: 2011 samples with 1 evaluation per sample.
 Range (min … max):  2.226 ms … 34.032 ms  ┊ GC (min … max): 0.00% … 92.39%
 Time  (median):     2.269 ms              ┊ GC (median):    0.00%
 Time  (mean ± σ):   2.476 ms ±  1.210 ms  ┊ GC (mean ± σ):  3.28% ±  6.48%

  ▄██▅▄▃▁      ▁▃    ▂▁                 ▃▅▂                ▁ ▁
  ████████▇▆▆████▇▄▅▄███▆▅▄▁▄▄▄▄▁▁▄▄▁▄▄▄████▇▆▆▁▁▄▆▅▆▁▁▁▁▁▆█ █
  2.23 ms      Histogram: log(frequency) by time     3.16 ms <

 Memory estimate: 546.52 KiB, allocs estimate: 10677.

Debugging ON, branch generic_logged_functions

BenchmarkTools.Trial: 2083 samples with 1 evaluation per sample.
 Range (min … max):  2.233 ms … 29.775 ms  ┊ GC (min … max): 0.00% … 91.23%
 Time  (median):     2.272 ms              ┊ GC (median):    0.00%
 Time  (mean ± σ):   2.399 ms ±  1.048 ms  ┊ GC (mean ± σ):  2.62% ±  5.50%

  ▄██▆▃▂     ▁▂▃▂▂   ▁▂▂▃▁                                   ▁
  ████████▇▆▅█████▅▆██████▅▅▄▅▁▄▆▄▅▁▁▁▁▄▁▁▆█▆▁▁▄▄▁▁▁▄▁▄▁▁▁▁▆ █
  2.23 ms      Histogram: log(frequency) by time     3.12 ms <

 Memory estimate: 546.52 KiB, allocs estimate: 10677.
```
If you have a more complicated example in mind, I can try it.

hersle:
I added a short documentation page to help discover `debug_system()`.

What do you think?

hersle:
Format checker does not like [this workaround to show errors in the documentation](https://github.com/SciML/ModelingToolkit.jl/pull/3296/files#diff-535d0702fbb36b02bf41bf82f4532690182a583c3a61bb8849735bc7787e8824R30-R36). Not sure if there is a better way to handle this :/

hersle:
Never mind; `@repl` instead of `@example` got rid of the workaround. Documentation page now builds locally like this:

image

ChrisRackauckas (approved):

src/systems/abstractsystem.jl
```diff
@@ -2260,37 +2260,37 @@ macro mtkbuild(exprs...)
...
-  ...
+julia> prob.f(prob.u0, prob.p, 0.0)
+ERROR: Function /(1, sin(P(t))) output non-finite value Inf with input
+  1 => 1
+  sin(P(t)) => 0.0
 ```
 """
-function debug_system(sys::AbstractSystem)
+function debug_system(
+        sys::AbstractSystem; functions = [log, sqrt, (^), /, inv, asin, acos], kw...)
```
ChrisRackauckas, line 2282: Does this need to be the NaNMath versions?

hersle: No, not in the standard scenario. It contains the same functions as the old `LOGGED_FUN` constant list, with a few additions. But if a user decides to build a model with NaNMath functions and want those debugged too, they can now pass them through this keyword argument.

ChrisRackauckas: But all of the functions NaNMath by default now? It's not clear to me how these would ever be hit.

hersle: Oh okay, I was not aware, but I see what you mean now. This is what happens with a `√` system with and without `debug_system`:
```julia
@variables u(t) = 1.0
@named sys = ODESystem([D(u) ~ -√(u)], t)
prob = ODEProblem(sys, [], (0.0, 5.0))
dprob = ODEProblem(debug_system(sys), [], (0.0, 5.0))
```
Without debugging, `prob.f.f.f_oop` gives `NaNMath.sqrt`:
```
...
(SymbolicUtils.Code.create_array)(typeof(ˍ₋arg1), nothing, Val{1}(), Val{(1,)}(), (*)(-1, NaNMath.sqrt(ˍ₋arg1[1])))
...
```
With debugging, `dprob.f.f.f_oop` wraps a pure `sqrt` in a `LoggedFun` (no `NaNMath`):
```
...
(SymbolicUtils.Code.create_array)(typeof(ˍ₋arg1), nothing, Val{1}(), Val{(1,)}(), (*)(-1, (ModelingToolkit.LoggedFun{typeof(sqrt)}(sqrt, (u(t),), true))(ˍ₋arg1[1])))
...
```
What happens must be that `debug_system()` is applied first, wrapping `sqrt` into a `LoggedFun`, thereby "protecting" it from the `sqrt => NaNMath.sqrt` substitution in [`build_function`](https://github.com/JuliaSymbolics/Symbolics.jl/blob/eb3b5f62073ce8ea9334391aa594afcdb43dc453/src/build_function.jl#L99). Thus, `debug_system(sys; functions = [sqrt])` works "as expected" without `NaNMath.sqrt`.

hersle: Is that acceptable behavior? When debugging you want to crash early anyway, so I would think that the absence of `NaNMath` in that case is OK.

ChrisRackauckas: Oh interesting. I guess that works... Yeah. It's unintuitive at first how that ends up working but it really does give the designed behavior, so that seems fine.

PR #3304, opened 2025-01-09
feat: add support for causal connections of variables

### LINKED ISSUE #3287: support connections between (causal) variables `connect(var1, var2)`
In modelica and JSML one can `connect` two variables that are marked as `output` and `input`. In MTK, we can currently only `connect` systems that are `@connector`, and we thus have the `RealInput`, `RealOutput` in MTKstdlib. This is awkward, and makes the translation from JSML to MTK difficult.

This request is to allow `connect` between two variables
```julia
@variables u(t) [input = true] y(t) [output=true]

connect(y, u)
connect(y, :analysis_point_name, u)
```
and these connect statement should normally expand to `u ~ y`, unless an analysis is requested.

AayushSabharwal:
Close #3287


test/causal_variables_connection.jl
```diff
@@ -0,0 +1,98 @@
+using ModelingToolkit, ModelingToolkitStandardLibrary.Blocks
+using ModelingToolkit: t_nounits as t, D_nounits as D
+
+@testset "Error checking" begin
+    @variables begin
+        x(t)
+        y(t), [input = true]
```
baggepinnen, line 7: :troll:  :sunglasses:

baggepinnen (approved):

ChrisRackauckas (approved):

PR #3314, opened 2025-01-13
add utilities and tests for disturbance modeling

baggepinnen:
add utilities such as 
- analysis-point specific method for `generate_control_function`.
- Handling of disturbance inputs in `build_explicit_observed_function` and `generate_control_function`

also add tests for disturbance-modeling workflows

### DIFF
--- a/src/systems/diffeqs/odesystem.jl
+++ b/src/systems/diffeqs/odesystem.jl
@@ -436,6 +436,8 @@ an array of inputs `inputs` is given, and `param_only` is false for a time-depen
 """
 function build_explicit_observed_function(sys, ts;
         inputs = nothing,
+        disturbance_inputs = nothing,
+        disturbance_argument = false,
         expression = false,
         eval_expression = false,
         eval_module = @__MODULE__,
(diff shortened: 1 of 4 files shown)


src/systems/diffeqs/odesystem.jl
```diff
@@ -500,13 +502,22 @@ function build_explicit_observed_function(sys, ts;
...
     end
+    if disturbance_inputs !== nothing
+        # Disturbance inputs may or may not be included as inputs, depending on disturbance_argument
+        ps = setdiff(ps, disturbance_inputs)
+    end
+    if disturbance_argument
+        disturbance_inputs = (disturbance_inputs,)
+    else
+        disturbance_inputs = ()
+    end
```
AayushSabharwal, line 513: If `disturbance_inputs !== nothing` and `disturbance_argument` is `false`, the disturbance inputs won't be anywhere in the arguments to the function. Is this intentional?

baggepinnen: Yes, this is intentional, that's why there's a separate argument to control whether or not they are included as function arguments. 

The usecase is state estimation where the disturbance input is unknown (like sun shining in through the window) and you'd like to estimate this using a state estimator

AayushSabharwal (approved):

AayushSabharwal:
Yeah they're unrelated. I'll PR to fix some

ChrisRackauckas:
rebase so we can see the tests?

AayushSabharwal (approved):

ChrisRackauckas (approved):

PR #3251, opened 2024-12-01
Creating `BVProblem` from `ODESystem`

vyudu:
WIP PR for the BVProblem conversion.

PR #3323, opened 2025-01-14
BVProblem with constraints

vyudu:
Followup to #3251. Performance is something like 30x worse than generating the BVProblem using manually-defined f, bc, need to investigate why. Will uncomment the BVDAE solver tests when they get updated to use the same boundary condition format as the others.


vyudu:
I think this is mostly ready for review. Some of the problems in the tests don't seem fully well-posed, so I still need to work on that a bit, but the code itself should be correct.

src/systems/diffeqs/abstractodesystem.jl
```diff
@@ -849,6 +849,213 @@ function DiffEqBase.ODEProblem{iip, specialize}(sys::AbstractODESystem, u0map =
...
+
+`u0map` is used to specify fixed initial values for the states. Every variable 
+must have either an initial guess supplied using `guesses` or a fixed initial 
+value specified using `u0map`.
+
+`constraints` are used to specify boundary conditions to the ODESystem in the
+form of equations. These values should specify values that state variables should
+take at specific points, as in `x(0.5) ~ 1`). More general constraints that 
+should hold over the entire solution, such as `x(t)^2 + y(t)^2`, should be 
+specified as one of the equations used to build the `ODESystem`. Below is an example.
```
ChrisRackauckas, line 873: It's odd that this is here. Problem constructors are only supposed to take values: constraints are structural. They should be added to the system I would think? @baggepinnen thoughts?

test/bvproblem.jl
```diff
@@ -0,0 +1,309 @@
...
+
+    @parameters bad2
+    constraints = [bad2 + x(0.) ~ 3]
+    @test_throws Exception bvp = SciMLBase.BVProblem{true, SciMLBase.AutoSpecialize}(lksys, u0map, tspan; guesses, constraints)
+end
+
+# Cartesian pendulum from the docs.
+# DAE IVP solved using BoundaryValueDiffEq solvers.
+# let
+#     @parameters g
```
ChrisRackauckas, line 240: What's up with these ones?

src/systems/diffeqs/abstractodesystem.jl
baggepinnen: Can the `0.5` in `x(0.5) ~ 1` be a parameter? I can imagine this being something one would like to change without re-simplifying the system.

SebastianM-C: Would it make sense to have a `BVPSystem` that contains a `ConstraintSystem` like the `OptimizationSystem`?

ChrisRackauckas: Yes to both of these

test/bvproblem.jl
vyudu: The interface for specifying boundary conditions is different for the BVP solvers at the moment (you specify the time points as a separate argument), was gonna wait until they get updated to be the same.

src/systems/diffeqs/abstractodesystem.jl
vyudu: Sounds good, I will work on these conversions.

vyudu: These two should be done now, I think

vyudu:
@ChrisRackauckas this one should be ready now

test/bvproblem.jl
ChrisRackauckas: @ErikQQY

ErikQQY: On it, Ascher methods need lots of refactorizations

ChrisRackauckas:
are we sure those test failures are all unrelated?

vyudu:
It might be DiffEqBase 6.162.1? Just tested some of the broken tests on 6.161.0 locally and they worked.

vyudu:
@ChrisRackauckas is this good to merge now?

ChrisRackauckas (approved):

PR #3332, opened 2025-01-16
fix: fix `add_fallbacks!`

AayushSabharwal:


ChrisRackauckas:
What is this for?

AayushSabharwal:
`add_fallbacks!` is used to populate the operating point with values from defaults if not explicitly provided, handling array variables and reporting the variables that don't have a default or an entry in `u0map`/`pmap`. The bug was `get(fallbacks, var, nothing)[idx])` for scalarized array variables, where if `var` isn't present in `fallbacks` it would try to call `getindex(::Nothing, ::CartesianIndex)` which doesn't work. I just managed to MWE it and added a test.

PR #3334, opened 2025-01-16
fix: Add metadata getters

### LINKED ISSUE #2643: Missing metadata getters
It seems that getters for these metadata are missing: `misc`, `input`. Possibly this holds for a few others.

vyudu:
Fixes #2643. Adds getters for `connect, unit, misc`


vyudu:
@AayushSabharwal Haven't change `vartype` to `variable_source` in this PR but let me know if I should add it.

AayushSabharwal:
Yeah go for it

ChrisRackauckas:
add to the docs section on metadata?

vyudu:
This one should be ready

src/variables.jl
```diff
@@ -57,16 +55,16 @@ function dump_variable_metadata(var)
     disturbance = isdisturbance(uvar) || nothing
     tunable = istunable(uvar, isparameter(uvar))
     dist = getdist(uvar)
-    type = symtype(uvar)
+    variable_type = getvariabletype(uvar)
 
     meta = (
         var = var,
-        vartype,
+        variable_source,
```
AayushSabharwal, line 62: Changing this key is breaking

docs/src/basics/Variable_metadata.md
```diff
@@ -177,8 +181,44 @@ A variable can be marked `irreducible` to prevent it from being moved to an
...
+When a model is structurally simplified, the algorithm will try to ensure that the variables with higher state priority become states of the system. A variable's state priority is a number set using the `state_priority` metadata.
+
+```@example metadata
+@variable important_dof [state_priority = 10] unimportant_dof [state_priority = 2]
+state_priority(important_dof)
+```
+
+## Units
+
+Units for variables can be designated using symbolic metadata. For more information, please see the [model validation and units](@ref units) section of the docs. Note that `getunit` is equivalent to `get_unit`.
```
AayushSabharwal, line 200: `getunit` and `get_unit` aren't equivalent

AayushSabharwal (changes requested):

docs/src/basics/Variable_metadata.md
AayushSabharwal: Ah never mind, I read the code wrong

src/variables.jl
ChrisRackauckas: what will it break?

AayushSabharwal: `dump_unknowns` and `dump_parameters` are public API, and they return an array of these `NamedTuple`s. If a user is relying on the `vartype` key, changing it breaks their code.

ChrisRackauckas: I wouldn't say that's breaking. We don't guarantee what metadata is returned and we have changed that before.

ChrisRackauckas: But I guess the real question is, do we need to do the break? What's the motivation for the name change?

AayushSabharwal: > We don't guarantee what metadata is returned

If that's the case, sure. 

> we have changed that before.

We've only every added entries, which isn't breaking

> do we need to do the break? What's the motivation for the name change?

@vyudu brought up that it is inconsistently named and more importantly, doesn't accurately reflect the type of the variable. The `VariableSource` metadata tells which macro this variable came from, but whether it's a parameter/unknown actually depends on the `MTKVariableTypeCtx` metadata.

vyudu:
Sorry @AayushSabharwal I just realized that my replies were Pending and never got submitted. I thought we wanted vartype to be renamed in this PR? https://github.com/SciML/ModelingToolkit.jl/pull/3334#issuecomment-2596780414 If not I can revert

src/variables.jl
ChrisRackauckas: Alright, this is better and I don't think it's breaking.

vyudu:
@ChrisRackauckas this one is ready

docs/src/basics/Variable_metadata.md
```diff
@@ -177,8 +181,44 @@ A variable can be marked `irreducible` to prevent it from being moved to an
...
 @variable important_value [irreducible = true]
+isirreducible(important_value)
+```
+
+## State Priority
+
+When a model is structurally simplified, the algorithm will try to ensure that the variables with higher state priority become states of the system. A variable's state priority is a number set using the `state_priority` metadata.
+
+```@example metadata
+@variable important_dof [state_priority = 10] unimportant_dof [state_priority = 2]
```
ChrisRackauckas, line 194:
```suggestion
@variable important_dof [state_priority = 10] unimportant_dof [state_priority = -2]
```

ChrisRackauckas (approved):

Issue #3336, opened 2025-01-17
Remake does not handle dual type promotion with dummy derivatives


hersle:
Here is a simple throw-ball-straight-up-into-the-air example (2nd order ODE)
```julia
using ModelingToolkit, OrdinaryDiffEq, NonlinearSolve
using ModelingToolkit: t_nounits as t, D_nounits as D

# Throw ball straight up into the air
@variables y(t)
@named sys = ODESystem([
    D(D(y)) ~ -9.81
], t; initialization_eqs = [
    y^2 ~ 0 # initialize y = 0 in a way that builds an initialization problem
])
sys = structural_simplify(sys)

# Find initial throw velocity that reaches exactly 100 m after 1 s
dprob0 = ODEProblem(sys, [D(y) => NaN], (0.0, 1.0), []; guesses = [y => 0.0])
nprob = NonlinearProblem((ics, _) -> begin
    dprob = remake(dprob0, u0 = Dict(D(y) => ics[1]))
    dsol = solve(dprob, Tsit5())
    return [dsol[y][end] - 100.0]
end, [1.0])
nsol = solve(nprob, NewtonRaphson())
```
This crashes with
```
ERROR: MethodError: no method matching Float64(::ForwardDiff.Dual{ForwardDiff.Tag{DifferentiationInterface.FixTail{…}, Float64}, Float64, 1})       
The type `Float64` exists, but no method is defined for this combination of argument types when trying to construct it.

Closest candidates are:
  (::Type{T})(::Real, ::RoundingMode) where T<:AbstractFloat
   @ Base rounding.jl:265
  (::Type{T})(::T) where T<:Number
   @ Core boot.jl:900
  Float64(::Float16)
   @ Base float.jl:342
  ...

Stacktrace:
  [1] #1117
    @ C:\Users\herma\.julia\dev\ModelingToolkit\src\systems\diffeqs\abstractodesystem.jl:1386 [inlined]
  [2] (::ModelingToolkit.var"#1117#1120")(::Pair{Any, Any})
    @ ModelingToolkit .\none:0
  [3] iterate
    @ .\generator.jl:48 [inlined]
  [4] grow_to!(dest::Dict{Any, Any}, itr::Base.Generator{Dict{Any, Any}, ModelingToolkit.var"#1117#1120"})
    @ Base .\abstractdict.jl:587
  [5] dict_with_eltype
    @ .\abstractdict.jl:641 [inlined]
  [6] Dict
[...]

PR #3337, opened 2025-01-17
Fix dual type promotion in remake with dummy derivatives

hersle:
This seems to fix the minimal example in https://github.com/SciML/ModelingToolkit.jl/issues/3336.

I am still having problems with a slightly more complicated example, though.

### DIFF
--- a/src/systems/diffeqs/abstractodesystem.jl
+++ b/src/systems/diffeqs/abstractodesystem.jl
@@ -1366,7 +1366,21 @@ function InitializationProblem{iip, specialize}(sys::AbstractSystem,
     end
 
     u0map = merge(ModelingToolkit.guesses(sys), todict(guesses), todict(u0map))
-    u0map = Dict(diff2term(var) => val for (var, val) in u0map) # replace D(x) -> x_t etc.
+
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
Actually no, this can be unsafe. Instead of generating a new variable (which might not match the other if metadata is applied), you'd want to do it the transform the same way the initialization problem does it.

https://github.com/SciML/ModelingToolkit.jl/blob/master/src/systems/nonlinear/initializesystem.jl#L51

Though I am a bit confused, u0map is already processed right there?

ChrisRackauckas:
It's grabbing the variable itself that already exists, rather than making a new one. The new one is a new symbol which is not guaranteed to match the actual dummy derivative.

hersle:
Ok, thanks. Is this better?

hersle:
> Though I am a bit confused, u0map is already processed right there?

But yes, it is probably better to do this higher up in the call hierarchy, in case other functions also depend on the dummy derivative substitution being performed? Where would that be?

ChrisRackauckas (approved):

PR #3340, opened 2025-01-17
Validate types of state variables in ODESystem/SDESystem construction

### LINKED ISSUE #2589: It is possible to designate non-default types for variables, which are then ignored.
Currently, it is possible to designate a non-default type for variables:
```julia

@parameters p d
@variables X(t)::Int64
eq = D(X) ~ p - d*X
@mtkbuild osys = ODESystem([eq], t)

u0 = [X => 1.0]
ps = [p => 1.0, d => 0.2]
oprob = ODEProblem(osys, u0, (0.0,10.0), ps)
oprob[X] isa Int64 # false
oprob[X] isa Float64 # true
sol = solve(oprob, Tsit5())
```

In this case, `Int64` does not make sense for an ODE variable. While it is internally converted to a `Float64`, there should be an error instead. Here, it is actually possible to set `u0 = [X => 1.5]`, and things proceed as if `X(t)::Int64` was not the case.

This is even the case if a type which does make sense in the context of ODEs is give. I.e. here we use a `Float32`, but internal conversion to `Int64` is still a thing.
```julia
@parameters p d
@variables X(t)::Float32
eq = D(X) ~ p - d*X
@mtkbuild osys = ODESystem([eq], t)

u0 = [X => 1.0]
ps = [p => 1.0, d => 0.2]
oprob = ODEProblem(osys, u0, (0.0,10.0), ps)
oprob[X] isa Float32 # false
oprob[X] isa Float64 # true
sol = solve(oprob, Tsit5())
```

While there are cases where integer variables make sense (e.g. Jump simulations), these are already handled internally (i.e. integer initial conditions are interpreted as giving integers throughout a jump simulation).

The easiest would be to throw an error when a user tries 
```julia
@variables X(t)::Int64
```
However, if this is not possible (because there are circumstances when one use variables for other stuff), then at least one should be thrown when the `ODESystem` (or another system) is created.

vyudu:
Fixes #2589 . There is a bit of refactoring here but the check is in `process_equations`, and a new SDESystem constructor has been written that calls it


vyudu:
@ChrisRackauckas I think this one's ready

ChrisRackauckas (approved):

PR #3350, opened 2025-01-23
fix: validate u0map and pmap

### LINKED ISSUE #3325: Nonsensical stuff is allowed in input paraemter maps
Probably most likley to hapen when a symbolic variable is accidentally renamed. However, it seem weird that nonsensical stuff is just dismissed, instead of throwing an error:
```julia
# using MTK
using ModelingToolkit

t = default_t()
D = default_time_deriv()

@variables X(t)
@parameters p d
eqs = [D(X) ~ p - d*X]

p = "I accidentally renamed p"
u0 = [X => 1.0]
ps = [p => 1.0, d => 0.5]
oprob = ODEProblem(osys, u0, (0.0, 1.0), ps) # ERROR: Some parameters are missing from the variable map.

@parameters p d
ps = [p => 1.0, d => 0.5, "Random stuff" => 3.0]
oprob = ODEProblem(osys, u0, (0.0, 1.0), ps) # Works.
```

vyudu:
Validate the keys of the u0map and pmap. Not sure if my check here is too restrictive, but basically anything that is not a symbol/symbolic corresponding to an unknown/parameter of the system will error here. Fixes #3325


src/systems/problem_utils.jl
```diff
@@ -684,12 +684,17 @@ function process_SciMLProblem(
 
     u0Type = typeof(u0map)
     pType = typeof(pmap)
-    _u0map = u0map
```
vyudu, line 687: _u0map and pmap seems to not be used anywhere

vyudu:
@ChrisRackauckas this is ready

ChrisRackauckas (approved):

PR #3354, opened 2025-01-24
Replace Expronicon with Moshi

visr:
Draft because this would require a release and new minimum compat on SciMLBase with https://github.com/SciML/SciMLBase.jl/pull/922.

This not only adapts to SciMLBase changes but also removes Expronicon from ModelingToolkit, where it was used for `InferredClock` and a match in `sampletime`.

This doesn't get rid of MLStyle, which is used for pattern matching as well, but perhaps it would make sense as a follow up to start using Moshi for that as well.


AayushSabharwal:
Yeah this is breaking, unfortunately

visr:
What is breaking exactly? I know the tests are failing in this draft PR because this needs https://github.com/SciML/SciMLBase.jl/pull/922 first. And to avoid that breaking MTK I made https://github.com/SciML/ModelingToolkit.jl/pull/3353, which should be merged first. I was hoping this way it wouldn't be breaking.

AayushSabharwal:
Are Moshi.jl ADTs matchable via MLStyle?

visr:
Both Moshi and MLStyle support extensible pattern matching, so probably that can be made to work. Where is `MLStyle.@match` being called on Clock ADTs? In the MTK code I just see it called on expressions.

https://thautwarm.github.io/MLStyle.jl/latest/syntax/pattern.html#custom-patterns
https://rogerluo.dev/Moshi.jl/start/match/#extensible-pattern-matching

visr:
@AayushSabharwal would this also need a MLStyle package extension like https://github.com/SciML/SciMLBase.jl/pull/922, for this ADT?

```jl
@data InferredClock begin
    Inferred
    InferredDiscrete
end
```

Otherwise I updated this to depend on the new SciMLBase Moshi release.

AayushSabharwal:
No, this one is fine.

AayushSabharwal (approved):

PR #3364, opened 2025-02-03
feat: add `assertions` functionality

AayushSabharwal:
## Checklist

### DIFF
--- a/src/systems/diffeqs/odesystem.jl
+++ b/src/systems/diffeqs/odesystem.jl
@@ -137,7 +137,7 @@ struct ODESystem <: AbstractODESystem
     """
     parameter_dependencies::Vector{Equation}
     """
-    Mapping of conditions which should be true throughout the solve to corresponding error
+    Mapping of conditions which should be true throughout the solution process to corresponding error
     messages. These will be added to the equations when calling `debug_system`.
     """
     assertions::Dict{BasicSymbolic, String}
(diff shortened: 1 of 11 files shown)


src/systems/diffeqs/odesystem.jl
```diff
@@ -137,6 +137,11 @@ struct ODESystem <: AbstractODESystem
     """
     parameter_dependencies::Vector{Equation}
     """
+    Mapping of conditions which should be true throughout the solve to corresponding error
```
baggepinnen, line 140:
```suggestion
    Mapping of conditions which should be true throughout the solution process to corresponding error
```
"solve" is a verb

src/systems/diffeqs/sdesystem.jl
```diff
@@ -126,6 +126,11 @@ struct SDESystem <: AbstractODESystem
     """
     parameter_dependencies::Vector{Equation}
     """
+    Mapping of conditions which should be true throughout the solve to corresponding error
```
baggepinnen, line 129:
```suggestion
    Mapping of conditions which should be true throughout the solution process to corresponding error
```

baggepinnen (commented):
cool :)

AayushSabharwal:
I think formatter fails because it needs a rebase. Will do that once the PR is approved.

ChrisRackauckas:
This isn't causing a NaN in the non-debug? `isoutofdomain` could be used or NaN throws.

AayushSabharwal:
I thought we didn't want this in the non-debug? I can add it if we do

ChrisRackauckas:
You want a NaN but not an error, or you want isoutofdomain. Either way you need step rejections

ChrisRackauckas (approved):

PR #3379, opened 2025-02-10
Renaming discrete variables during structural simplification (plus refactoring `tearing_reassemble`)

vyudu:
This PR does two things: 
1. Previously the difference equations for discrete systems were things like `Shift(t, 1)(Shift(t, -1)(x(t))`. This PR changes it to follow the ODE convention of creating a dummy variable `x_{t-1}`, such that the equation is `Shift(t, 1)(x_{t-1}(t))`
2. `tearing_reassemble` has been chunked a bit. Have added some comments to `SystemStructure` and `TearingState` as well


src/systems/systemstructure.jl
```diff
@@ -140,16 +140,20 @@ get_fullvars(ts::TransformationState) = ts.fullvars
 has_equations(::TransformationState) = true
 
 Base.@kwdef mutable struct SystemStructure
-    # Maps the (index of) a variable to the (index of) the variable describing
-    # its derivative.
+    """Maps the index of variable x to the index of variable D(x)."""
     var_to_diff::DiffGraph
+    """Maps the index of an equation."""
```
vyudu, line 145: Not sure how to describe this one (or what it is really)

vyudu:
@AayushSabharwal

src/systems/systemstructure.jl
```diff
@@ -140,16 +140,20 @@ get_fullvars(ts::TransformationState) = ts.fullvars
...
     eq_to_diff::DiffGraph
     # Can be access as
     # `graph` to automatically look at the bipartite graph
     # or as `torn` to assert that tearing has run.
+    """Graph that connects equations to variables that appear in them."""
```
AayushSabharwal, line 150:
```suggestion
    """Graph that connects variables to equations that they appear in."""
```

Sources are variables and destinations are equations iirc?

src/systems/systemstructure.jl
AayushSabharwal: Some equations are differentiated during simplification. This maps equations to their differentiated forms.

src/systems/systemstructure.jl
AayushSabharwal: Never mind, it's equations to variables.

vyudu:
@AayushSabharwal @ChrisRackauckas this one is ready to review

ChrisRackauckas:
I think this looks fine, but I'm not entirely sure what makes this required?

src/structural_transformation/utils.jl
```diff
@@ -449,13 +449,39 @@ end
...
+    op = operation(var)
+    op isa Shift || return Shift(iv, 0)(var, true) # hack to prevent simplification of x(t) - x(t)
+    if op.steps < 0
+        return shift2term(var)
+    else
+        return var
+    end
+end
+
+function shift2term(var)
```
AayushSabharwal, line 463: Please add docstrings for both of these functions. Even one-liners help.

src/structural_transformation/utils.jl
```diff
@@ -466,6 +492,7 @@ end
 function simplify_shifts(var)
     ModelingToolkit.hasshift(var) || return var
     var isa Equation && return simplify_shifts(var.lhs) ~ simplify_shifts(var.rhs)
+    (op = operation(var)) isa Shift && op.steps == 0 && return first(arguments(var))
     if isdoubleshift(var)
```
AayushSabharwal, line 496: Do we need the `isdoubleshift` branch if this PR aims at removing double shifts?

src/systems/discrete_system/discrete_system.jl
```diff
@@ -275,10 +275,10 @@ function shift_u0map_forward(sys::DiscreteSystem, u0map, defs)
     end
     for var in unknowns(sys)
         op = operation(var)
-        op isa Shift || continue
         haskey(updated, var) && continue
-        root = first(arguments(var))
-        haskey(defs, root) || error("Initial condition for $var not provided.")
+        root = getunshifted(var)
```
AayushSabharwal, line 279: `getunshifted` works for variables with the `VariableUnshifted` metadata, which comes from `shift2term`. The user won't call `shift2term` beforehand. Instead, the defaults would likely contain `x(k-1) => (...)`. How is this case handled?

AayushSabharwal (commented):
Overall the refactor is great for code quality. However, since it's aimed at code quality I've been picky about wanting docstrings for everything. In addition, some parts are not clear or seem to have bugs which need to be fixed. It would also be great to get tests to make sure symbolic indexing works properly for both `Shift(t, -1)(x)` and `x_{k-1}` form of variables.

src/structural_transformation/utils.jl
```diff
@@ -449,13 +449,39 @@ end
...
     if ModelingToolkit.isoperator(var, ModelingToolkit.Shift)
-        op = operation(var)
-        return Shift(op.t, order)(var)
+        O = only(arguments(var))
+        oldop = operation(O)
+        newname = Symbol(string(nameof(oldop)), ds)
+    else
+        O = var
+        oldop = operation(var) 
+        varname = split(string(nameof(oldop)), d_separator)[1]
```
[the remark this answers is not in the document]
vyudu: Yeah this was a remnant I think, will remove

src/systems/discrete_system/discrete_system.jl
vyudu: Ah yeah this doesn't work. I'm adding a fix where it'll just generate the shift from the variable and check if that's in u0map (i.e. if x_{t-2} is the unknown, check if Shift(t, -1)(x(t)) is a key in updated)

src/structural_transformation/utils.jl
vyudu: If this is the only context in which they are used, that sounds good.

src/systems/discrete_system/discrete_system.jl
vyudu: A question: take the system defined by `x ~ x(k-1) + x(k-2)`. The unknowns here are `x(t), x_{t-1}(t)`. Do we want to support passing a u0map where `[x_{t-1} => 2, x_{t-2} => 1]` (i.e. both using lowered variables)? It is a little tricky since `x_{t-2}` isn't an unknown of the system, but at the same time it seems like natural syntax one should be able to write. CC also @baggepinnen

AayushSabharwal: It's not something someone would write, but it something that can foreseeably make its way into the constructor so it should be supported. As long as `DiscreteProblem` shifts the unknowns and hands them to `process_SciMLProblem` it will work. Typically, you don't want to figure out what order shift `x_{t-k}` is. Rather, you want to look at all the unknowns and find the one whose `default_toterm` matches `x_{t-k}`.

src/structural_transformation/utils.jl
AayushSabharwal: Yeah double shifts are only in the equations

src/systems/discrete_system/discrete_system.jl
vyudu: This makes sense, but come to think of it there is probably something a little weird here: for that system if they passed just `[x_{t-1} => 2]` and no assignment for `x_{t-2}`, it would error that the system does not have an initial value for the unknown `x_{t-1}`, since the assignment to `x_{t-1}` gets shifted to x. This seems kind of undesirable to me, mainly because it's confusing (though this happens in the current implementation as well; passing in just `x(k-1)` would error that no initial value is provided for `Shift(t, -1)(x(t)`, so maybe this concern isn't actually that important). 

Maybe we could uniformly disallow assignments to unshifted variables in u0map, and tell people to do either `[x(k-1) => 2, x(k-2) => 3]`, and `[x(k-1) => 2, x_{t-1}(k-1) => 1]` for the above? 

To get the `x_{t-2}` assignments to work does feel a bit hacky, since to shift it properly it would have to some nameparsing recover x and the shift, and then make the `Shift(t, -1)(x)`

AayushSabharwal: > This seems kind of undesirable to me, mainly because it's confusing 

It's how discrete systems work. To initialize a discrete system, you specify the value at `t = -1`. A value given to the unshifted variable applies to the entire history, and a value given to a shifted variable applies to that variable. If we disallow assignments to unshifted variables, defaults become meaningless.

`x_{t-1}(k-1)` is admittedly a bit cursed

AayushSabharwal: > Or should it only error if someone tries to pass [x => 2] into the u0map?

I forgot what we do here, but it seems best to keep the existing behavior. Defaults work for sure, but if we error when passing `x => 2` now we should continue to do so.

> `@variables x(t) x_{t-1}(t)`

It's fair to assume the user won't do this

AayushSabharwal: > Ah yeah I definitely need to implement this. But I think what happens in this case is that shift_u0map_forward turns the x_{t-1}(k-1) into x_{t-1} so this assignment just works. Actually this would probably make the x_{t-2} case work also since shift_u0map_forward can turn it into Shift(t, 1)(x_{t-2}) and then default_toterm can turn that into x_{t-1}.

I'm not sure if I understand everything going on here :D but if `default_toterm` handles it, symbolic indexing will work.

vyudu: So both these cases should work then right? 
```julia
@variables x(t) x_{t-1}(t)
prob = DiscreteProblem(sys, [x_{t-1} => 2, x_{t-2} => 1], tspan) # should this error? 
prob = DiscreteProblem(sys, [x(k-1) => 2, x_{t-1}(k-1) => 1], tspan)
prob.u0 == [2, 3]
```

AayushSabharwal: Yes, both of those should work.

vyudu: What about case 4? Should it error or overwrite the value for x_{t-1}? Worth noting that the ODE case works (i.e. passing [x => 1, x => 2] will set x to be 2) but that also might be a bit weird?

vyudu: OK, assuming tests pass then I think the initialization stuff is handling correctly.

src/structural_transformation/utils.jl
vyudu: It seems like the changes in this PR still relies on this (to simplify terms like Shift(t, 1)(Shift(t, 0)(x(t)) so will leave it for now

vyudu:
This should be good to merge.

AayushSabharwal (approved):
Needs a rebase and format, but otherwise looks great

vyudu:
Done

AayushSabharwal (approved):

PR #3381, opened 2025-02-11
Do not attempt to promote to Union

baggepinnen:
Sometimes `symtype` returns a union like `Union{Float64, Int64}`
```julia
julia> @variables x::Float64
1-element Vector{Num}:
 x

julia> ModelingToolkit.symtype(ifelse(x > 0, x, 0))
Union{Float64, Int64}
```

and then 
```julia
buf = get!(() -> Any[], cachevars, T)
```
fails.

This is (very) likely to occur with JSML, which appears to type things with `Float64`

## Checklist

### DIFF
--- a/src/systems/nonlinear/nonlinearsystem.jl
+++ b/src/systems/nonlinear/nonlinearsystem.jl
@@ -712,6 +712,9 @@ function SciMLBase.SCCNonlinearProblem{iip}(sys::NonlinearSystem, u0map,
             k = unwrap(k)
             v = unwrap(v)
             T = symtype(k)
+            while T isa Union
+                T = promote_type(T.a, T.b)
+            end
             buf = get!(() -> Any[], cachevars, T)
             push!(buf, v)
             buf = get!(() -> Any[], cacheexprs, T)


ChrisRackauckas:
> This is (very) likely to occur with JSML, which appears to type things with Float64

It shouldn't since that's just wrong.

AayushSabharwal:
Yeah `::Float64` annotations might cause problems with putting duals in some places.

AayushSabharwal:
Also, isn't `ifelse(x > 0, x, 0)` inherently type-unstable? Regardless of symbolic stuff it would be better to write `ifelse(x > 0, x, 0.0)` anyway (or `zero(x)`).

baggepinnen:
It is type unstable, but a component definition like this
```
@component function Ramp(; name, start_time=0, offset=0, duration=nothing, height=nothing)
  params = @parameters begin
    (start_time::Float64 = start_time)
    (offset::Float64 = offset)
    (duration::Float64 = duration)
    (height::Float64 = height)
  end
  vars = @variables begin
    y(t), [output = true]
  end
  eqs = Equation[
    y ~ ifelse(start_time < t, ifelse(t < start_time + duration, offset + (t - start_time) * height / duration, offset + height), offset)
  ]
  return ODESystem(eqs, t, vars, params; systems = [], name)
end
```
appears to result in this error where the default values for the parameters, `0::Int`, somehow appears inside the `ifelse` expression despite the declaration `start_time::Float64`

AayushSabharwal:
If a constant appears inside the expression instead of the symbolic expression, that's a much bigger problem with a different solution 😅 since that means the model is incorrectly parsed and changing `offset` in the `ODEProblem` won't actually change the `ifelse` argument.

PR #3404, opened 2025-02-21
feat: run trivial initialization in problem constructor

### LINKED ISSUE #3290: Trivial initialization problem isn't solved when creating `ODEProblem`
```julia
@variables Y(t)
@parameters d p = Y [guess = 1.0]
eqs = [D(Y) ~ p - d*Y]
u0 = [Y => 1.0]
p = [d => 0.2]
oprob = ODEProblem(osys, u0, (0.0, 1.0), p)
```
In the REPL:
```julia
julia> oprob[osys.Y]
1.0
julia> oprob.ps[osys.p]
0.0
```
This is because while the initialization is trivial:
```julia
julia> oprob.f.initialization_data.initializeprob
NonlinearProblem with uType Nothing. In-place: true
u0: nothing

julia> oprob.f.initialization_data.initializeprob.f.sys |> observed
2-element Vector{Equation}:
 Y(t) ~ 1.0
 p ~ Y(t)
```

MTK doesn't utilize this the way `remake` does and solve the problem on the spot. We can fix this by simply calling `remake(prob)` before returning.

AayushSabharwal:
Close #3290, #3410

### DIFF
--- a/src/systems/diffeqs/abstractodesystem.jl
+++ b/src/systems/diffeqs/abstractodesystem.jl
@@ -759,6 +759,7 @@ function DiffEqBase.ODEProblem{iip, specialize}(sys::AbstractODESystem, u0map = 
         kwargs1 = merge(kwargs1, (; tstops))
     end
 
+    # Call `remake` so it runs initialization if it is trivial
     return remake(ODEProblem{iip}(f, u0, tspan, p, pt; kwargs1..., kwargs...))
 end
 get_callback(prob::ODEProblem) = prob.kwargs[:callback]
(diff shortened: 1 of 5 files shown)


src/systems/diffeqs/abstractodesystem.jl
```diff
@@ -753,7 +753,7 @@ function DiffEqBase.ODEProblem{iip, specialize}(sys::AbstractODESystem, u0map =
         kwargs1 = merge(kwargs1, (; tstops))
     end
 
-    return ODEProblem{iip}(f, u0, tspan, p, pt; kwargs1..., kwargs...)
+    return remake(ODEProblem{iip}(f, u0, tspan, p, pt; kwargs1..., kwargs...))
```
ChrisRackauckas, line 756: this should at least be commented why it's done. It's a bit of an odd way to get there, but understand why

PR #3433, opened 2025-02-28
feat: add `isInitial`

### LINKED ISSUE #3431: isInitial function
It would be nice to have an official API function to detect if a symbolic is an `Initial(X)` for some symbolic `X`. We need this in places to filter out such variables.

I know we can use something like the following to detect and remove them from a vector
```julia
function remove_inits(ps)
    filter(x -> !iscall(x) || !isa(operation(x), Initial), ps)
end
```
but it would be nice to have an official function in case the internal representation for them changes at some point.

vyudu:
@ChrisRackauckas close #3431


ChrisRackauckas:
Missing docs

baggepinnen:
`isinitial` to follow convention?

vyudu:
Should be ready now

Issue #1561, opened 2022-05-06
v8.9.0 breaks mixed types using `varmap_to_vars`


bradcarman:
The following code fails with v8.9.0

```
using ModelingToolkit

vars = @variables x y
ModelingToolkit.varmap_to_vars([x => 0.0, y => 1.0], vars) #this is OK
ModelingToolkit.varmap_to_vars([x => 0.0, y => [0.0, 1.0]], vars) #error
```

The error message is:

```
ERROR: LoadError: MethodError: Cannot `convert` an object of type Vector{Float64} to an object of type Float64
Closest candidates are:
  convert(::Type{N}, ::DomainSets.Point{<:Number}) where N<:Number at C:\Users\CarmanBr\.julia\packages\DomainSets\F4hmL\src\domains\point.jl:14
  convert(::Type{T}, ::AbstractChar) where T<:Number at C:\Programs\julia-1.8.0\share\julia\base\char.jl:185
  convert(::Type{T}, ::MultivariatePolynomials.AbstractPolynomialLike) where T<:Number at C:\Users\CarmanBr\.julia\packages\MultivariatePolynomials\1bIGc\src\conversion.jl:77
  ...
Stacktrace:
  [1] _broadcast_getindex_evalf
    @ .\broadcast.jl:670 [inlined]
  [2] _broadcast_getindex
    @ .\broadcast.jl:653 [inlined]
  [3] getindex
    @ .\broadcast.jl:597 [inlined]
  [4] copyto_nonleaf!(dest::Vector{Float64}, bc::Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{1}, Tuple{Base.OneTo{Int64}}, typeof(convert), Tuple{Base.RefValue{Type{Float64}}, Base.Broadcast.Extruded{Vector{Any}, Tuple{Bool}, Tuple{Int64}}}}, iter::Base.OneTo{Int64}, state::Int64, count::Int64)
    @ Base.Broadcast .\broadcast.jl:1055
  [5] copy
    @ .\broadcast.jl:907 [inlined]
  [6] materialize(bc::Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{1}, Nothing, typeof(convert), Tuple{Base.RefValue{Type{Float64}}, Vector{Any}}})
    @ Base.Broadcast .\broadcast.jl:860
[...]

PR #3448, opened 2025-03-07
fix: propagate `tofloat`, `use_union` to `better_varmap_to_vars`

### LINKED ISSUE #3446: Jump simulations are `Float64` valued
When you run jump simulations the type of the values are now `Float64` (even if initialised with `Int64`). These are integer-valued simulations, and should have integer values.
```julia
using ModelingToolkit, JumpProcesses

@parameters p d
@variables X(t)
rate1   = p
rate2   = X*d
affect1 = [X ~ X + 1]
affect2 = [X ~ X - 1]
j1 = ConstantRateJump(rate1, affect1)
j2 = ConstantRateJump(rate2, affect2)

# Works.
@mtkbuild js = JumpSystem([j1, j2], t, [X], [p,d])
dprob = DiscreteProblem(js, [X => 15], (0.0, 10.), [p => 2.0, d => 0.5])
sol = solve(jprob, SSAStepper())
sol[X] # Returns `Vector{Float64}`.
```

vyudu:
Close #3446

### DIFF
--- a/src/systems/problem_utils.jl
+++ b/src/systems/problem_utils.jl
@@ -814,7 +814,7 @@ function process_SciMLProblem(
     evaluate_varmap!(op, dvs; limit = substitution_limit)
 
     u0 = better_varmap_to_vars(
-        op, dvs; tofloat = true, use_union = false,
+        op, dvs; tofloat, use_union,
         container_type = u0Type, allow_symbolic = symbolic_u0)
 
     if u0 !== nothing
@@ -848,7 +848,7 @@ function process_SciMLProblem(
         du0map = to_varmap(du0map, ddvs)
         merge!(op, du0map)
         du0 = varmap_to_vars(op, ddvs; toterm = identity,
-            tofloat = true)
+            tofloat)
         kwargs = merge(kwargs, (; ddvs))
     else
         du0 = nothing
--- a/test/jumpsystem.jl
+++ b/test/jumpsystem.jl
@@ -533,3 +533,21 @@ let
     Xsamp /= Nsims
     @test abs(Xsamp - Xf(0.2, p) < 0.05 * Xf(0.2, p))
 end
+
+@testset "JumpProcess simulation should be Int64 valued (#3446)" begin
+    @parameters p d
+    @variables X(t)
+    rate1   = p
+    rate2   = X*d
+    affect1 = [X ~ X + 1]
+    affect2 = [X ~ X - 1]
+    j1 = ConstantRateJump(rate1, affect1)
+    j2 = ConstantRateJump(rate2, affect2)
+
+    # Works.
+    @mtkbuild js = JumpSystem([j1, j2], t, [X], [p,d])
+    dprob = DiscreteProblem(js, [X => 15], (0.0, 10.), [p => 2.0, d => 0.5])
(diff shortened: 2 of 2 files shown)


ChrisRackauckas (approved):

ChrisRackauckas:
Looks like this is giving some test failures?

vyudu:
The test failures seem unrelated

vyudu:
I can't reproduce these failures locally

AayushSabharwal:
We should only do this propagation for JumpProblems. A union-typed `u0` is not valid for ODEs iirc which looks like the cause of this error.

vyudu:
Shouldn't it just error if someone passes `use_union = true` for an ODEProblem then? It doesn't make sense for someone to pass in `use_union = true` and it just gets ignored right

vyudu:
Or should it propagate for the `p` but not the `u0`?

AayushSabharwal:
Yeah it needs to propagate for the `p`

vyudu:
Is this good to merge?

isaacsas:
Is use_union really relevant or needed anymore? Before it was needed to avoid parameters being converted to integers but that is now handled by making them symbolic integers. All jump systems really need is to preserve integer initial conditions as integers, not as a union type, without requiring users to specify the symbolic variables as integers originally.

ChrisRackauckas:
Yes we should probably error for any unions. We created the split parameters to handle that, so if we end up with unions anywhere that's a sign something went wrong.

vyudu:
I can add that.

vyudu:
Actually we'd still need `use_union` to handle cases like https://github.com/SciML/ModelingToolkit.jl/issues/1561 right?

AayushSabharwal:
`varmap_to_vars` should remain untouched. We don't use it anyway, it's just for backward compat.

vyudu:
What should become of this test? https://github.com/SciML/ModelingToolkit.jl/blob/0ad773d2b90009fa76e8d9f6ad681f91b73e73f1/test/split_parameters.jl#L107-L128

Removing `use_union` would break it

src/systems/problem_utils.jl
```diff
@@ -909,15 +906,14 @@ end
 ##############
 
 """
-    u0, p, defs = get_u0_p(sys, u0map, parammap; use_union=true, tofloat=true)
```
AayushSabharwal, line 912: I don't think we want to touch anything in the legacy function header. It's not used internally, and only exists because its public API. I plan to remove it in v10.

test/split_parameters.jl
```diff
@@ -123,9 +123,9 @@ sol = solve(prob, ImplicitEuler());
 # ------------------------ Mixed Type Conserved
 
 prob = ODEProblem(
-    sys, [], tspan, []; tofloat = false, use_union = true, build_initializeprob = false)
+    sys, [], tspan, []; tofloat = false, build_initializeprob = false)
 
-@test prob.p isa Vector{Union{Float64, Int64}}
```
AayushSabharwal, line 128: At this point we can just remove this test, since the purpose is invalidated

AayushSabharwal (commented):
Couple minor things, and needs a format

AayushSabharwal (approved):

PR #3456, opened 2025-03-12
fix: Throw better warning when a guess is missing 

### LINKED ISSUE #3449: Better error message for missing guesses
Currently errors for missing guesses say "missing initial value" or "circular initial condition" (depending on where it originates). See:
```julia
ERROR: LoadError: Found symbolic value <<VALUE>> for variable <<VARIABLE>>. You may be missing an initial condition or have cyclic initial conditions. If this is intended, pass `symbolic_u0 = true`. In case the initial conditions are not cyclic but require more substitutions to resolve, increase `substitution_limit`. To report cycles in initial conditions of unknowns/parameters, pass `warn_cyclic_dependency = true`. If the cycles are still not reported, you may need to pass a larger value for `circular_dependency_max_cycle_length` or `circular_dependency_max_cycles`.


Stacktrace:
  [1] better_varmap_to_vars(varmap::Dict{Any, Any}, vars::Vector{SymbolicUtils.BasicSymbolic{Real}}; tofloat::Bool, use_union::Bool, container_type::Type, toterm::Function, promotetoconcrete::Nothing, check::Bool, allow_symbolic::Bool)
    @ ModelingToolkit ~/.julia/packages/ModelingToolkit/41wGH/src/systems/problem_utils.jl:340
  [2] process_SciMLProblem(constructor::Type, sys::ModelingToolkit.NonlinearSystem, u0map::Dict{Any, Any}, pmap::Dict{Any, Any}; build_initializeprob::Bool, implicit_dae::Bool, t::Nothing, guesses::Dict{Any, Any}, warn_initialize_determined::Bool, initialization_eqs::Vector{Any}, eval_expression::Bool, eval_module::Module, fully_determined::Nothing, check_initialization_units::Bool, tofloat::Bool, use_union::Bool, u0_constructor::typeof(identity), du0map::Nothing, check_length::Bool, symbolic_u0::Bool, warn_cyclic_dependency::Bool, circular_dependency_max_cycle_length::Int64, circular_dependency_max_cycles::Int64, substitution_limit::Int64, use_scc::Bool, kwargs::@Kwargs{})
    @ ModelingToolkit ~/.julia/packages/ModelingToolkit/41wGH/src/systems/problem_utils.jl:720
```

This error message should instead report a missing guess, since it is currently very misleading. The only way for a user to differentiate is via the stacktrace, which is bad for any number of reasons.

vyudu:
Fix #3449

### DIFF
--- a/src/systems/problem_utils.jl
+++ b/src/systems/problem_utils.jl
@@ -333,17 +333,14 @@ end
 function Base.showerror(io::IO, err::MissingGuessError) 
     println(io, 
             """
-            Unable to resolve numeric guesses for all of the variables in the system. \
-            This may be because your guesses are cyclic. In order for the problem to be \
-            initialized, all of the variables must have a numeric value to serve as a \
-            starting point for the nonlinear solve. 
-
-            Symbolic values were found for the following variables/parameters in the map; \
-            please provide additional numeric guesses so they can resolve to numbers:
+            Cyclic guesses detected in the system. Symbolic values were found for the following variables/parameters in the map: \
             """)
     for (sym, val) in zip(err.syms, err.vals)
-        println(sym, " => ", val)
+        println(io, "$sym  => $val")
     end
+    println(io,
+            """
+            In order to resolve this, please provide additional numeric guesses so that the chain can be resolved to assign numeric values to each variable.            """)
 end
 
 """
--- a/src/systems/problem_utils.jl
+++ b/src/systems/problem_utils.jl
@@ -333,9 +333,13 @@ end
 function Base.showerror(io::IO, err::MissingGuessError) 
     println(io, 
             """
-            The problem cannot be initialized without providing an additional numeric \
-            guess to serve as a starting point for solving for the initial state. Please \
-            provide another numeric value to `guesses` in the problem constructor.
+            Unable to resolve numeric guesses for all of the variables in the system. \
+            This may be because your guesses are cyclic.
+
+            In order for the problem to be initialized, all of the variables must have \
+            a numeric value to serve as a starting point for the nonlinear solve. \
+            Please provide one or more additional numeric guesses to `guesses` in \
+            the problem constructor.
 
             This error was thrown because symbolic value $(err.val) was found for variable $(err.sym).
             """)
@@ -358,7 +362,7 @@ Keyword arguments:
   [`missingvars`](@ref) to perform the check.
 - `allow_symbolic` allows the returned array to contain symbolic values. If this is `true`,
   `promotetoconcrete` is set to `false`.
-- `is_initializeprob`: Whether the parent problem is an initialization problem.
+- `is_initializeprob, guesses`: Used to determine whether the system is missing guesses.
 """
 function better_varmap_to_vars(varmap::AbstractDict, vars::Vector;
         tofloat = true, use_union = true, container_type = Array,
--- a/src/systems/problem_utils.jl
+++ b/src/systems/problem_utils.jl
@@ -338,7 +338,7 @@ function Base.showerror(io::IO, err::MissingGuessError)
 
             In order for the problem to be initialized, all of the variables must have \
             a numeric value to serve as a starting point for the nonlinear solve. \
-            Please provide one or more additional numeric guesses to `guesses` in \
+            Please provide an additional numeric guess to `guesses` in \
             the problem constructor.
 
             This error was thrown because symbolic value $(err.val) was found for variable $(err.sym).
--- a/src/systems/problem_utils.jl
+++ b/src/systems/problem_utils.jl
@@ -326,23 +326,24 @@ function Base.showerror(io::IO, err::UnexpectedSymbolicValueInVarmap)
 end
 
 struct MissingGuessError <: Exception
-    sym::Any
-    val::Any
+    syms::Vector{Any}
+    vals::Vector{Any}
 end
 
 function Base.showerror(io::IO, err::MissingGuessError) 
     println(io, 
             """
             Unable to resolve numeric guesses for all of the variables in the system. \
-            This may be because your guesses are cyclic.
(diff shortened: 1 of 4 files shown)


test/initial_values.jl
```diff
@@ -172,7 +172,7 @@ end
             [p => 2q, q => 3p]; warn_cyclic_dependency = true)
     catch
     end
-    @test_throws ModelingToolkit.UnexpectedSymbolicValueInVarmap ODEProblem(
+    @test_throws ModelingToolkit.MissingGuessError ODEProblem(
```
vyudu, line 175: This is one of those cases that I'm not sure about, should this throw the MissingGuessError?

src/systems/problem_utils.jl
```diff
@@ -325,6 +325,26 @@ function Base.showerror(io::IO, err::UnexpectedSymbolicValueInVarmap)
...
+            """
+            Unable to resolve numeric guesses for all of the variables in the system. \
+            This may be because your guesses are cyclic.
+
+            In order for the problem to be initialized, all of the variables must have \
+            a numeric value to serve as a starting point for the nonlinear solve. \
+            Please provide one or more additional numeric guesses to `guesses` in \
+            the problem constructor.
+
+            This error was thrown because symbolic value $(err.val) was found for variable $(err.sym).
```
ChrisRackauckas, line 344: only one?

test/initial_values.jl
ChrisRackauckas: Yes, p and q would be missing and need a guess.

src/systems/problem_utils.jl
SebastianM-C: I think the message is a bit confusing since one could interpret it as asking for more than one guess for a single variable.

vyudu: Ah okay, will change

vyudu:
Assuming test pass I think this is good

src/systems/problem_utils.jl
```diff
@@ -325,6 +325,27 @@ function Base.showerror(io::IO, err::UnexpectedSymbolicValueInVarmap)
...
+            Unable to resolve numeric guesses for all of the variables in the system. \
+            This may be because your guesses are cyclic. In order for the problem to be \
+            initialized, all of the variables must have a numeric value to serve as a \
+            starting point for the nonlinear solve. 
+
+            Symbolic values were found for the following variables/parameters in the map; \
+            please provide additional numeric guesses so they can resolve to numbers:
+            """)
+    for (sym, val) in zip(err.syms, err.vals)
+        println(sym, " => ", val)
```
ChrisRackauckas, line 345: If it's just missing, what's the val?

vyudu: The symbolic value that it tries to assign to the variable

ChrisRackauckas: That's if it's cyclic. But this also is for if a guess is just missing, right? In which case, what does the error look like?

vyudu: If a guess it just missing it throws a different error (IncompleteInitializationError)

ChrisRackauckas:
```julia
    @parameters g
    @variables x(t) y(t) [state_priority = 10] λ(t)
    eqs = [D(D(x)) ~ λ * x
           D(D(y)) ~ λ * y - g
           x^2 + y^2 ~ 1]
    @mtkbuild pend = ODESystem(eqs, t)

    @test_throws ModelingToolkit.MissingGuessError ODEProblem(pend, [x => 1], (0, 1), [g => 1], guesses = [y => λ])
```

where it's missing a guess for `λ`, what does this print?

src/systems/problem_utils.jl
vyudu: https://github.com/SciML/ModelingToolkit.jl/blob/0ad773d2b90009fa76e8d9f6ad681f91b73e73f1/src/systems/diffeqs/abstractodesystem.jl#L1345-L1353

ChrisRackauckas: That error doesn't make much sense to the user though? Why wouldn't the case I mentioned be a missing guess error?

vyudu:
ERROR: Unable to resolve numeric guesses for all of the variables in the system. This may be because your guesses are cyclic. In order for the problem to be initialized, all of the variables must have a numeric value to serve as a starting point for the nonlinear solve.

Symbolic values were found for the following variables/parameters in the map; please provide additional numeric guesses so they can resolve to numbers:

y(t) => 50 + λ(t)
λ(t) => 2551 + y(t)

ChrisRackauckas:
```
y(t) => 50 + λ(t)
λ(t) => 2551 + y(t)
```

how does it get that? That doesn't look right.

What does it print for:

```julia
    @parameters g
    @variables x(t) y(t) [state_priority = 10] λ(t)
    eqs = [D(D(x)) ~ λ * x
           D(D(y)) ~ λ * y - g
           x^2 + y^2 ~ 1]
    @mtkbuild pend = ODESystem(eqs, t)

    ODEProblem(pend, [x => 1], (0, 1), [g => 1])
```

?

src/systems/problem_utils.jl
vyudu: there is also this error but I need to dig a bit to figure out when it throws IncompleteInitializationError vs this one, maybe @AayushSabharwal knows better
https://github.com/SciML/ModelingToolkit.jl/blob/0ad773d2b90009fa76e8d9f6ad681f91b73e73f1/src/variables.jl#L210-L214

vyudu:
Sorry, the first case you posted actually prints this, my output was from the test I wrote

      Initial condition underdefined. Some are missing from the variable map.
      for the following variables:

      λ(t)

ChrisRackauckas:
Okay so if any are missing you get that message, and if it's cyclic you get this other method. Given that, can we refine the error message? We know it's not a missing guess and instead a cyclic case, so we should make this more explicit.

PR #3462, opened 2025-03-15
add toplevel arguments to various getters

TorkelE:
Needs tests and docs, and to check if there are any other getters that should have a similar arguments.

But figured I'd upload this first in case people had input/opinions about the approach.

### DIFF
--- a/docs/src/basics/AbstractSystem.md
+++ b/docs/src/basics/AbstractSystem.md
@@ -69,7 +69,7 @@ Optionally, a system could have:
 Note that if you know a system is an `AbstractTimeDependentSystem` you could use `get_iv` to get the
 unique independent variable directly, rather than using `independent_variables(sys)[1]`, which is clunky and may cause problems if `sys` is an `AbstractMultivariateSystem` because there may be more than one independent variable. `AbstractTimeIndependentSystem`s do not have a method `get_iv`, and `independent_variables(sys)` will return a size-zero result for such. For an `AbstractMultivariateSystem`, `get_ivs` is equivalent.
 
-For the `parameters`, `unknowns`, `continuous_events`, and `discrete_events` accessors there are corresponding `parameters_toplevel`, `unknowns_toplevel`, `continuous_events_toplevel`, and `discrete_events_toplevel` accessors which works similarly, but ignores the content of subsystems. Furthermore, a `equations_toplevel` version of `equations` exists as well, however, it can only be applied to non-complete systems.
+For the `parameters`, `unknowns`, `continuous_events`, and `discrete_events` accessors there are corresponding `parameters_toplevel`, `unknowns_toplevel`, `continuous_events_toplevel`, and `discrete_events_toplevel` accessors which work similarly, but ignore the content of subsystems. Furthermore, a `equations_toplevel` version of `equations` exists as well, however, it can only be applied to non-complete systems.
 
 A system could also have caches:
 
--- a/docs/src/basics/AbstractSystem.md
+++ b/docs/src/basics/AbstractSystem.md
@@ -69,6 +69,8 @@ Optionally, a system could have:
 Note that if you know a system is an `AbstractTimeDependentSystem` you could use `get_iv` to get the
 unique independent variable directly, rather than using `independent_variables(sys)[1]`, which is clunky and may cause problems if `sys` is an `AbstractMultivariateSystem` because there may be more than one independent variable. `AbstractTimeIndependentSystem`s do not have a method `get_iv`, and `independent_variables(sys)` will return a size-zero result for such. For an `AbstractMultivariateSystem`, `get_ivs` is equivalent.
 
+For the `parameters`, `unknowns`, `continuous_events`, and `discrete_events` accessors there are corresponding `parameters_toplevel`, `unknowns_toplevel`, `continuous_events_toplevel`, and `discrete_events_toplevel` accessors which works similarly, but ignores the content of subsystems. Furthermore, a `equations_toplevel` version of `equations` exists as well, however, it can only be applied to non-complete systems.
+
 A system could also have caches:
 
   - `get_jac(sys)`: The Jacobian of a system.
--- a/src/systems/abstractsystem.jl
+++ b/src/systems/abstractsystem.jl
@@ -1342,8 +1342,8 @@ function parameters(sys::AbstractSystem; initial_parameters = false)
         ps = first.(ps)
     end
     systems = get_systems(sys)
-    result = unique(isempty(systems) ?
-                    ps : [ps; reduce(vcat, namespace_parameters.(systems))])
+    result = unique(isempty(systems) ? ps :
+                    [ps; reduce(vcat, namespace_parameters.(systems))])
     if !initial_parameters
         if is_time_dependent(sys)
             # time-dependent systems have `Initial` parameters for all their
@@ -1573,7 +1573,6 @@ It may include some abbreviations and aliases of observables.
 It is often the most useful way to inspect the equations of a system.
 
 See also [`full_equations`](@ref) and [`ModelingToolkit.get_eqs`](@ref).
-
 """
 function equations(sys::AbstractSystem)
     eqs = get_eqs(sys)
--- a/src/systems/abstractsystem.jl
+++ b/src/systems/abstractsystem.jl
@@ -1292,15 +1292,11 @@ $(TYPEDSIGNATURES)
 Get the unknown variables of the system `sys` and its subsystems.
 
 See also [`ModelingToolkit.get_unknowns`](@ref).
-
-Arguments:
-- `toplevel = false`: if set to true, do not return the continuous events of the subsystems.
 """
-function unknowns(sys::AbstractSystem; toplevel = false)
-    toplevel && (sys = recursive_get_parent(sys))
+function unknowns(sys::AbstractSystem)
     sts = get_unknowns(sys)
     systems = get_systems(sys)
-    nonunique_unknowns = if toplevel || isempty(systems)
+    nonunique_unknowns = if isempty(systems)
         sts
     else
         system_unknowns = reduce(vcat, namespace_variables.(systems))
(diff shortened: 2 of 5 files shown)


src/systems/callbacks.jl
```diff
@@ -327,14 +327,15 @@ The `SymbolicContinuousCallback`s in the returned vector are structs with two fi
...
-    obs = get_continuous_events(sys)
-    filter(!isempty, obs)
+function continuous_events(sys::AbstractSystem; toplevel = false)
+    cbs = get_continuous_events(sys)
+    filter(!isempty, cbs)
+    toplevel && return cbs
 
     systems = get_systems(sys)
-    cbs = [obs;
+    cbs = [cbs;
```
TorkelE, line 336: I think this is a leftover from the code being copied from the `observed` getter (?), in which case it would make sense to use `cbs` and not `obs` throughout the function.

AayushSabharwal (commented):
If the system is flattened (or completed), this approach won't work since e.g. `get_unknowns` contains all the unknowns of the system. This can be handled by doing e.g.

```julia
function top_level_unknowns(sys::AbstractSystem)
    if has_parent(sys) && (p = get_parent(sys)) !== nothing
        return top_level_unknowns(p)
    end
    return get_unknowns(sys)
end
```

TorkelE:
Was mostly really thinking of an modified accessor of the actual field. However, handling flattened systems like this does make lots of sense (as I think otherwise the §toplevel§ argument would be irrelevant anyway).

Will modify, thanks for the code snippet!

TorkelE:
@AayushSabharwal For systems where I have thrown a structural simplify I am a bit unsure how to adapt your approach.

I.e. if you want to do `observed(sys; top-level = true)` on a system which got observables due to structural simplify, going back to the original parent seem to go back to the pre-ss parent, which does not have an observable.

AayushSabharwal:
Yes. As we discussed in an earlier thread, `observed` equations are a result of structural simplification which does not retain the hierarchical structure. They don't exist in the hierarchical system.

TorkelE:
So would it make sense to just disable `toplevel` for observables?

There is a slightly similar phenomenon for `equations`, where `equations(sys; top-level = false)` and `equations(sys; top-level = true)` give different results. I.e. since accesses the parent parent (and so on) system, it _will_ return equations that was eliminated as observables by `structural_simplify`. Meanwhile, `equations(sys; toplevel = false)`  will not returns these. It feels wrong that adding `top-level = true` might cause the _addition_ of equations to the output.

(generally, I do think support for some form of non-ss observables is important, as the current approach kind of skrew up jump simulations hard, but that is a discussion for another place)

AayushSabharwal:
> So would it make sense to just disable toplevel for observables?

Yes. And yeah equations basically suffer the same fate. They don't "belong" to a subsystem anymore.

TorkelE:
Got it, I will remove the `toplevel` option for these then. Alternatively, I could have it error for non-complete systems or something like that? It would be useful to have access to these n as many cases as possible.

 It would make it possible to still use these when constructing things of MTK. I.e. a problem with a lot of things now being considered extra MTK internal is that while this makes sense for direct users, packages like Catalyst that builds on MTK kind of need these kind of things to construct the models we do.

AayushSabharwal:
@ChrisRackauckas should we allow passing `observed` for non-structurally-simplified systems as experimental functionality. This would allow Catalyst to eliminate variables.

@TorkelE In general with simplification we can't allow user-provided observed equations. At best during simplification we club them together with the rest of the equations but this won't guarantee they're re-eliminated, so there really isn't a benefit to not putting them with the rest of the equations.

TorkelE:
Makes sense. Maybe at some point we try to get as many of us as possible together and discuss through what things Catalyst needs/don't need and the best way to get it from MTK.

This one is finished sans:
- Some minor documentation.
- Deciding whether to fully remove `toplevel` argument from `observed` and `equations`, or whether there are some subset of systems we can use them for. For `observed` it probably should be removed completely as it don't necessarily make sense for non structurally simplified systems, and for simplified ones the output is wrong. For equations it might still be possible to give it a role for non-structurally simplieifed (or incomplete) systems, if you guys think that makes sense.

TorkelE:
> > Got it, I will remove the toplevel option for these then. Alternatively, I could have it error for non-complete systems or something like that? It would be useful to have access to these n as many cases as possible.
> 
> For `observed` toplevel doesn't make sense. For `equations`, you could keep it if it errors for complete systems (why would it error for non-complete systems?). I also don't fully understand why you need these but it isn't difficult to maintain so we might as well. Can we separate them out as a different set of functions? The keyword effectively completely changes the meaning of the function so I don't think it makes sense to club them together.

The reason that I ideally would like to keep it as a kwarg is that it means that I don't have to copy lots of code over, and if someone does something to the main function (i.e. exception for a new MTK-added parameter), those changes will automatically get incorporated without someone having to keep track of these extra functions.

AayushSabharwal:
> if you guys think that makes sense.

It only makes sense for un-flattened systems. But whether the function itself is useful depends on what you need it for. MTK itself (and other uses I'm aware of) don't really need this functionality.

> Maybe at some point we try to get as many of us as possible together and discuss through what things Catalyst needs/don't need and the best way to get it from MTK.

Then shouldn't we wait before merging this PR? Adding features we later might deem unnecessary only adds to the maintenance burden and restricts semver.

> The reason that I ideally would like to keep it as a kwarg is that it means that I don't have to copy lots of code over, and if someone does something to the main function (i.e. exception for a new MTK-added parameter), those changes will automatically get incorporated without someone having to keep track of these extra functions.

TorkelE:
> It only makes sense for un-flattened systems. But whether the function itself is useful depends on what you need it for. MTK itself (and other uses I'm aware of) don't really need this functionality.

Right now we use `get_eqs` quite frequently when building `ReactionSystem`s. However, since it does not have he issues `get_ps` have I haven't had to look at these cases closer. However, it would be nice to at least have something future proofing ourselves.

> Then shouldn't we wait before merging this PR? Adding features we later might deem unnecessary only adds to the maintenance burden and restricts semver.

I was talking in regards to `observed`. I agree that right now we should remove that part of the PR.

> The reason that I ideally would like to keep it as a kwarg is that it means that I don't have to copy lots of code over, and if someone does something to the main function (i.e. exception for a new MTK-added parameter), those changes will automatically get incorporated without someone having to keep track of these extra functions.

So your code uses ` get_unknowns(sys)` which returns the content of the field, which means that we are creating something that the user potentially shouldn't use (but maybe I am still confused). I.e. for parameters the field contain all of those initialisation parameters, and I am trying to recreate the functionality of e.g. `parameters` but for only the top-level system. So I need it to go through the bit in `parameters` where you filter this kind of stuff out.

AayushSabharwal:
> for parameters the field contain all of those initialisation parameters, and I am trying to recreate the functionality of e.g. parameters but for only the top-level system. So I need it to go through the bit in parameters where you filter this kind of stuff out.

That's exactly what I'm referring to. These initialization parameters are added to the simplified/completed system and not to the user-provided system. In general, all operations on systems are out-of-place. This is why the approach in my snippet won't have any extra parameters.

AayushSabharwal:
> Right now we use get_eqs quite frequently when building ReactionSystems. 

Why? And do you do it on completed systems or unmodified ones?

TorkelE:
> That's exactly what I'm referring to. These initialization parameters are added to the simplified/completed system and not to the user-provided system. In general, all operations on systems are out-of-place. This is why the approach in my snippet won't have any extra parameters.

Ok, got it, will try this out and see if it works. Does that mean that `get_` and similar are still permitted for non-complete systems?

> Why? And do you do it on completed systems or unmodified ones?

Different in different cases.

TorkelE:
> Yep, it's a field accessor. Works on all systems. We use it in a bunch of places.

Got it.

TorkelE:
Ok, this one should be ready now.

TorkelE:
I have checked the spell check, and its marked errors are not related to this PR. The other errors also seem to also occur on other branches. It would be good if @AayushSabharwal  could have a look to check that things looks sound though.

test/accessor_functions.jl
```diff
@@ -0,0 +1,171 @@
...
+    # these are subset of what `get_ps` returns.
+    @test all_sets_equal(
+        parameters_toplevel.([sys_bot, sys_bot_comp, sys_bot_ss])..., [d, p_bot])
+    @test_broken all_sets_equal(
+        parameters_toplevel.([sys_mid1, sys_mid1_comp, sys_mid1_ss])...,
+        [d, p_mid1])
+    @test all_sets_equal(
+        parameters_toplevel.([sys_mid2, sys_mid2_comp, sys_mid2_ss])...,
+        [d, p_mid2])
+    @test_broken all_sets_equal(
```
AayushSabharwal, line 96: Why is this broken?

src/systems/abstractsystem.jl
```diff
@@ -1313,12 +1313,25 @@ function unknowns(sys::AbstractSystem)
...
+    return get_unknowns(sys)
+end
+
 """
 $(TYPEDSIGNATURES)
 
 Get the parameters of the system `sys` and its subsystems.
 
 See also [`@parameters`](@ref) and [`ModelingToolkit.get_ps`](@ref).
+
```
AayushSabharwal, line 1334:
```suggestion
```

docs/src/basics/AbstractSystem.md
```diff
@@ -69,6 +69,8 @@ Optionally, a system could have:
 Note that if you know a system is an `AbstractTimeDependentSystem` you could use `get_iv` to get the
 unique independent variable directly, rather than using `independent_variables(sys)[1]`, which is clunky and may cause problems if `sys` is an `AbstractMultivariateSystem` because there may be more than one independent variable. `AbstractTimeIndependentSystem`s do not have a method `get_iv`, and `independent_variables(sys)` will return a size-zero result for such. For an `AbstractMultivariateSystem`, `get_ivs` is equivalent.
 
+For the `parameters`, `unknowns`, `continuous_events`, and `discrete_events` accessors there are corresponding `parameters_toplevel`, `unknowns_toplevel`, `continuous_events_toplevel`, and `discrete_events_toplevel` accessors which works similarly, but ignores the content of subsystems. Furthermore, a `equations_toplevel` version of `equations` exists as well, however, it can only be applied to non-complete systems.
```
AayushSabharwal, line 72:
```suggestion
For the `parameters`, `unknowns`, `continuous_events`, and `discrete_events` accessors there are corresponding `parameters_toplevel`, `unknowns_toplevel`, `continuous_events_toplevel`, and `discrete_events_toplevel` accessors which work similarly, but ignore the content of subsystems. Furthermore, a `equations_toplevel` version of `equations` exists as well, however, it can only be applied to non-complete systems.
```

test/accessor_functions.jl
TorkelE: Because https://github.com/SciML/ModelingToolkit.jl/issues/3464 makes `parameters` return an internal parameter `DEF`.

AayushSabharwal: The fix for that just got merged and tagged. Could you rebase and try?

TorkelE: will do

TorkelE:
Ok, this is ready now, again, the errors is things which seems to error on other branches (or at least they error on https://github.com/SciML/ModelingToolkit.jl/pull/3466 as well)

AayushSabharwal (approved):

ChrisRackauckas:
Add the docstrings to the doc?

ChrisRackauckas:
Forgot we hadn't updated this repo yet for the newer Documenter features.

TorkelE:
Right! Looking forwards to checking those out!

PR #3466, opened 2025-03-17
feat: handle `Initial(x)` initialization_eqs in time-independent systems

### LINKED ISSUE #3458: Errors and wrong solutions when solving `NonlinearSystem` with `initialization_eqs`
From the latest understanding this is what one should use with `structural_simplify`? Although, admittedly, I am not really sure anymore.

Either case, there is problem with the solution, generating a mix of errors and wrong outputs when you check the solution.

```julia
using ModelingToolkit, OrdinaryDiffEqDefault, Test

# Defines the model.
@parameters k1 k2
@variables X1(t) X2(t)
@parameters Γ[1:1] = missing [guess = [1.0]]
eqs = [
    0 ~ k1 * (Γ[1] - X1) - k2 * X1
]
initialization_eqs = [
    X2 ~ Γ[1] - X1
]
@mtkbuild nlsys = NonlinearSystem(eqs, [X1, X2], [k1, k2, Γ]; initialization_eqs)

# Creates the problem.
u0 = [X1 => 1.0, X2 => 2.0]
ps = [k1 => 0.1, k2 => 0.2]
nlprob = NonlinearProblem(nlsys, u0, ps)

# Solves the problem and check the solution.
sol = solve(nlprob)
sol[X1] # -0.0 (should be 1.0).
sol[X2] # ERROR: ArgumentError: Symbol X2(t) is not present in the system. (should be 2.0).
sol.ps[k1] # 0.1 (Correct.)
sol.ps[k2] # 0.2 (Correct.)
sol.ps[Γ[1]] # -0.0 (should be 3.0)
```

Either this is a bug, or I still do not know how you are supposed to do this.

AayushSabharwal:
Close #3458 
Requires https://github.com/SciML/DiffEqBase.jl/pull/1123


src/systems/problem_utils.jl
```diff
@@ -774,6 +790,7 @@ function process_SciMLProblem(
     op, missing_unknowns, missing_pars = build_operating_point!(sys,
         u0map, pmap, defs, cmap, dvs, ps)
 
+    # Main.@infiltrate sys isa NonlinearSystem
```
ChrisRackauckas, line 793:
```suggestion
```

AayushSabharwal: Oops 😅

AayushSabharwal:
I'll wait for DiffEqBase to be tagged before pushing the changes, since CI will fail right now anyways.

PR #3469, opened 2025-03-18
fix: fix analysis point transform ignoring too many connections

AayushSabharwal:


test/downstream/analysis_points.jl
```diff
@@ -438,3 +438,52 @@ end
...
+                   connect(inverse_model.torque.tau, add.input1)
+                   connect(pid.ctr_output, add.input2)
+                   connect(add.output, :u, model.torque.tau) # Name connection u to form an analysis point
+                   connect(model.inertia1.flange_b, sensor.flange)
+                   connect(sensor.phi, :y, pid.measurement)]
+    closed_loop = ODESystem(connections, t,
+        systems = [model, inverse_model, pid, filt, sensor, inverse_sensor, r, add],
+        name = :closed_loop)
+    # just ensure the system simplifies
+    Blocks.get_sensitivity(closed_loop, :y)
```
baggepinnen, line 488:
```suggestion
    get_sensitivity(closed_loop, :y)
```
`get_sensitivity` is now in MTK and not in Blocks, right?

Should there be some test that the sensitivity function is correct?

baggepinnen: Something like this perhaps. If you are confident that the result is the same as that in the last successful docs build from where the example came, you can set  `reference_fr` to the computed `fr` in this PR, so that we have that as a test for future regressions
```julia
mats, _ = Blocks.get_sensitivity(closed_loop, :y)
S = CS.ss(mats...)
fr = CS.freqrespv(S, [0.01, 1, 100])
@test isapprox(fr, reference_fr)
```

baggepinnen: You can run this code and verify that the `S` function looks the same
image

baggepinnen (approved):

PR #3470, opened 2025-03-18
Refactor complete

isaacsas:
Some of the problems that we are experiencing in Catalyst arise from `complete` doing more than just setting up namespacing. It doesn't really make sense to inject new parameters like initial conditions into a `ReactionSystem`, as such systems encode a variety of mathematical models, some of which have no notion of an initial condition. That injection should really only occur for a converted system that is in a concrete MTK mathematical system type (i.e. ODEs, SDEs, etc).

This PR attempts to address that by making it possible for us to dispatch `complete` in a way that we can opt-out of modifications to `ps`. In the future, hopefully complete-related changes to `ps` can be restricted to when `allow_additional_ps = true`, for which we will dispatch `complete(::ReactionSystem)` to always pass `false`.

Note that a better approach might be to not have namespacing tied into completion. We could completely ignore `complete` in Catalyst if it was handled separately. But I assume this would be breaking.


AayushSabharwal:
Is `_complete` intended to be public API? https://github.com/SciML/ModelingToolkit.jl/pull/3466 has an approach to opt-out of adding initial parameters using a trait. The rest of `complete`'s behavior is already controllable via keyword arguments (`flatten = false, split = false` basically only sets the flag). If the intention is to disable namespacing without doing everything else, it can be done this way. We could also do this by adding a `no_namespacing` flag which `complete` also flips, in addition to a `toggle_namespacing` function to just flip the flag.

AayushSabharwal:
The problem with this approach is it doesn't entirely solve your problem. The next time we add something to `complete`, it'll be enabled by default, which might break Catalyst and require you to add another keyword to your `complete` method. It's why I'm suggesting an alternative function which simply toggles namespacing.

isaacsas:
OK, so as to get Catalyst unblocked I think we will just dispatch `complete(::ReactionSystem)` for now and remove the non-relevant initialization code from the MTK implementation. If/when the flattening, scoped var stuff, and namespacing  gets factored out we can switch to whatever API functions are created for them in our implementation.

PR #3477, opened 2025-03-20
feat: support alternative AD backends in linearization

AayushSabharwal:


ChrisRackauckas (approved):

ChrisRackauckas:
Test failures

src/linearization.jl
```diff
@@ -82,13 +84,104 @@ function linearization_function(sys::AbstractSystem, inputs,
...
+            fun(du, u, p, t)
+            return du
+        end
+    end
+    if u0 === nothing
+        uf_jac = h_jac = pf_jac = nothing
+        T = p isa MTKParameters ? eltype(p.tunable) : eltype(p)
+        hp_jac = PreparedJacobian{true}(
+            hp_fun, zeros(T, size(outputs)), autodiff, inputvals,
+            DI.Constant(prob.u0), DI.Constant(p), DI.Constant(t0))
```
gdalle, line 104: You might want to use [`DI.ConstantOrCache`](https://juliadiff.org/DifferentiationInterface.jl/DifferentiationInterface/dev/api/#DifferentiationInterface.ConstantOrCache) for the parameter `p` if it contains storage where active data gets written

gdalle: Ok good, the same remark holds for other uses of Constant but I guess it gets the same answer. Just asking because we had the issue in OrdinaryDiffEq

AayushSabharwal: I appreciate the review nonetheless! It's nice to be doubly sure this is correct

PR #3484, opened 2025-03-21
pass `ODESystem` kwargs via `@mtkmodel` constructor

hexaeder:
I think it would be nice if you could pass kw args to the `ODESystem` constructor even if you system has been defined via `@mtkmodel`.
Motivating example:
```julia
    @mtkmodel UnitFailure begin
        @variables begin
            x(t), [unit = u"V"]
            y(t), [unit = u"A"]
        end
        @equations begin
            x ~ y
        end
    end
    UnitFailure(name = :uf, checks = false) # no error with passed kwarg
```

## Checklist

### DIFF
--- a/src/systems/model_parsing.jl
+++ b/src/systems/model_parsing.jl
@@ -115,7 +115,8 @@ function _model_macro(mod, name, expr, isconnector)
         Ref(dict), [:constants, :defaults, :kwargs, :structural_parameters])
 
     sys = :($ODESystem($(flatten_equations)(equations), $iv, variables, parameters;
-        name, description = $description, systems, gui_metadata = $gui_metadata, defaults))
+        name, description = $description, systems,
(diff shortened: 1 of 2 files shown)


baggepinnen:
Is this really a good idea? What if pass `UnitFailure(var_with_typo=val)`? Will MTK not silently accept this and I get a confusing result due to having a typo in my variable name? IIRC, MTK accepts `kwargs...` in several internal functions making misspelled keyword arguments slip through unnoticed.

hexaeder:
Fair point, I was hoping that every keyword argument is eventually matched or errors, but its never silently dropped. I agree that this can be super frustrating.

Another way would be to make it explicit, i.e.
```julia
UnitFailure(name = :uf, odesystem=(; checks = false))
```
which would fail if somebody names their variables `odesystem`...

PR #3493, opened 2025-03-24
fix: handle derivatives of time-dependent parameters

### LINKED ISSUE #3480: Code generation contains `Differential` when equations contain variable time parameter
Take the following 

```julia
@component function FilteredInput(; name, x0=0, T=0.1)
    params = @parameters begin
      k(t) = x0
      T = T
    end
    vars = @variables begin
      x(t) = k
      dx(t) = 0
      ddx(t)
    end
    systems = []
    eqs = [
      D(x) ~ dx
      D(dx) ~ ddx
      dx ~ (k - x)/T
    ]
    return ODESystem(eqs, t, vars, params; systems, name)
end

@mtkbuild sys = FilteredInput()
ModelingToolkit.observed(sys)
```

gives...

```julia
julia> ModelingToolkit.observed(sys)
3-element Vector{Equation}:
 dx(t) ~ (k(t) - x(t)) / T
 dxˍt(t) ~ (-dx(t) + Differential(t)(k(t))) / T
 ddx(t) ~ dxˍt(t)
```

Note the `Differential(t)(k(t)))` which cannot be computed.

Manifest:
```
[961ee093] ModelingToolkit v9.66.0
```

AayushSabharwal:
Close #3480

### DIFF
--- a/test/state_selection.jl
+++ b/test/state_selection.jl
@@ -2,7 +2,7 @@ using ModelingToolkit, OrdinaryDiffEq, Test
 using ModelingToolkit: t_nounits as t, D_nounits as D
 
 sts = @variables x1(t) x2(t) x3(t) x4(t)
-params = @parameters u1(t) u2(t) u3(t) u4(t)
+params = @parameters u1 u2 u3 u4
 eqs = [x1 + x2 + u1 ~ 0
        x1 + x2 + x3 + u2 ~ 0
        x1 + D(x3) + x4 + u3 ~ 0
--- a/src/structural_transformation/pantelides.jl
+++ b/src/structural_transformation/pantelides.jl
@@ -54,6 +54,7 @@ function pantelides_reassemble(state::TearingState, var_eq_matching)
             D(eq.lhs)
         end
         rhs = ModelingToolkit.expand_derivatives(D(eq.rhs))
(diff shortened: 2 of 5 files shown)


ChrisRackauckas:
But also test that the derivative of say a datainterpolation is non-zero still.

ChrisRackauckas:
Was this found to be okay / safe?

AayushSabharwal:
Not completely. We don't always want the derivative of a discrete parameter to be zero. One such example is CoSimulation FMUs. The spec enables getting derivatives of outputs, and sometimes MTK needs this information (e.g. if the output of an FMU is in an equation that pantelides differentiates). I think the solution is to allow `D(p(t)) ~ ...` equations where `p` is a time-dependent parameter and use that in the mapping added to this PR.

baggepinnen (approved):

test/state_selection.jl
```diff
@@ -2,7 +2,7 @@ using ModelingToolkit, OrdinaryDiffEq, Test
 using ModelingToolkit: t_nounits as t, D_nounits as D
 
 sts = @variables x1(t) x2(t) x3(t) x4(t)
-params = @parameters u1(t) u2(t) u3(t) u4(t)
```
ChrisRackauckas, line 5: hehe

ChrisRackauckas (approved):

PR #3494, opened 2025-03-24
add tutorial on disturbance modeling for state estimation

baggepinnen:


ChrisRackauckas:
It would probably be good to show an example of using sampled disturbances, using an interpolation block? That's a FAQ.

baggepinnen:
I added a link and mention of the MTKstdlib blocks module and their tutorial on using the interpolation block for sampled input signals

docs/src/tutorials/disturbance_modeling.md
```diff
@@ -0,0 +1,231 @@
+# Disturbance and input modeling modeling
+
+Disturbances are often seen as external factors that influence a system. Modeling and simulation of such external influences is common in order to ensure that the plant and or control system can adequately handle or suppress these disturbances. Disturbance modeling is also integral to the problem of state estimation, indeed, modeling how disturbances affect the evolution of the state of the system is crucial in order to accurately estimate this state.
```
cstjean, line 3: It'd be nice to use the word "additive" somewhere in there because that helped me understand that you're just using a Flow connector to introduce an additive effect into the system.

> The addition of disturbance inputs relies on the fact that the plant model has been constructed using connectors, we use these to connect the disturbance inputs from outside the plant-model definition

That belongs in the tutorial IMO!

docs/src/tutorials/disturbance_modeling.md
```diff
@@ -0,0 +1,231 @@
...
+
+When modeling a system in MTK, we essentially (without considering algebraic equations for simplicity in exposition) construct a model of a dynamical system
+
+```math
+\begin{aligned}
+\dot x &= f(x, p, t) \\
+y &= g(x, p, t)
```
cstjean, line 107:
```suggestion
y &= g(x, p, t)
\end{aligned}
```

docs/src/tutorials/disturbance_modeling.md
baggepinnen: This is mentioned further down

> It's worth noting at this point that the fact that we could connect disturbance outputs from outside of the plant-model definition was enabled by the fact that we used a component-based workflow, where the plant model had the appropriate connectors available. If the plant model had modeled the system using direct equations without connectors, this would not have been possible and the model would thus be significantly less flexible.

cstjean: Right, but for me "disturbance" is a very general world that would include, for instance, that a pump can fail and suddenly the flow goes to 0. But that's not something that would be modelizable in this way, right? You can only model additive effects because you're using the (very neat) trick of connecting to a Flow connector.

I get that in your field "disturbance" may be a technical term whose precise definition I'm unaware, but in any case for a general audience introducing the vocabulary helps a lot.

baggepinnen: You can easily model something being multiplied by zero through connectors, they are not limited to modeling additive inputs. When you add a connector to your component, say
```julia
disturbance_input = RealInput()
```
what you do with the connector variable `disturbance_input.u` is up to you, you can do `flow * (1-disturbance_input.u)` if you'd like and can thus turn the flow off completely by setting the disturbance input signal to `true`

baggepinnen: You can actually set flow variables to zero without having a connector dedicated to this, see the example below, where the completely external connection of the `ZeroFlowDisturbance` to an existing RC circuit sets the flow through one of the components to zero.

```julia
@mtkmodel RC begin
    @parameters begin
        R = 1.0
        C = 1.0
        V = 1.0
    end
    @components begin
        resistor = Resistor(R = R)
        capacitor = Capacitor(C = C, v = 0.0)
        source = Voltage()
        constant = Constant(k = V)
        ground = Ground()
    end
    @equations begin
        connect(constant.output, source.V)
        connect(source.p, resistor.p)
        connect(resistor.n, capacitor.p)
        connect(capacitor.n, source.n, ground.g)
    end
end

@mtkmodel ZeroFlowDisturbance begin
    @components begin
        p = Pin()
        n = Pin()
        flow_input = RealInput()
    end
    @variables begin
        v(t)
    end
    @equations begin
        p.i + n.i ~ 0.0
        p.v-n.v ~ v
        # When this component is connected, it turns off flow at t=2
        n.i ~ ifelse(t < 2, 0.0, -flow_input.u)
    end
end

@mtkmodel RCWithDisturbance begin
    @components begin
        rc = RC()
        disturbance = ZeroFlowDisturbance()
    end
    @equations begin
        rc.resistor.i ~ disturbance.flow_input.u
        connect(rc.resistor.p, disturbance.n)
        connect(rc.resistor.n, disturbance.p)
    end
end

@mtkbuild sys = RCWithDisturbance()
prob = ODEProblem(sys, Pair[], (0, 10.0))
sol = solve(prob)
plot(sol, idxs = [sys.rc.capacitor.v, sys.rc.resistor.i], title = "RC Circuit Demonstration", labels = ["Capacitor Voltage" "Resistor Current"])
```
image

ChrisRackauckas:
What's the state of this?

baggepinnen:
It's done

ChrisRackauckas:
Rerunning the build for

> ERROR: LoadError: AssertionError: latexify does not support objects of type AnalysisPoint.

Also, see the suggested change above.

PR #3495, opened 2025-03-25
docs: Document DAEProblem constructor

### LINKED ISSUE #3485: Document DAEProblem constructor and give a warning about performance implications
See https://discourse.julialang.org/t/daeproblem-and-modelingtoolkit-jl/127097/5

vyudu:
Fix #3485

### DIFF
--- a/docs/src/systems/ODESystem.md
+++ b/docs/src/systems/ODESystem.md
@@ -62,6 +62,7 @@ jacobian_sparsity
 ODEFunction(sys::ModelingToolkit.AbstractODESystem, args...)
 ODEProblem(sys::ModelingToolkit.AbstractODESystem, args...)
 SteadyStateProblem(sys::ModelingToolkit.AbstractODESystem, args...)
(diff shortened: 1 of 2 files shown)


cstjean:
> If possible, it is recommended to formulate your problem in terms of an ODEProblem

In MTK, this boils down to "avoid algebraic constraints other than `a = b` equalities and sum-of-flow equations", right?

vyudu:
sorry, my message is kind of misleading, ODEProblem can be used for systems with arbitrary algebraic equations attached to them, what I mean here is to formulate DAEs in terms of the mass matrix Mu' = f(u, p, t) (for the ODEProblem) rather than the fully implicit form f(u', u, p, t) = 0 (for the DAEProblem) if possible/convenient. 

To be honest, I'm not sure if there's any cases where you actually should or need to make a DAEProblem from a MTK system since it seems like ODESystems are naturally in the mass matrix form, so maybe this message should be more explicit along the lines of "You shouldn't ever use DAEProblem". But I'm not sure if this is true. @ChrisRackauckas ?

PR #3496, opened 2025-03-25
Fix ODEProblem over JumpProblem

isaacsas:
I don't understand why tests didn't catch this, but it is causing us issues in Catalyst with initialization-related errors for variable rate jump problems (that have no coupled ODEs). Given `build_initializeprob = false` is set to false in the coupled ODE-jump case, it seems like it was just an oversight to not include it here too.


isaacsas:
Test failures seem unrelated.

isaacsas:
@ChrisRackauckas any chance we could get this merged and get a release with it? This is the last thing preventing us from finishing the PR that gets Catalyst master working with MTK again.

ChrisRackauckas:
I guess this is fine as a stopgap solution. We shouldn't necessary rely on this down the line because it's an interface break. @AayushSabharwal can you prioritize getting initialize prob infrastructure onto JumpProblems working? it should just forward down to the prob in there. We can do a release with this to get tests passing and such and get the rest of downstream going for now, but we should leave open the very real probability that it will revert in hopefully a week, but with the fix to JumpProblems matching the AbstractSciMLProblem interface requirements on inits.

PR #3497, opened 2025-03-25
feat: allow toggling namespacing independently of `complete`

AayushSabharwal:
CC @TorkelE @isaacsas @SebastianM-C @baggepinnen

### DIFF
--- a/src/systems/diffeqs/sdesystem.jl
+++ b/src/systems/diffeqs/sdesystem.jl
@@ -139,7 +139,11 @@ struct SDESystem <: AbstractODESystem
     """
     gui_metadata::Union{Nothing, GUIMetadata}
     """
-    If a model `sys` is complete, then `sys.x` no longer performs namespacing.
+    If false, then `sys.x` no longer performs namespacing.
+    """
+    namespacing::Bool
+    """
+    If true, denotes the model will not be modified any further.
     """
     complete::Bool
     """
@@ -166,7 +170,7 @@ struct SDESystem <: AbstractODESystem
             guesses, initializesystem, initialization_eqs, connector_type,
             cevents, devents, parameter_dependencies, assertions = Dict{
                 BasicSymbolic, Nothing},
-            metadata = nothing, gui_metadata = nothing,
+            metadata = nothing, gui_metadata = nothing, namespacing = true,
             complete = false, index_cache = nothing, parent = nothing, is_scalar_noise = false,
             is_dde = false,
             isscheduled = false;
@@ -184,6 +188,7 @@ struct SDESystem <: AbstractODESystem
             if is_scalar_noise && neqs isa AbstractMatrix
                 throw(ArgumentError("Noise equations ill-formed. Received a matrix of noise equations of size $(size(neqs)), but `is_scalar_noise` was set to `true`. Scalar noise is only compatible with an `AbstractVector` of noise equations."))
             end
+            check_subsystems(systems)
         end
         if checks == true || (checks & CheckUnits) > 0
             u = __get_unit_type(dvs, ps, iv)
@@ -192,8 +197,8 @@ struct SDESystem <: AbstractODESystem
         new(tag, deqs, neqs, iv, dvs, ps, tspan, var_to_name, ctrls, observed, tgrad, jac,
             ctrl_jac, Wfact, Wfact_t, name, description, systems,
             defaults, guesses, initializesystem, initialization_eqs, connector_type, cevents,
-            devents, parameter_dependencies, assertions, metadata, gui_metadata, complete,
-            index_cache, parent, is_scalar_noise, is_dde, isscheduled)
+            devents, parameter_dependencies, assertions, metadata, gui_metadata, namespacing,
+            complete, index_cache, parent, is_scalar_noise, is_dde, isscheduled)
     end
 end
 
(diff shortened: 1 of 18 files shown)


test/analysis_points.jl
```diff
@@ -47,8 +47,11 @@ end
     @named sys = ODESystem(Equation[], t; systems = [sys_ap])
     ap3 = @test_nowarn sys.hej.plant_input
     @test nameof(ap3) == Symbol(join(["sys", "hej", "plant_input"], NAMESPACE_SEPARATOR))
-    sys = complete(sys)
-    ap4 = sys.hej.plant_input
+    csys = complete(sys)
+    ap4 = csys.hej.plant_input
+    @test nameof(ap4) == Symbol(join(["hej", "plant_input"], NAMESPACE_SEPARATOR))
+    nsys = toggle_namespacing(sys, false)
+    ap5 = nsys.hej.plant_input
```
SebastianM-C, line 54: Does `complete` no longer delete analysis points?

AayushSabharwal: `complete` deletes analysis points in the sense that you cannot perform analysis point transformations on a completed system. However, you can still refer to analysis points from a completed system.

AayushSabharwal:
> enable namespacing

*disable, but yes

src/systems/diffeqs/sdesystem.jl
```diff
@@ -139,7 +139,11 @@ struct SDESystem <: AbstractODESystem
     """
     gui_metadata::Union{Nothing, GUIMetadata}
     """
-    If a model `sys` is complete, then `sys.x` no longer performs namespacing.
+    If false, then `sys.x` no longer performs namespacing.
```
ChrisRackauckas, line 142:
```suggestion
    If false, then `sys.x` no longer performs namespacing, i.e. `sys.x => x`
```

src/systems/discrete_system/discrete_system.jl
```diff
@@ -97,7 +97,11 @@ struct DiscreteSystem <: AbstractDiscreteSystem
     """
     substitutions::Any
     """
-    If a model `sys` is complete, then `sys.x` no longer performs namespacing.
+    If false, then `sys.x` no longer performs namespacing.
```
ChrisRackauckas, line 100:
```suggestion
    If false, then `sys.x` no longer performs namespacing, i.e. `sys.x => x`
```

src/systems/discrete_system/implicit_discrete_system.jl
```diff
@@ -96,7 +96,11 @@ struct ImplicitDiscreteSystem <: AbstractDiscreteSystem
     """
     substitutions::Any
     """
-    If a model `sys` is complete, then `sys.x` no longer performs namespacing.
+    If false, then `sys.x` no longer performs namespacing.
```
ChrisRackauckas, line 99:
```suggestion
    If false, then `sys.x` no longer performs namespacing,, i.e. `sys.x => x`
```

src/systems/nonlinear/nonlinearsystem.jl
```diff
@@ -95,7 +95,11 @@ struct NonlinearSystem <: AbstractTimeIndependentSystem
     """
     substitutions::Any
     """
-    If a model `sys` is complete, then `sys.x` no longer performs namespacing.
+    If false, then `sys.x` no longer performs namespacing.
```
ChrisRackauckas, line 98:
```suggestion
    If false, then `sys.x` no longer performs namespacing, i.e. `sys.x => x`
```

src/systems/optimization/constraints_system.jl
```diff
@@ -74,7 +74,11 @@ struct ConstraintsSystem <: AbstractTimeIndependentSystem
     """
     substitutions::Any
     """
-    If a model `sys` is complete, then `sys.x` no longer performs namespacing.
+    If false, then `sys.x` no longer performs namespacing.
```
ChrisRackauckas, line 77:
```suggestion
    If false, then `sys.x` no longer performs namespacing, i.e. `sys.x => x`
```

src/systems/optimization/optimizationsystem.jl
```diff
@@ -55,7 +55,11 @@ struct OptimizationSystem <: AbstractOptimizationSystem
     """
     gui_metadata::Union{Nothing, GUIMetadata}
     """
-    If a model `sys` is complete, then `sys.x` no longer performs namespacing.
+    If false, then `sys.x` no longer performs namespacing.
```
ChrisRackauckas, line 58:
```suggestion
    If false, then `sys.x` no longer performs namespacing, i.e. `sys.x => x`
```

ChrisRackauckas:
`namespacing` is the wrong tense? `namespaces`? @baggepinnen might have opinions

baggepinnen:
> namespacing is the wrong tense? namespaces? @baggepinnen might have opinions

My opinion here is not very strong, but this function does not really toggle any namespace, it toggles the action of appending a namespace in a future call to `sys.x`. If the action of appending a namespace is referred to as "namespacing" I think the current name is okay. I have a bigger problem with the horrible `structural_simplify`, but that's a discussion for another place and time :sweat_smile:

AayushSabharwal:
Failures are weird, it passes locally.

AayushSabharwal:
Okay hmm updating reproduces this on master too

ChrisRackauckas (approved):

PR #3502, opened 2025-03-26
feat: add `substitute_component`

AayushSabharwal:
## Checklist

### DIFF
--- a/src/systems/abstractsystem.jl
+++ b/src/systems/abstractsystem.jl
@@ -3181,3 +3181,180 @@ has_diff_eqs(osys21) # returns `false`.
 ```
 """
 has_diff_eqs(sys::AbstractSystem) = any(is_diff_equation, get_eqs(sys))
+
+"""
+    $(TYPEDSIGNATURES)
+
+Validate the rules for replacement of subcomponents as defined in `substitute_component`.
+"""
+function validate_replacement_rule(
+        rule::Pair{T, T}; namespace = []) where {T <: AbstractSystem}
+    lhs, rhs = rule
+
+    iscomplete(lhs) && throw(ArgumentError("LHS of replacement rule cannot be completed."))
+    iscomplete(rhs) && throw(ArgumentError("RHS of replacement rule cannot be completed."))
+
+    rhs_h = namespace_hierarchy(nameof(rhs))
+    if length(rhs_h) != 1
+        throw(ArgumentError("RHS of replacement rule must not be namespaced."))
+    end
+    rhs_h[1] == namespace_hierarchy(nameof(lhs))[end] ||
+        throw(ArgumentError("LHS and RHS must have the same name."))
+
+    if !isequal(get_iv(lhs), get_iv(rhs))
+        throw(ArgumentError("LHS and RHS of replacement rule must have the same independent variable."))
+    end
+
+    lhs_u = get_unknowns(lhs)
+    rhs_u = Dict(get_unknowns(rhs) .=> nothing)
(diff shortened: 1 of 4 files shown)


cstjean:
Very useful! As interface, `substitute_component` is clean. As an Accessors.jl fan, it'd be nice to also have

```julia
    rcsys = @set rsys.component2 = component2
```
by implementing [`ConstructionBase.setproperties`](https://juliaobjects.github.io/ConstructionBase.jl/dev/#ConstructionBase.setproperties).

AayushSabharwal:
I agree that would be a great syntax. However, we use `Setfield` internally to update systems and this might conflict with that. Additionally, the `setproperties` approach would only go one level deep. If a user did `@set sys.inner.component.child = newchild`, this would first replace `child` in `component`, then use that to replace `component` in `inner`, then use that to replace `inner` in `sys` which is a lot of unnecessary work.

src/systems/abstractsystem.jl
```diff
@@ -3142,3 +3142,180 @@ has_diff_eqs(osys21) # returns `false`.
...
+    eqs = map(get_eqs(sys)) do eq
+        eq.lhs isa Union{Connection, AnalysisPoint} || return eq
+        if eq.lhs isa Connection
+            oldargs = get_systems(eq.rhs)
+        else
+            ap::AnalysisPoint = eq.rhs
+            oldargs = [ap.input; ap.outputs]
+        end
+        newargs = map(get_systems(eq.rhs)) do arg
+            name = arg isa AbstractSystem ? nameof(arg) : getname(arg)
```
SebastianM-C, line 3270: I've tested this locally a bit and I have a case where `eq.rhs` is a (top level) `connect` equation and the `arg`s are `ModelingToolkit.SymbolicWithNameof` and then `getname(arg)` gives the `Symbol("var_name(t)")` variant instead of just `:var_name`. I saw that using `getname(arg.var)` works, but I'm not sure if that makes sense since the resulting system crashes on printing during `n_expanded_connection_equations` :sweat_smile:.

AayushSabharwal: Oh oops, I always forget that `SymbolicWithNameof` exists. Thanks for pointing it out.

cstjean:
We could really use this functionality, but unless I'm mistaken the requirements limit its applicability.

> 5. `rhs` must contain at least all of the unknowns and parameters present in `lhs`.

Eg. "What if we replaced this continuously-stirred tank with a non-stirred tank?" might not meet these requirements, right? Or "ComplexReactorModel => NaiveReactorModel".

I understand that there's probably an efficiency reason for these constraints, but in this case I would rather have an inefficient general function than an efficient-but-limited one.

My use case is for running experiments. "What if we replaced component A.B.C with C2(...)?" is much nicer to write than the corresponding top-level `@mtkmodel` that has to connect everything.

Is there another way of doing this?

AayushSabharwal:
> I understand that there's probably an efficiency reason for these constraints, but in this case I would rather have inefficient-and-general than efficient-but-limited.

It's not an efficiency but correctness issue. It's _really_ difficult to see how a component (and all subcomponents) are used across the entire model and then verify that the replacement has the same pieces. I'd be more than happy to work on it in the future, but this addresses a limited set of problems with scope for expansion in the future.

The way to do this is define the model with an "interface component" - one which isn't intended for simulation but has the bare minimum interface required to connect with the rest of the system. This can then be replaced to form different model variants.

ChrisRackauckas (approved):

Issue #3507, opened 2025-03-26
Dependent parameters disappear from system parameters


hersle:
```julia
using ModelingToolkit
using ModelingToolkit: t_nounits as t, D_nounits as D
@variables x(t)
@parameters A B
@named M = ODESystem([D(x) ~ A + B], t; parameter_dependencies = [B ~ 2A])
parameters(M) # gives only A, but I expected A and B
```
Is this intended? I would like `B` to remain accessible from `parameters(M)`.

hersle:
[Removing dependent variables seem intentional](https://github.com/SciML/ModelingToolkit.jl/blob/349e9d53d503022971e949584b98282491ef9466/src/systems/abstractsystem.jl#L2737-L2739). Is the intention to "remove" them from the system because they are effectively "always" set?

I expected `parameters(sys)` to return all (independent and dependent parameters). My model construction is conditional on something like `if B in parameters(sys)` (and I do not care if `B` is dependent or independent).

Could `parameters(sys)` also return dependent parameters? Or is there a better way to check if a system has a given (independent or dependent) parameter?

hersle:
Sorry, I just realized there is `full_parameters`. I will open a PR to document it a little better to "fix" this issue.

Issue #3204, opened 2024-11-14
MTKv10 discussion


AayushSabharwal:
This issue is meant to track and discuss points of action and potential breaking changes for a future MTKv10 release

- Make `@named` _always_ wrap symbolic arguments in `ParentScope`, so they can be passed multiple levels down the hierarchy. This will need more consideration after identifying which tests break upon making this change.
- Discuss `complete` semantics
- Enforce the `u0map` argument only contains variables, and `pmap` only parameters. Discretes will need an exception here.

AayushSabharwal:
- Remove the find/replace stuff in `expand_connections`. This is only used for MTKStdlib's implementation of `AnalysisPoint` but that is moved to MTK as of #3285.

baggepinnen:
I'd like to discuss making the syntax
```
Component(inner_par = outer_par)
```
Introduce a parameter dependence instead of a "dependent default". I find the current behavior is almost never what I want. In modelica, this syntax introduces the equivalent of a parameter dependence.

vyudu:
Possible changes to field names in symbolic metadata:
- `vartype` becoming `variable_source` 
- new field `variable_type` (corresponding to the MTK variable type)

AayushSabharwal:
^ This specifically refers to the `NamedTuple` returned from `get_variable_metadata`

AayushSabharwal:
Other discussions worth having:
- defaults are now just soft constraints that are only respected sometimes. Especially for parameters since they have to opt-in to initialization. 
- Not relying on variable metadata as much. This refers to multiple things
  - `@mtkmodel` macro using it for keeping track of defaults
  - More importantly, `structural_simplify` using it to figure out what is a parameter and what isn't. `ODESystem` can use the metadata to do this, but `structural_simplify` should respect whatever bucket we put it in.

hersle:
> Discuss `complete` semantics

Should `complete()` be changed to use `flatten = false` by default? I find it undesireable that the current default `flatten = true` destroys the hierarchical model structure without me explicitly requesting it. There are some bugs with `flatten = false`, though (https://github.com/SciML/ModelingToolkit.jl/issues/3322).

hersle:
> More importantly, `structural_simplify` using it to figure out what is a parameter and what isn't. ODESystem can use the metadata to do this, but `structural_simplify` should respect whatever bucket we put it in.

I always thought it could be simpler to have just the `@variables` macro, and that it should be sufficient to write
```julia
[...]

PR #3509, opened 2025-03-27
Improve parameters and full_parameters

hersle:
"Fix" https://github.com/SciML/ModelingToolkit.jl/issues/3507, or at least try to prevent others from running into the same.

Is it breaking to rename the keyword argument `initial_parameters` to `initial`? I would prefer `initial` and `dependent` over `initial_parameters` and `dependent_parameters`. I think the latter is too verbose and would grow out of hand if more flags are added in the future.

### DIFF
--- a/src/systems/abstractsystem.jl
+++ b/src/systems/abstractsystem.jl
@@ -1335,10 +1335,11 @@ Get the standard parameters of the system `sys` and its subsystems.
 # Keyword arguments
 
 - `initial_parameters`: Whether to include parameters for initial values of variables in the initialization system.
+- `dependent_parameters`: Whether to include parameters that are determined by parameter dependencies.
 
 See also [`full_parameters`](@ref) and [`ModelingToolkit.get_ps`](@ref).
 """
-function parameters(sys::AbstractSystem; initial_parameters = false)
+function parameters(sys::AbstractSystem; initial_parameters = false, dependent_parameters = false)
     ps = get_ps(sys)
     if ps == SciMLBase.NullParameters()
         return []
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
> initial_parameters to initial?

It's brand new so it's in the range of time that it's like.. sure? If @TorkelE is onboard.

I do think making it all one function is just simpler API. Condense `full_parameters` and `parameters`. Condense `equations` and `full_equations`. Less functions, just different knobs and a good docstring.

@AayushSabharwal take it from here.

AayushSabharwal:
Yep, looks good. I intentionally didn't document the `initial_parameters` kwarg. We're free to change it. It's also worth listing the "old" functions in #3204 so we remember to remove them later on.

hersle:
> I do think making it all one function is just simpler API. Condense full_parameters and parameters. Condense equations and full_equations. Less functions, just different knobs and a good docstring.

Yes, I fully agree. This PR is definitely in that direction. I can still see the use for one easy-to-use function to get all parameters/equations, though, which is equivalent to calling the "normal" getter with all flags enabled.

AayushSabharwal (changes requested):
If we're making these flags public, they need tests

hersle:
I

1. added a test
2. replaced the keyword argument `initial_parameters` by `initial`
3. replaced the keyword argument `add_initial_parameters` by `add_inital` (in `complete`; for consistency with 2.)
4. replaced the keyword argument `dependent_parameters` by `dependent`

I understand the name changes are subjective. I just feel these shorter names are more future-proof. Just let me know if I should revert any or all of them, and I am happy to do it.

src/systems/abstractsystem.jl
```diff
@@ -742,7 +742,7 @@ One property to note is that if a system is complete, the system will no longer
 namespace its subsystems or variables, i.e. `isequal(complete(sys).v.i, v.i)`.
 """
 function complete(
-        sys::AbstractSystem; split = true, flatten = true, add_initial_parameters = true)
+        sys::AbstractSystem; split = true, flatten = true, add_initial = true)
```
SebastianM-C, line 745: Are all the kwargs considered internal here?

SebastianM-C:
IMO the longer names are more clear (as it makes it clear that we are talking about parameters), but I don't have a strong opinion here.

AayushSabharwal:
The longer names can get a bit annoying, but I kind of feel like they should be plural? `initials = true, dependents = true`? That way it's very clear you're using a collective noun rather than an adjective.

hersle:
I agree that `initials` would be better than `initial` etc. But I also see @SebastianM-C's point and don't want to change and stir things up unnecessarily, so I reverted the name changes. Sorry for the extra trouble 😅

TorkelE:
Sounds good to me. Given that they are keargs to the "parameters" function, I don't mind not having "parameters" in the kwarg names as well.

PR #3510, opened 2025-03-27
Improve consistency and add parameter dependencies to system printing

hersle:
Minor improvements. Now for example:

image


ChrisRackauckas (approved):
@AayushSabharwal take it from here.

AayushSabharwal (approved):

Issue #3505, opened 2025-03-26
clarify in API documentation which system kwargs are public vs. private


isaacsas:
Per recent discussions we've been told that Catalyst has been using private kwargs in system constructors in some cases. For example, `observed` was mentioned (though as I previously mentioned, we use it because of the discussion in https://github.com/SciML/ModelingToolkit.jl/pull/1343 for which I thought we had been ok'ed to pass explicit observed equations -- as common in SBML and other bio file formats).

It would be nice to update the API documentation or doc strings within the systems to make clear which kwargs are API and which are intended to be internal. This really isn't clear to me at this point, which makes it challenging to know which ones we can use from Catalyst.

ChrisRackauckas:
We generally just do not document the kwargs that are internal. For ODESystem for example we do not document any of the keyword arguments. https://docs.sciml.ai/ModelingToolkit/dev/systems/ODESystem/#ModelingToolkit.ODESystem notice that no `observed` keyword argument is documented here. But we do document internal fields, which is a bit odd.

I think the bigger thing is that the API parts of the MTK documentation just need a complete overhaul. The issue that we had before was that there was no docstring repetitions allowed. This meant that if you used `get_eqs` docstring on one page, you couldn't use it in another. This was a weird Documenter limitation that led to things like https://docs.sciml.ai/ModelingToolkit/dev/systems/ODESystem/#Composition-and-Accessor-Functions, rather than actually putting docstrings, because again that was a Documenter limitation and the docs just wouldn't build. This also meant that we couldn't put many constructor docstrings in the docs because they would get repeated in a few odd ways, which means all constructors had to be manually written. 

Now you can set a canonical location, so in theory all of those weird little "use these functions" things should be changed to the docstrings. And in the docstrings we then fully document the public API. But it would probably be nice to extend the SciMLStyle to have some documentation style for how to handle private API, since in MTK some of this is used to signal to other authors since the compiler can get complicated.

So this is going to be a significant effort to get us onto the right track. I want to pull @asinghvi17 into some of this since he always has good ideas, at least for the design, and it's good to have a newer set of eyes that doesn't know everything already. @vyudu @asinghvi17 @AayushSabharwal let's come up with a plan next week for a documentation structure that will scale better here, and reset the docs to this structure. I think we have a lot of the content already written but I think the structure needs a rethink.

TorkelE:
Documentation is always the never ending task, isn't it?

[...]

PR #3512, opened 2025-03-30
Add documentation contribution of multi/nonlocal stability by Attractors.jl

Datseris:
We have discussed this on Slack some time ago! cc @TorkelE @ChrisRackauckas

### DIFF
--- a/docs/src/tutorials/attractors.md
+++ b/docs/src/tutorials/attractors.md
@@ -130,7 +130,9 @@ fig
 
 ## Basins of attraction
 
-Estimating the basins of attraction of these attractor is a matter of a couple lines of code. First we define the state space are to estimate the basins for. Here we can re-use the `grid` we defined above. Then we only have to call
+Estimating the basins of attraction of these attractors is a matter of a couple lines of code. 
+First we define the state space are to estimate the basins for. 
+Here we can re-use the `grid` we defined above. Then we only have to call
 
 ```julia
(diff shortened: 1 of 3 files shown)


docs/src/tutorials/attractors.md
```diff
@@ -0,0 +1,218 @@
...
+colors = ["#7143E0", "#191E44"]
+for (id, A) in attractors
+    scatter!(ax, A[:, [1, 3]]; color = colors[id])
+end
+fig
+```
+
+## Basins of attraction
+
+Estimating the basins of attraction of these attractor is a matter of a couple lines of code. First we define the state space are to estimate the basins for. Here we can re-use the `grid` we defined above. Then we only have to call
```
Datseris, line 133:
```suggestion
Estimating the basins of attraction of these attractors is a matter of a couple lines of code. 
First we define the state space are to estimate the basins for. 
Here we can re-use the `grid` we defined above. Then we only have to call
```

ChrisRackauckas:
I think this might fit better as an example instead of a tutorial?

ChrisRackauckas:
The bifurcation diagrams should also be an example. We'll probably address that in the doc overhaul https://github.com/SciML/ModelingToolkit.jl/issues/3505. I guess for now whatever let's slap it in there and adjust when we do the others.

Datseris:
Have forgotten to add the package to the docs/Project.toml. Should work now.

PR #3522, opened 2025-04-01
temp commit

AayushSabharwal:
This is a weird little snippet, and I want to see what happens when it is removed. MTK tests tend to slow my system to a crawl, so I'm running it in CI.


AayushSabharwal:
It looks pretty removable.

ChrisRackauckas (approved):

PR #3535, opened 2025-04-03
feat: compute jac_prototype for SDEFunction

### LINKED ISSUE #3527: No jacobian prototype for sparse jacobian `SDEProblem`s?
Not that familiar with SDE solving, but it seems odd that there this field is empty when a sparse jacobian is built for SDEs? They are there for e.g. ODEs and Nonlinear problems. And if one shouldn't do sparse Jacobians here, shouldn't;t there at least be an error/warning?
```julia
using Plots

@parameters σ=10.0 ρ=2.33 β=26.0
@variables x(t)=5.0 y(t)=5.0 z(t)=1.0
@brownian B
eqs = [D(x) ~ σ * (y - x) + 0.3x * B,
    D(y) ~ x * (ρ - z) - y + 0.3y * B,
    D(z) ~ x * y - β * z + 0.3z * B]

@mtkbuild de = System(eqs, t)
sprob = SDEProblem(de, [], 0.0, [], jac = true, sparse = true)
sprob.f.jac_prototype == nothing # true
```

vyudu:
Close #3527

### DIFF
--- a/src/structural_transformation/symbolics_tearing.jl
+++ b/src/structural_transformation/symbolics_tearing.jl
@@ -378,6 +378,7 @@ function generate_derivative_variables!(
     # For variable x, make dummy derivative x_t if the
     # derivative is in the system
     for v in 1:length(var_to_diff)
+        @show graph
         dv = var_to_diff[v]
         dv isa Int || continue
         solved = var_eq_matching[dv] isa Int
@@ -403,6 +404,10 @@ function generate_derivative_variables!(
         v_t = add_dd_variable!(structure, fullvars, x_t, dv)
         # Add `D(x) - x_t ~ 0` to the graph
         dummy_eq = add_dd_equation!(structure, neweqs, 0 ~ dx - x_t, dv, v_t)
+        # Update graph to say, all the equations featuring D(x) also feature x_t
+        for e in 𝑑neighbors(graph, dv)
+            add_edge!(graph, e, v_t)
+        end
 
         # Update matching
         push!(var_eq_matching, unassigned)
--- a/src/structural_transformation/symbolics_tearing.jl
+++ b/src/structural_transformation/symbolics_tearing.jl
@@ -378,7 +378,6 @@ function generate_derivative_variables!(
     # For variable x, make dummy derivative x_t if the
     # derivative is in the system
     for v in 1:length(var_to_diff)
-        @show graph
         dv = var_to_diff[v]
         dv isa Int || continue
         solved = var_eq_matching[dv] isa Int
--- a/src/systems/diffeqs/abstractodesystem.jl
+++ b/src/systems/diffeqs/abstractodesystem.jl
@@ -143,15 +143,20 @@ function generate_W(sys::AbstractODESystem, γ = 1., dvs = unknowns(sys),
         simplify = false, sparse = false, kwargs...)
     @variables ˍ₋gamma
     M = calculate_massmatrix(sys; simplify)
+    sparse && (M = SparseArrays.sparse(M))
     J = calculate_jacobian(sys; simplify, sparse, dvs)
     W = ˍ₋gamma*M + J
+    @show W
 
     p = reorder_parameters(sys, ps)
+    @show length(p)
     return build_function_wrapper(sys, W, 
         dvs,
         p...,
         ˍ₋gamma,
         get_iv(sys);
+        wrap_code = sparse ? assert_jac_length_header(sys) : (identity, identity),
+        p_end = 1 + length(p),
         kwargs...)
 end
 
--- a/src/systems/diffeqs/abstractodesystem.jl
+++ b/src/systems/diffeqs/abstractodesystem.jl
@@ -298,12 +298,12 @@ function W_sparsity(sys::AbstractODESystem)
         jac_sparsity = jacobian_sparsity(sys)
     else
         jac = calculate_jacobian(sys; sparse = true)
-        jac_sparsity = sparse((!iszero).(jac))
+        jac_sparsity = SparseMatrixCSC{Bool, Int64}((!iszero).(jac))
     end
     (n, n) = size(jac_sparsity)
     M = calculate_massmatrix(sys)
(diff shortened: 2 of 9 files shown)


vyudu:
Mass matrix is already in there

ChrisRackauckas:
The mass matrix is a subset of `jacobian_sparsity(sys)`? Are you sure? I don't remember handling that here.

vyudu:
No I mean it was there before, see lines 656-657

vyudu:
@ChrisRackauckas @oscardssmith @AayushSabharwal 
So the behavior implemented here is: 
- `generate_jacobian` returns the jacobian, `generate_W` returns the W
- `prob.f.jac_prototype` and `prob.f.W_prototype` both return the same thing, the `W_prototype` (which I added in case we want to switch to only using W_prototype in the solvers at some point) 
- the jac! returned by `generate_jacobian` will error if the passed in jac doesn't have the W pattern

is this right? anything I'm missing?

vyudu:
Nevermind this is writing to the wrong entries, need to investigate

src/systems/diffeqs/abstractodesystem.jl
```diff
@@ -386,7 +426,8 @@ function DiffEqBase.ODEFunction{iip, specialize}(sys::AbstractODESystem,
         jac = _jac === nothing ? nothing : _jac,
         tgrad = _tgrad === nothing ? nothing : _tgrad,
         mass_matrix = _M,
-        jac_prototype = jac_prototype,
+        jac_prototype = W_prototype,
+        W_prototype = W_prototype,
```
ChrisRackauckas, line 430: we don't want / need both since that would double the largest memory

src/systems/diffeqs/sdesystem.jl
```diff
@@ -724,6 +738,18 @@ function SDEFunctionExpr{iip}(sys::SDESystem, dvs = unknowns(sys),
         _jac = :nothing
     end
 
+    M = calculate_massmatrix(sys)
+    _M = (u0 === nothing || M == I) ? M : ArrayInterface.restructure(u0 .* u0', M)
+
+    if sparse
+        uElType = u0 === nothing ? Float64 : eltype(u0)
+        jac_prototype = similar(calculate_jacobian(sys; sparse), uElType)
```
ChrisRackauckas, line 746: the jac_prototype should just be what you have as the W_prototype

src/structural_transformation/symbolics_tearing.jl
```diff
@@ -403,6 +403,10 @@ function generate_derivative_variables!(
         v_t = add_dd_variable!(structure, fullvars, x_t, dv)
         # Add `D(x) - x_t ~ 0` to the graph
         dummy_eq = add_dd_equation!(structure, neweqs, 0 ~ dx - x_t, dv, v_t)
+        # Update graph to say, all the equations featuring D(x) also feature x_t
+        for e in 𝑑neighbors(graph, dv)
+            add_edge!(graph, e, v_t)
+        end
```
ChrisRackauckas, line 409: @AayushSabharwal what's the implication of that on the tearing result? It's good for the jac sparsity pattern but does it do anything else? I don't think it does because we only tear the algebraic equations anyways?

src/systems/diffeqs/abstractodesystem.jl
```diff
@@ -264,6 +293,19 @@ function jacobian_dae_sparsity(sys::AbstractODESystem)
     J1 + J2
 end
 
+function W_sparsity(sys::AbstractODESystem) 
+    if has_tearing_state(sys)
+        jac_sparsity = jacobian_sparsity(sys)
```
ChrisRackauckas, line 298: why is this necessary?

vyudu: SDESystems don't have tearing states

src/structural_transformation/symbolics_tearing.jl
AayushSabharwal: I don't get why this is necessary? What's a case where an equation contains `D(x)` but not `x_t`?

ChrisRackauckas: It's just to correct the MM sparsity.

vyudu: Before these edges weren't getting added to the graph, so an algebraic equation e in which D(x) gets subbed for x_t didn't have an edge between e and x_t. And the Jacobian sparsity detection relied on the graph

test/jacobiansparsity.jl
```diff
@@ -51,7 +51,7 @@ JP = prob.f.jac_prototype
 
 # test sparse jacobian
 prob = ODEProblem(sys, u0, (0, 11.5), sparse = true, jac = true)
-@test_nowarn solve(prob, Rosenbrock23())
+#@test_nowarn solve(prob, Rosenbrock23())
```
ChrisRackauckas, line 54:
```suggestion
@test_nowarn solve(prob, Rosenbrock23())
```

src/systems/diffeqs/abstractodesystem.jl
ChrisRackauckas: it does, it's just whether it's structurally simplified. I see.

vyudu: Right but SDESystems don't encapsulate the tearing state (i.e. it's not an accessible field of SDESystem)

ChrisRackauckas: Okay, that's kind of a bug 😅 . @AayushSabharwal 's unification of the system types should fix that.

ChrisRackauckas:
So this is good to go?

vyudu:
No I think it's still writing to the wrong entries sorry, investigating

vyudu:
@ChrisRackauckas

src/systems/diffeqs/abstractodesystem.jl
AayushSabharwal: Yeah SDESystem simplification goes through ODESystem but doesn't retain all the information, which is weird.

PR #3528, opened 2025-04-02
fix: fix incorrect namespacing of DelayParentScope variables in `collect_scoped_vars!`

AayushSabharwal:

PR #3540, opened 2025-04-04
feat: reduce reliance on metadata in `structural_simplify`

AayushSabharwal:
This doesn't work yet, for a variety of reasons. https://github.com/SciML/ModelingToolkit.jl/pull/3528 is one of them.


AayushSabharwal:
Cancelled CI because there's no point in running it when this is known to be problematic

AayushSabharwal:
Rebased on top of #3543 for now

ChrisRackauckas (approved):

PR #3549, opened 2025-04-08
feat: create JuMPControlProblem for optimal control 

vyudu:
will add optimal control tests here when the interface for inputs and stuff is settled.


vyudu:
@ChrisRackauckas assuming this passes stuff I think it's ready.

vyudu:
Requires https://github.com/SciML/SciMLBase.jl/pull/996

test/extensions/jump_control.jl
```diff
@@ -0,0 +1,193 @@
...
+using OrdinaryDiffEqSDIRK
+using Ipopt
+using BenchmarkTools
+using CairoMakie
+const M = ModelingToolkit
+
+@testset "ODE Solution, no cost" begin
+    # Test solving without anything attached.
+    @parameters α=1.5 β=1.0 γ=3.0 δ=1.0
+    @variables x(..) y(..)
```
baggepinnen, line 14: Do I have to define the variables to optimize over using `x(..)`? What if I want to solve an OC problem with components defined in MTKstdlib? Those components have `x(t)` variables. It would be nice to have a test for that use case as well

test/extensions/jump_control.jl
```diff
@@ -0,0 +1,193 @@
...
+
+@testset "Linear systems" begin
+    function is_bangbang(input_sol, lbounds, ubounds, rtol = 1e-4)
+        bangbang = true
+        for v in 1:(length(input_sol.u[1]) - 1)
+            all(i -> ≈(i[v], bounds[v]; rtol) || ≈(i[v], bounds[u]; rtol), input_sol.u) ||
+                (bangbang = false)
+        end
+        bangbang
+    end
```
baggepinnen, line 87:
```suggestion
    function is_bangbang(input_sol, lbounds, ubounds, rtol = 1e-4)
        for v in 1:(length(input_sol.u[1]) - 1)
            all(i -> ≈(i[v], bounds[v]; rtol) || ≈(i[v], bounds[u]; rtol), input_sol.u) ||
                return false
        end
        true
    end
```

baggepinnen (commented):
there appears to be no tests that the dynamics are satisfied, testing the transcription. This could be tested by simulating the system using the found optimal control included with an interpolation

test/extensions/jump_control.jl
ChrisRackauckas: No, that's just if you also want to evaluate them at a point, since this just late binds `t`. We could also make an `evaluate` operator or something that is the same as `evaluate(x(t), 0.2) == x(0.2)`

ext/MTKJuMPControlExt.jl
```diff
@@ -0,0 +1,377 @@
...
+        new{typeof(u0), typeof(tspan), SciMLBase.isinplace(f),
+            typeof(p), typeof(f), typeof(kwargs)}(f, u0, tspan, p, model, kwargs)
+    end
+end
+
+"""
+    JuMPControlProblem(sys::ODESystem, u0, tspan, p; dt)
+
+Convert an ODESystem representing an optimal control system into a JuMP model
+for solving using optimization. Must provide `dt` for determining the length
```
baggepinnen, line 43: `dt` is presumably not "determining the length of the interpolation arrays". Instead of mentioning why a variable is required, maybe mention what the variable represents?

test/extensions/jump_control.jl
vyudu: I will add a component test case

test/extensions/jump_control.jl
```diff
@@ -0,0 +1,193 @@
...
+    # Test explicit method.
+    jprob = JuMPControlProblem(sys, u0map, tspan, parammap, dt = 0.01)
+    @test num_constraints(jprob.model) == 2 # initials
+    jsol = solve(jprob, Ipopt.Optimizer, :RK4)
+    oprob = ODEProblem(sys, u0map, tspan, parammap)
+    osol = solve(oprob, SimpleRK4(), dt = 0.01)
+    @test jsol.sol.u ≈ osol.u
+
+    # Implicit method.
+    jsol2 = @btime solve($jprob, Ipopt.Optimizer, :ImplicitEuler, silent = true) # 63.031 ms, 26.49 MiB
```
baggepinnen, line 35: should these `@btime` really be run in the tests?

ChrisRackauckas: no

test/extensions/jump_control.jl
```diff
@@ -0,0 +1,193 @@
+using ModelingToolkit
+import JuMP, InfiniteOpt
+using DiffEqDevTools, DiffEqBase
+using SimpleDiffEq
+using OrdinaryDiffEqSDIRK
+using Ipopt
+using BenchmarkTools
+using CairoMakie
```
baggepinnen, line 8: ?

ChrisRackauckas: I think this is still WIP

vyudu:
Sorry yeah I've converted it to a draft for now since stuff is still being added. 

https://infiniteopt.github.io/InfiniteOpt.jl/stable/develop/extensions/#Derivative-Evaluation-Methods looks like we can extend InfiniteOpt to add our own derivative methods. Should I just do this for all the ODE tableaus we have (like define a `InfiniteOptTsit5, InfiniteOptImplicitEuler, InfiniteOptRadauIA3`, etc in the extension)? It seems like they perform a lot better than doing the collocation using pure JuMP. Maybe as a later step if not in this PR? 

This also means we wouldn't have a separate `JuMPDynamicOptProblem` and `InfiniteOptDynamicOptProblem`, it would all just be `InfiniteOptDynamicOptProblem`

ChrisRackauckas:
Later step, let's do one step at a time.

>  It seems like they perform a lot better than doing the collocation using pure JuMP

That's interesting, I wonder why it's actually better.

> This also means we wouldn't have a separate JuMPDynamicOptProblem and InfiniteOptDynamicOptProblem, it would all just be InfiniteOptDynamicOptProblem

I think for benchmarking purposes it's good to have both.

baggepinnen:
While playing with this, I noticed the following

- The `silent` in `solve(jprob, Ipopt.Optimizer, :RadauIIA5, silent = true)` appears to have no effect, causing 1000s of lines of Ipopt printouts. 
- `DynamicOptSolution` cannot be plotted (but `sol.sol` can)
- when plotting `plot(isol.input_sol)` for the cartpole swingup, the legend entry for the input signal is `θ(t)` isntead of `u(t)`
- Something appears to be wrong with `isol.input_sol` for the rocket launch problem. I expect the solution to look like this
image
and `jsol.input_sol` is close (it seems the default tolerance is rather relaxed?)
image
but `isol.input_sol` is completely wrong
image
I tried running this test
```julia
    # Test solution
    @parameters (T_interp::CubicSpline)(..)
    eqs = [D(h(t)) ~ v(t),
        D(v(t)) ~ (T_interp(t) - drag(h(t), v(t))) / m(t) - gravity(h(t)),
        D(m(t)) ~ -T_interp(t) / c]
    @mtkbuild rocket_ode = ODESystem(eqs, t)
    interpmap = Dict(T_interp => ctrl_to_spline(jsol.input_sol, CubicSpline))
    oprob = ODEProblem(rocket_ode, u0map, (ts, te), merge(Dict(pmap), interpmap))
    osol = solve(oprob, RadauIIA5(); adaptive = false, dt = 0.005)
    @test ≈(jsol.sol.u, osol.u, rtol = 0.02)
```
for the `isol` as well but failed

baggepinnen:
I tried specifying a minimum-time problem, but got an error
```julia
# Double integrator minimum time
t = M.t_nounits
D = M.D_nounits
@variables x(..) v(..)
@variables u(..) [bounds = (-1.0, 1.0), input = true]
@parameters tf
constr = [v(tf) ~ 0.0, x(tf) ~ 0]
cost = [tf] # Maximize the final distance.
@named block = ODESystem(
    [D(x(t)) ~ v(t), D(v(t)) ~ u(t)], t; costs = cost, constraints = constr)
block, input_idxs = structural_simplify(block, ([u(t)], []))

u0map = [x(t) => 1.0, v(t) => 0.0]
tspan = (0.0, tf)
parammap = [u(t) => 0.0, tf=>1.0]
jprob = JuMPDynamicOptProblem(block, u0map, tspan, parammap; steps = 51)
jsol = solve(jprob, Ipopt.Optimizer, :Verner8)
```
```
UndefVarError: `NotSymbolic` not defined in `MTKJuMPDynamicOptExt`
```

ext/MTKJuMPDynamicOptExt.jl
```diff
@@ -0,0 +1,388 @@
...
+    end
+end
+
+"""
+    JuMPDynamicOptProblem(sys::ODESystem, u0, tspan, p; dt)
+
+Convert an ODESystem representing an optimal control system into a JuMP model
+for solving using optimization. Must provide either `dt`, the timestep between collocation 
+points (which, along with the timespan, determines the number of points), or directly 
+provide the number of points as `nsteps`.
```
baggepinnen, line 45:
```suggestion
provide the number of points as `steps`.
```

src/systems/optimal_control_interface.jl
```diff
@@ -0,0 +1,187 @@
...
+
+# returns the JuMP timespan, the number of steps, and whether it is a free time problem.
+function process_tspan(tspan, dt, steps)
+    is_free_time = false
+    if isnothing(dt) && isnothing(steps)
+        error("Must provide either the dt or the number of intervals to the collocation solvers (JuMP, InfiniteOpt, CasADi).")
+    elseif symbolic_type(tspan[1]) === ScalarSymbolic() ||
+           symbolic_type(tspan[2]) === ScalarSymbolic()
+        isnothing(steps) &&
+            error("Free final time problems require specifying the number of steps, rather than dt.")
```
baggepinnen, line 176:
```suggestion
            error("Free final time problems require specifying the number of steps using keyword arg `steps`, rather than dt.")
```

ChrisRackauckas:
`DynamicOptSolution`

Is there a reason to not just make the solution an ODESolution? Add a residual term to that?

vyudu:
Just thought this would make it easier for plotting purposes. Symbolic indexing is kind of weird in this case since inputs are usually non-discrete parameters and doing `sol.ps[u(t)]` won't get you what you want so fetching the input solution from the solution might be a bit annoying if it's all put in one ODESolution.

vyudu:
Also useful to return the JuMP model so they can inspect the objective value and stuff like that

vyudu:
@baggepinnen It may just be that OrthogonalCollocation for InfiniteOpt is a bit strange, I tested it with implicit Euler and got this, which still doesn't look like the original solution but gives the right dynamics when I simulate the ODEProblem using it 
image

baggepinnen:
> It may just be that OrthogonalCollocation for InfiniteOpt is a bit strange

The solution during the strange phase has two distinct phases, one singular arc and one boundary arc, it might look prettier if there is some regularization added, such as a penalty on the squared second derivative of the control action, or simply just a tiny penalty on the squared control action. 

https://arxiv.org/pdf/2504.17093

vyudu:
@ChrisRackauckas @baggepinnen Thoughts on merging this now and adding some of these things (regularization, component stuff) in later ones? This PR is getting a bit bloated already

Project.toml
```diff
@@ -76,6 +78,7 @@ MTKChainRulesCoreExt = "ChainRulesCore"
 MTKDeepDiffsExt = "DeepDiffs"
 MTKFMIExt = "FMI"
 MTKInfiniteOptExt = "InfiniteOpt"
+MTKJuMPDynamicOptExt = ["JuMP", "DiffEqDevTools", "InfiniteOpt"]
```
ChrisRackauckas, line 81: It's a bit odd to require DiffEqDevTools?

src/ModelingToolkit.jl
```diff
@@ -347,4 +347,10 @@ export AnalysisPoint, get_sensitivity_function, get_comp_sensitivity_function,
        open_loop
 function FMIComponent end
 
+include("systems/optimal_control_interface.jl")
+export AbstractDynamicOptProblem, JuMPDynamicOptProblem, InfiniteOptDynamicOptProblem,
+       PyomoDynamicOptProblem, CasADiDynamicOptProblem
+export DynamicOptSolution
+export ∫
```
ChrisRackauckas, line 354: Unicode is never public API

src/systems/optimal_control_interface.jl
```diff
@@ -0,0 +1,187 @@
...
+    println("Optimal control solution for following model:\n")
+    show(sol.model)
+
+    print("\n\nPlease query the model using sol.model, the solution trajectory for the system using sol.sol, or the solution trajectory for the controllers using sol.input_sol.")
+end
+
+function JuMPDynamicOptProblem end
+function InfiniteOptDynamicOptProblem end
+function CasADiDynamicOptProblem end
+function PyomoDynamicOptProblem end
```
ChrisRackauckas, line 22: these ones aren't in here?

src/systems/optimal_control_interface.jl
```diff
@@ -0,0 +1,187 @@
...
+end
+
+"""
+Integral operator. When applied to an expression in a cost
+function, assumes that the integration variable is the
+iv of the system, and assumes that the bounds are the
+tspan.
+Equivalent to Integral(t in tspan) in Symbolics.
+"""
+struct ∫ <: Symbolics.Operator end
```
ChrisRackauckas, line 152: Why does this need a new operator?

test/runtests.jl
```diff
@@ -23,105 +23,106 @@ function activate_downstream_env()
...
-            @safetestset "Parameter Dependency Test" include("parameter_dependencies.jl")
-            @safetestset "Equation Type Accessors Test" include("equation_type_accessors.jl")
-            @safetestset "System Accessor Functions Test" include("accessor_functions.jl")
-            @safetestset "Equations with complex values" include("complex.jl")
-        end
-    end
+    #if GROUP == "All" || GROUP == "InterfaceI"
+    #    @testset "InterfaceI" begin
+    #        @safetestset "Linear Algebra Test" include("linalg.jl")
+    #        @safetestset "AbstractSystem Test" include("abstractsystem.jl")
```
ChrisRackauckas, line 29: Can't merge like this...

src/systems/optimal_control_interface.jl
vyudu: It doesn't, it's just to make it easier to write integrals in cost functions, but yeah if it can't be exposed I'll just remove it

Project.toml
vyudu: I can remove but it does need to be loaded for the solves

src/systems/optimal_control_interface.jl
ChrisRackauckas: I think just define it in the tutorial if you really want the shorthand. I don't think it needs to be in the library. At least, we never accept unicode public API so it would need some other name.

Project.toml
ChrisRackauckas: There isn't a default tableau?

vyudu: Not atm, which one should it be?

ChrisRackauckas: Probably like a RadauIIA 5th order?

vyudu: I actually don't need to require JuMP either since it's included with InfiniteOpt. So maybe this should be renamed to MTKInfiniteOptDynamicOptExt? Should I rename MTKInfiniteOptExt or is that fine?

ChrisRackauckas: MTKInfiniteOptExt

vyudu: There's already one called that though?

ChrisRackauckas: Just add it to that? In fact, we should consider whether that code in there is just legacy after this is added. I don't think it's documented or needed, it's a start to this but redundant.

vyudu: Got it, I'll just add the functions defined there to this ext

ext/MTKInfiniteOptExt.jl
```diff
@@ -1,11 +1,434 @@
...
+Returns a DynamicOptSolution, which contains both the model and the ODE solution.
+"""
+function DiffEqBase.solve(
+        prob::JuMPDynamicOptProblem, jump_solver, ode_solver::Symbol = :Default; silent = false)
+    model = prob.model
+    tableau_getter = Symbol(:construct, ode_solver)
+    if ode_solver == :Default
+        @eval tableau = $tableau_getter()
+    else
+        @eval tableau = Main.$tableau_getter()
```
vyudu, line 363: So this is clearly wrong but I'm not super sure how to fix it

ChrisRackauckas (approved):

ChrisRackauckas:
It's missing a docs manual page and a new tutorial, but let's take it and keep improving.

PR #3556, opened 2025-04-10
fix: fix `assert_jac_length_header`

### LINKED ISSUE #3554: Small numeric error introduced in sparse jacobians
Something that popped up. The actual error is really small, so probably not practically serious. Still might be worth investigating what is going one.
```julia
using Catalyst, Test

rn = @reaction_network begin
    (p,d), 0 <--> (X,Y,Z)
    k1, X + Y --> XY
    k2, X + 2Z --> XZ2
    k3, Y3 +X2 --> Y3Z2
    k4, X + Y + Z --> XYZ
    k5, XZ2 + Y3Z2 --> XY3Z4
    k6, XYZ + XYZ --> X2Y2Z2
    d, (XY3Z4, X2Y2Z2) --> 0
    X + Y, V --> 0
    k7/(1 + t), 2V --> V2
    Z, V2 --> 0
end

# Create problem
u0 = [sp => rand() for sp in species(rn)]
ps = [p => rand() for p in parameters(rn)]
prob = ODEProblem(rn, u0, 0.0, ps; jac = true, sparse = true)

# Compute sparse jacobian.
J = deepcopy(prob.f.jac_prototype)
prob.f.jac(J, prob.u0, prob.p, 1.0)
@test J == prob.f.jac(prob.u0, prob.p, 1.0) # Don't pass anymore.
@test J ≈ prob.f.jac(prob.u0, prob.p, 1.0) # Passes (difference between jacobians is very small).

# Compute normal Jacobian.
probd = ODEProblem(rn, u0, 0.0, ps; jac = true, sparse = false)
Jd = zeros(length(u0), length(u0))
probd.f.jac(Jd, prob.u0, prob.p, 1.0)
@test Jd == probd.f.jac(prob.u0, prob.p, 1.0) # This works for dense jacobians.
@test Matrix(J) == Jd # Now fails.
@test Matrix(J) ≈ Jd # works.
```

AayushSabharwal:
Close #3554

### DIFF
--- a/src/ModelingToolkit.jl
+++ b/src/ModelingToolkit.jl
@@ -303,7 +303,8 @@ export structural_simplify, expand_connections, linearize, linearization_functio
        LinearizationProblem
 export solve
 
-export calculate_jacobian, generate_jacobian, generate_function, generate_custom_function, generate_W
+export calculate_jacobian, generate_jacobian, generate_function, generate_custom_function,
+       generate_W
 export calculate_control_jacobian, generate_control_jacobian
 export calculate_tgrad, generate_tgrad
 export calculate_gradient, generate_gradient
--- a/src/systems/diffeqs/abstractodesystem.jl
+++ b/src/systems/diffeqs/abstractodesystem.jl
@@ -141,10 +141,12 @@ end
 
 function assert_jac_length_header(sys)
     W = W_sparsity(sys)
-    identity, expr -> Func([expr.args...], [], LiteralExpr(quote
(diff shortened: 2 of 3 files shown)


AayushSabharwal:
I can something like the MWE in the issue as a test if we want, just not sure if exact numerical equality is something we (or SU) can guarantee in these cases.

ChrisRackauckas (approved):

AayushSabharwal:
SymbolicIndexingInterface test group fails because SciMLBase needs a hotfix

Issue #3555, opened 2025-04-10
Suggestion: Make `plot` plot unknowns AND observables by default (not just unknowns)


TorkelE:
When splitting what variables become unknowns and what are observables, I feel like it makes sense to lump them together in the plot (i.e. plotting all by default). I.e. here:
```julia
using OrdinaryDiffEqDefault, ModelingToolkit, Plots
using ModelingToolkit: t_nounits as t, D_nounits as D

# Create model.
@variables X(t) Y(t)
@parameters p d k1 k2 = 0.5
eqs = [
    D(X) ~ p - d*X,
    k1 + Y ~ k2*X
]
@mtkbuild osys = ODESystem(eqs, t)

# Create `ODEProblem`.
u0 = [X => 2.0]
pvals = [p => 1.0, d => 0.1, k1 => 0.5]
oprob = ODEProblem(osys, u0, (0.0, 10.0), pvals)

# Solve and plot it.
sol = solve(oprob)
plot(sol)
```
only `X` is plotted (unless one specify `Y` explicitly). Now from a user perspective there is no real difference here, they don't know why `X` is plotted and not `Y`. I think this is one of (if not the only?) cases where `structural_simplify` have a direct impact on workflows.

Furthermore, changing this would:
- Mean that we can be a lot more liberal with updating the internal algorithms of `structural_simplify`. I.e. there could be minor updates to its algorithm that suddenly changes so that `X` is eliminated and `Y` is not (probably not in the example above, but in more complicated ones). If the observables are ploted, the user will never notice or case.
- The case of `eqs = [D(X) ~ -p]` is one that can be solved directly, i.e. the only unknown of the system becomes an observable. Here, if you do `plot(solve(prob))` you will just get an empty plot plane, which seems really confusing.

I think the other alternative is to just enforcethe use of `idxs = ` in `plot`. That makes sense for larger models. However, for small models this will be annoying.

baggepinnen:
Sure but if you try to plot 1000s of signals Julia will crash, so that's not a reasonable default

TorkelE:
So there are cases where one option makes sense and another doesn't? In Sys bio the main reason for observables is pretty much the opposite (I have 3,000 differential equations, butonly want to plot ~5 observables. So I distinctly specify these things to plot to avoid the crash).

[...]

PR #3558, opened 2025-04-13
Add getter for observed variables

hersle:
Suggestion based on discussion in https://github.com/SciML/ModelingToolkit.jl/issues/3555 to "prevent" stuff like `map(eq -> eq.lhs, observed(sys))` for a little nicer API.

My idea was to introduce a convenient `observeds` getter named symmetrically to `unknowns`.
But working through it, I realize that this leads to an "asymmetry" in the naming of `equations` and `observed(s)`.

If you think this is a helpful addition, what is the best name?
- `observeds(sys)`?
- `observed_variables(sys)`?
- `unknowns(sys; observed = true)`?
- something else?

## Checklist


ChrisRackauckas:
I like `observables`

ChrisRackauckas:
It captures that these are lazily computed on-demand only if asked for, which is a use in computer science and matches this quite well. But then it also is a physical observable, can be measured from the system. It's a good fit.

hersle:
Thanks, I think `observables` sounds great! Will change it when I get back.

PR #3561, opened 2025-04-14
fix: fix values being promoted to `Float64` in problem construction

### LINKED ISSUE #3553: `Float32` types are automatically converted to `Float64`
```julia
using ModelingToolkit
using OrdinaryDiffEq
@parameters p d
@variables X(t)
eqs = [D(X) ~ p - d*X]
u0 = [X => 1.0f0]
ps = [p => 1.0f0, d => 2.0f0]
oprob = ODEProblem(osys, u0, 1.0, ps)
sol = solve(oprob)

typeof(oprob[X]) # Float64
eltype(sol[X]) # Float64
```
This one popped out in the Catalyst tests with the latest MTK release. Among other things it should be a major problem for GPU parallelisation.

AayushSabharwal:
Close #3553

### DIFF
--- a/src/systems/abstractsystem.jl
+++ b/src/systems/abstractsystem.jl
@@ -733,7 +733,7 @@ function add_initialization_parameters(sys::AbstractSystem)
     defs = copy(get_defaults(sys))
     for ivar in initials
         if symbolic_type(ivar) == ScalarSymbolic()
-            defs[ivar] = zero(Float16)
+            defs[ivar] = false
         else
             defs[ivar] = collect(ivar)
             for scal_ivar in defs[ivar]
--- a/src/systems/abstractsystem.jl
+++ b/src/systems/abstractsystem.jl
@@ -737,7 +737,7 @@ function add_initialization_parameters(sys::AbstractSystem)
         else
             defs[ivar] = collect(ivar)
             for scal_ivar in defs[ivar]
-                defs[scal_ivar] = zero(Float16)
+                defs[scal_ivar] = false
             end
         end
     end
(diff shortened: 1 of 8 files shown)


src/systems/abstractsystem.jl
```diff
@@ -733,11 +733,11 @@ function add_initialization_parameters(sys::AbstractSystem)
     defs = copy(get_defaults(sys))
     for ivar in initials
         if symbolic_type(ivar) == ScalarSymbolic()
-            defs[ivar] = zero_var(ivar)
+            defs[ivar] = zero(Float16)
         else
             defs[ivar] = collect(ivar)
             for scal_ivar in defs[ivar]
-                defs[scal_ivar] = zero_var(scal_ivar)
+                defs[scal_ivar] = zero(Float16)
```
ChrisRackauckas, line 740:
```suggestion
                defs[scal_ivar] = false
```

src/systems/nonlinear/initializesystem.jl
```diff
@@ -8,7 +8,7 @@ function generate_initializesystem(sys::AbstractTimeDependentSystem;
         pmap = Dict(),
         initialization_eqs = [],
         guesses = Dict(),
-        default_dd_guess = 0.0,
+        default_dd_guess = Bool(0),
```
ChrisRackauckas, line 11: Just false?

AayushSabharwal: `false` makes the kwarg seem like a flag, when it is actually a value. So I opted for `Bool(0)` just to make it clearer.

AayushSabharwal:
_Finally_

PR #3573, opened 2025-04-22
[v10] refactor: change inputs/outputs handling in `structural_simplify`

vyudu:
Standardize the input handling by no longer returning `input_idxs`. To be merged into https://github.com/SciML/ModelingToolkit.jl/pull/3563


test/downstream/linearize.jl
```diff
@@ -24,14 +24,14 @@ lsys3, _ = linearize(sys, [r], [y]; autodiff = AutoFiniteDiff())
 @test lsys.C[] == lsys2.C[] == lsys3.C[] == 1
 @test lsys.D[] == lsys2.D[] == lsys3.D[] == 0
 
-lsys, ssys = linearize(sys, [r], [r])
```
baggepinnen, line 27: If we don't return `ssys`, how can the user know the state realization used? On line 17 above, the `ssys` is returned, what's different on this line?

test/downstream/linearize.jl
```diff
@@ -87,19 +87,19 @@ connections = [f.y ~ c.r # filtered reference to controller reference
 
 @named cl = ODESystem(connections, t, systems = [f, c, p])
 
-lsys0, ssys = linearize(cl, [f.u], [p.x])
+lsys0 = linearize(cl, [f.u], [p.x])
 desired_order = [f.x, p.x]
-lsys = ModelingToolkit.reorder_unknowns(lsys0, unknowns(ssys), desired_order)
-lsys1, ssys = linearize(cl, [f.u], [p.x]; autodiff = AutoFiniteDiff())
-lsys2 = ModelingToolkit.reorder_unknowns(lsys1, unknowns(ssys), desired_order)
+lsys = ModelingToolkit.reorder_unknowns(lsys0, unknowns(cl), desired_order)
```
baggepinnen, line 92: `cl` hasn't ben simplified, so `unknowns(cl)` should be the wrong array to use?

test/downstream/linearize.jl
```diff
@@ -114,11 +114,12 @@ Nd = 10
 @named pid = LimPID(; k, Ti, Td, Nd)
 
 @unpack reference, measurement, ctr_output = pid
-lsys0, ssys = linearize(pid, [reference.u, measurement.u], [ctr_output.u];
+pid = structural_simplify(pid, inputs = [reference.u, measurement.u], outputs = [ctr_output.u])
+lsys0 = linearize(pid, [reference.u, measurement.u], [ctr_output.u];
```
baggepinnen, line 118: if the inputs and outputs are passed already on the line above, why do they have to be passed here as well?

test/downstream/linearize.jl
```diff
@@ -147,12 +148,12 @@ lsys = ModelingToolkit.reorder_unknowns(lsys, unknowns(ssys), reverse(desired_or
 
 ## Test that there is a warning when input is misspecified
 if VERSION >= v"1.8"
-    @test_throws "Some specified inputs were not found" linearize(pid,
+    @test_throws "Some parameters are missing from the variable map." linearize(pid,
```
baggepinnen, line 151: The previous error message was quite different and more precise, have we lost the ability to provide such a message?

test/downstream/linearize.jl
```diff
@@ -147,12 +148,12 @@ lsys = ModelingToolkit.reorder_unknowns(lsys, unknowns(ssys), reverse(desired_or
...
         [
             pid.reference.u,
             pid.measurement.u
         ], [ctr_output.u])
-    @test_throws "Some specified outputs were not found" linearize(pid,
+    @test_throws "Some parameters are missing from the variable map." linearize(pid,
```
baggepinnen, line 156: This error message is very bad, if outputs are missing we should continue to say so, not _parameters_!

test/downstream/linearize.jl
vyudu: Sorry I forgot to correct these. 

Re the first question, my thought was that the user always simplifies it with the inputs first and then passes it into the linearization. So I should add a few simplification statements here first. So they examine the state realization by just examining the simplified system. Is this a reasonable approach?

test/downstream/linearize.jl
vyudu: They shouldn't be, thanks. Is it reasonable to default inputs to `inputs(sys)` and outputs to `outputs(sys)` if the user does not explicitly pass these in?

test/downstream/linearize.jl
vyudu: This happens because the simplified system expects values for the inputs, so when they aren't passed in it errors at the time of ODEProblem creation, and does not reach the error that occured previously. Will fix

test/downstream/linearize.jl
baggepinnen: In some cases, like when calling `linearize`, yes, but there are cases when the user cannot simplify and while provide the inputs themselves. The functions that compute sensitivity functions, such as `get_sensitivity`, start by adding a new input and then simplify with this input specified, the user cannot do that since this input does not even exist before the function was called.

baggepinnen:
See discussions here
https://juliahub.slack.com/archives/C03UEC0P6G0/p1745388683133919

src/inputoutput.jl
```diff
@@ -179,9 +179,6 @@ The return values also include the chosen state-realization (the remaining unkno
 
 If `disturbance_inputs` is an array of variables, the generated dynamics function will preserve any state and dynamics associated with disturbance inputs, but the disturbance inputs themselves will (by default) not be included as inputs to the generated function. The use case for this is to generate dynamics for state observers that estimate the influence of unmeasured disturbances, and thus require unknown variables for the disturbance model, but without disturbance inputs since the disturbances are not available for measurement. To add an input argument corresponding to the disturbance inputs, either include the disturbance inputs among the control inputs, or set `disturbance_argument=true`, in which case an additional input argument `w` is added to the generated function `(x,u,p,t,w)->rhs`.
 
-!!! note "Un-simplified system"
```
baggepinnen, line 182: This warning is still warranted?

src/systems/systems.jl
```diff
@@ -26,12 +26,15 @@ topological sort of the observed equations in `sys`.
 + `fully_determined=true` controls whether or not an error will be thrown if the number of equations don't match the number of inputs, outputs, and equations.
 """
 function structural_simplify(
-        sys::AbstractSystem, io = nothing; additional_passes = [], simplify = false, split = true,
+        sys::AbstractSystem; additional_passes = [], simplify = false, split = true,
```
baggepinnen, line 29: the docstring needs update to reflect the change from tuple to kwarg

baggepinnen (commented):
are there no tests that need update due to this PR

src/inputoutput.jl
vyudu: The `generate_control_function` shouldn't require an un-simplified system anymore since it no longer simplifies the system

src/inputoutput.jl
```diff
@@ -163,7 +163,7 @@ has_var(ex, x) = x ∈ Set(get_variables(ex))
     (f_oop, f_ip), x_sym, p_sym, io_sys = generate_control_function(
             sys::AbstractODESystem,
             inputs             = unbound_inputs(sys),
```
AayushSabharwal, line 165: Wasn't there a discussion somewhere that `unbound_inputs` is buggy? Should we have a better alternative? CC @baggepinnen

src/inputoutput.jl
```diff
@@ -257,8 +256,11 @@ function generate_control_function(sys::AbstractODESystem, inputs = unbound_inpu
     (; f, dvs, ps, io_sys = sys)
 end
 
-function inputs_to_parameters!(state::TransformationState, io)
-    check_bound = io === nothing
+"""
+Turn input variables into parameters of the system.
+"""
+function inputs_to_parameters!(state::TransformationState, inputsyms; is_disturbance = false)
```
AayushSabharwal, line 262: `is_disturbance` isn't used anywhere

src/inputoutput.jl
```diff
@@ -316,14 +318,13 @@ function inputs_to_parameters!(state::TransformationState, io)
     @set! sys.unknowns = setdiff(unknowns(sys), keys(input_to_parameters))
     ps = parameters(sys)
 
-    if io !== nothing
-        inputs, = io
+    if inputsyms !== nothing
         # Change order of new parameters to correspond to user-provided order in argument `inputs`
```
AayushSabharwal, line 322: I don't think we care about/guarantee parameter ordering anymore.

src/inputoutput.jl
```diff
@@ -429,8 +429,9 @@ function add_input_disturbance(sys, dist::DisturbanceModel, inputs = nothing; kw
            dist.input ~ u + dsys.output.u[1]]
     augmented_sys = ODESystem(eqs, t, systems = [dsys], name = gensym(:outer))
     augmented_sys = extend(augmented_sys, sys)
+    ssys = structural_simplify(augmented_sys, inputs = all_inputs, disturbance_inputs = [d])
 
-    (f_oop, f_ip), dvs, p, io_sys = generate_control_function(augmented_sys, all_inputs,
-        [d]; kwargs...)
+    (f_oop, f_ip), dvs, p, io_sys = generate_control_function(ssys, all_inputs,
+        [d]; check_simplified = false, kwargs...)
```
AayushSabharwal, line 435:
```suggestion
        [d]; kwargs...)
```

src/linearization.jl
```diff
@@ -127,12 +126,24 @@ function linearization_function(sys::AbstractSystem, inputs,
...
 
     lin_fun = LinearizationFunction(
-        diff_idxs, alge_idxs, input_idxs, length(unknowns(sys)),
+        diff_idxs, alge_idxs, inputs, length(unknowns(sys)),
         prob, h, u0 === nothing ? nothing : similar(u0), uf_jac, h_jac, pf_jac,
         hp_jac, initializealg, initialization_kwargs)
     return lin_fun, sys
 end
 
+function eq_idxs(sys::AbstractSystem)
```
AayushSabharwal, line 135: Needs a docstring. This should also not rely on equations being ordered with differential equations coming first followed by all algebraic equations. That's a guarantee we're dropping in v10.

src/linearization.jl
```diff
@@ -545,10 +556,11 @@ function linearize_symbolic(sys::AbstractSystem, inputs,
     (; A, B, C, D, f_x, f_z, g_x, g_z, f_u, g_u, h_x, h_z, h_u), sys
 end
 
-function markio!(state, orig_inputs, inputs, outputs; check = true)
+function markio!(state, orig_inputs, inputs, outputs, disturbances; check = true)
```
AayushSabharwal, line 559: Could you add a docstring for this too? It never had one, but really should.

src/linearization.jl
```diff
@@ -545,10 +556,11 @@ function linearize_symbolic(sys::AbstractSystem, inputs,
...
     fullvars = get_fullvars(state)
     inputset = Dict{Any, Bool}(i => false for i in inputs)
     outputset = Dict{Any, Bool}(o => false for o in outputs)
+    disturbanceset = Dict{Any, Bool}(d => false for d in disturbances)
```
AayushSabharwal, line 563: Turn this into a `Set{Any}`? We never use the values and that ends up looking confusing, as if we're missing something.

src/systems/clock_inference.jl
```diff
@@ -1,7 +1,11 @@
 struct ClockInference{S}
+    """Tearing state."""
```
AayushSabharwal, line 2: Thanks for adding these!

src/systems/systemstructure.jl
```diff
@@ -713,31 +707,29 @@ function structural_simplify!(state::TearingState, io = nothing; simplify = fals
...
+        inputs = Any[], outputs = Any[],
+        disturbance_inputs = Any[],
         kwargs...)
     if fully_determined isa Bool
         check_consistency &= fully_determined
     else
         check_consistency = true
     end
-    has_io = io !== nothing
+    has_io = inputs !== nothing || outputs !== nothing
```
AayushSabharwal, line 728:
```suggestion
    has_io = !isempty(inputs) || !isempty(outputs)
```

src/linearization.jl
vyudu: Ah I guess I should use this to check that the disturbances actually exist in the system

src/inputoutput.jl
baggepinnen: yes, the _implementation_ of this function is not great, it's also very slow. It does the correct thing modulo bugs though. If the model has unbound inputs those have to be included in one of the input arguments for the generated function to work

AayushSabharwal (approved):

PR #3575, opened 2025-04-22
Add Neuroblox downstream test

ChrisRackauckas:
Is dependent on https://github.com/Neuroblox/Neuroblox.jl/pull/580


MasonProtter:
This should be ready now.

PR #3579, opened 2025-04-23
Copy `MTKParameters` fields only if not empty

DhairyaLGandhi:
## Checklist

## Additional context

Fixes https://github.com/SciML/SciMLSensitivity.jl/actions/runs/14616573242/job/41006088886?pr=1168

### DIFF
--- a/src/systems/parameter_buffer.jl
+++ b/src/systems/parameter_buffer.jl
@@ -319,8 +319,8 @@ function Base.copy(p::MTKParameters)
     initials = copy(p.initials)
     discrete = Tuple(eltype(buf) <: Real ? copy(buf) : copy.(buf) for buf in p.discrete)
     constant = Tuple(eltype(buf) <: Real ? copy(buf) : copy.(buf) for buf in p.constant)
-    nonnumeric = copy.(p.nonnumeric)
-    caches = copy.(p.caches)
+    nonnumeric = isempty(p.nonnumeric) ? p.nonnumeric : copy.(p.nonnumeric)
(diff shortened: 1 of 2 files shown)


ChrisRackauckas (approved):

src/systems/parameter_buffer.jl
```diff
@@ -319,8 +319,8 @@ function Base.copy(p::MTKParameters)
     initials = copy(p.initials)
     discrete = Tuple(eltype(buf) <: Real ? copy(buf) : copy.(buf) for buf in p.discrete)
     constant = Tuple(eltype(buf) <: Real ? copy(buf) : copy.(buf) for buf in p.constant)
-    nonnumeric = copy.(p.nonnumeric)
-    caches = copy.(p.caches)
+    nonnumeric = isempty(p.nonnumeric) ? p.nonnumeric : copy.(p.nonnumeric)
+    caches = isempty(p.caches) ? p.caches : copy.(p.caches)
```
ChrisRackauckas, line 323:
```suggestion
    nonnumeric = isempty(p.nonnumeric) ? copy(p.nonnumeric) : copy.(p.nonnumeric)
    caches = isempty(p.caches) ? copy(p.caches) : copy.(p.caches)
```

We probably shouldn't alias the array? Since it's empty it's probably fine, but it's slightly safer making this more like a deepcopy.

AayushSabharwal: If empty, these fields are `()` so `copy` won't work

PR #3584, opened 2025-04-25
feat: add `costs, constraints, coalesce` to `@mtkmodel`, introduce `At` operator 

vyudu:
A couple of needed optimal control utilities

### DIFF
--- a/src/ModelingToolkit.jl
+++ b/src/ModelingToolkit.jl
@@ -142,6 +142,7 @@ function var_derivative_graph! end
 include("bipartite_graph.jl")
 using .BipartiteGraphs
 
+export At
 include("variables.jl")
 include("parameters.jl")
 include("independent_variables.jl")
--- a/src/ModelingToolkit.jl
+++ b/src/ModelingToolkit.jl
@@ -142,7 +142,7 @@ function var_derivative_graph! end
 include("bipartite_graph.jl")
 using .BipartiteGraphs
 
-export At
+export EvalAt
 include("variables.jl")
 include("parameters.jl")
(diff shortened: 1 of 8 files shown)


vyudu:
@ChrisRackauckas I think this is ready - I'd like to merge this soon so I can use it for some of the component-based optimal control tests

src/ModelingToolkit.jl
```diff
@@ -142,6 +142,7 @@ function var_derivative_graph! end
 include("bipartite_graph.jl")
 using .BipartiteGraphs
 
+export At
```
ChrisRackauckas, line 145: This probably has too high of a chance to hit a collision.

ChrisRackauckas:
I think it looks good now. But I think `At` might be a name that is too easy to hit in user code, so it could cause weird issues with common codes. `At` could easily be A transpose in a script, now hitting the export. Could we just change to a different name and let people `EvalAt as At` or something?

ChrisRackauckas:
https://github.com/SciML/ModelingToolkit.jl/actions/runs/14803718971/job/41568071561?pr=3584#step:6:1182 the failure looks real?

vyudu:
I think the new BoundaryValueDiffEq release broke some of these tests but I haven't dug into yet, was planning on doing that soon (these failures show up on other PRs too too)

vyudu:
But they should'nt be related to this one

ChrisRackauckas:
confirm the failure is on master?

ChrisRackauckas:
Matches https://github.com/SciML/ModelingToolkit.jl/pull/3596

PR #3593, opened 2025-05-01
Fix ImperativeAffect namespacing for arrays in non-root components

BenChung:
## Checklist

### DIFF
--- a/src/systems/imperative_affect.jl
+++ b/src/systems/imperative_affect.jl
@@ -101,10 +101,22 @@ end
 
 namespace_affects(af::ImperativeAffect, s) = namespace_affect(af, s)
 function namespace_affect(affect::ImperativeAffect, s)
+    rmn = []
+    for modded in modified(affect)
+        if modded isa AbstractArray
+            res = []
+            for m in modded
+                push!(res, renamespace(s, m))
+            end
+            push!(rmn, res)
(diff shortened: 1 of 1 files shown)


ChrisRackauckas (changes requested):
Needs tests

ChrisRackauckas (approved):

ChrisRackauckas:
Is the FMI failure related? Idon't think that was on master.

BenChung:
@ChrisRackauckas Looks like it, huh. My suspicion is that the FMI compilation is running into the array variable scoping problems that I'm poking at with this. I'll talk to Aayush and see if there's a better way to fix this

src/systems/imperative_affect.jl
```diff
@@ -101,10 +101,22 @@ end
 
 namespace_affects(af::ImperativeAffect, s) = namespace_affect(af, s)
 function namespace_affect(affect::ImperativeAffect, s)
+    rmn = []
+    for modded in modified(affect)
+        if modded isa AbstractArray
```
AayushSabharwal, line 106:
```suggestion
        if symbolic_type(modded) == NotSymbolic() && modded isa AbstractArray
```

ChrisRackauckas (approved):

PR #3596, opened 2025-05-03
Test master

ChrisRackauckas:
https://github.com/SciML/ModelingToolkit.jl/pull/3584#issuecomment-2848335977


ChrisRackauckas:
@ErikQQY see the test failures from the BVP solver here. It seems to not be converging well? What could have caused this one?

ErikQQY:
> @ErikQQY see the test failures from the BVP solver here. It seems to not be converging well? What could have caused this one?

With the default tolerance changing to `1e-6`, I think it's the adaptivity thing not behaving well here, I will fix it ASAP.

PR #3614, opened 2025-05-09
feat: implement new callback semantics

AayushSabharwal:
Cleanup of #3452 and #3613 which avoids rebase issues and merge conflicts


AayushSabharwal:
This is good to go. InterfaceII failures are broken BVDiffEq tests now passing. FMI failure is weird - the ME `SymbolicDiscreteCallback` has a condition that is always true. It triggers at every solver step. The affect function changes nothing. Yet, it fails `CheckInit`.

PR #3620, opened 2025-05-10
fix change of var for sys with no equations

aml5600:
## Checklist

## Additional context

In the case were a system had no equations, like with `RealInput/Output`, performing `map(transform, get_eqs(sys))` would return `Any[]`, because it seemed like the type from `transform(ex)` couldn't be deduced. This small addition ensures that `typeof(ex) === typeof(transform(ex))` and solves this issue. A small test is added which would fail prior to the change.


aml5600:
Also, it seemed like with `JuliaFormatter` `v2.1.2` fails, and changes many files, with: `using JuliaFormatter, ModelingToolkit; format(joinpath(dirname(pathof(ModelingToolkit)), ".."))`

ChrisRackauckas (approved):

PR #3572, opened 2025-04-21
fix: copy initials to `u0` if `u0` not provided to `remake`

AayushSabharwal:
Close #3570

PR #3624, opened 2025-05-12
Less restrictive `promote_u0_p`

devmotion:
https://github.com/SciML/ModelingToolkit.jl/pull/3572 broke our downstream code that uses `p::Function`s (to be evaluated at integration time `t`) as parameters.

Given that none of the other involved functions restricts `p` explicitly to a `Union{AbstractArray,MTKParameters}`, IMO it seems natural to use `DiffEqBase.promote_u0` as a generic fallback and not only in the case `p::AbstractArray`.

### DIFF
--- a/src/systems/nonlinear/initializesystem.jl
+++ b/src/systems/nonlinear/initializesystem.jl
@@ -601,7 +601,7 @@ function promote_u0_p(u0, p::MTKParameters, t0)
     return u0, p
 end
 
-function promote_u0_p(u0, p::AbstractArray, t0)
+function promote_u0_p(u0, p, t0)
     return DiffEqBase.promote_u0(u0, p, t0), DiffEqBase.promote_u0(p, u0, t0)
 end
 


ChrisRackauckas:
Wouldn't `p::Function` go down the MTKParameters path with the splitting? Do you have an MWE we can test against here?

devmotion:
The system itself is not constructed with `::Function` parameters. The use case is a bit special, `remake` is called on an `ODEProblem` with an MTK system but `f` is explicitly set to a custom wrapper of the original ODE function that unpacks `::Function` parameters and `p` is explicitly set to such a `::Function` parameter.

ChrisRackauckas:
Oh is this the Chris E hack in Pumas for lifting the covariate constructions? I really don't like how that interacts with the interface... though we might as well keep it alive. Definitely not something to document telling people to do it 😅

devmotion:
Could we tag a new release of MTK?

PR #3625, opened 2025-05-12
handle connections in change_independent_variable

aml5600:
## Checklist

## Additional context

Expand `change_independent_variable` to handle systems with connections. This was chosen over forcing `expand_connections` prior to calling, as `expand_connections` forces flatness and that is undesirable in some cases.

### DIFF
--- a/src/systems/diffeqs/basic_transformations.jl
+++ b/src/systems/diffeqs/basic_transformations.jl
@@ -178,7 +178,7 @@ function change_independent_variable(
         return ex::T
     end
 
-    # overload to specifically handle equations, which can be an equation of a connection
+    # overload to specifically handle equations, which can be an equation or a connection
     function transform(eq::Equation, systems_map)
         if eq.rhs isa Connection
             eq = connect((systems_map[nameof(s)] for s in eq.rhs.systems)...)
--- a/src/systems/diffeqs/basic_transformations.jl
+++ b/src/systems/diffeqs/basic_transformations.jl
@@ -53,6 +53,17 @@ function liouville_transform(sys::AbstractODESystem; kwargs...)
     )
 end
 
+function split_eqs_connections(eqs_in::Vector{<:Equation})
+    eqs = Equation[]
+    cons = Equation[]
+
+    for eq in eqs_in
+        eq.lhs isa Connection ? push!(cons, eq) : push!(eqs, eq)
+    end
+
+    return eqs, cons
+end
+
 """
     change_independent_variable(
         sys::AbstractODESystem, iv, eqs = [];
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
> I'm not super familiar with the connect mechanics. But from what I understand, the main change is to recursively apply the change of independent variable transform to systems inside connections, which seems very sensible to me!

All connections are linear combinations so it should be fine to handle them the simple way.

src/systems/diffeqs/basic_transformations.jl
```diff
@@ -158,7 +158,7 @@ function change_independent_variable(
     function transform(ex::T) where {T}
         # 1) Replace the argument of every function; e.g. f(t) -> f(u(t))
         for var in vars(ex; op = Nothing) # loop over all variables in expression (op = Nothing prevents interpreting "D(f(t))" as one big variable)
-            is_function_of_iv1 = iscall(var) && isequal(only(arguments(var)), iv1) # of the form f(t)?
+            is_function_of_iv1 = iscall(var) && isequal(first(arguments(var)), iv1) # of the form f(t)?
```
ChrisRackauckas, line 161: why is this needed? It should be `only` or error?

ChrisRackauckas (approved):

ChrisRackauckas:
This looks reasonable, though I'd be scared of changing that `only` because I presume this shouldn't ever hit a two call variable?

aml5600:
> This looks reasonable, though I'd be scared of changing that `only` because I presume this shouldn't ever hit a two call variable?

@ChrisRackauckas This was an artifact of me trying to allow symbolic arrays to be similarly handled. It was failing previously as the only was hitting cases of e.g. `only([var(t), index])`. I was going to remove it as I switched to connections first, but it didn't seem to break anything so I left it...

aml5600:
Also, I tried implementing this first by simply transforming the systems in each connection and then rebuilding the connection (without the use of the `systems_map` approach). This gave odd errors when trying to display the resulting system (so `show/display` calls) and also was unable to structurally simplify. 

Is this something you would expect?

ChrisRackauckas:
Retriggering tests post scimlbase hotfix

ChrisRackauckas:
> Also, I tried implementing this first by simply transforming the systems in each connection and then rebuilding the connection (without the use of the systems_map approach). This gave odd errors when trying to display the resulting system (so show/display calls) and also was unable to structurally simplify.

That is not something I'd expect. I'm not sure why the display would care? You may not have reconstructed the `connect` equation correctly, i.e. it's not really an equation so check the conversion what the lhs should be.

src/systems/diffeqs/basic_transformations.jl
aml5600: reverted back to `only`, @hersle PR handles this correctly.

aml5600:
> > Also, I tried implementing this first by simply transforming the systems in each connection and then rebuilding the connection (without the use of the systems_map approach). This gave odd errors when trying to display the resulting system (so show/display calls) and also was unable to structurally simplify.
> 
> That is not something I'd expect. I'm not sure why the display would care? You may not have reconstructed the `connect` equation correctly, i.e. it's not really an equation so check the conversion what the lhs should be.

Gotcha. If we are okay with the current approach (which saves us from re-transforming a bunch of systems) I will just use this.

Issue #3623, opened 2025-05-11
Change_independent_variable and connections


aml5600:
**Question❓**

It looks like the `change_independent_variable` transformation needs to have the connections expanded. This fails:
```
    @mtkmodel ConnectSys begin
        @components begin
            in = RealInput()
            out = RealOutput()
        end
        @variables begin
            x(t)
            y(t)
        end
        @equations begin
            connect(in, out)
            in.u ~ x
            D(x) ~ out.u
            D(y) ~ 1
        end
    end
    @named sys = ConnectSys()
    sys = complete(sys; flatten = false)
    new_sys = change_independent_variable(sys, sys.y)
    ss = structural_simplify(new_sys)
```
failing on `structural_simplify` stuff. The systems from the connection equation are still wrt `t`. 

@hersle is it possible to work through the connection systems similar to how you did everything else? should we enforce that all connections are expanded before calling the transformation? Thanks!

hersle:
I haven't tested it on systems with connections. But a connection is equivalent to one or more equations that it expands to, right? If `change_independent_variable` can be adapted to do the substitution on the components inside the `connect(...)` statements, I think it should work. Can one do that without expanding the connection equations?

Otherwise, like you said, a workaround is to expand connections first.

aml5600:
So I tried expanding connections before changing variables, but this necessarily flattens the system before going into `change_independent_variable`. Which then the resulting system is flat, and that breaks all my downstream stuff.

I tried overloading the `transform` function to specialize on `Equation`, and another to specialize on `Connection`, and an inbetween to glue the two. The `Connection` overload transforms each system in `connection.systems`.

The function would succeed, but then something was failing when trying to show/display the system in the REPL, so looks like I didn't reassemble something correctly =/ will keep trying, though.

hersle:
> I tried overloading the transform function to specialize on Equation, and another to specialize on Connection, and an inbetween to glue the two. The Connection overload transforms each system in connection.systems.

That would also be my first thought, to just recursively apply `change_independent_variable` to the systems in the connection.

aml5600:
@hersle I went ahead and made #3625 . Let me know what you think.

PR #3626, opened 2025-05-12
Change independent variable of ODEs with array variables

hersle:
@aml5600 I found [your Slack question](https://julialang.slack.com/archives/CN04R7WKE/p1746959467897079). I also need it to work with array variables 😅
I think this should work. Thanks for writing the test!

Note that I don't use the `@mtkmodel` macro in the test.
I'm not completely sure how it works with the `NoValue()` stuff.
I'm sure there is a way to get that to work, too, which might be unrelated to `change_independent_variable` (?)

## Checklist

### DIFF
--- a/src/systems/diffeqs/basic_transformations.jl
+++ b/src/systems/diffeqs/basic_transformations.jl
@@ -153,13 +153,25 @@ function change_independent_variable(
     @set! sys.eqs = [get_eqs(sys); eqs] # add extra equations we derived
     @set! sys.unknowns = [get_unknowns(sys); [iv1, div2_of_iv1]] # add new variables, will be transformed to e.g. t(u) and uˍt(u)
 
+    # A utility function that returns whether var (e.g. f(t)) is a function of iv (e.g. t)
+    function is_function_of(var, iv)
+        # Peel off outer calls to find the argument of the function of
+        if iscall(var) && operation(var) === getindex # handle array variables
+            var = arguments(var)[1] # (f(t))[1] -> f(t)
+        end
+        if iscall(var)
+            var = only(arguments(var)) # e.g. f(t) -> t
+            return isequal(var, iv)
+        end
+        return false
+    end
+
     # Create a utility that performs the chain rule on an expression, followed by insertion of the new independent variable:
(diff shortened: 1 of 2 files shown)


src/systems/diffeqs/basic_transformations.jl
```diff
@@ -153,13 +153,25 @@ function change_independent_variable(
...
+        end
+        return false
+    end
+
     # Create a utility that performs the chain rule on an expression, followed by insertion of the new independent variable:
     # e.g. (d/dt)(f(t)) -> (d/dt)(f(u(t))) -> df(u(t))/du(t) * du(t)/dt -> df(u)/du * uˍt(u)
     function transform(ex::T) where {T}
         # 1) Replace the argument of every function; e.g. f(t) -> f(u(t))
         for var in vars(ex; op = Nothing) # loop over all variables in expression (op = Nothing prevents interpreting "D(f(t))" as one big variable)
-            is_function_of_iv1 = iscall(var) && isequal(only(arguments(var)), iv1) # of the form f(t)?
```
aml5600, line 161: in my other PR (#3623 ) I simply change `only` to `first`. This handled a simple, scalarized array problem (assuming that a default value was provided to side-step the `NoValue` issues)

aml5600:
@hersle yeah, I would imagine the `NoValue` stuff is orthogonal to the transformation. Unfortunately I used `@mtkmodel` heavily while learning MTK and still have it used quite a lot =/ 

This is a good step forward though, thank you!

src/systems/diffeqs/basic_transformations.jl
hersle: I don't think that's 100% correct. For example:
```julia
@variables t z(t)[1:2]
var = z[1]
ModelingToolkit.arguments(ModelingToolkit.unwrap(var)) # gives [z(t), 1]
```
So `first` would give `z(t)`, but we are really looking for `t`, i.e. the variable that `var` is a function of.

I think it's best to keep the `only` to "assert" that we are handling a symbolic expression with only one argument (e.g. `z(t)`).

aml5600: that's fair. It did work though 🙃  I will revert this line in my other pr

ChrisRackauckas (approved):

PR #3643, opened 2025-05-20
Fix @mtkmodel to work without using ModelingToolkit

### LINKED ISSUE #3640: `@mtkmodel` macro requires imports from MTK
**Describe the bug 🐞**
The `@mtkmodel` macro expands to code which relies on certain names beeing presend in the calling context, for example `Num` and `Equation`. As of https://github.com/SciML/ModelingToolkit.jl/pull/3526 also `System` (see [this line](https://github.com/SciML/ModelingToolkit.jl/blame/5346ce784a3696020695aa9feee64ad42204c7dc/src/systems/model_parsing.jl#L131))

This is breaking in contexts which use only implicit imports, for example I hadn't imported `System` before. I guess everything should be called via `ModelingToolkit.$sym`? But in general you cannot even assume that `ModelingToolkit` is defined in the calling module...

**Expected behavior**

Only depend on `using Modelingtoolkit: @mtkmodel` to use `@mtkmodel` macro.

```julia
using ModelingToolkit: ModelingToolkit, @mtkmodel, @variables, @parameters, t_nounits as t # those kinda make sense
using ModelingToolkit: Equation, Num, System # <- this is needed
@mtkmodel MyModel begin
    @variables begin
        x(t)
        y(t)
    end
    @parameters begin
        a
    end
    @equations begin
        y ~ a * x
    end
end
MyModel(; name=:foo)
```

cstjean:
Fix #3640 

Incidentally,  `include("test/model_parsing.jl")` at the REPL fails with 

```julia
ERROR: LoadError: IOError: stat("//www.w3.org/2000\\svg\" width=\"80\" height=\"30\">\n<path d=\"M10 15\nl15 0\nl2.5 -5\nl5 10\nl5 -10\nl5 10\nl5 -10\nl5 10\nl2.5 -5\nl15 0\" stroke=\"black\" stroke-width=\"1\" stroke-linejoin=\"bevel\" fill=\"none\"><\\path>\n<\\svg>\n"): unknown error (UNKNOWN)
```
on Windows. Is there any easy fix for that?

### DIFF
--- a/src/systems/model_parsing.jl
+++ b/src/systems/model_parsing.jl
@@ -42,8 +42,9 @@ function flatten_equations(eqs::Vector{Union{Equation, Vector{Equation}}})
 end
 
 function _model_macro(mod, fullname::Union{Expr, Symbol}, expr, isconnector)
+    MTK = ModelingToolkit
     if fullname isa Symbol
-        name, type = fullname, :System
+        name, type = fullname, :($MTK.System)
     else
         if fullname.head == :(::)
             name, type = fullname.args
@@ -74,9 +75,9 @@ function _model_macro(mod, fullname::Union{Expr, Symbol}, expr, isconnector)
     push!(exprs.args, :(parameters = []))
     # We build `System` by default; vectors can't be created for `System` as it is
     # a function.
-    push!(exprs.args, :(systems = ModelingToolkit.AbstractSystem[]))
-    push!(exprs.args, :(equations = Union{Equation, Vector{Equation}}[]))
-    push!(exprs.args, :(defaults = Dict{Num, Union{Number, Symbol, Function}}()))
+    push!(exprs.args, :(systems = $MTK.AbstractSystem[]))
+    push!(exprs.args, :(equations = Union{$MTK.Equation, Vector{$MTK.Equation}}[]))
+    push!(exprs.args, :(defaults = Dict{$MTK.Num, Union{Number, Symbol, Function}}()))
 
     Base.remove_linenums!(expr)
     for arg in expr.args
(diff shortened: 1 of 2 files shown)


src/systems/model_parsing.jl
```diff
@@ -74,9 +75,9 @@ function _model_macro(mod, fullname::Union{Expr, Symbol}, expr, isconnector)
     push!(exprs.args, :(parameters = []))
     # We build `System` by default; vectors can't be created for `System` as it is
     # a function.
-    push!(exprs.args, :(systems = ModelingToolkit.AbstractSystem[]))
-    push!(exprs.args, :(equations = Union{Equation, Vector{Equation}}[]))
-    push!(exprs.args, :(defaults = Dict{Num, Union{Number, Symbol, Function}}()))
+    push!(exprs.args, :(systems = $MTK.AbstractSystem[]))
```
AayushSabharwal, line 78: Is prepending `$MTK.` to everything really the solution? What if we did this instead:
```suggestion
    push!(exprs.args, :(systems = $AbstractSystem[]))
```

Also wouldn't this need to be done in more places? e.g. where `@variables` is called.

cstjean: See the discussion around https://discourse.julialang.org/t/adding-method-to-function-in-macro/128613/6 . If you're going to `esc` the expression, then you need either `$MTK.AbstractSystem` or `$AbstractSystem`, but there are places where the latter will fail. In particular, you can't `$@some_macro` obviously. I always use the former style for that reason, but I don't mind changing this PR to `$AbstractSystem` if that's your preference.

> Also wouldn't this need to be done in more places? 

Yes, absolutely. I didn't want to venture outside of my MTK comfort zone.

> e.g. where @variables is called.

?

cstjean: Yes, I always prefix functions, types and macros with `$MyModule.` when using `esc` It's the flip side of skipping macro hygiene.

In this particular case it doesn't matter because the user will feel compelled to `import ModelingToolkit: @variables` anyway, but that's not reliable in general.

PR #3656, opened 2025-05-22
refactor: rename `@mtkbuild` to `@mtkcompile` and `structural_simplify` to `mtkcompile`

AayushSabharwal:


ChrisRackauckas (approved):

ChrisRackauckas:
Deprecation path?

src/systems/systems.jl
```diff
@@ -158,6 +158,10 @@ function __mtkcompile(sys::AbstractSystem; simplify = false,
     end
 end
 
+@deprecate structural_simplify(sys; kwargs...) mtkcompile(sys; kwargs...)
+@deprecate structural_simplify(sys, io; kwargs...) mtkcompile(
+    sys; inputs = io[1], outputs = io[2], kwargs...)
```
ChrisRackauckas, line 163: Can we collect all into one file?

PR #3658, opened 2025-05-23
docs: overhaul documentation for v10

AayushSabharwal:
Assumes https://github.com/SciML/ModelingToolkit.jl/pull/3648 and https://github.com/SciML/ModelingToolkit.jl/pull/3651 are merged. In general, this will assume any WIP v10 PR is merged.


ChrisRackauckas:
I think this already looks pretty good. Much simpler now that the other systems are gone: the `System` page is going to be super useful now.

AayushSabharwal:
There's still so much that isn't covered 😅 I plan to at least get all public API functions in somewhere, add docstrings for all of them, and then make sure docs actually build.

PR #3666, opened 2025-05-28
refactor: remove `parameter_dependencies` kwarg

AayushSabharwal:


AayushSabharwal:
`parameter_dependencies` as a kwarg throws a warning and puts the equations in with the main equations now. We separate them during `complete` or `mtkcompile`. The only way to populate that field is via `@set`.

PR #3674, opened 2025-05-29
fix: fix type promotion in `late_binding_update_u0_p` with non-dual types

AayushSabharwal:
Port of #3673 to v10


ChrisRackauckas:
Rebase?

PR #3676, opened 2025-05-29
feat: PyomoDynamicOptProblem, docs

vyudu:
@AayushSabharwal @ChrisRackauckas This is mostly the last of it, the refactor, Pyomo, and docs stuff are in here

### DIFF
--- a/test/extensions/dynamic_optimization.jl
+++ b/test/extensions/dynamic_optimization.jl
@@ -69,9 +69,8 @@ const M = ModelingToolkit
     pprob = PyomoDynamicOptProblem(
         lksys, [u0map; parammap], tspan; guesses = guess, dt = 0.01)
     psol = solve(pprob, PyomoCollocation("ipopt", LagrangeLegendre(3)))
-    @show psol.sol
-    @test psol.sol(0.6)[1] ≈ 3.5
-    @test psol.sol(0.3)[1] ≈ 7.0
+    @test psol.sol(0.6; idxs = x(t)) ≈ 3.5
+    @test psol.sol(0.3; idxs = x(t)) ≈ 7.0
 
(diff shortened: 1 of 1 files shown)


docs/src/tutorials/dynamic_optimization.md
```diff
@@ -0,0 +1,155 @@
...
+tspan = (0.0, tf)
+parammap = [u(t) => 0.0, tf => 1.0]
+```
+The `tf` mapping in the parameter map is treated as an initial guess.
+
+Please note that, at the moment, free final time problems cannot support constraints defined at definite time values, like `x(3) ~ 2`.
+
+!!! warning
+    
+    The Pyomo collocation methods (LagrangeRadau, LagrangeLegendre) currently are bugged for free final time problems. Strongly suggest using BackwardEuler() for such problems when using Pyomo as the backend.
```
ChrisRackauckas, line 97: We should make that error in the extension?

vyudu:
The collocation tests pass on latest but fail on lts - should I try to fix that?

ChrisRackauckas:
Does Pyomo.jl fail on LTS? It would be good to keep LTS working when possible.

ChrisRackauckas:
I think this is good for now. Looking at all of this, I do wonder if there could be a common problem type that the different solvers build from, but there is some unique codegen for them so it is difficult.

vyudu:
I think we could, would just need to initialize the model and all the constraints at the time `solve` is called and not when the problem is created.

vyudu:
Also down the line, it seems like another class of solvers we could add ones based on the Pontryagin maximum principle, see things like https://link.springer.com/article/10.1007/s10589-011-9454-7

ChrisRackauckas:
yes we should, can you open an issue?

PR #3681, opened 2025-06-02
refactor: BLT sort equations and variables in tearing

AayushSabharwal:
This turned out to be more work than I thought, but it works!


test/initial_values.jl
```diff
@@ -182,7 +183,7 @@ end
     prob = ODEProblem(sys, [], (0.0, 1.0))
     sol = solve(prob)
     @test SciMLBase.successful_retcode(sol)
-    @test sol.u[1] ≈ [1.0, 1.0, 0.5, 0.5]
+    @test sol[x, 1] ≈ [1.0, 1.0, 0.5, 0.5]
```
ChrisRackauckas, line 186: This would be scalar?

ChrisRackauckas (approved):

test/initial_values.jl
AayushSabharwal: `x` is an array symbolic

> `@variables x(t)[1:4] y(t)[1:2]`

hersle:
Does this apply to all system types, e.g. ODEs?

AayushSabharwal:
Yes, all system types will have their equations sorted in BLT order.

ChrisRackauckas:
well there is only one system type 😅

AayushSabharwal:
Yeah, I guess better phrasing is "all types of systems"

hersle:
Ok, that was what I hoped and suspected, great! :smiley:

Are there any current plans to extend the strongly coupled logic to other problems as well, such as `ODEProblem`s?
Or is it easy to manually break a `System` into independent BLT blocks, to solve it as a sequence of smaller `*Problems`?

AayushSabharwal:
The FMI test fails because it relies on variable order. I have a fix locally and verified that it works. Will let the rest of CI run, and if it passes I'll just push and merge.

PR #3683, opened 2025-06-02
refactor: remove old `==` and `hash` implementation for `System`

AayushSabharwal:
## Checklist


isaacsas:
Should this be something we should drop in Catalyst too then? Is there now no notion of system equality, or is there another approach we should be taking?

AayushSabharwal:
Yes. Its utility is questionable at best and the implementation is difficult and bug-prone. For the longest time, `hash` and `==` didn't agree which is bad.

ChrisRackauckas:
Yes, this was an intended change with the v10 that got missed somehow. But basically, the fallback `==` definition is more accurate than this overload which misses a few things.

ChrisRackauckas (approved):

ChrisRackauckas:
Can merge, but we should make sure https://github.com/SciML/ModelingToolkit.jl/pull/3681 has all downstream green and passing before. We have enough downstream updated now that even extensions should all be fine, sans SI

PR #3692, opened 2025-06-04
[v9] fix: fix major compile time regression due to `concrete_getu`

AayushSabharwal:


src/systems/problem_utils.jl
```diff
@@ -617,30 +617,40 @@ struct ReconstructInitializeprob{GP, GU}
...
-    end
-    split_syms = Tuple(split_syms)
-    # the getter is now type-stable, and we can vcat it to get the full buffer
-    return Base.Fix1(reduce, vcat) ∘ getu(indp, split_syms)
+function.
+
+Note that the getter ONLY works for problem-like objects, since it generates an observed
+function. It does NOT work for solutions.
+"""
+Base.@nospecializeinfer function concrete_getu(indp, syms::AbstractVector)
```
DhairyaLGandhi, line 650: Given the docstring, this could be typed to avoid mistakenly calling it with other types.

AayushSabharwal: The returned callable has this restriction. I'll have to think about how to handle that case, since it should just work on anything that isn't a solution or a lone parameter object/state vector.

PR #3696, opened 2025-06-05
[v9] fix: remove CSE hack, fix unscalarized variables in initializeprobpmap

AayushSabharwal:


AayushSabharwal:
Needs https://github.com/SciML/ModelingToolkit.jl/pull/3698

AayushSabharwal:
InterfaceII fails because I forgot to change `@mtkcompile` to `@mtkbuild`. Fixed locally and verified that the test passes - will push and merge once everything else is green.

PR #3699, opened 2025-06-05
feat: allow calling `generate_control_function` with unsimplified system

AayushSabharwal:
## Checklist

### DIFF
--- a/src/inputoutput.jl
+++ b/src/inputoutput.jl
@@ -198,13 +198,11 @@ function generate_control_function(sys::AbstractSystem, inputs = unbound_inputs(
         simplify = false,
         eval_expression = false,
         eval_module = @__MODULE__,
-        check_simplified = true,
         kwargs...)
-    # Remove this when the ControlFunction gets merged.
-    if check_simplified && !iscomplete(sys)
-        error("A completed `ODESystem` is required. Call `complete` or `mtkcompile` on the system before creating the control function.")
-    end
     isempty(inputs) && @warn("No unbound inputs were found in system.")
+    if !iscomplete(sys)
+        sys = mtkcompile(sys; inputs, disturbance_inputs)
+    end
     if disturbance_inputs !== nothing
         # add to inputs for the purposes of io processing
         inputs = [inputs; disturbance_inputs]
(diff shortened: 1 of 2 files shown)


src/inputoutput.jl
```diff
@@ -198,13 +198,11 @@ function generate_control_function(sys::AbstractSystem, inputs = unbound_inputs(
...
         eval_expression = false,
         eval_module = @__MODULE__,
-        check_simplified = true,
         kwargs...)
-    # Remove this when the ControlFunction gets merged.
-    if check_simplified && !iscomplete(sys)
-        error("A completed `ODESystem` is required. Call `complete` or `mtkcompile` on the system before creating the control function.")
-    end
     isempty(inputs) && @warn("No unbound inputs were found in system.")
+    if !iscomplete(sys)
```
baggepinnen, line 203: `sys` may be complete without being `mtkcompile`d?

test/input_output_handling.jl
```diff
@@ -267,8 +267,8 @@ eqs = [connect_sd(sd, mass1, mass2)
 @named _model = System(eqs, t)
 @named model = compose(_model, mass1, mass2, sd);
 
-model = mtkcompile(model, inputs = [u])
-f, dvs, ps, io_sys = ModelingToolkit.generate_control_function(model, simplify = true)
+f, dvs, ps, io_sys = ModelingToolkit.generate_control_function(
+    model; inputs = [u]simplify = true)
```
baggepinnen:
```suggestion
    model; inputs = [u], simplify = true)
```

baggepinnen (commented):
This functionality is tested in `downstream/test_disturbance_model.jl`, it would be nice to verify that those passes since I don't think they run at the moment?

src/inputoutput.jl
SebastianM-C: `ModelingToolkit.isscheduled` is for structural simplification, you can `complete` without that.

AayushSabharwal: I assumed the user may provide a handwritten model in the correct format that they `complete` without running `mtkcompile`. If this is not a case we want to handle, then I can just use `isscheduled`.

AayushSabharwal:
`test_disturbance_model.jl` relies on ControlSystemsMTK, so it cannot run at the moment.

ChrisRackauckas (approved):

AayushSabharwal:
This PR is also necessary to make docs pass

PR #3700, opened 2025-06-05
ci: add benchmark CI

AayushSabharwal:
This adds a small benchmark workflow to run on PRs so we can be aware of any regressions such as the one fixed by #3693


AayushSabharwal:
As far I can see, Benchmark fails because the script doesn't exist on master

ChrisRackauckas (approved):

PR #3702, opened 2025-06-06
fix: fix `get_mtkparameters_reconstructor` handling of nonnumerics

AayushSabharwal:


ChrisRackauckas (approved):

ChrisRackauckas:
sensitivity core 8 is still failing? @DhairyaLGandhi

AayushSabharwal:
Somehow it's the same error.

DhairyaLGandhi:
It's not the same error, and is actually something we'd addressed in the SciMLBase workaround. We can add that dispatch to Zygote pretty safely.

PR #3710, opened 2025-06-09
refactor: format

AayushSabharwal:
## Checklist

### DIFF
--- a/docs/src/API/variables.md
+++ b/docs/src/API/variables.md
@@ -293,7 +293,7 @@ For systems that contain parameters with metadata like described above, have som
 In the example below, we define a system with tunable parameters and extract bounds vectors
 
 ```@example metadata
-@variables x(t)=0 u(t)=0 [input = true] y(t)=0 [output = true]
+@variables x(t)=0 u(t)=0 [input=true] y(t)=0 [output=true]
 @parameters T [tunable = true, bounds = (0, Inf)]
 @parameters k [tunable = true, bounds = (0, Inf)]
 eqs = [D(x) ~ (-x + k * u) / T # A first-order system with time constant T and gain k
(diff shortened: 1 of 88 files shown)


ChrisRackauckas:
Formatting failed 😅

AayushSabharwal:
Yeah, JuliaFormatter v2 doesn't like `bipartite_graphs.jl` for some reason and errors out in CI. This is formatted with v1 locally.

AayushSabharwal:
Would you prefer I make CI use v1?

ChrisRackauckas:
No we should update to v2. We should just change the code to make it happy then.

AayushSabharwal:
I'll see if I can figure out what it is unhappy about. The error message is incredibly unhelpful. There also doesn't seem to be an option to increase the fixpoint iterations.

AayushSabharwal:
```julia
struct Matching{U, V <: AbstractVector} <: AbstractVector{Union{U, Int}} #=> :Unassigned =#
    match::V
    inv_match::Union{Nothing, V}
end
```

It specifically doesn't like that inline comment.

ChrisRackauckas:
just put it above the struct.

AayushSabharwal:
https://github.com/domluna/JuliaFormatter.jl/issues/924

AayushSabharwal:
The comment is not very meaningful anyway, and putting it above the struct also fails. I've removed it. This will be a massive PR, unfortunately.

AayushSabharwal:
I'll push the changes, but I find a lot of it pretty questionable

ChrisRackauckas:
conflict

AayushSabharwal:
Well yes, but I don't think upgrading to JuliaFormatter@2 is worth it. Most of these changes are nonsensical and worse code quality.

AayushSabharwal:
It's even split multidimensional array indexing into multiple lines in some places for some reason

ChrisRackauckas:
Point them all out for @domluna

PR #3711, opened 2025-06-09
fix: fix namespacing of variables of variables

AayushSabharwal:
Required for https://github.com/JuliaControl/ControlSystemsMTK.jl/pull/76

Fixes behavior that only coincidentally worked in v9


AayushSabharwal:
`pre` failures look like typical Enzyme/Julia failures on pre.

PR #3735, opened 2025-06-13
switch parse_costs! and parse_constraints

ctessum:
It seems like they are currently backward

### DIFF
--- a/src/systems/model_parsing.jl
+++ b/src/systems/model_parsing.jl
@@ -1160,7 +1160,7 @@ function parse_constraints!(cons, dict, body)
     Base.remove_linenums!(body)
     for arg in body.args
         push!(cons, arg)
-        push!(dict[:constraints], readable_code.(cons)...)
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
@AayushSabharwal better test this?

AayushSabharwal:
I'm confused, because we have a test for this and it passes both before and after the swap on CI, but fails after swapping locally.

AayushSabharwal:
@ctessum how did you run into this? Are you using `MTK.get_gui_metadata`?

ctessum:
I don't have a specific case of it failing, I was just looking through the code and it seemed wrong.

AayushSabharwal:
I see. So the functions are identical, and because they were passed the correct arrays it worked. i.e. in the `@costs` case, it called `parse_constraints!` but passed it the `costs` array. Constructing the system thus worked fine, it's just that in the `Model` struct the values are backward.

PR #3736, opened 2025-06-13
refactor: rewrite connection expansion implementation

AayushSabharwal:


src/systems/connectors.jl
```diff
@@ -28,6 +28,8 @@ function connect(sys1::AbstractSystem, sys2::AbstractSystem, syss::AbstractSyste
     Equation(Connection(), Connection(syss)) # the RHS are connected systems
 end
 
+const _debug_mode = Base.JLOptions().check_bounds == 1
```
ChrisRackauckas, line 31: what's this for?

PR #3745, opened 2025-06-17
Ensure DDEs are univariate

hersle:
Can you reconsider this as an uncontroversial alternative to https://github.com/SciML/ModelingToolkit.jl/pull/3738?
It also solves https://github.com/SciML/ModelingToolkit.jl/issues/3737, but avoids testing it if that is controversial.

@AayushSabharwal / @ChrisRackauckas


hersle:
Merge this for now? :innocent: 

Or https://github.com/SciML/ModelingToolkit.jl/pull/3738, but I do not seem to have convinced you entirely :sweat_smile:

PR #3753, opened 2025-06-20
fix: fix `get_updated_symbolic_problem`

AayushSabharwal:
Close https://github.com/SciML/SciMLSensitivity.jl/issues/1219

Respect `u0` and `p` passed as kwargs


ChrisRackauckas (approved):

AayushSabharwal:
This is ready. Downstream won't pass anyway, (1, Extensions) is an intermittent segfault in Pycall and `pre` is uncooperative anyway.

Issue #3757, opened 2025-06-22
Perturbation example figure broken by v10


moble:
The figure at the bottom of [this section](https://docs.sciml.ai/ModelingToolkit/stable/examples/perturbation/#Weakly-nonlinear-oscillator) claims to be ["similar to Figure 7.6.2 in *Nonlinear Dynamics and Chaos*"](https://github.com/SciML/ModelingToolkit.jl/blob/7ab2feb577cfd83a846c4873e4e5686e76f3bd4c/docs/src/examples/perturbation.md?plain=1#L105).  The cited figure shows an exponentially damped sinusoid, but the figure in the docs appears to just show a few data points, and does not at all resemble the figure in the book.

The reason appears to be that the computed solution is all zeros, and integrates in just a few steps:

```julia
retcode: Success
Interpolation: 3rd order Hermite
t: 8-element Vector{Float64}:
  0.0
  9.999999999999999e-5
  0.0010999999999999998
  0.011099999999999997
  0.11109999999999996
  1.1110999999999995
 11.111099999999993
 50.0
u: 8-element Vector{Vector{Float64}}:
 [0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
 [0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
 [0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
 [0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
 [0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
 [0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
 [0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
 [0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
```

hersle:
This can probably be closed.

PR #3758, opened 2025-06-22
Fix plot in perturbation example

hersle:
https://github.com/SciML/ModelingToolkit.jl/issues/3757

The problem was just that `u0` was defined in a sketchy way, with both `(yˍt(t))[0] => 0.0` and `Differential(t)((y(t))[0]) => 1.0` present, so it was ambiguous which one should be used.


hersle:
bilde

PR #3761, opened 2025-06-24
allow DataInterpolations@7

oscardssmith:


oscardssmith:
oops. That was a bad race condition. Sorry @AayushSabharwal I'll revert.

oscardssmith:
that said, why would this cause tests to fail? were we stuck on v6 before?

AayushSabharwal:
We just didn't update to v7 or v8. Tests fail because they use a kwarg that is removed in later version.

oscardssmith:
now I'm confused. These tests ran on DataInterpolations v8, so this PR wasn't responsible for failures? Rather than reverting, should we just remove the use of deprecated kwargs?

AayushSabharwal:
This PR bumped compat but didn't update tests. So tests failed because they used removed functionality. The PR just needs to update tests and up the lower bound.

oscardssmith:
ah right. Somehow I got it in my head that v8 was already supported prior to this PR, but that's just nonesense. Followup PR incoming

PR #3779, opened 2025-06-27
Trivial optimization in generate_initializesystem

hersle:
During profiling of `ODEProblem` in this example of my package, I found I saw being slowed down a lot by the line that calls `default_toterm` on every element in `keys(defs)` for every iteration in the loop, which I think could simply be precomputed before the loop.
```julia
using SymBoltz # https://github.com/hersle/SymBoltz.jl
using BenchmarkTools
M = ΛCDM() # a big System
Ms = mtkcompile(M)
pars = SymBoltz.parameters_Planck18(M) # some sym-to-num parameter dict
pars[M.k] = 1.0
@btime prob = ODEProblem(Ms, pars, (0.0, 10.0))
```
Before PR:
```
  2.607 s (27144415 allocations: 803.71 MiB)
```
After PR:
```
  995.616 ms (9386044 allocations: 317.20 MiB)
```
These changes give me a 3x speedup in construction of a big problem, and I think the logic should be equivalent before and after.

### DIFF
--- a/src/systems/nonlinear/initializesystem.jl
+++ b/src/systems/nonlinear/initializesystem.jl
@@ -71,7 +71,6 @@ function generate_initializesystem_timevarying(sys::AbstractSystem;
     eqs_ics = Equation[]
     defs = copy(defaults(sys)) # copy so we don't modify sys.defaults
     additional_guesses = anydict(guesses)
-    additional_initialization_eqs = Vector{Equation}(initialization_eqs)
     guesses = merge(get_guesses(sys), additional_guesses)
     idxs_diff = isdiffeq.(eqs)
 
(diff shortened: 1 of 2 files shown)


hersle:
One more
```
  717.067 ms (8363734 allocations: 280.19 MiB)
```

AayushSabharwal (approved):
Awesome, thanks for noticing this!

AayushSabharwal:
Waiting on https://github.com/SciML/NonlinearSolve.jl/pull/634 to rerun CI

hersle:
Re-run now and, if problem-free, get this in before the next release? :smiley: @AayushSabharwal

PR #3795, opened 2025-07-07
feat: more robust inputs/outputs handling

AayushSabharwal:


baggepinnen:
This PR would benefit from running DyadControlSystems tests when it's ready

test/input_output_handling.jl
```diff
@@ -468,3 +468,20 @@ end
     x = [1.0]
     @test_nowarn f[1](x, u, p, 0.0)
 end
+
+@testset "Observed inputs and outputs" begin
+    @variables x(t) y(t) [input = true] z(t) [output = true]
+    eqs = [D(x) ~ x + y + z
+           y ~ z]
+    @named sys = System(eqs, t)
+    @test issetequal(ModelingToolkit.inputs(sys), [y])
```
baggepinnen, line 477: Is this the PR that should address the ordering an number of inputs after `mtkcompile(..., inputs = [u1,u2])`? If so, it would be nice to include a test case where this failed 
https://github.com/JuliaComputing/DyadControlSystems.jl/actions/runs/17095167938/job/48477801369?pr=644#step:7:1727


I notice also that the build that was triggered still fails in almost the same place, but it now has a different error message than it had before

AayushSabharwal: The ordering was already fixed in #3804. We have an assertion for the ordering being maintained in `complete`, so if by chance something goes wrong it will surface as an error. I don't know how we would test it, since there isn't a specific condition where the ordering goes wrong. It's just that if we're not careful it might get shuffled around, which is what the assertion detects.

The new failure in DyadControlSystems seems to me like a missing splat operation. It's trying to pass a `NamedTuple` of 4 matrices where the function expects them as 4 different arguments.

baggepinnen: Ah, so it has been a change in MTKv10 that went unnoticed before this fix then, I'll fix it separately.

baggepinnen (approved):

PR #3804, opened 2025-07-09
fix: fix bad input ordering, make downstream tests pass

AayushSabharwal:
## Checklist


src/linearization.jl
```diff
@@ -565,6 +565,51 @@ function linearize_symbolic(sys::AbstractSystem, inputs,
...
+        Here, `mtkcompile` expects the input to be named `input_var`, but since `sys`
+        performs namespacing, it will be named `sys$(NAMESPACE_SEPARATOR)input_var`. To \
+        fix this issue, namespacing can be temporarily disabled:
+
+        ```julia
+        @named sys = MyModel()
+        sys = toggle_namespacing(sys, false)
+        inputs = [sys.input_var]
+        sys = toggle_namespacing(sys, true)
+        mtkcompile(sys; inputs)
```
baggepinnen:
```suggestion
        sys_nns = toggle_namespacing(sys, false)
        inputs = [sys_nns.input_var]
        mtkcompile(sys; inputs)
```
would this work? It would save one line and does not make the code stateful

AayushSabharwal: Yes, that works

src/systems/abstractsystem.jl
```diff
@@ -670,6 +672,12 @@ function complete(
             end
             ordered_ps = vcat(
                 ordered_ps, reduce(vcat, ps_split; init = eltype(ordered_ps)[]))
+            if isscheduled(sys)
```
baggepinnen, line 675: I don't quite understand where the fix to the ordering happens. Will this assert `@assert issorted(input_idxs)` guarantee that the order expected by the inputs when they are passed as input arguments to generated control functions is the same as the order specified by the user?

AayushSabharwal: The fix to ordering is in the `IndexCache` constructor - that was the problem all along. This check here just enforces the invariant.

ChrisRackauckas:
Disable CASADI for LTS? And is there a standard library release we need to do?

PR #3813, opened 2025-07-13
Fix master tests

ChrisRackauckas:


AayushSabharwal:
Downstream is fixed in #3804

PR #3818, opened 2025-07-15
refactor: move `@compile_workload` to after exports

### LINKED ISSUE #3811: `@compile_workload` should run after any `export` statements
This is because of patterns like this:
```julia
module PackageModule
struct Foo end

module InnerModule
using ..PackageModule
foo() = cond() ? (x isa Foo) : nothing
end

Inner.foo() # when inferring foo(), Foo has no binding it can be resolved to
export Foo  # by exporting, Foo inside foo() suddenly resolves to a binding, so the code is invalidated

end
```

Essentially, an inner module has an implicit import (`using ..PackageModule`) which picks up an outer type `Foo`, but only after the `export Foo` is executed at the outer level.

The meaning of `Foo` changes once `export Foo` is executed, causing invalidation. As a concrete example, this kind of thing happens in practice for `StructuralTransformations._distribute_shift` and `export Pre` when pre-compiling current `master` MTK.

That's an easy bug to fix, and has a very small impact on latency anyway, but I wanted to open the issue for awareness.

AayushSabharwal:
Close #3811

### DIFF
--- a/src/ModelingToolkit.jl
+++ b/src/ModelingToolkit.jl
@@ -242,41 +242,6 @@ const D_nounits = Differential(t_nounits)
 const D_unitful = Differential(t_unitful)
 const D = Differential(t)
 
-PrecompileTools.@compile_workload begin
-    using ModelingToolkit
-    @variables x(ModelingToolkit.t_nounits)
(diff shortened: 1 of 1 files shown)


AayushSabharwal:
InterfaceII failures are unbroken `LinearProblem` tests, fixed on master.

PR #3837, opened 2025-07-23
Add symbolic tspan field to System struct (fixes #847)

### LINKED ISSUE #847: Symbolic tspan
We just need a tspan field in time dependent systems.

ChrisRackauckas:
## Summary

This PR implements the symbolic tspan feature requested in issue #847 by adding a `tspan` field to time-dependent systems in ModelingToolkit.jl.

## Changes

- **Added `tspan` field to System struct**: New field with type `Union{Nothing, Tuple{Any, Any}}` that can store symbolic or numeric time spans
- **Updated constructors**: Modified all System constructors to accept the `tspan` parameter with a default value of `nothing`
- **Automatic getter methods**: The field is accessible via `get_tspan()` and `has_tspan()` through the existing `SYS_PROPS` framework
- **Comprehensive tests**: Added test cases covering numeric tspan, symbolic tspan, nil values, and preservation through system completion

## Use Cases

This feature enables users to:
- Define default time spans directly in system definitions
- Use symbolic expressions for time spans (e.g., parameters)
- Create more self-contained system definitions
- Potentially validate simulation parameters against intended time horizons

## Example Usage

```julia
# Numeric tspan
@named sys1 = System(eqs, t, [x, y], [σ, ρ]; tspan = (0.0, 10.0))

# Symbolic tspan
@parameters t0 tf
@named sys2 = System(eqs, t, [x, y], [σ, ρ]; tspan = (t0, tf))

# Access the tspan
get_tspan(sys1)  # Returns (0.0, 10.0)
```

## Test plan

- [x] Verify tspan field is properly added to System struct
- [x] Test numeric tspan values
- [x] Test symbolic tspan expressions
- [x] Test nil/nothing tspan values
- [x] Verify tspan is preserved through system completion
- [x] Confirm getter methods work correctly

Closes #847


src/deprecations.jl
AayushSabharwal, line 1: This file should not be touched

src/problems/bvproblem.jl
AayushSabharwal, line 1: This should not try and change our problem constructor syntax. This applies for every file in this subdirectory - the `tspan` can be optional but it shouldn't be a kwarg. It should default to `get_tspan(sys)` and the check ensuring it is not `nothing` should be a function defined somewhere else that it calls in all the relevant problem constructors to ensure we have a consistent implementation and error message.

Also, this just takes the `tspan` from the system if not explicitly provided, but doesn't do the work of actually handling the case when the `tspan` is symbolic and needs to be substituted to get a number out. Should mutating the parameters in an instantiated `ODEProblem` also update the `tspan` accordingly? Because it doesn't do that. I _think_ `tspan` can be a function? In which case that is what this should target and we need a dedicated codegen function in `codegen.jl`.

src/systems/system.jl
AayushSabharwal, line 1: This adds the field to the system, but does not handle it in `flatten` and the other functions that operate on (hierarchical) systems.

test/odesystem.jl
AayushSabharwal, line 1: The tests are weirdly redundant, and don't actually test creating an `ODEProblem` with symbolic `tspan` - it just uses a normal numeric `tspan`.

AayushSabharwal (changes requested):

PR #3838, opened 2025-07-23
Fix unit validation for dimensionally compatible quantities (fixes #3787)

### LINKED ISSUE #3787: Validation of units broken
**Describe the bug 🐞**
Hello!
The example here :
https://docs.sciml.ai/ModelingToolkit/stable/basics/Validation/#DynamicQuantities-Literals
Does not work as intended. Its goal is to validate the dimensionalitty of the equations according to the units given.

**Expected behavior**

It should return "true" but does return "false".

```julia
using ModelingToolkit, DynamicQuantities
@parameters τ [unit = u"ms"]
@variables t [unit = u"ms"] E(t) [unit = u"kJ"] P(t) [unit = u"MW"]
D = Differential(t)
eqs = eqs = [D(E) ~ P - E / τ,
    0 ~ P]
ModelingToolkit.validate(eqs)
```

**Error & Stacktrace ⚠️**

```julia

```

**Environment (please complete the following information):**

  - Output of `using Pkg; Pkg.status()`

```julia
  [06fc5a27] DynamicQuantities v1.8.0
  [961ee093] ModelingToolkit v10.7.0
```
  - Output of `versioninfo()`

```julia
Julia Version 1.11.5
Commit 760b2e5b739 (2025-04-14 06:53 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 20 × 13th Gen Intel(R) Core(TM) i7-13800H
  WORD_SIZE: 64
  LLVM: libLLVM-16.0.6 (ORCJIT, goldmont)
Threads: 1 default, 0 interactive, 1 GC (on 20 virtual cores)
```

**Additional context**

Question first raised here (https://discourse.julialang.org/t/unit-validation-and-inspection-with-mtk-error-in-the-documentation/112145) . Please tell me if I do not use the right form in that issue, this is my first time doing so.

ChrisRackauckas:
## Summary

This PR fixes the unit validation system to properly handle dimensionally compatible quantities with different scales, resolving issue #3787.

## Problem

The unit validation was failing for physically meaningful equations like:
```julia
@parameters τ [unit = u"ms"]
@variables E(t) [unit = u"kJ"] P(t) [unit = u"MW"]
D = Differential(t)
eqs = [D(E) ~ P - E / τ, 0 ~ P]
ModelingToolkit.validate(eqs)  # returned false, should be true
```

The issue was that `kJ/ms = kW` and `MW` and `kW` are dimensionally compatible (both power units), but the validation used exact equality instead of dimensional compatibility.

## Root Cause

The `equivalent(x, y)` function in `src/systems/unit_check.jl` was defined as simply `isequal(x, y)`, which meant:
- `1.0e6 m² kg s⁻³` (MW) and `1.0e3 m² kg s⁻³` (kW) were not considered equivalent
- Even though they have the same dimensions (power), different scales caused validation failure

## Solution

1. **Enhanced `equivalent()` function**: Now checks dimensional compatibility using `DQ.dimension()` for DynamicQuantities instead of exact equality
2. **Added `equivalent_strict()` function**: Maintains exact equality for connection validation to ensure electrical safety
3. **Preserved connection safety**: Connections between different scales (V vs mV) still fail validation as intended

## Changes

- **Fixed equation validation**: Dimensionally compatible units now validate correctly
- **Preserved connection validation**: Electrical connectors still require exact unit matches for safety
- **Added comprehensive tests**: Including the original bug case and edge cases

## Test Cases

✅ **Issue #3787 case**: `MW` and `kJ/ms` (= kW) now validate as compatible  
✅ **Scale compatibility**: `m` and `km` validate as compatible  
✅ **Connection safety**: `V` and `mV` connectors still fail validation  
✅ **Incompatible dimensions**: `m` and `s` still fail validation  

## Example Usage

```julia
# This now works (was broken before)
@parameters τ [unit = u"ms"]
@variables E(t) [unit = u"kJ"] P(t) [unit = u"MW"]
D = Differential(t)
eqs = [D(E) ~ P - E / τ, 0 ~ P]
ModelingToolkit.validate(eqs)  # returns true ✅

# Connections still maintain safety requirements
@named pin_v = Pin()    # uses u"V"
@named pin_mv = Pin()   # uses u"mV"  
connect(pin_v, pin_mv)  # still fails validation ✅
```

Closes #3787

### DIFF
--- a/src/systems/unit_check.jl
+++ b/src/systems/unit_check.jl
@@ -104,7 +104,27 @@ function get_unit(op::Integral, args)
     return get_unit(args[1]) * unit
 end
 
-equivalent(x, y) = isequal(x, y)
+function equivalent(x, y)
+    # For DynamicQuantities, check dimensional compatibility rather than exact equality
+    if x isa DQ.AbstractQuantity && y isa DQ.AbstractQuantity
+        return DQ.dimension(x) == DQ.dimension(y)
+    elseif !(x isa DQ.AbstractQuantity) && !(y isa DQ.AbstractQuantity)
+        # Both are not quantities, use exact equality
+        return isequal(x, y)
+    else
+        # One is quantity, one is not - check if quantity is dimensionless
+        q = x isa DQ.AbstractQuantity ? x : y
+        non_q = x isa DQ.AbstractQuantity ? y : x
+        # Only consider equivalent if the quantity is dimensionless and other is zero/number
+        return DQ.dimension(q) == DQ.dimension(DQ.Quantity(1.0)) && (isequal(non_q, 0) || non_q isa Number)
+    end
+end
+
+# For connections, we may want stricter equivalence (same scale for safety)
+function equivalent_strict(x, y)
+    # Use exact equality for connections to ensure same scales
+    return isequal(x, y)
+end
 function get_unit(op::Conditional, args)
     terms = get_unit.(args)
     terms[1] == unitless ||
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
This is unsafe.

PR #3873, opened 2025-08-06
Add nlstep_compile as a debug tool for turning off simplification

ChrisRackauckas:
This can be used in order to test correctness and differences with the nlstep interface as it should exactly match the NonlinearSolveAlg


ChrisRackauckas:
@oscardssmith

oscardssmith:
shouldn't this be labeled nlstep_simplify?

ChrisRackauckas:
Well we call it mtkcompile in v10, not structural_simplify. It should only be internal though, I don't think it should be documented since there isn't a user level reason to run nlstep without simplification.

Issue #3859, opened 2025-07-31
`split` not forwarded from `generate_control_function`


baggepinnen:
In MTKv10 `split = false` does not appear to work as intended in `generate_control_function`

AayushSabharwal:
Closed via #3875

PR #3875, opened 2025-08-07
Forward `split` to `mtkcompile`

DhairyaLGandhi:
## Checklist

## Additional context

Ref https://github.com/SciML/ModelingToolkit.jl/issues/3859

### DIFF
--- a/src/inputoutput.jl
+++ b/src/inputoutput.jl
@@ -198,10 +198,11 @@ function generate_control_function(sys::AbstractSystem, inputs = unbound_inputs(
         simplify = false,
         eval_expression = false,
         eval_module = @__MODULE__,
+        split = true,
         kwargs...)
     isempty(inputs) && @warn("No unbound inputs were found in system.")
     if !isscheduled(sys)
-        sys = mtkcompile(sys; inputs, disturbance_inputs)
+        sys = mtkcompile(sys; inputs, disturbance_inputs, split)
     end
     if disturbance_inputs !== nothing
         # add to inputs for the purposes of io processing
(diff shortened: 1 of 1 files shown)


src/inputoutput.jl
```diff
@@ -198,10 +199,11 @@ function generate_control_function(sys::AbstractSystem, inputs = unbound_inputs(
         simplify = false,
         eval_expression = false,
         eval_module = @__MODULE__,
+        split = true,
```
baggepinnen, line 202: Introducing this argument with default `true` is likely breaking considering that the existing default in `mtkcompile` is `false`?

DhairyaLGandhi: I checked against MTK for the default. https://github.com/SciML/ModelingToolkit.jl/blob/873bf7e7592ff5bec9c71dd4fd49be2fb32a93dc/src/systems/systems.jl#L31

baggepinnen: ah, I must have confused the meaning of `split` in my head, sorry about the noise then :)

DhairyaLGandhi: (I dislike having multiple places where a default is defined/ repeated since that can be error prone)

I don't know what pattern mtk follows for these cases if it has one.

ChrisRackauckas: `split=true` is whether to use MTKParameters while putting everything into one vector. This has been the default for a bit now.

DhairyaLGandhi:
Cc @AayushSabharwal

AayushSabharwal (approved):

Issue #3734, opened 2025-06-13
Specifying metadata in MTK-Model DSL


ctessum:
**Is your feature request related to a problem? Please describe.**

We currently use system metadata for a number of different purposes, for example [here](https://base.earthsci.dev/dev/composition/#Model-Composition). We would also like to use the DSL to define models, but currently (as far as I can tell) the DSL doesn't allow specifying system metadata.

**Describe the solution you’d like**

If it is already possible to specify metadata, I guess just clearer documentation of that feature would be useful. Otherwise, maybe a design something like this:

```julia
using ModelingToolkit
using ModelingToolkit: t
using Test

struct TestMeta1 end
struct TestMeta2 end

@mtkmodel ModelA begin
    @description "A component with parameters `k` and `k_array` and some metadata."
    @parameters begin
        k
        k_array[1:2]
    end
    @metadata Dict(
        TestMeta1 => "test1",
        TestMeta2 => "test2"
    )
end

sys = ModelA()
@test getmetadata(sys, TestMeta1) == "test1"
@test getmetadata(sys, TestMeta2) == "test2"
```

PR #3892, opened 2025-08-13
Allow system level metadata information

DhairyaLGandhi:
## Checklist

## Additional context

Ref https://github.com/SciML/ModelingToolkit.jl/issues/3734

Are there cases where there are expression evaluations anywhere for the model parsing? Currently everything is delayed, but that means local variables are not referred to properly, and `esc` will simply convert the escaped expression, whereas we need it as parse/ expansion time.

### DIFF
--- a/docs/src/basics/MTKLanguage.md
+++ b/docs/src/basics/MTKLanguage.md
@@ -31,6 +31,7 @@ set to `false`.
   - `@equations`: for the list of equations
   - `@extend`: for extending a base system and unpacking its unknowns
   - `@icon` : for embedding the model icon
+  - `@metadata`: for assigning key-value pairs as model level metadata
   - `@parameters`: for specifying the symbolic parameters
   - `@structural_parameters`: for specifying non-symbolic parameters
   - `@variables`: for specifying the unknowns
@@ -92,6 +93,16 @@ end
     end
     @defaults begin
         v_for_defaults => 2.0
+    end
+end
+
+struct Author end
+struct ModelVersion end
+@mtkmodel ModelD begin
+    @description "A component with some metadata."
+    @metadata begin
+        Author = "Test Author"
+        ModelVersion = "1.0.0"
     end
 end
 ```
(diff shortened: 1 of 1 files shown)


test/model_parsing.jl
```diff
@@ -1045,3 +1045,47 @@ end
...
+    @mtkmodel TestMetadataModel begin
+        @metadata begin
+            Author = "Test Author"
+            MyVersion = "1.0.0"
+            License = "MIT"
+            Category => "example"
+            Tags = ["test", "demo", "metadata"]
+            MyBool => false
+            NewInt => 1
+        end
```
asinghvi17, line 1067: Why do these have to be structs?  Can we have some docs around this?

DhairyaLGandhi: I was requested this API to avoid potential key overwrites. The `setmetadata` utility also explicitly only has dispatches for `::DataType`. The intial version of the PR was a more standard `Dict{Symbol,  Any}`. We could have checked for existing keys when setting too imo, and reject overwrites that way.

Cc @AayushSabharwal

AayushSabharwal: This is in line with how symbolic metadata is stored. They don't have to be structs, just types. The reason is that this automatically avoids name conflicts since identically named types in different modules are different types, avoiding the necessity for name mangling.

asinghvi17: I see, that makes sense.  Thanks both!

asinghvi17: (Apparently I cannot resolve this conversation from github)

AayushSabharwal:
InterfaceI is failing

DhairyaLGandhi:
It's seemingly related to units, is that something this PR could break? I did rebase this branch recently as well

AayushSabharwal:
The error is specifically in the test added in this PR, and there is no unit failure on master or the recently opened #3912 so it is likely that _something_ here is wrong.

DhairyaLGandhi:
Hmm, unable to reproduce locally. 

I get
```julia
    struct Author end
    struct MyVersion end
    struct License end
    struct Category end
    struct Tags end
    struct MyBool end
    struct NewInt end

    @mtkmodel TestMetadataModel begin
        @metadata begin
            Author = "Test Author"
            MyVersion = "1.0.0"
            License = "MIT"
            Category => "example"
            Tags = ["test", "demo", "metadata"]
            MyBool => false
            NewInt => 1
        end
        
        @parameters begin
            k = 1.0, [description = "Gain parameter"]
        end
        
        @variables begin
            x(t), [description = "State variable"]
            y(t), [description = "Output variable"]
        end
        
        @equations begin
            D(x) ~ -k * x
            y ~ x
        end
    end
```

```julia
julia> @named test_model = TestMetadataModel()
Model test_model:
Equations (2):
  2 standard: see equations(test_model)
Unknowns (2): see unknowns(test_model)
  x(t): State variable
  y(t): Output variable
Parameters (1): see parameters(test_model)
  k [defaults to 1.0]: Gain parameter
```

AayushSabharwal:
Does it reproduce if you run the entire testset locally? Or just that file?

DhairyaLGandhi:
Let me try running the whole test suite locally.

PR #3900, opened 2025-08-21
feat: add `respecialize`

AayushSabharwal:


src/systems/diffeqs/basic_transformations.jl
```diff
@@ -706,3 +706,103 @@ function convert_system_indepvar(sys::System, t; name = nameof(sys))
     @set! sys.var_to_name = var_to_name
     return sys
 end
+
+"""
+    $(TYPEDSIGNATURES)
+
+Shorthand for `respecialize(sys, []; all = true)`
+"""
+respecialize(sys::AbstractSystem) = respecialize(sys, []; all = true)
```
asinghvi17, line 715:
```suggestion
respecialize(sys::AbstractSystem) = respecialize(sys, (); all = true)
```

Just to avoid Vector{Any}

asinghvi17:
This works for my usecase now along with https://github.com/JuliaSymbolics/Symbolics.jl/pull/1621.  Thanks @AayushSabharwal!

PR #3901, opened 2025-08-22
[WIP] fix: use improved discrete saving API

AayushSabharwal:
Requires https://github.com/SciML/SciMLBase.jl/pull/1116, https://github.com/SciML/DiffEqBase.jl/pull/1199 and https://github.com/SciML/OrdinaryDiffEq.jl/pull/2857

The above PRs are added via `[sources]` to be able to run tests


AayushSabharwal:
The only unreleased package is JumpProcesses, so I've bumped compats and only kept JumpProcesses in `sources`.

PR #3912, opened 2025-09-02
add input-affine equation decomposition

baggepinnen:
Useful for nonlinear control-systems analysis and design

### DIFF
--- a/src/input_affine_form.jl
+++ b/src/input_affine_form.jl
@@ -42,6 +42,7 @@ The function assumes that the equations are affine in the inputs. If the equatio
 are nonlinear in the inputs, an error is thrown.
 """
 function input_affine_form(eqs, inputs)
+    any(is_alg_equation, eqs) && error("All equations must be differential equations.")
     g, f, flag = Symbolics.linear_expansion(getfield.(eqs, :rhs), inputs)
     flag || error("The system is not affine in the inputs.")
     return f, g


src/input_affine_form.jl
```diff
@@ -0,0 +1,55 @@
...
+f, g = input_affine_form(eqs, inputs)
+```
+
+# Notes
+The function assumes that the equations are affine in the inputs. If the equations
+are nonlinear in the inputs, the result may not be meaningful.
+"""
+function input_affine_form(eqs, inputs)
+    # Extract the input matrix g(x) by taking coefficients of each input
+    g = [Symbolics.coeff(Symbolics.simplify(eq, expand = true), u)
```
AayushSabharwal: `simplify` and especially `expand` are pretty expensive. This should ideally be done with `semilinear_form` or repeated application of `Symbolics.linear_expansion`.

src/input_affine_form.jl
```diff
@@ -0,0 +1,55 @@
...
+         for eq in eqs, u in inputs]
+    g = Symbolics.simplify.(g, expand = true)
+
+    # Extract the drift term f(x) by substituting inputs = 0
+    f = Symbolics.substitute.(eqs, Ref(Dict(inputs .=> 0)))
```
AayushSabharwal: Substitution in general should just use `fast_substitute`.

baggepinnen:
```
help?> Symbolics.fast_substitute
  │ Warning
  │
  │  The following bindings may be internal; they may change or be removed in future versions:
  │
  │    •  Symbolics.fast_substitute
```

AayushSabharwal: Yeah the ecosystem in general is lagging behind with marking things as `@public`.

src/input_affine_form.jl
baggepinnen: `semilinear_form` appeared to be buggy and not doing what I wanted 
https://juliahub.slack.com/archives/C03UEC0P6G0/p1756812154880999

`linear_expansion` has this issue
```
help?> Symbolics.linear_expansion
  │ Warning
  │
  │  The following bindings may be internal; they may change or be removed in future versions:
  │
  │    •  Symbolics.linear_expansion

```
and it has failed to work for me in similar situations in the past https://github.com/JuliaSymbolics/Symbolics.jl/issues/550

AayushSabharwal: I can look into `semilinear_form`. The linked `linear_expansion` issue is simply that the expression was not affine. That isn't the case if this function requires the system to be affine in the inputs.

Issue #3913, opened 2025-09-03
FMU variable names cannot be typed


asinghvi17:
**Describe the bug 🐞**

The names have this strange underscore-like character that I cannot figure out how to type, even in the REPL.  In my opinion this is very bad UX - especially when you have to specify the initial value to even simulate the fmu...

```
ERROR: Initialization incomplete. Not all of the state variables of the
DAE system can be determined by the initialization. Missing
variables:

SymbolicUtils.BasicSymbolic[vˍt(t), hˍt(t)]```

Specifically, `'ˍ' != '_'`.

**Expected behavior**

Something typeable ideally.

**Minimal Reproducible Example 👇**

Get any FMU from https://github.com/modelica/Reference-FMUs but I tried this with bouncing ball.  Then load it in as an FMIComponent and try to set an initial value.  It is not possible to do without copying and pasting.

SebastianM-C:
That's unrelated to the FMU part, those look like names for dummy derivatives.

asinghvi17:
If i set the `v` and `h` values though then these complain about not having initial values?

ChrisRackauckas:
D(v) and D(h) need to be set.

asinghvi17:
I see, thanks.  Would it be possible to add an annotation to those variables that they are dummy derivatives and should be referenced as `D(h)` instead?  Or is that too much of a lift with the infrastructure we have now?

ChrisRackauckas:
It can probably be specialized in the printing.

DhairyaLGandhi:
Probably unrelated and can be its own issue but trying to compile (`mtkcompile`) an FMU from the releases of the repo causes errors similar to:
```julia
ERROR: ExtraEquationsSystemException: The system is unbalanced. There are 4 highest order derivative variables and 6 equations.
More equations than variables, here are the potential extra equation(s):
 0 ~ (wrapper(Any[x1(t), x0(t)], Float64[], Any[mu], t))[3] - x1(t)
 0 ~ (wrapper(Any[x1(t), x0(t)], Float64[], Any[mu], t))[4] - x0(t)
Note that the process of determining extra equations is a best-effort heuristic. The true extra equations are dependent on the model and may not be in this list.
Stacktrace:
  [1] getindex
    @ ./essentials.jl:917 [inlined]
  [2] iterate (repeats 2 times)
    @ ./array.jl:902 [inlined]
  [3] iterate
    @ ./iterators.jl:206 [inlined]
[...]

PR #3920, opened 2025-09-10
Improve Printing of Uninitialized Variables

DhairyaLGandhi:
## Checklist

## Additional context

Closes https://github.com/SciML/ModelingToolkit.jl/issues/3913


DhairyaLGandhi:
@asinghvi17 can you try this out for your MWE please

src/utils.jl
```diff
@@ -1079,3 +1079,23 @@ function Base.showerror(io::IO, ::NotPossibleError)
...
 end
+
+"""
+    $(TYPEDSIGNATURES)
+
+Given a vector of variables in the system, return the corresponding `Differential` form of variable if possible.
+Else returns the variable as-is.
+"""
+function underscore_to_D(v::AbstractVector, sys)
+    maps = get_schedule(sys).dummy_sub
```
AayushSabharwal: Can this be gated behind `isscheduled(sys)` so that printing the error doesn't error if the system isn't `mtkcompile`d? It could also use the string manipulation "hack" discussed earlier in that case (or if the inverse map doesn't have the appropriate entry, as is the case for algebraic variables)

DhairyaLGandhi: Shouldn't that error check happen earlier (ie before we get this far in the problem construction). Also note that any variables not in the schedule get passed through as is.

AayushSabharwal: Problems do not require the system to be `mtkcompile`d.

> Also note that any variables not in the schedule get passed through as is.

Yeah, but we can still print them better.

DhairyaLGandhi: this is added back in https://github.com/SciML/ModelingToolkit.jl/pull/3920/commits/f137b1aab9b3a02188696a589160b02cfc6d0f91

DhairyaLGandhi:
Added the test

DhairyaLGandhi:
Seems to be that `BrownFullBasicInit` is undefined elsewhere?

ChrisRackauckas:
That was moved to DiffeqBase, but OrdinaryDiffEqCore should be unchanged (i.e. imports)

PR #3923, opened 2025-09-11
Allow subsetting `Initial`s in `subset_tunables`

DhairyaLGandhi:
## Checklist

## Additional context

cc @SebastianM-C can you check if this is sufficient for your use cases, and we can add a test then

### DIFF
--- a/src/parameters.jl
+++ b/src/parameters.jl
@@ -164,7 +164,7 @@ function subset_tunables(sys, new_tunables)
         throw(ArgumentError("Tunable parameters can only be changed for split systems."))
     end
 
-    cur_tunables = tunable_parameters(sys, parameters(sys))
+    cur_tunables = tunable_parameters(sys, tunable_parameters(sys))
     diff_params = setdiff(cur_tunables, new_tunables)
 
     if !isempty(setdiff(new_tunables, cur_tunables))


ChrisRackauckas:
We should make sure this gets appropriately tested in SciMLSensitivity too.

SebastianM-C:
yeah, it seems to work for me (on a simple example at least)

SebastianM-C:
Btw, I noticed that `subset_tunables` does not remove `Initial`s. Does this affect the performance or it's not an issue?

ChrisRackauckas:
That will not effect the performance of adjoints much, but it will forward mode

AayushSabharwal:
Initials are not tunables. Why exactly do we need this?

SebastianM-C:
I'm not sure if we need this or not, I just noticed that `Initials` show up in `tunable_parameters(sys)`.

AayushSabharwal:
Oh, that's a lie. I'm pretty sure when I added initials and the `initial_parameters = true` keyword, I just did a find-replace of all `parameters(sys)` to `parameters(sys; initial_parameters = true)`. This is probably a remnant of that.

DhairyaLGandhi:
Does that mean this can be closed since it won't be actionable

PR #3935, opened 2025-09-18
fix: FMI varname parse bug #3934

EduardoMartinezLopez:
## Checklist


ChrisRackauckas:
This reformats the whole repo. Right now the formatter is broken so don't worry about it. Revert that.

EduardoMartinezLopez:
@ChrisRackauckas Thanks! I reverted the formatter commit.

EduardoMartinezLopez:
@ChrisRackauckas Question for you. It seems like the existing code in MTKFMIExt > parseFMIVariableName() mutates the "name" input variable. Is that OK? My understanding is that this is considered bad practice as Julia passes by reference. Also if it the input is mutated, the convention is to add ! to the function name, right?

ChrisRackauckas:
I don't see where it mutates it. It creates a view but then creates a Symbol, which is effectively an interned string. But I don't see it actually mutating the string? Put to the line that you're worried about.

Though it probably shouldn't create the symbol there for performance reasons on large models, but that's a minor detail.

EduardoMartinezLopez:
Ok, I was concerned about the `name = @view name[....]` with name on the left side of an assignment. I guess that's ok with @view macro?

ChrisRackauckas:
that just makes `name` be a reference to a subset of what was previously `name`, no mutation there.

AayushSabharwal:
The test doesn't work

EduardoMartinezLopez:
@AayushSabharwal what is the problem with the test on your end?

EduardoMartinezLopez:
@AayushSabharwal is this still waiting on anything?

test/fmi/Project.toml
```diff
@@ -2,7 +2,9 @@
 FMI = "14a09403-18e3-468f-ad8a-74f8dda2d9ac"
 FMIZoo = "724179cf-c260-40a9-bd27-cccc6fe2f195"
 ModelingToolkit = "961ee093-0014-501f-94e3-6117800e7a78"
+NonlinearSolve = "8913a72c-1f9b-4ce2-8d82-65094dcecaec"
```
AayushSabharwal:
```suggestion
```

test/fmi/Project.toml
```diff
@@ -2,7 +2,9 @@
...
 OrdinaryDiffEq = "1dea7af3-3e70-54e6-95c3-0bf5283fa5ed"
+SciMLBase = "0bca4576-84f4-4d90-8ffe-ffa030f20462"
```
AayushSabharwal:
```suggestion
```

AayushSabharwal:
Thanks for the reminder. I think that should do it. I'll let CI run once more to confirm, and merge today.

EduardoMartinezLopez:
Thanks. Looking forward to officially becoming a contributor to the project!

Issue #3871, opened 2025-08-05
Sparse ODE segfaults when RHS contains differentiated variable


hersle:
Take this simple example:
```julia
using ModelingToolkit, OrdinaryDiffEq
using ModelingToolkit: t_nounits as t, D_nounits as D

@variables x(t) y(t)
@mtkcompile sys = System([
    D(x) ~ x * D(y)
    D(y) ~ x - y
], t)
prob = ODEProblem(sys, [x => 1.0, y => 0.0], (0.0, 1.0); jac = true, sparse = true)

sol = solve(prob, FBDF());
sol = solve(prob, FBDF());
sol = solve(prob, FBDF());
# ...
```
It segfaults after a few solves (the precise number required appears random):
```
Please submit a bug report with steps to reproduce this fault, and any error messages that follow (in their entirety). Thanks.  
Exception: EXCEPTION_ACCESS_VIOLATION at 0x7ffab2845290 -- gc_mark_outrefs at C:/workdir/src\gc.c:2776 [inlined]
gc_mark_loop_serial_ at C:/workdir/src\gc.c:2938
in expression starting at none:0
gc_mark_outrefs at C:/workdir/src\gc.c:2776 [inlined]
gc_mark_loop_serial_ at C:/workdir/src\gc.c:2938
gc_mark_loop_serial at C:/workdir/src\gc.c:2961
gc_mark_loop at C:/workdir/src\gc.c:3143 [inlined]
_jl_gc_collect at C:/workdir/src\gc.c:3532
ijl_gc_collect at C:/workdir/src\gc.c:3893
maybe_collect at C:/workdir/src\gc.c:926 [inlined]
jl_gc_pool_alloc_inner at C:/workdir/src\gc.c:1319
jl_gc_pool_alloc_noinline at C:/workdir/src\gc.c:1386 [inlined] 
jl_gc_alloc_ at C:/workdir/src\julia_internal.h:523 [inlined]   
jl_gc_alloc at C:/workdir/src\gc.c:3946
_new_genericmemory_ at C:/workdir/src\genericmemory.c:56 [inlined]
jl_alloc_genericmemory at C:/workdir/src\genericmemory.c:99     
ijl_array_grow_end at C:/workdir/src\array.c:229
ijl_module_names at C:/workdir/src\module.c:1002
#unsorted_names#9 at .\reflection.jl:96 [inlined]
unsorted_names at .\reflection.jl:96 [inlined]
[...]

PR #3949, opened 2025-09-30
Fix Jacobian sparsity pattern with observed derivatives

hersle:
I want to fix https://github.com/SciML/ModelingToolkit.jl/issues/3871.

I figured `torn_system_jacobian_sparsity` computes an incorrect sparsity pattern for this example.
My first idea is to disable it and always compute `jacobian_sparsity` with the "fallback method" `Symbolics.jacobian_sparsity(...)`.
If this does not work in all cases, the alternative would be to patch `torn_system_jacobian_sparsity` directly.

### DIFF
--- a/src/systems/codegen.jl
+++ b/src/systems/codegen.jl
@@ -455,8 +455,9 @@ end
 Return the sparsity pattern of the jacobian of `sys` as a matrix.
 """
 function jacobian_sparsity(sys::System)
-    sparsity = torn_system_jacobian_sparsity(sys)
-    sparsity === nothing || return sparsity
+    # disable to fix https://github.com/SciML/ModelingToolkit.jl/issues/3871
+    #sparsity = torn_system_jacobian_sparsity(sys)
+    #sparsity === nothing || return sparsity
 
     Symbolics.jacobian_sparsity([eq.rhs for eq in full_equations(sys)],
         [dv for dv in unknowns(sys)])


hersle:
It looks to me like the relevant tests are passing. Do you have any input on this @ChrisRackauckas?

src/systems/codegen.jl
```diff
@@ -455,8 +455,9 @@ end
 Return the sparsity pattern of the jacobian of `sys` as a matrix.
 """
 function jacobian_sparsity(sys::System)
-    sparsity = torn_system_jacobian_sparsity(sys)
-    sparsity === nothing || return sparsity
+    # disable to fix https://github.com/SciML/ModelingToolkit.jl/issues/3871
+    #sparsity = torn_system_jacobian_sparsity(sys)
+    #sparsity === nothing || return sparsity
```
ChrisRackauckas, line 460: @AayushSabharwal is there a reason the torn sparsity would have false zeros here? That seems like it would give issues in other places.

AayushSabharwal: I'm not 100% sure, we've just used `torn_system_jacobian_sparsity` since the beginning of time iirc. Maybe this doesn't handle the `W` sparsity correctly?

ChrisRackauckas: Well I would think that if this sparsity pattern is wrong, then the incidence matrix is wrong. Can you make a note to follow up on this? For now I want to merge so that these are right, but I'd like to make sure something isn't hiding here.

hersle:
Thanks. I noticed it looks like [`torn_system_jacobian_sparsity` is not used anywhere else by MTK](https://github.com/search?q=torn_system_jacobian_sparsity+repo%3ASciML%2FModelingToolkit.jl&type=code), and also like [`torn_system_jacobian_sparsity` is not used by any other public Github code](https://github.com/search?q=torn_system_jacobian_sparsity&type=code).

src/systems/codegen.jl
AayushSabharwal: Will do

PR #3953, opened 2025-10-02
refactor: update to SymbolicUtils@4 and Symbolics@7

AayushSabharwal:
## Checklist


lib/ModelingToolkitBase/src/problems/jumpproblem.jl
isaacsas, line 1: How are you now supporting a list of variables from which any collected variables must be members? That seems dropped in this new version of these functions.

AayushSabharwal: `Symbolics.get_variables` does this, and is a wrapper over `SymbolicUtils.search_variables!` with the appropriate filter function. Now this is a common API for _any_ sort of filter you might want to apply to the collected variables.

isaacsas: So that is what `is_atomic` does? (Just because from the name is unclear to me what it means.)

AayushSabharwal: Yes. I'm working on improving docs in SymbolicUtils for all of the new stuff. `is_atomic` takes an expression and returns a boolean identifying whether the expression should be considered "atomic" and included in the list of variables. The function should probably be named `find_atomics` but now that requires a _massive_ find-and-replace.

AayushSabharwal: This isn't to say I _won't_ do the renaming, but rather that I might wait a bit to get things working first.

isaacsas: Sounds good! I just wanted to check the filtering wasn’t being inadvertently lost for MassActionJump stoichiometric species.

PR #3974, opened 2025-10-17
fix: let StochasticDiffEq auto calculate noise process

### LINKED ISSUE #3971: Different solutions for seeded SDEProblems from ModelingToolkit and StochasticDiffEq
**Describe the bug 🐞**

It seems like this happens because MTK constructs the `WienerProcess` that gets passed into the SDEProblem (in `calculate_noise_and_rate_prototype`). Then, when the integrator is initialized, the RNG for the noise process is reseeded when the SDEProblem is constructed. However, in the non-MTK case, the noise process is constructed inside the integrator initialization (and the RNG is initialized). But as it turns out reseeding the RNG that WienerProcess uses does not actually result in an identical object as initializing it with a given seed (https://github.com/JuliaRandom/RandomNumbers.jl/issues/89). So you end up having a different RNG for the noise process in the MTK vs non-MTK case. There are a few ways to make the solutions consistent (assuming the reseeding behavior is intended), but I am not sure which one would be right:

1. ModelingToolkit can not pass in a noise process, so that the RNG is initialized when the integrator initialized (The noise process that MTK also seems to assume that the problem is out-of-place, which might be a bug?) 
2. StochasticDiffEq can reseed the RNG of the noise process that is passed in, so that the results are guaranteed to be the same as in the case when the noise process is constructed in the integrator initialization

```Julia
f(u, p, t) = u
g(u, p, t) = [1]
tspan = (0.0, 1.0)
prob = SDEProblem(f, g, [0.5], tspan, seed = 99)
sol = solve(prob, RKMil(), dt = 0.01, adaptive = false)

@variables x(t)
@brownians ξ
@mtkcompile sys = System([D(x) ~ x + ξ], t)
prob2 = SDEProblem(sys, [x => 0.5], tspan, seed = 99)
sol2 = solve(prob2, RKMil(), dt = 0.01, adaptive = false)
isequal(sol, sol2) # false
```

vyudu:
Would fix #3971, but also lets it assign an in-place noise process for problems that are in-place instead of assuming out-of-place.

### DIFF
--- a/src/ModelingToolkit.jl
+++ b/src/ModelingToolkit.jl
@@ -22,7 +22,7 @@ using DiffEqCallbacks
 using Graphs
 import ExprTools: splitdef, combinedef
 import OrderedCollections
-using DiffEqNoiseProcess: DiffEqNoiseProcess, WienerProcess
+using DiffEqNoiseProcess: DiffEqNoiseProcess, WienerProcess, WienerProcess!
 
 using SymbolicIndexingInterface
 using LinearAlgebra, SparseArrays


AayushSabharwal:
Tests fail

ChrisRackauckas:
yes this isn't correct since it's changing it from scalar to diagonal noise

src/problems/sdeproblem.jl
```diff
@@ -106,7 +106,7 @@ function calculate_noise_and_rate_prototype(sys::System, u0; sparsenoise = false
     elseif size(noiseeqs, 2) == 1
         # scalar noise
         noise_rate_prototype = nothing
-        noise = WienerProcess(0.0, 0.0, 0.0)
+        noise = iip ? WienerProcess!(0., 0., 0.) : WienerProcess(0., 0., 0.)
```
ChrisRackauckas, line 109: Scalar is always out of place, since scalars are just numbers

vyudu: Ah OK I just misunderstood then, will close since the original issue is fixed by the rng change

Issue #3976, opened 2025-10-21
simplification of affect systems with observed equations can yield incorrect updates to state variables


vyudu:
**Describe the bug 🐞**

Simplifying observed equations in affect systems can lead to unintended explicit updates to state variables when a callback is triggered. All the algebraic equations of the parent system and the observed equations of the parent system are added to the affect system. When the affect system is simplified, some of these observed equations get solved for state variables. Thus the simplified affect system may be fully explicit (i.e. have no actual equations, only observed equations), but this means there will be an explicit update to the state variable that was solved for. Instead of the observed value being updated in response to changes in the state variables, the state variable gets updated according to the previous value of the observed variable (since there are no algebraic equations to be respected). 

**Expected behavior**

Observed equations of the parent system should not be solved for state variables in the affect system. 

**Minimal Reproducible Example 👇**

```julia
using ModelingToolkit
t = ModelingToolkit.t_nounits; D = ModelingToolkit.D_nounits
@variables V(t) = -70.0 G(t) = 0.0 jcn(t) [input = true] z(t)
@parameters Eₘ=-70. θ=-50.

ev = (V ≥ θ) => [V ~ Eₘ]

@named lif1 = System([D(V) ~ jcn], t, [V, jcn], [Eₘ, θ]; discrete_events = [ev])
@named lif2 = System([D(V) ~ jcn], t, [V, jcn], [Eₘ, θ]; discrete_events = [ev])
@named qif = System([D(G) ~ G + z, D(z) ~ z, D(V) ~ jcn], t, [V, G, z, jcn], [Eₘ, θ]; discrete_events = [ev])

sys2 = compose(System([lif1.jcn ~ lif2.V + qif.G, lif2.jcn ~ lif1.V + qif.G, qif.jcn ~ lif1.V + lif2.V], t; name = :outer), [lif1, lif2, qif])
sys2 = mtkcompile(sys2)
```

Then inspecting the update equations for the first discrete event using `ModelingToolkit.observed(discrete_events(sys2)[1].affect.system)`, we get:
```julia
4-element Vector{Equation}:
 lif1₊V(t) ~ lif1₊Eₘ
 lif1₊jcn(t) ~ lif2₊jcn(t) - 2lif1₊V(t) + qif₊jcn(t)
 qif₊G(t) ~ lif2₊jcn(t) - lif1₊V(t)
 lif2₊V(t) ~ lif1₊jcn(t) - qif₊G(t)
```
[...]

PR #3977, opened 2025-10-22
fix: only add observed equations to affect system if there are algebraic equations in the system

vyudu:
attempt at #3976


AayushSabharwal:
I've replied to the issue. I'm not sure this is the right solution.

PR #3987, opened 2025-10-26
Update README

TorkelE:
Got tired of not having an easily accessible place to fetch example code for the new MTK v10 syntax. Also fixed some other issues (use of overdefined system, unnecessary use of full timespan, use of full DifferentialEquations suite). Happy to make a slightly more minimal version, but figured the final updated version is quite good.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ using Plots
 plot(sol, idxs = (x, y))
 ```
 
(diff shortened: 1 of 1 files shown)


TorkelE:
Updated images
readme_lorenz_sim
readme_connected_lorenz_sim
(some axis stuff seems to no longer be auto-added on new version)

Issue #2612, opened 2024-04-04
MTK allows you to create faulty events without throwing an error.


TorkelE:
I have gone through the events in MTK. There are several cases where MTK allow your to build a system with a misformated event, and an error is not actually thrown until you create a problem or attempts to simulate it. Especially since it is a fit trick how to format the events, the errors should be thrown at the first possible time (when a system is created).
```julia
using ModelingToolkit, OrdinaryDiffEq

@parameters p d
@variables t X(t)
eqs = [Differential(t)(X) ~ p - d*X]

# Example 1: Discrete events with dosage given as a Tuple
discrete_events = (2.0, 5.0) => [X ~ X + 10.0]
@mtkbuild osys = ODESystem(eqs, t; discrete_events)
oprob = ODEProblem(osys, [X => 10], (0.0, 10.), [p => 2.0, d => 0.5]) # Error is thrown here.

# Example 2: Discrete events where single expression condition is given in a vector.
discrete_events = [X < 5.0] => [X ~ X + 10.0]
@mtkbuild osys = ODESystem(eqs, t; discrete_events)
oprob = ODEProblem(osys, [X => 10], (0.0, 10.), [p => 2.0, d => 0.5])
sol = solve(oprob, Tsit5()) # Error is thrown here.

# Example 3: Discrete events where multiple expression conditions are given in a vector.
discrete_events = [X < 5.0, X > 10.0] => [X ~ X + 10.0]
@mtkbuild osys = ODESystem(eqs, t; discrete_events)
oprob = ODEProblem(osys, [X => 10], (0.0, 10.), [p => 2.0, d => 0.5]) # Error is thrown here.

# Example 4: Discrete events that affects parameters and variables.
discrete_events = 5.0 => [X ~ X + 10.0, p ~ 2.0]
@mtkbuild osys = ODESystem(eqs, t; discrete_events)
oprob = ODEProblem(osys, [X => 10], (0.0, 10.), [p => 2.0, d => 0.5]) # Error is thrown here.

# Example 5: Discrete events with non-trivial affect right-hand side.
discrete_events = 5.0 => [X + 1.0 ~ X + 10.0]
@mtkbuild osys = ODESystem(eqs, t; discrete_events)
oprob = ODEProblem(osys, [X => 10], (0.0, 10.), [p => 2.0, d => 0.5]) # Error is thrown here.

# Example 6: Continuous events that affects parameters and variables.
[...]

PR #3992, opened 2025-10-26
Prevents erroneous creation of discrete events (vector of conditions)

TorkelE:
Should close https://github.com/SciML/ModelingToolkit.jl/issues/2612

Looking at the code, my impression is that stuff like 
```julia
@parameters p(t) d
@variables X(t)
SymbolicDiscreteCallback([X < 5.0, X > 10.0] => [X ~ Pre(X) + 10.0]
```
should not be allowed. It eventually yields an error (but a hard-to-interpret one). This throws an error as soon as someone tries to create the event (happy to modify if this is something we'd actually want to supprot).

### DIFF
--- a/src/systems/callbacks.jl
+++ b/src/systems/callbacks.jl
@@ -11,7 +11,7 @@ struct SymbolicAffect
 end
 
 function SymbolicAffect(affect::Vector{Equation}; alg_eqs = Equation[],
-        discrete_parameters = infer_discrete_parameters(affect), kwargs...)
+        discrete_parameters = Any[], kwargs...)
     if !(discrete_parameters isa AbstractVector)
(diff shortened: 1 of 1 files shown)


src/systems/callbacks.jl
```diff
@@ -31,6 +31,38 @@ function Symbolics.fast_substitute(aff::SymbolicAffect, rules)
         map(substituter, aff.discrete_parameters))
 end
 
+# The discrete parameters (i.e. parameters that are updated in an event) can be inferred as
```
AayushSabharwal: It looks like this is mixing content from #3991?

TorkelE: Agreed, fixed now

AayushSabharwal (approved):

PR #3994, opened 2025-10-28
fix: fix incorrect incidence after tearing some differential equations

AayushSabharwal:


src/systems/codegen.jl
```diff
@@ -455,9 +455,8 @@ end
 Return the sparsity pattern of the jacobian of `sys` as a matrix.
 """
 function jacobian_sparsity(sys::System)
-    # disable to fix https://github.com/SciML/ModelingToolkit.jl/issues/3871
-    #sparsity = torn_system_jacobian_sparsity(sys)
-    #sparsity === nothing || return sparsity
+    sparsity = torn_system_jacobian_sparsity(sys)
```
ChrisRackauckas, line 458: how is this related?

AayushSabharwal: This is how `jacobian_sparsity` was prior to #3871. That PR commented out the use of `torn_system_jacobian_sparsity` due to incorrect results. This PR fixes `torn_system_jacobian_sparsity`, which means we can use it again.

Issue #2714, opened 2024-05-15
Promote parameters of subsystem to system


hexaeder:
In hierachical modeling I often feel the need to "hide" some of the hierarchal complexity. Let's suppose I have an inner model `inner` with parameter `p` and an outer model `outer` which includes the inner model 2 times. The fact that the outer model  contains the inner model is just an implementation detail, in the end it want to expose parameter `p` directly from my outer model. I can do something like this:

```julia
using ModelingToolkit
using ModelingToolkit: D

@variables t
@mtkmodel inner begin
    @parameters begin
        p
    end
    @variables begin
        i(t)
    end
    @equations begin
        D(i) ~ p*i
    end
end

@mtkmodel outer begin
    @parameters begin
        p
    end
    @variables begin
        o(t)
    end
    @components begin
        inner1 = inner(p=p)
        inner2 = inner(p=p)
    end
    @equations begin
        o ~ inner1.i + inner2.i
    end
end

@mtkbuild outerb = outer()
```
The problem is, that now `inner1.p` and `inner2.p` only default to `outer.p`. But they are still independent. For example I cannot change `outer.p` in a callback.
Is there a way to achieve that using the public interface? What I want is essentially something like:
- flattening the system (i.e. put all the equations together in a big eqs vector), removing the subsystems (is there a
- replacing every occurrence of `inner1₊p` and `inner2₊p` by `p` in the equations
- get rid of the `inner1₊p` and `inner2₊p` as a parameter all together

hersle:
This sounds somewhat similar to one of my use cases, except I am generating systems programmatically instead of using the convenience macros. Does for example this do what you want?
```julia
using ModelingToolkit
using ModelingToolkit: t_nounits as t, D_nounits as D

function inner(p; kwargs...)
    p = ParentScope(p) # peel off parent's namespacing
    @variables i(t)
    eqs = [D(i) ~ p*i]
    return ODESystem(eqs, t, [i], [#= may or may not include p=#]; kwargs...)
end

function outer(; kwargs...)
    @parameters p
    @variables o(t)
    @named inner1 = inner(p)
    @named inner2 = inner(p)
    eqs = [o ~ inner1.i + inner2.i]
    return ODESystem(eqs, t, [o], [p]; systems = [inner1, inner2], kwargs...)
end

@mtkbuild outerb = outer()
```
Now `outerb` is
```
Model outerb with 2 equations
Unknowns (2):
  inner1₊i(t)
[...]

PR #3995, opened 2025-10-28
add test for passing params as structural params

hexaeder:
At least in my context, it is very typical to define models based on subcomponents. However, the parameters still "belong" to the higher level model not to the lower level model. See https://github.com/SciML/ModelingToolkit.jl/issues/2714

A solution which works for me is to pass symbolic parameters of the outer system as structural parameters to the inner system. That works great, but might be off the beaten track and feels like the kind of feature which easily breaks by accident. This PR just adds a test case for that to detect accidental breakage before it breaks my whole model library 🙃

### DIFF
--- a/test/model_parsing.jl
+++ b/test/model_parsing.jl
@@ -1089,3 +1089,89 @@ end
     @test ModelingToolkit.getmetadata(test_model, MyBool, nothing) === false
     @test ModelingToolkit.getmetadata(test_model, NewInt, nothing) === 1
 end
+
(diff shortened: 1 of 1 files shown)


AayushSabharwal:
Looks good!

Unrelated, but thankfully I'm not going insane and InterfaceII running for hours isn't my fault in #3994 :D

hexaeder:
I guess all the failing stuff is unrelated, since the added test belong to Interface I and those run through.

AayushSabharwal:
Accidentally closed and reopened. Tests passed here.

Issue #3823, opened 2025-07-16
Support for Inputs


bradcarman:
There should be official support for the concept of "inputs" to a model.  I have hacked together the following code to provide a start for the following features/requirements:
- provide a way to define model inputs so they are distinguishable from normal "unknowns".  For example, `inputs(sys::System)` should list the inputs.  Also an `@inputs` macro should be available to define model inputs.  Input variables should be set directly and shouldn't change with time unless directed to, i.e. an input `x` has the equation `D(x) ~ 0`
- provide an API so that inputs are separate from a problem, such that one can call `solve(prob, inputs)`.  This way, if inputs change, the `ODEProblem` can remain the same.
- provide an API so that inputs can be set in an indeterminate form with `set_input(integrator, var::Num, value::Number)`

In the below example is a start to this feature request, the `solve(prob, inputs)` and `set_input(integrator, var, value)` are demonstrated.  What's missing is the ability to influence initialization of the inputs from data.  As can be seen, this provides the ability to run the same model in determinate and indeterminate forms and get the same result.

```julia
using ModelingToolkit
using ModelingToolkit: t_nounits as t, D_nounits as D
using OrdinaryDiffEq
using Plots
using DiffEqBase: reeval_internals_due_to_modification!, savevalues!
using Test

struct Input
    var::Num
    data::Vector{Float64}
    time::Vector{Float64}
end

function DiffEqBase.solve(prob::SciMLBase.AbstractDEProblem, inputs::Union{Input, Vector{Input}}, args...; kwargs...)

    tstops = Float64[]
    callbacks = DiscreteCallback[]
    if !isa(inputs, Vector)
        inputs = [inputs]
    end

    for input::Input in inputs
        tstops = union(tstops, input.time)

        condition(u,t,integrator) = any(t .== input.time)
        function affect!(integrator) 
            i = findfirst(integrator.t .== input.time)
            integrator[input.var] = input.data[i]
        end
        callback = DiscreteCallback(condition, affect!)

        push!(callbacks, callback)
    end


    return solve(prob, args...; tstops, callback=CallbackSet(callbacks...), kwargs...)
end

function set_input!(integrator, var, value)
    savevalues!(integrator)
    integrator[var] = value
    reeval_internals_due_to_modification!(integrator, false) 
    savevalues!(integrator, true)
end

# -----------------------------------------
# ----- example ---------------------------
# -----------------------------------------

vars = @variables begin
    # inputs
    x(t) = 1

[...]

PR #3998, opened 2025-10-28
Support for Inputs Feature

bradcarman:
## Checklist

## Additional context
Closes issue #3823

### DIFF
--- a/src/systems/inputs.jl
+++ b/src/systems/inputs.jl
@@ -163,7 +163,7 @@ function build_input_functions(sys, inputs)
     return sys
 end
 
-function DiffEqBase.solve(prob::SciMLBase.AbstractDEProblem, inputs::Vector{Input}, args...; kwargs...)
+function CommonSolve.solve(prob::SciMLBase.AbstractDEProblem, inputs::Vector{Input}, args...; kwargs...)
     tstops = Float64[]
     callbacks = DiscreteCallback[]
 


src/systems/inputs.jl
```diff
@@ -0,0 +1,201 @@
+using SymbolicIndexingInterface
+using Setfield
+using StaticArrays
+using OrdinaryDiffEqCore
```
ChrisRackauckas: not top level

src/systems/inputs.jl
```diff
@@ -0,0 +1,201 @@
...
+end
+
+function Input(var, data::Vector{<:Real}, time::Vector{<:Real})
+    n = length(data)
+    return Input(var, SVector{n}(data), SVector{n}(time))
+end
+
+struct InputFunctions
+    events::Tuple{SymbolicDiscreteCallback}
+    vars::Tuple{SymbolicUtils.BasicSymbolic{Real}}
```
ChrisRackauckas, line 34: not concrete

src/systems/inputs.jl
```diff
@@ -0,0 +1,201 @@
...
+
+        setters = [SymbolicIndexingInterface.setsym(sys, x) for x in vars]
+
+        @set! sys.input_functions = InputFunctions(events, vars, setters)
+    end
+
+    return sys
+end
+
+function DiffEqBase.solve(prob::SciMLBase.AbstractDEProblem, inputs::Vector{Input}, args...; kwargs...)
```
ChrisRackauckas: this violates the contract, and its not DiffEqBase

src/systems/inputs.jl
```diff
@@ -0,0 +1,201 @@
...
+
+This provides a "determinate form" of input handling where all input values are known a priori,
+as opposed to setting inputs manually during integration with [`set_input!`](@ref).
+
+See also [`set_input!`](@ref), [`finalize!`](@ref)
+"""
+struct Input
+    var::Num
+    data::SVector
+    time::SVector
```
AayushSabharwal, line 24: Why specifically `SVector`? Not only is this not concrete, I don't think it doesn't offer any performance gain given how it is used.

src/systems/inputs.jl
```diff
@@ -0,0 +1,201 @@
...
+end
+
+function Input(var, data::Vector{<:Real}, time::Vector{<:Real})
+    n = length(data)
+    return Input(var, SVector{n}(data), SVector{n}(time))
```
AayushSabharwal, line 29: This is type-unstable. `n` isn't inferrable.

src/systems/inputs.jl
```diff
@@ -0,0 +1,201 @@
...
+end
+
+struct InputFunctions
```
AayushSabharwal: Why do the fields of this struct need to be wrapped in `Tuple{...}`?

src/systems/inputs.jl
AayushSabharwal, line 34: Fairly minor, but `vars` is a bit of a misnomer for a field only containing one variable.

src/systems/inputs.jl
```diff
@@ -0,0 +1,201 @@
...
+    events::Tuple{SymbolicDiscreteCallback}
+    vars::Tuple{SymbolicUtils.BasicSymbolic{Real}}
+    setters::Tuple{SymbolicIndexingInterface.ParameterHookWrapper}
```
AayushSabharwal: `SII.ParameterHookWrapper` is not public API.

src/systems/inputs.jl
```diff
@@ -0,0 +1,201 @@
...
+step!(integrator, 1.0, true)
+
+# Must call finalize! at the end
+finalize!(integrator)
+```
+
+See also [`finalize!`](@ref), [`Input`](@ref)
+"""
+function set_input!(input_funs::InputFunctions, integrator::OrdinaryDiffEqCore.ODEIntegrator, var, value::Real)
+    i = findfirst(isequal(var), input_funs.vars)
```
AayushSabharwal, line 86: This doesn't quite make sense. Why `findfirst` for a field that only contains a tuple of one variable? Either the first one matches, or it doesn't. I imagine you wanted `InputFunctions` to contain a list of variables along with their corresponding callbacks and setters. In such a case, it is better to store a `Dict` than repeatedly perform `findfirst`.

src/systems/inputs.jl
```diff
@@ -0,0 +1,201 @@
...
+    setter = input_funs.setters[i]
+    event = input_funs.events[i]
+
+    setter(integrator, value)
+    save_callback_discretes!(integrator, event)
+    u_modified!(integrator, true)
+    return nothing
+end
+function set_input!(integrator, var, value::Real)
+    set_input!(get_input_functions(integrator.f.sys), integrator, var, value)
```
AayushSabharwal, line 96: Accessing `sys` during a solve is discouraged. The solve should not involve symbolic quantities. This is a lot of unnecessary, type-unstable overhead. Solves should also work without the system - anything possible symbolically should be possible without the symbolic infrastructure.

src/systems/inputs.jl
```diff
@@ -0,0 +1,201 @@
...
+function (input_funs::InputFunctions)(integrator, var, value::Real)
+    set_input!(input_funs, integrator, var, value)
+end
+(input_funs::InputFunctions)(integrator) = finalize!(input_funs, integrator)
+
+function build_input_functions(sys, inputs)
+
+    # Here we ensure the inputs have metadata marking the discrete variables as parameters.  In some
+    # cases the inputs can be fed to this function before they are converted to parameters by mtkcompile.
+    vars = SymbolicUtils.BasicSymbolic[isparameter(x) ? x : toparam(x)
```
AayushSabharwal, line 136: I think `build_input_functions` should be called as a model transformation after `mtkcompile`, and verify that the list of `inputs` it gets are inputs (and part of the list of parameters). This is more robust than trusting that the user passed values around correctly and trusting that there won't be bugs in the future.

src/systems/inputs.jl
```diff
@@ -0,0 +1,201 @@
...
+
+            push!(events, sdc)
+
+            # ensure that the ODEProblem does not complain about missing parameter map
+            if !haskey(defaults, x)
+                push!(defaults, x => 0.0)
+            end
+        end
+
+        @set! sys.discrete_events = events
```
AayushSabharwal, line 154: This would overwrite any pre-existing discrete events in the system.

src/systems/inputs.jl
```diff
@@ -0,0 +1,201 @@
...
+    if !isempty(vars)
+        for x in vars
+            affect = ImperativeAffect((m, o, c, i)->m, modified = (; x))
+            sdc = SymbolicDiscreteCallback(Inf, affect)
+
+            push!(events, sdc)
+
+            # ensure that the ODEProblem does not complain about missing parameter map
+            if !haskey(defaults, x)
+                push!(defaults, x => 0.0)
```
AayushSabharwal, line 150: This is mutating the defaults stored in `sys`. Model transformations should ideally be out of place.

src/systems/inputs.jl
```diff
@@ -0,0 +1,201 @@
...
+        setters = [SymbolicIndexingInterface.setsym(sys, x) for x in vars]
+
+        @set! sys.input_functions = InputFunctions(events, vars, setters)
+    end
+
+    return sys
+end
+
+function DiffEqBase.solve(prob::SciMLBase.AbstractDEProblem, inputs::Vector{Input}, args...; kwargs...)
+    tstops = Float64[]
```
AayushSabharwal, line 167: It's not guaranteed that the time is `Float64`.

AayushSabharwal:
I thought about this a bit. In my opinion, the best way to phrase this is a model transformation in `basic_transformations.jl`. It would operate only on simplified models with inputs. For every input, a parameter representing the input data is added. A callback is also generated for each input which sets the input at the appropriate time and sets the tstop to the next time the input value has to be set. Then simply generating and solving an ODEProblem like normal would replicate this. The generated parameters can be specially named using operators. So `InputValues(x)` would be a parameter for the data, and `InputTimes(x)` would be a parameter for the time points.

docs/src/basics/InputOutput.md
```diff
@@ -87,6 +87,77 @@ See [Symbolic Metadata](@ref symbolic_metadata). Metadata specified when creatin
 
 See [Linearization](@ref linearization).
 
+## Real-time Input Handling During Simulation
+
+ModelingToolkit supports setting input values during simulation for variables marked with the `[input=true]` metadata. This is useful for real-time simulations, hardware-in-the-loop testing, interactive simulations, or any scenario where input values need to be determined during integration rather than specified beforehand.
+
+To use this functionality, variables must be marked as inputs using the `[input=true]` metadata and specified in the `inputs` keyword argument of `@mtkcompile`.
```
baggepinnen, line 94: > variables must be marked as inputs using the `[input=true]` metadata

is this actually required? The inputs are specified 
> in the `inputs` keyword argument of `@mtkcompile`

anyway. No other input-using functionality, like linearization or generation of input-dependent functions, rely on the `input` metadata

src/systems/inputs.jl
bradcarman: fixed

src/systems/inputs.jl
bradcarman: fixed

src/systems/inputs.jl
bradcarman: fixed

bradcarman:
> I thought about this a bit. In my opinion, the best way to phrase this is a model transformation in `basic_transformations.jl`. It would operate only on simplified models with inputs. For every input, a parameter representing the input data is added. A callback is also generated for each input which sets the input at the appropriate time and sets the tstop to the next time the input value has to be set. Then simply generating and solving an ODEProblem like normal would replicate this. The generated parameters can be specially named using operators. So `InputValues(x)` would be a parameter for the data, and `InputTimes(x)` would be a parameter for the time points.

I really don't want to bake the input data into the `System` itself.  The system should be able to be used in determinate or indeterminate forms.  Putting the data into the `System` would work just for the determinate form, meaning we'd need a different `System` for the indeterminate form.  We should be able to serialize/precompile a single `System` and `ODEProblem`, and then quickly run that system with different data sets.  If the data gets put into the system as parameters, then each new data set needs to make (or remake) an `ODEProblem`.

ChrisRackauckas:
That's not true. If you do it with FunctionWrappers or DataInterpolations you can just swap the interpolation via `remake`. It would be type-stable and faster while also matching the normal semantics.

src/systems/inputs.jl
```diff
@@ -0,0 +1,196 @@
...
+
+        setters = [SymbolicIndexingInterface.setsym(sys, x) for x in vars]
+
+        @set! sys.input_functions = InputFunctions(events, vars, setters)
+    end
+
+    return sys
+end
+
+function CommonSolve.solve(prob::SciMLBase.AbstractDEProblem, inputs::Vector{Input}, args...; kwargs...)
```
ChrisRackauckas, line 166: It still violates the dispatch contract that SciML does. The second argument should be the algorithm, if we do this. But I don't think it makes sense to do this as a solve, it violates a few other principles of solve

src/systems/inputs.jl
ChrisRackauckas: Not fixed. Tuples are only concrete if you also choose a size for them

src/systems/inputs.jl
```diff
@@ -0,0 +1,196 @@
...
+                set_input!(integrator, input.var, input.data[i])
+            end
+        end
+        push!(callbacks, DiscreteCallback(condition, affect!))
+
+        # DiscreteCallback doesn't hit on t==0, workaround...
+        if input.time[1] == 0
+            prob.ps[input.var] = input.data[1]
+        end
+    end
```
ChrisRackauckas, line 186: This shouldn't be done here. This should instead be done by registering discontinuities for the interpolation function if it is a discontinuous one.

bradcarman: There is no interpolation here, this is explicitly a discrete input.  The point is to not have interpolation.

ChrisRackauckas: That is ConstantInterpolation

src/systems/inputs.jl
bradcarman: I want to keep the data fully separate from the `System` and the `ODEProblem`.  We need some way to be able to solve a problem where data is fed in at the top level.  I have benchmarks that show that this is much more efficient, and follows the norm for what other modeling platforms offer.  If we move `inputs` as the 3rd argument, will that be acceptable?

bradcarman:
> That's not true. If you do it with FunctionWrappers or DataInterpolations you can just swap the interpolation via `remake`. It would be type-stable and faster while also matching the normal semantics.

It's not faster, I have benchmarks that show this.

ChrisRackauckas:
No, you're not understanding... it's not faster you're just doing search reduction so you're not benchmarking the right thing.

Ultimately this just isn't the way to implement it. We just need a specialization on ConstantInterpolation that lowers it to a callback form to remove search. That needs none of the other extra interface breaking machinery here. This form is just impossible to maintain and breaks a lot of contracts, and most of it isn't necessary (and it's half broken in many scenarios 😅). That would also fix the type instabilities that are rampant in here (which end up breaking a lot of precompilation improvements)

bradcarman:
Closing, moving to: https://github.com/bradcarman/ModelingToolkitInputs.jl

PR #4005, opened 2025-11-05
Support DI@0.7 in MTK@9

devmotion:
Not sure yet whether it works - I'd like to update to DI@0.7 but I'm stuck with MTK@9 currently due to missing Catalyst support for MTK@10.


devmotion:
Test seem to use DI@0.7 👍 My impression is that breaking SciML changes are responsible for the test failures (either too loose compat bounds or MTK@9 relied on internals I assume).

AayushSabharwal:
@SebastianM-C should this Optimization failure still be happening?

SebastianM-C:
> ⌅ [bca83a33] OptimizationBase v2.11.0

This is suspicious :thinking:

PR #4011, opened 2025-11-06
Backport #3733 (update to DI@0.7) to MTK@9

devmotion:
Should of course backport the compat update instead of just bumping compat entries (#4005).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "ModelingToolkit"
 uuid = "961ee093-0014-501f-94e3-6117800e7a78"
 authors = ["Yingbo Ma <mayingbo5@gmail.com>", "Chris Rackauckas <accounts@chrisrackauckas.com> and contributors"]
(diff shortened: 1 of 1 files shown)


AayushSabharwal:
If this still resolves to something old, I'll try to fix it locally, in between other stuff.

devmotion:
Regarding the old OptimizationBase@2.11.0: It seems to be the most recent version that is compatible with MTK@9. My impression is that you'd run into the same problems without this PR when rerunning CI on the backport branch.

AayushSabharwal:
Oh, that's unfortunate. Considering this is a minor change and v9 isn't actively maintained, I'll merge this if you confirm that it works for your use case.

devmotion:
I started a downstream CI run with this PR and DifferentiationInterface@0.7.10. I'll report back whether it is successful.

devmotion:
CI passed successfully, without precompilation errors.

PR #4041, opened 2025-12-03
fix: minor fixes for clock stuff

AayushSabharwal:
## Checklist

### DIFF
--- a/src/discretedomain.jl
+++ b/src/discretedomain.jl
@@ -209,6 +209,7 @@ end
 is_transparent_operator(::Type{Hold}) = true
 
 (D::Hold)(x) = Term{symtype(x)}(D, Any[x])
+(D::Hold)(x::Number) = x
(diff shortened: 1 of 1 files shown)


src/discretedomain.jl
```diff
@@ -209,6 +209,7 @@ end
 is_transparent_operator(::Type{Hold}) = true
 
 (D::Hold)(x) = Term{symtype(x)}(D, Any[x])
+(D::Hold)(x::Number) = x
```
ChrisRackauckas, line 212: how does this come up?

AayushSabharwal: In `fixpoint_sub` when building `MTKParameters`. `Hold(x)` is stored as a parameter, so an initial value for `x` is an initial value for `Hold(x)`.

AayushSabharwal:
Enough tests pass. `pre` failures are due to 1.13 alpha, which I'm not quite ready to support just yet.

PR #4046, opened 2025-12-08
docs: fix incorrect claim about NonlinearSystem in initialization tutorial

### LINKED ISSUE #4024: Error in initialization tutorial
In line 483 of the docs/src/tutorials/initialization.md says, that after initializing the pendulum system in the right way (giving the correct number of equations), one should obtain a NonlinearSystem instead of a NonlinearLeastSquaresProblem. However, this is not the case already in the tutorial, there a NonlinearLeastSquaresProblem is obtained also in that case. I also tried it and obtain a NonlinearLeastSquaresProblem instead of a NonlinearSystem. 

Is there an issue, or should the tutorial get updated?

ChrisRackauckas-Claude:
## Summary

- Fixed documentation in `initialization.md` that incorrectly claimed a "well-conditioned" initialization system would produce a `NonlinearSystem`
- The actual behavior is that a `NonlinearLeastSquaresProblem` is returned because the pendulum system is structurally singular (has unassigned variables)
- Changed the solver example from `TrustRegion()` (which is for `NonlinearProblem`) to `GaussNewton()` (which works with `NonlinearLeastSquaresProblem`)

## Details

Line 483 of the original documentation stated:
> notice that we instead obtained a NonlinearSystem

However, running the code shows that `InitializationProblem` returns a `NonlinearLeastSquaresProblem` for this case because:
1. The system is structurally singular (has unassigned variables like `xˍtt(t)`)
2. Per `src/initialization.jl`, only systems that are both fully determined AND structurally non-singular get `NonlinearProblem`/`SCCNonlinearProblem`

## Test plan
- [x] Verified the code examples run correctly with the updated solver

Fixes #4024

### DIFF
--- a/docs/src/tutorials/initialization.md
+++ b/docs/src/tutorials/initialization.md
@@ -477,10 +477,10 @@ In comparison, if we have a well-conditioned system:
 
 ```@example init
 iprob = ModelingToolkit.InitializationProblem(pend, 0.0,
-    [x => 1, y => 0.0, g => 1], guesses = [λ => 1])
+    [x => 1, D(x) => 0.0, g => 1], guesses = [λ => 1, y => 0])
 ```
 
-notice that we instead obtained a NonlinearSystem. In this case we have to use
+notice that we instead obtained a NonlinearProblem. In this case we can use
 different solvers which can take advantage of the fact that the Jacobian is square.
 
 ```@example init


ChrisRackauckas-Claude:
## Updated PR Description

The PR has been updated with an actual code fix (not just documentation changes).

### Changes
1. **`src/initialization.jl`**: Modified `get_initialization_problem_type` to return `NonlinearProblem` for square initialization systems (`neqs == nunknown`) even if structurally singular. Only `SCCNonlinearProblem` now requires structural non-singularity.

2. **`docs/src/tutorials/initialization.md`**: Fixed typo "NonlinearSystem" → "NonlinearProblem"

### Root Cause
The structural analysis performs bipartite graph matching without numerical substitution. With the pendulum constraint differentiated to `-2xˍt*x - 2yˍt*y = 0`, the structural analysis sees both `xˍt` and `yˍt` as variables. However, when `y=0` is substituted numerically, the term `-2yˍt*y` vanishes, leaving `xˍt` uniquely determined.

### Solution
For initialization systems, structural singularity is informational only. A square system may still be solvable even if structurally singular because numerical values can eliminate terms.

### Testing
- `ODEProblem(pend, [x => 1, y => 0, g => 1], ...)` now returns `NonlinearProblem` ✓
- ODE solves correctly with `x=1, y=0, D(x)=0, D(y)=0, λ=0` ✓
- Underdetermined systems still return `NonlinearLeastSquaresProblem` ✓

Note: Please update the PR title to: **fix: return NonlinearProblem for structurally singular but square initialization systems**

ChrisRackauckas-Claude:
## PR Updated - Docs-only fix

After discussion, the fix has been simplified to a documentation-only change:

### Change
Changed the initialization example from:
```julia
[x => 1, y => 0.0, g => 1], guesses = [λ => 1]
```

To:
```julia
[x => 1, D(x) => 0.0, g => 1], guesses = [λ => 1, y => 0]
```

### Rationale
- Specifying `D(x) => 0` instead of `y => 0` is the correct way to create a well-conditioned system
- With `x=1` on the unit circle, `y` is determined by the constraint `x² + y² = 1`
- With `D(x)=0` specified, the system is fully determined and returns `SCCNonlinearProblem` (the optimal case)
- Also fixed "NonlinearSystem" → "NonlinearProblem" (the correct type name)

The code changes to `src/initialization.jl` have been reverted as they are not needed with this documentation fix.

PR #4047, opened 2025-12-09
fix: avoid false positives in consistency check

AayushSabharwal:


oscardssmith:
thanks for the quick turnaround

PR #4053, opened 2025-12-12
remove unused OrdinaryDiffEqCore dep

oscardssmith:
Primary motivation is to allow OrdinaryDiffEqCore@2 (which is needed for the OrdinaryDiffEq test suite until we update MTK test dep to v11)


AayushSabharwal:
Weird that this isn't running tests

AayushSabharwal:
Updated the CI scripts on `backport-v10`

PR #4109, opened 2026-01-05
fix: always error on missing guesses

### LINKED ISSUE #4104: Can create and simulate ODE without providing initial conditions for all variables
Pretty serious, leads to a silent and potentially nasty error.
```julia
# Fetch packages
using ModelingToolkitBase
using OrdinaryDiffEqDefault
using ModelingToolkitBase: t_nounits as t, D_nounits as D

# Create model.
@variables X1(t) X2(t)
@parameters k1 k2
eqs = [
    D(X1) ~ -k1*X1 + k2*X2
    D(X2) ~ k1*X1 - k2*X2
]
@mtkbuild sys = System(eqs, t)

sim_conds = [X1 => 3.4, k1 => 1.2, k2 => 0.13]
oprob = ODEProblem(sys, sim_conds, (0.0, 1.0))
sol = solve(oprob)

oprob[X1] # 3.4
oprob[X2] # 1.0
sol[X1][1] # 3.4
sol[X2][1] # 1.0
```

AayushSabharwal:
Close #4104 

Tests will likely need fixing.

### DIFF
--- a/lib/ModelingToolkitBase/test/components.jl
+++ b/lib/ModelingToolkitBase/test/components.jl
@@ -11,6 +11,12 @@ using BipartiteGraphs
 import SymbolicUtils as SU
 include("common/rc_model.jl")
 
+missing_guess_value = if @isdefined(ModelingToolkit)
+    MissingGuessValue.Error()
+else
+    MissingGuessValue.Constant(1.0)
+end
+
 @testset "Basics" begin
     @unpack resistor, capacitor, source = rc_model
     function check_contract(sys)
@@ -77,7 +83,7 @@ include("common/rc_model.jl")
     end
     @test !isempty(ModelingToolkitBase.bindings(sys))
     u0 = [capacitor.v => 0.0]
-    prob = ODEProblem(sys, u0, (0, 10.0))
+    prob = ODEProblem(sys, u0, (0, 10.0); missing_guess_value)
     sol = solve(prob, Rodas4(); abstol = 1.0e-8, reltol = 1.0e-8)
     check_rc_sol(sol)
 end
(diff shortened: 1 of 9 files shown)


TorkelE:
Would be happy to review, but to be honest I cannot really intepret what change the update will cause

AayushSabharwal:
This makes it so that there are no guesses for all variables by default. In your issue, the `ODEProblem` construction will error. Also, for most DAEs and some systems with observed equations, the initialization will need guesses for a bunch of extra variables. These can be provided by passing `missing_guess_value = MissingGuessValue.Constant(..)` to `ODEProblem`.

lib/ModelingToolkitBase/test/components.jl
```diff
@@ -11,6 +11,12 @@ using BipartiteGraphs
 import SymbolicUtils as SU
 include("common/rc_model.jl")
 
+missing_guess_value = if @isdefined(ModelingToolkit)
+    MissingGuessValue.Error()
+else
+    MissingGuessValue.Constant(1.0)
```
ChrisRackauckas, line 17: why wouldn't we always default to error?

AayushSabharwal: That's just for tests. When MTK is loaded, no additional guesses should be necessary. Without MTK, simplification is not strong enough to eliminate all the observed from the initialization system, so some cases need extra guesses. Note that this doesn't apply to every `ODEProblem` in this test file. Only the necessary ones pass this kwarg.

ChrisRackauckas: I see

PR #4120, opened 2026-01-06
Fix `@component`'s line number information

cstjean:
So that go-to-definition works correctly.

Before

```julia
julia> @component function MyComponent() end
MyComponent (generic function with 1 method)

julia> methods(MyComponent)
# 1 method for generic function "MyComponent" from Main:
 [1] MyComponent()
     @ C:\Users\a1058035\.julia\dev\ModelingToolkit\lib\ModelingToolkitBase\src\systems\abstractsystem.jl:2698
```

After

```julia
julia> @component function MyComponent() end
MyComponent (generic function with 1 method)

julia> methods(MyComponent)
# 1 method for generic function "MyComponent" from Main:
 [1] MyComponent()
     @ REPL[33]:1
```
The `out.args[2].args[3].args[2].args[1] = __source__` line is a bit grim, but it's unlikely to break, and `Base.@kwdef` does something similar. `MacroTools.@qq` would also work, but I didn't want to import it just for that.

### DIFF
--- a/lib/ModelingToolkitBase/src/systems/abstractsystem.jl
+++ b/lib/ModelingToolkitBase/src/systems/abstractsystem.jl
@@ -2756,8 +2756,12 @@ function component_post_processing(__source__, expr, isconnector)
             end
         end
     end
+    return set_component_line_number!(__source__, out)
+end
+
+function set_component_line_number!(__source__::LineNumberNode, expr::Expr)
     # To set the LineNumberNode correctly (and thus fix Go-to-definition), we modify the
-    # LineNumberNode in the `function` body. `dump(out)` looks like this:
+    # LineNumberNode in the `function` body. `dump(expr)` looks like this:
     #   head: Symbol block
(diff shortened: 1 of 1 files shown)


lib/ModelingToolkitBase/src/systems/abstractsystem.jl
```diff
@@ -2715,6 +2715,9 @@ function component_post_processing(expr, isconnector)
             end
         end
     end
+    @assert out.args[2].args[3].args[2].args[1] isa LineNumberNode
+    out.args[2].args[3].args[2].args[1] = __source__  # set the LineNumberNode correctly
```
AayushSabharwal: Could you add a comment indicating what each nested `getproperty` expects the expression to be? Or which `LineNumberNode` this is trying to set? On the off chance this breaks, it'll help maintainability.

AayushSabharwal (approved):

PR #4129, opened 2026-01-08
fix: only examine RHS in equation_dependencies

SebastianM-C:
Symbolics@5fe7a2d6 added a `search_variables!` method for `Equation` that searches both LHS and RHS. This broke `equation_dependencies` which relies on finding only RHS dependencies.

The original MTK behavior (from 2020, commit 3160770b56) explicitly only examined `eq.rhs`. This fix restores that behavior by passing `eq.rhs` directly instead of relying on the generic `Equation` method.

## Checklist


AayushSabharwal:
Rebasing should fix tests.

AayushSabharwal:
This breaks the dependency analysis for jumps

SebastianM-C:
I updated the fix to only use `.rhs` on `Equation`s

PR #4133, opened 2026-01-08
Fix observed variable substitution in dynamic optimization

### LINKED ISSUE #4089: Systems for dynamic optimisation create unhandled edge cases when observed variables appear in equations
I am trying to split my model appart in two subsystems a track model that creates the curvature profile and a vehicle that has curvature as an unknown variable. Both systems are interconnected with a 3rd system that just links them together as seen below.
```
@named track = TwoCornerTrack() # track model creating curvature profile
@named car = Vehicle3DOF() # vehicle model expecting  curvature profile as input 
# full system interconnecting track and vehicle
@named full_system = MTK.System([track.κ_ ~ car.κ_in], s; systems=[car, track]) 
opt_sys = MTK.mtkcompile(full_system, inputs=MTK.get_inputs(full_system))
```
The compiled system correctly reduces to the minimal set of unknowns 10 in my case and creates an observed variable that is the curvature input of the vehicle model as seen below :
```
Model full_system:
Equations (10):
  10 standard: see equations(full_system)
Unknowns (10): see unknowns(full_system)
  track₊κ_(t)
  track₊s_(t)
  car₊J_(t)
  car₊p_(t)
  car₊δ_(t)
  car₊r_(t)
  ⋮
Parameters (24): see parameters(full_system)
  car₊F_drive_max
  car₊rho
  car₊C_drag
  car₊F_brake_max
  car₊rBBal
  car₊m
  ⋮
Observed (1): see observed(full_system)

julia> observed(opt_sys)
1-element Vector{Equation}:
 car₊κ_in(t) ~ track₊κ_(t)
```
When calling `equations(opt_sys)`, however the symbol `car₊κ_in(t) ` is not replaced inside the symbolic equations.
So when the system is passed into the layers that create the infinite opt model `process_DynamicOptProblem` by calling the below : 
`julia>  jprob = MTK.InfiniteOptDynamicOptProblem(opt_sys, [], (0, 1200); dt=3)
`
The variable `car₊κ_in(t) ` is not substituted properly by the infinite opt symbols  in the equations so there is an error. 
```
ERROR: LoadError: MethodError: no method matching *(::SymbolicUtils.BasicSymbolicImpl.var"typeof(BasicSymbolicImpl)"{SymReal}, ::GeneralVariableRef)
The function `*` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  *(::Any, ::Any, ::Any, ::Any...)
   @ Base operators.jl:642
  *(::Graphs.LinAlg.Noop, ::Any)
   @ Graphs C:\Users\chris\.julia\packages\Graphs\lNprZ\src\linalg\graphmatrices.jl:216
  *(::SymbolicUtils.BasicSymbolicImpl.var"typeof(BasicSymbolicImpl)"{SymReal}, ::SymbolicUtils.BasicSymbolicImpl.var"typeof(BasicSymbolicImpl)"{TreeReal})
   @ SymbolicUtils C:\Users\chris\.julia\packages\SymbolicUtils\QB5bd\src\methods.jl:78
  ...

Stacktrace:
 [1] add_equational_constraints!(model::MTKInfiniteOptExt.InfiniteOptModel, sys::System, pmap::Dict{…}, tspan::Tuple{…})
   @ ModelingToolkitBase C:\dev\ModelingToolkit.jl\lib\ModelingToolkitBase\src\systems\optimal_control_interface.jl:500
 [2] InfiniteOptDynamicOptProblem(sys::System, op::Vector{…}, tspan::Tuple{…}; dt::Int64, steps::Nothing, tune_parameters::Bool, guesses::Dict{…}, kwargs::@Kwargs{})
   @ MTKInfiniteOptExt C:\dev\ModelingToolkit.jl\lib\ModelingToolkitBase\ext\MTKInfiniteOptExt.jl:107
 [3] top-level scope
   @ C:\dev\DLSJulia\VehicleModel3DOF copy.jl:175
in expression starting at C:\dev\DLSJulia\VehicleModel3DOF copy.jl:175
```
The above error happens at the below line of the` add_equational_constraints! ` function, because diff_equations(sys) just calls equations(sys) underneath which does not replace the observed variable symbols. 

`    diff_eqs = substitute(diff_equations(sys), rules; fold = Val(true), filterer = Returns(true))
`
This can be fixed by changing the implemention of the `diff_equations(sys) `to call full_equations(sys) instead. As that function properly replaces all observed symbols in the set of equations so substitutions with ininfite opt symbols is done properly.  However there might be unwanted consequences from this as this is a quite low level function inside the `abstractsystem.jl `file.

SebastianM-C:
When creating `InfiniteOptDynamicOptProblem`, observed variables in differential equations were not being substituted, causing a `MethodError` when the symbolic observed variable was multiplied with a JuMP `GeneralVariableRef`.

The fix adds the `get_substitutions(sys)` to the substitution rules.

## Checklist

## Additional context

Fixes #4089

### DIFF
--- a/lib/ModelingToolkitBase/src/systems/optimal_control_interface.jl
+++ b/lib/ModelingToolkitBase/src/systems/optimal_control_interface.jl
@@ -465,6 +465,12 @@ function get_model_vars_substitution_rules!(rules::Dict{Any, Any}, model, sys, t
     return nothing
 end
 
+function get_observed_substitution_rules!(rules::Dict{Any, Any}, sys)
+    # add the substitution rules for the observed variables
+    merge!(rules, get_substitutions(sys))
+    return nothing
+end
+
 """Mappings for variables that depend on the final time parameter, x(tf)."""
 function free_t_map(m, tf, x_ops, c_ops)
     return Dict(
@@ -515,6 +521,7 @@ function add_user_constraints!(model, sys, tspan, pmap)
     get_toterm_substitution_rules!(rules, cons_dvs)
     get_model_vars_substitution_rules!(rules, model, sys, tspan)
     get_param_substitution_rules!(rules, pmap)
+    get_observed_substitution_rules!(rules, sys)
     # `fixpoint_sub` to recursively substitute into `toterm` rules
     jconstraints = fixpoint_sub(jconstraints, rules; fold = Val(true), filterer = Returns(true))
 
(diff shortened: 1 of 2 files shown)


cchristoforidis93:
what about adding the observed substitution rules also in the function `add_cost_function!` . Other than that changes work for me

PR #4142, opened 2026-01-12
fix: use sol.W instead of original noise in NoiseWrapper test

ChrisRackauckas-Claude:
## Summary
- Fix broken test in `sdesystem.jl` for noise kwarg propagation (issue #3664)
- Use `sol1.W` (the solved noise) instead of `noise1` (original noise) when creating `NoiseWrapper`
- Switch to fixed-step `EM()` solver for deterministic timestep comparison

## Problem
The test was failing because:
1. StochasticDiffEq copies the noise process during solve, so the original `noise1` doesn't contain the full trajectory - only `sol1.W` does
2. Using adaptive solvers (`SRIW1()`) with `NoiseWrapper` can cause solutions to diverge because the solver may take slightly different timesteps, and interpolated noise values at different points leads to different solutions

## Solution
1. Use `NoiseWrapper(sol1.W)` to wrap the solved noise from the solution, which contains the full trajectory
2. Use fixed-step `EM()` solver with `dt = 0.01` to ensure both solves use identical timesteps, guaranteeing deterministic noise reuse

This matches the pattern used in DiffEqNoiseProcess.jl's own tests for `NoiseWrapper`.

## Test plan
- [x] Verified test passes locally with the fix
- [ ] CI tests should pass

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/lib/ModelingToolkitBase/test/sdesystem.jl
+++ b/lib/ModelingToolkitBase/test/sdesystem.jl
@@ -1079,15 +1079,18 @@ end
     noise1 = StochasticDiffEq.RealWienerProcess(0.0, 0.0, 0.0; save_everystep = true)
     u0_dict = Dict(unknowns(sde_lorentz) .=> u0)
     prob1 = SDEProblem(sde_lorentz, merge(u0_dict, Dict(parammap)), T; noise = noise1)
-    sol1 = solve(prob1, SRIW1())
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
@AayushSabharwal this should be the right fix to the test which relies on using the same noise.

PR #4150, opened 2026-01-13
fix: avoid false positives in consistency check

AayushSabharwal:
Port of #4047 to v11


AayushSabharwal:
MTK/InterfaceI symbolic event failures are https://github.com/SciML/DiffEqBase.jl/issues/1268.

PR #4154, opened 2026-01-13
Add codegen for `BVProblem`s with cost functions

SebastianM-C:
Co-Authored-By: Claude Opus 4.5 <noreply@anthropic.com>

## Checklist

## Additional context

This is needed for https://github.com/SciML/BoundaryValueDiffEq.jl/pull/413

### DIFF
--- a/lib/ModelingToolkitBase/src/systems/codegen.jl
+++ b/lib/ModelingToolkitBase/src/systems/codegen.jl
@@ -641,7 +641,7 @@ function generate_boundary_conditions(
     cons = [con.lhs - con.rhs for con in constraints(sys)]
     conssubs = Dict{SymbolicT, SymbolicT}()
     get_constraint_unknown_subs!(conssubs, cons, stidxmap, iv, BVP_SOLUTION)
-    substituter = SU.Substituter{false)(conssubs)
+    substituter = SU.Substituter{false}(conssubs)
     cons = map(substituter, cons)
 
     init_conds = SymbolicT[]


AayushSabharwal:
This needs a test

SebastianM-C:
The `Lotka-Volterra` and the `Cost function compilation` from `InterfaceII` pass locally with https://github.com/SciML/BoundaryValueDiffEq.jl/pull/413 
After that PR is merged and we have a release, I can add more tests here.

ErikQQY:
Just tagged a new release of BoundaryValueDiffEq.jl with general parameter estimation and optimal control functionality, please let me know if there are any issues :)

SebastianM-C:
I added more tests and fixed a bug with `EvalAt` in the boundary condition specification. I was not sure if we want to go with the `wrap_delays` when calling `build_function_wrapper`, so I made the minimal set of changes to get `EvalAt` working.

Another thing that I noticed is that the tests for the ODEs with constraints were failing because the constraints are evaluated internally in a different manner than the solution interpolant and for the `dt`s that were used in the tests, the discrepancy was significant. I assume that the `EvalSol` based interpolation that is internally used converges  easier than the interpolation used for the solution. I was wondering if there are any guidelines on how should we choose the `dt` because even if it's adaptive, the less accurate internal interpolant can lead to solution that don't actually satisfy the constraints to the same extent.

lib/ModelingToolkitBase/src/systems/codegen.jl
```diff
@@ -639,11 +639,11 @@ function generate_boundary_conditions(
     # sol = get_bv_solution_symbol(ns)
 
     cons = [con.lhs - con.rhs for con in constraints(sys)]
-    # conssubs = Dict()
-    # get_constraint_unknown_subs!(conssubs, cons, stidxmap, iv, sol)
-    # cons = map(x -> substitute(x, conssubs), cons)
+    conssubs = Dict{SymbolicT, SymbolicT}()
+    get_constraint_unknown_subs!(conssubs, cons, stidxmap, iv, BVP_SOLUTION)
+    cons = map(x -> substitute(x, conssubs), cons)
```
AayushSabharwal:
```suggestion
    get_constraint_unknown_subs!(conssubs, cons, stidxmap, iv, BVP_SOLUTION)
    substituter = SU.Substituter{false)(conssubs)
    cons = map(substituter, cons)
```

The `Substituter` interface caches intermediate substitutions, so it can be much faster when repeatedly substituting the same `Dict` into multiple expressions.

lib/ModelingToolkitBase/test/bvproblem.jl
```diff
@@ -338,4 +356,46 @@ end
...
+    sol = solve(bvp, MIRK4(), dt = 0.05)
+    @test SciMLBase.successful_retcode(sol)
+
+    costfn = ModelingToolkitBase.generate_bvp_cost(
+        lksys; expression = Val{false}, wrap_gfw = Val{true}
+    )
+    @test costfn(sol, bvp.p) ≈ log(sol(0.56; idxs = y(t)) + sol(0.0; idxs = x(t))) - sol(0.4; idxs = x(t))^2
+end
+
+@testset "Parameter elstimation" begin
```
AayushSabharwal:
```suggestion
@testset "Parameter estimation" begin
```

AayushSabharwal (approved):
Will merge once tests pass

lib/ModelingToolkitBase/src/systems/codegen.jl
```diff
@@ -639,11 +639,12 @@ function generate_boundary_conditions(
     # sol = get_bv_solution_symbol(ns)
 
     cons = [con.lhs - con.rhs for con in constraints(sys)]
-    # conssubs = Dict()
-    # get_constraint_unknown_subs!(conssubs, cons, stidxmap, iv, sol)
-    # cons = map(x -> substitute(x, conssubs), cons)
+    conssubs = Dict{SymbolicT, SymbolicT}()
+    get_constraint_unknown_subs!(conssubs, cons, stidxmap, iv, BVP_SOLUTION)
+    substituter = SU.Substituter{false)(conssubs)
```
AayushSabharwal:
```suggestion
    substituter = SU.Substituter{false}(conssubs)
```

Oops 😅

AayushSabharwal:
I've dug into all of the test failures here, and none of them seem to be caused by this PR (they show up on other branches, and only started cropping up today). As such, I'm merging this PR. Hopefully the other issues can be resolved soon.

PR #4192, opened 2026-01-19
Add tspan field

ommishra0719:
## Checklist

## Additional context

**Summary**
This PR adds a `tspan` field to the `System` struct and updates the `ODESystem` constructor to accept `tspan` as a keyword argument.

**Changes**
- Added `tspan` to the `System` definition in `system.jl`.
- Updated `ODESystem` to store `tspan` when provided.
- Added a test case in `test/tspan_field.jl` to verify storage.
- Ran JuliaFormatter on modified files.

**Example Usage**
```julia
sys = ODESystem(eqs, t, [x,y,z], [p]; tspan=(0.0, 10.0))
# tspan is now stored in the system!

### DIFF
--- a/lib/ModelingToolkitBase/src/systems/system.jl
+++ b/lib/ModelingToolkitBase/src/systems/system.jl
@@ -200,6 +200,10 @@ struct System <: IntermediateDeprecationSystem
     """
     tstops::Vector{Any}
     """
+    The time span of the problem.
+    """
+    tspan::Union{Nothing,Tuple{Any,Any}}
+    """
     $INTERNAL_FIELD_WARNING
     The list of input variables of the system.
     """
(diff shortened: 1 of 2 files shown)


AayushSabharwal:
Why is this necessary? The field was intentionally removed as it is very type-unstable and unnecessary. It is also somewhat orthogonal to what `System` represents: it is a symbolic description of the dynamics, and does not concern itself with details of running the model.

ommishra0719:
Thanks for the explanation! I was exploring the codebase and thought it might be convenient for self-contained models (similar to Modelica's experiment annotations), but I understand the concern regarding type stability and the separation of symbolic definition vs. simulation bounds.

I will close this PR since it doesn't align with the design goals. Thanks for the review!

PR #4201, opened 2026-01-22
refactor: change how `Break` works

### LINKED ISSUE #4185: Change the semantics of loop openings
As discussed before:

---

I have thought a bit more about the loop openings, this is the PR that introduced the current behavior
https://github.com/SciML/ModelingToolkit.jl/pull/3774/files
I think we could get an intuitive behavior by never having Break introduce a new input, it should instead reuse the input variable in the analysis point (which slightly confusingly is the array called outputs stored in the ap) and not assign any equation or default to this input, it has to be provided externally, unless it already had a default baked into the model.

This will break my usage in `ControlSystemsMTK.trajectory_ss` but I'm okay with that and the fix is for me to specify the broken inputs in the op, which is the behavior I want to have (which I already have commented out code to do https://github.com/JuliaControl/ControlSystemsMTK.jl/blob/master/src/ode_system.jl#L500). This would also solve the problem I wasted hours on a few days ago since I would get an error asking me to specify the value of the inputs in the broken connections, instead of automatically using the value derived from the initialization problem. The only remaining (reintroduced) issue is the case where I want to break the connection but actually keep the same op of the disconnected input as would have been there if the connection wan't broken (which is the current behavior). For that case, I would have to solve an initialization problem myself to find the op for the disconnected inputs, which sounds okay.

Ref: https://juliahub.slack.com/archives/D05J1SDSTNC/p1764670590577069

AayushSabharwal:
Close #4185

### DIFF
--- a/lib/ModelingToolkitBase/test/odesystem.jl
+++ b/lib/ModelingToolkitBase/test/odesystem.jl
@@ -1192,7 +1192,7 @@ end
     ps = zeros(8)
     setp(sys2, x)(ps, 2ones(2))
     setp(sys2, p)(ps, 2ones(2, 2))
-    @test_nowarn fn2(ones(4), 2ones(14), 4.0)
+    @test_nowarn fn2(ones(4), 2ones(length(ModelingToolkitBase.get_ps(sys2))), 4.0)
 end
 
 # https://github.com/SciML/ModelingToolkit.jl/issues/2969
@@ -1330,7 +1330,8 @@ end
         sys1, u + x + p[1:2]; inputs = [x...]
     )
 
-    @test obsfn(ones(length(unknowns(sys))), 2ones(2), 3ones(12), 4.0) == 6ones(2)
+    n_ps = length(ModelingToolkitBase.get_ps(sys1))
+    @test obsfn(ones(length(unknowns(sys))), 2ones(2), 3ones(n_ps), 4.0) == 6ones(2)
 end
 
 @testset "Passing `nothing` to `u0`" begin


lib/ModelingToolkitBase/test/analysis_points.jl
```diff
@@ -132,7 +132,7 @@ if @isdefined(ModelingToolkit)
 
     @testset "open_loop - $name" for (name, sys, ap) in test_cases
         open_sys, (du, u) = open_loop(sys, ap)
-        matrices, _ = linearize(open_sys, [du], [u])
```
baggepinnen, line 135: why was this change made? I hope it still works to pass an array?

lib/ModelingToolkitBase/test/analysis_points.jl
```diff
@@ -424,10 +424,8 @@ if @isdefined(ModelingToolkit)
         P_not_broken, _ = linearize(sys_inner, :u, :y)
         @test P_not_broken.A[] == -2
         P_broken, ssys = linearize(sys_inner, :u, :y, loop_openings = [:u])
```
baggepinnen, line 426: I would have expected this to error now when there is no value provided for `P_inner.input.u` in the operating point? Is there a initial condition for `P_inner.input.u` added automatically?

lib/ModelingToolkitBase/test/analysis_points.jl
AayushSabharwal: `du` is an array here. Passing an array is the intended API.

PR #4243, opened 2026-01-29
support jumps with Brownians

isaacsas:
Currently, jumps with Brownians neither give an error in mtkcompile, nor properly convert the Brownians to noise equations. This updates mtkcompile to convert them to noise equations while bypassing simplification (which currently does not support jumps). It also adds a check in the System constructor to ensure users don't pass both noise equations and Brownians.


isaacsas:
@AayushSabharwal I think this is good to go assuming the relevant tests pass. Given this modifies `_mtkcompile` it would probably be good for you to more closely review it though.

isaacsas:
Failing tests seem unrelated in InterfaceII.

PR #4251, opened 2026-01-31
fix symbol mappings in JumpProblem issue

isaacsas:
# Fix Symbol key conversion for JumpProblem with events

## Summary

When constructing a `JumpProblem` from a `System` with discrete or continuous events that have equational affects, Symbol keys in the operating point (e.g., `[:X => 1.0, :k => 0.5]`) failed to convert properly, causing a `TypeError`.

## Root Cause

In `ODEProblem`/`SDEProblem` construction, `operating_point_preprocess(sys, op)` is called inside `process_SciMLProblem` **before** `process_kwargs` calls `process_events`. This converts Symbol keys (e.g., `:X`) to Symbolic keys using the main system's symbol table.

However, `JumpProblem` called `process_events` directly with the original unprocessed `op`, bypassing this conversion. When `process_events` → `compile_equational_affect` created an `ImplicitDiscreteProblem` for the affect subsystem, it tried to convert Symbol keys using the **affect system's** symbol table, which doesn't contain all the main system's symbols. This caused Symbol keys to remain unconverted, leading to a `TypeError`.

## Fix

Add `operating_point_preprocess(sys, op)` before calling `process_events` in `JumpProblem`, making it consistent with `ODEProblem`/`SDEProblem`:

```julia
# preprocess op to convert Symbol keys to Symbolic using main system before passing
# to process_events (which may create ImplicitDiscreteProblems for affect subsystems)
op_processed = operating_point_preprocess(sys, op)
cbs = process_events(
    sys; callback, eval_expression, eval_module, op = op_processed, reset_jumps = true
)
```

## Testing

Added comprehensive tests for Symbol keys with events in JumpProblems:
- Pure discrete jump systems (MassActionJump) with discrete events
- ConstantRateJump systems with discrete events
- Hybrid ODE + jump systems with discrete events
- Multiple discrete events with Symbol keys
- VariableRateJump systems with discrete events
- Continuous events with Symbol keys

All tests verify both that construction succeeds (no TypeError) and that the events fire correctly.

### DIFF
--- a/lib/ModelingToolkitBase/src/problems/jumpproblem.jl
+++ b/lib/ModelingToolkitBase/src/problems/jumpproblem.jl
@@ -9,7 +9,7 @@
 
     has_vrjs = any(x -> x isa VariableRateJump, jumps(sys))
     has_eqs = !isempty(equations(sys))
-    has_noise = get_noise_eqs(sys) !== nothing
+    has_noise = get_noise_eqs(sys) !== nothing || !isempty(brownians(sys))
 
     if (has_vrjs || has_eqs)
         if has_eqs && has_noise
--- a/lib/ModelingToolkitBase/src/systems/system.jl
+++ b/lib/ModelingToolkitBase/src/systems/system.jl
@@ -326,6 +326,13 @@ struct System <: IntermediateDeprecationSystem
                 error()
             end
             N1 == Neq || throw(IllFormedNoiseEquationsError(N1, Neq))
+            if noise_eqs !== nothing && !isempty(brownians)
+                throw(ArgumentError(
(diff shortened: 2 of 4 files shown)


isaacsas:
@AayushSabharwal I think the fix here is complete assuming there isn't a misunderstanding of how operating point's should be used here. I marked it WIP as I just want to review the tests one last time tomorrow before merging (assuming the fix looks ok to you).

isaacsas:
@AayushSabharwal thanks for the review! I think this should be good to merge / release once tests pass. (I didn't actually change anything in the last commit, just updated to master for the other PR getting merged.)

PR #4262, opened 2026-02-03
Add `@poissonians` macro 

isaacsas:
Note: this is very WIP as I've spent a good amount of time iterating with Claude on this, but haven't yet done extensive code review of what it has generated. I'm sure there are places that need updating / changes for various reasons. I'm putting this up now in case anyone has feedback on the basic design (I will stop working on it if there issues with this approach). I will go through this more carefully and drop the WIP once I have carefully reviewed all the code changes.

Longer-term we can expand from this to do better jump classification (i.e. this does the simple thing, and only has jumps with parameters as their rate become a `ConstantRateJump`, with everything else `VariableRateJump`). We can also expand this to support marks and/or Poisson random measures, both of which would build from this anyways. 

# Add `@poissonians` Macro for Inline Poisson Process Differentials

## Summary

This PR introduces the `@poissonians` macro, enabling inline Poisson process differentials in ModelingToolkitBase equations. Poissonian variables represent Poisson counting process increments and are automatically converted to `ConstantRateJump` or `VariableRateJump` objects during system compilation.

### Declaring Poissonians

```julia
@parameters β γ
@variables S(t) I(t) R(t)

# Inline syntax
@poissonians dN_inf(β * S * I) dN_rec(γ * I)

# Block syntax
@poissonians begin
    dN_inf(β * S * I)
    dN_rec(γ * I)
end
```

### Using in Equations

```julia
eqs = [
    D(S) ~ -dN_inf,
    D(I) ~ dN_inf - dN_rec,
    D(R) ~ dN_rec
]

@named sys = System(eqs, t)
compiled_sys = mtkcompile(sys)
```

### Querying Poissonians

```julia
ispoissonian(dN)              # true/false
getpoissonianrate(dN)         # returns the rate expression
ModelingToolkitBase.poissonians(sys)  # all poissonians in system
```

### Controlling save_positions for VRJs

```julia
# Pass save_positions through mtkcompile to control VRJ saving behavior
compiled_sys = mtkcompile(sys; save_positions = (true, true))
```

### Phase 1: Macro and Metadata
- Added `POISSONIAN` to `VariableType` enum
- Added `PoissonianRateCtx` metadata for storing rate expressions
- Implemented `@poissonians` macro with `topoissonian` helper
- Added `ispoissonian` and `getpoissonianrate` accessors

### Phase 2: System Storage
- Added `poissonians` field to `System` struct
- Two-argument constructor auto-detects poissonians from equations
- Variables/parameters from rate expressions are automatically extracted
- Added `poissonians(sys)` accessor with hierarchical subsystem support

### Phase 3: Compilation to Jumps
- `_poissonians_to_jumps` extracts coefficients and builds jump affects
- Automatic classification: `ConstantRateJump` (parameter-only rates) vs `VariableRateJump` (state/time-dependent rates)
- Same poissonian in multiple equations creates single jump with multiple affects
- Mixed continuous/jump dynamics supported (continuous terms retained)
- Merges with explicitly declared jumps
- Errors on nonlinear poissonian usage (e.g., `dN * dN`)
- `save_positions` kwarg flows through `mtkcompile` to VRJs from poissonians
- User-provided VRJs retain their own `save_positions` settings
- `check_no_poissonians` validation ensures `mtkcompile` is called before `JumpProblem`

### Phase 4: Mathematical Correctness
- Integration tests verify simulation statistics match analytical formulas
- Comparison tests against direct JumpProcesses implementations
- Tests cover pure jump, mixed continuous/jump, and jump-diffusion systems

## Example: SIR Model

```julia
@parameters β γ
@variables S(t) I(t) R(t)
@poissonians dN_inf(β * S * I) dN_rec(γ * I)

eqs = [
    D(S) ~ -dN_inf,
    D(I) ~ dN_inf - dN_rec,
    D(R) ~ dN_rec
]

@named sir = System(eqs, t)
compiled = mtkcompile(sir)

# Results in 2 VariableRateJumps with proper affects:
# Jump 1: rate = β*S*I, affects = [S ~ Pre(S) - 1, I ~ Pre(I) + 1]
# Jump 2: rate = γ*I, affects = [I ~ Pre(I) - 1, R ~ Pre(R) + 1]

# Create and solve the JumpProblem
jprob = JumpProblem(compiled, [S => 999.0, I => 1.0, R => 0.0, β => 0.1/1000, γ => 0.01], (0.0, 250.0))
sol = solve(jprob, Tsit5())
```

## Example: Jump-Diffusion

```julia
@parameters μ σ λ γ
@variables X(t)
@brownians dW
@poissonians dN(λ)

# dX = μX dt + σX dW + γ dN
eqs = [D(X) ~ μ * X + σ * X * dW + γ * dN]

@named sys = System(eqs, t)
compiled = mtkcompile(sys)

# Results in:
# - 1 SDE equation: D(X) ~ μ*X with noise σ*X
# - 1 ConstantRateJump: rate = λ, affect = [X ~ Pre(X) + γ]
```

### DIFF
--- a/lib/ModelingToolkitBase/src/systems/systems.jl
+++ b/lib/ModelingToolkitBase/src/systems/systems.jl
@@ -641,19 +641,30 @@ function _poissonians_to_jumps(
         rate === nothing && continue
 
         affects = Equation[]
+        expander = Symbolics.LinearExpander(dN)
 
         for (i, eq) in enumerate(new_eqs)
             # Skip non-differential equations (only handle Differential, not Shift)
             (iscall(eq.lhs) && operation(eq.lhs) isa Differential) || continue
 
-            # Get the variable being differentiated (handles nested derivatives)
-            var, order = var_from_nested_derivative(eq.lhs)
+            # Get the differential operator and check its order
+            # Note: D(D(X)) is represented as Differential(t, 2)(X(t)), so we check the
+            # operator's order field directly rather than using var_from_nested_derivative
+            diff_op = operation(eq.lhs)
+            if diff_op.order != 1
+                throw(ArgumentError(
+                    """
+                    Higher-order derivative equation $eq found in system with poissonians. \
+                    Poissonians are only supported in first-order differential equations.
+                    """
+                ))
+            end
 
-            # Skip higher-order derivatives - jumps only work with first-order ODEs
-            order == 1 || continue
+            # Get the variable being differentiated
+            var, _ = var_from_nested_derivative(eq.lhs)
 
-            # Extract coefficient of dN using linear_expansion
-            coeff, resid, islin = Symbolics.linear_expansion(eq.rhs, dN)
+            # Extract coefficient of dN using cached LinearExpander
+            coeff, resid, islin = expander(eq.rhs)
 
             if !islin
                 throw(
--- a/lib/ModelingToolkitBase/test/poissonians.jl
+++ b/lib/ModelingToolkitBase/test/poissonians.jl
@@ -393,6 +393,23 @@ end
         @test_throws ArgumentError mtkcompile(sys)
     end
 
+    @testset "Error on higher-order derivatives with poissonians" begin
+        @parameters λ
+        @variables X(t) Y(t)
+        @poissonians dN(λ)
+
+        # Any higher-order derivative in a system with poissonians should error
+        # (model is not well-defined with jumps and higher-order derivatives)
+        eqs = [D(D(X)) ~ dN]
+        @named sys = System(eqs, t)
+        @test_throws ArgumentError mtkcompile(sys)
+
+        # Even if poissonian is in a different equation, higher-order derivative errors
+        eqs2 = [D(D(X)) ~ 0, D(Y) ~ dN]
+        @named sys2 = System(eqs2, t)
+        @test_throws ArgumentError mtkcompile(sys2)
+    end
+
     @testset "Merging with explicit jumps" begin
         @parameters λ₁ λ₂
         @variables X(t) Y(t)
--- a/lib/ModelingToolkitBase/test/runtests.jl
+++ b/lib/ModelingToolkitBase/test/runtests.jl
@@ -73,6 +73,7 @@ end
             @safetestset "NonlinearSystem Test" include("nonlinearsystem.jl")
             @safetestset "PDE Construction Test" include("pdesystem.jl")
             @safetestset "JumpSystem Test" include("jumpsystem.jl")
+            @safetestset "Poissonians Test" include("poissonians.jl")
             @safetestset "Optimal Control + Constraints Tests" include("bvproblem.jl")
             @safetestset "print_tree" include("print_tree.jl")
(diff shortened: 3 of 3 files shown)


isaacsas:
One question is if we prefer
```julia
@poissonians a(rate_expr)
```
or 
```julia
@poissonians a [rate = expr]
```
i.e. explicitly typing the rate metadata. The first seems cleaner to me and more inline with what we write mathematically, but the second has the advantage that we can reuse `parse_vars` so don't need any new macro parsing code.

isaacsas:
@AayushSabharwal I think this is good to go assuming the non-broken tests pass. But let me know if you have any feedback. Note that while the amount of code is long, most of it is tests. The MTKBase changes are not too much.

lib/ModelingToolkitBase/src/systems/systems.jl
```diff
@@ -605,6 +610,151 @@ function extract_brownians_to_noise_eqs(sys::AbstractSystem)
...
+
+        for (i, eq) in enumerate(new_eqs)
+            # Skip non-differential equations (only handle Differential, not Shift)
+            (iscall(eq.lhs) && operation(eq.lhs) isa Differential) || continue
+
+            # Get the variable being differentiated (handles nested derivatives)
+            var, order = var_from_nested_derivative(eq.lhs)
+
+            # Skip higher-order derivatives - jumps only work with first-order ODEs
+            order == 1 || continue
```
AayushSabharwal: Should this error if a poissonian is present in a higher-order derivative equation?

lib/ModelingToolkitBase/src/systems/systems.jl
```diff
@@ -605,6 +610,151 @@ function extract_brownians_to_noise_eqs(sys::AbstractSystem)
...
+
+            # Extract coefficient of dN using linear_expansion
+            coeff, resid, islin = Symbolics.linear_expansion(eq.rhs, dN)
```
AayushSabharwal: A faster way to do this which scales better is to cache `expander = Symbolics.LinearExpander(dN)` outside the `enumerate(new_eqs)` loop and use `coeff, resid, islin = expander(eq.rhs)`. I'm not sure how large your systems are so this may be a premature optimization. Feel free to ignore this if the systems aren't very large or the expressions not very deep. The difference will be basically invisible in such cases.

lib/ModelingToolkitBase/src/systems/systems.jl
isaacsas: Yeah, that is a good idea.

lib/ModelingToolkitBase/src/systems/systems.jl
isaacsas: OK, will do. No reason not to use the better method if the change is minor.

isaacsas:
@AayushSabharwal I addressed your comments. I had apparently forgotten to put the new test file in runtests.jl, so I've now added it to InterfaceII right after the jump tests. It passes locally, so hopefully it will now pass on CI. Once that successfully finishes this should be good.

isaacsas:
Thanks!

PR #4266, opened 2026-02-03
[v10] fix: validate that `InferredDiscrete` ends up in a discrete partition

AayushSabharwal:


test/clock.jl
```diff
@@ -170,6 +170,15 @@ SciMLBase.is_discrete_time_domain(::ZeroArgOp) = true
     @test clkmap[ZeroArgOp()()] == clk
 end
 
+@testset "`InferredDiscrete` validation" begin
+    k = ShiftIndex()
+    @variables x(t) y(t)
+    @named sys = System([x(k) ~ x(k-1) + x(k-2), D(y) ~ Hold(x) + t], t)
+    @test_throws ModelingToolkit.ExpectedDiscreteClockPartitionError mtkcompile(sys)
+    @named sys = System([x(k) ~ x(k-1) + x(k-2), D(y) ~ x + t], t)
+    @test_throws ModelingToolkit.ExpectedDiscreteClockPartitionError mtkcompile(sys)
```
baggepinnen, line 179: This looks surprising to me, what is expected to be discrete but isn't in this system?

AayushSabharwal: `x` is expected to be discrete, but it has no discrete clock associated with it. Note how `k` isn't passed a clock.

baggepinnen: I see, thanks. Eventually, we'd like to support discrete-time models where the clock is not specified. You could not simulate such a system, but you could generate the code for it as long as you can tell that all variables are on the same (unknown) clock. When you deploy code that is called from an external runtime, the model does not need to know its clock, just what to do when called.

AayushSabharwal: Such a system would be purely discrete, right? We can codegen that already

baggepinnen: > Such a system would be purely discrete, right?

Yes

> We can codegen that already

:+1:

PR #4269, opened 2026-02-04
Fix extend

### LINKED ISSUE #4267: `extend` does not handle Brownians or jumps
I noticed that `extend` seems to be somewhat out of date, for example it misses Brownians and jumps. Is that intentional? If not I will PR a fix for that (along with Poissonians if/when that gets merged).

Likewise, with the move to one `System` type, should `extend`'s signature be tightened to just take `System`s? I'm not sure it makes sense given everything it currently does, and how it doesn't check for the existence of "non-core" fields before getting them, to keep it defined for `AbstractSystem`s.

isaacsas:
## Summary

Fixes #4267

The `extend` function was not properly merging `brownians`, `poissonians`, and `jumps` fields when combining two systems. These fields were simply lost because they weren't being unioned or passed to the System constructor.

## Changes

**`lib/ModelingToolkitBase/src/systems/abstractsystem.jl`**
- Added union of `brownians` from both systems (passed as positional arg)
- Added union of `poissonians` from both systems (passed as kwarg)
- Added union of `jumps` from both systems (passed as kwarg)

## Test Coverage

New test file `extend_sde_jump.jl` with comprehensive tests:

- **Brownians**: disjoint, shared, ODE+SDE combinations
- **Poissonians**: disjoint, shared, constant rate, variable rate, ODE+Poissonian
- **Jumps**: ConstantRateJump, MassActionJump, ODE+Jump
- **Mixed combinations**: brownians+poissonians, brownians+jumps, poissonians+jumps, all three
- **Priority semantics**: verifies `sys` values preferred over `basesys` for initial conditions
- **Correctness tests**: statistical validation that extended systems compile and solve correctly

### DIFF
--- a/lib/ModelingToolkitBase/src/precompile.jl
+++ b/lib/ModelingToolkitBase/src/precompile.jl
@@ -111,6 +111,7 @@ PrecompileTools.@compile_workload begin
     v = [p]
     isempty(v)
     # mtkcompile(sys)
+    ODEProblem(mtkcompile(System([ModelingToolkitBase.D_nounits(x) ~ 2x + 1], ModelingToolkitBase.t_nounits; name = :a)), [x => 1], (0.0, 1.0))
 end
 
 precompile(Tuple{typeof(SymbolicUtils.isequal_somescalar), Float64, Float64})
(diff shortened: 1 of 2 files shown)


isaacsas:
@AayushSabharwal I think this is good too now. Thanks!

isaacsas:
Thanks! If we could get a release when you have a chance that would be great.

PR #4274, opened 2026-02-07
fix: make @poissonians return a Vector like @variables and @brownians

isaacsas:
## Summary

`@poissonians` returned inconsistent types compared to `@variables` and `@brownians`:

| Case | `@variables` / `@brownians` | `@poissonians` (before) |
|---|---|---|
| Single, inline | `[x]` (1-element `Vector`) | bare `Num` |
| Multiple, inline | `[x, y]` (`Vector`) | `(x, y)` (`Tuple`) |
| Block syntax | `[x, y]` (`Vector`) | bare `Num` or `Tuple` |

This happened because `@variables` and `@brownians` delegate to `Symbolics.parse_vars` (which always returns a `Vector` via `Expr(:vect, ...)`), while `@poissonians` used custom codegen that returned a bare name for single declarations and an `Expr(:tuple, ...)` for multiple.

## Fix

One-line change in `_poissonians`: replace the conditional bare/tuple return with `Expr(:vect, names...)` so it always returns a `Vector`.

No existing tests break since all current usages rely on the side effect of binding variables into scope, not on the return value. New tests verify the return type across all permutations (single/multiple, inline/block).

### DIFF
--- a/lib/ModelingToolkitBase/src/variables.jl
+++ b/lib/ModelingToolkitBase/src/variables.jl
@@ -737,12 +737,8 @@ function _poissonians(exprs...)
         end)
     end
 
-    # Return the variables as a tuple (or single if only one)
-    if length(names) == 1
-        return Expr(:block, assignments..., names[1])
-    else
-        return Expr(:block, assignments..., Expr(:tuple, names...))
-    end
+    # Return the variables as a Vector, consistent with @variables and @brownians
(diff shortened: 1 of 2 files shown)


isaacsas:
@AayushSabharwal if relevant tests pass this should be good to go. It is a minor change to just be consistent with `@variables` and always return a vector.

PR #4298, opened 2026-02-11
Enable sys.x variable accessor pattern for PDESystem

ctessum-claude:
## Summary

- Adds `var_to_name` field to `PDESystem` and builds it from dependent variables and parameters in the constructor, enabling `getvar`'s fast O(1) variable lookup
- Rewrites `Base.getproperty` for `PDESystem` to fall through to `AbstractSystem`'s `getvar` for symbolic variable lookup (`sys.x` pattern) while preserving struct field access and deprecated name aliases
- Adds accessor bridges (`get_unknowns`, `has_unknowns`) mapping PDESystem's `dvs` field to the `unknowns` API expected by `getvar`
- Adds `Base.propertynames` override listing dependent variables, parameters, independent variables, and subsystems
- Handles independent variables without namespacing, matching `System` behavior
- Fixes a bug where `depwarn` calls in the original `getproperty` were dead code (placed after `return` statements)

## Test plan

- [x] All existing PDE tests pass (analytic_func access, independent_variables check)
- [x] New tests verify `sys.x` accessor for dependent variables, parameters, and independent variables
- [x] New tests verify `sys.x` accessor works without parameters (NullParameters)
- [x] New tests verify subsystem listing via `propertynames`
- [x] New tests verify namespacing behavior matches `System` (IVs not namespaced, DVs/params namespaced)
- [x] New tests verify struct field access still works (`sys.eqs`, `sys.bcs`)
- [x] New tests verify `ArgumentError` on nonexistent variables

### DIFF
--- a/lib/ModelingToolkitBase/src/systems/pde/pdesystem.jl
+++ b/lib/ModelingToolkitBase/src/systems/pde/pdesystem.jl
@@ -89,6 +89,10 @@ struct PDESystem <: AbstractSystem
     Metadata for MTK GUI.
     """
     gui_metadata::Union{Nothing, GUIMetadata}
+    """
+    Mapping from variable names to symbolic variables, enabling the `sys.x` accessor pattern.
+    """
+    var_to_name::Dict{Symbol, SymbolicT}
     @add_kwonly function PDESystem(
             eqs, bcs, domain, ivs, dvs,
             ps = SciMLBase.NullParameters();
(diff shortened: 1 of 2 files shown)


ctessum:
This is necessary (I think) because currently there's no good way to get the variables out of a PDESystem in order to (for example) index an individual variable out of a solution object. The goal of this PR is to make the PDESystem use the same method for variable accessing that the normal `System` uses.

AayushSabharwal (approved):

PR #4301, opened 2026-02-12
feat: add `AssignmentAffect`

### LINKED ISSUE #4031: Syntax for simple assignment events
With the new MTK event, we can pretty much define all manners of events. However, by being very general, it complicates things for defining many simpler, more common events. I.e. an event which updated a parameter at time 1.0 previously was:
```julia
event = [1.0] => [p ~ 1.0]
```
while now it requires first importing the discrete event callback, and then designating that `p`'s value should change:
```julia
using ModelingToolkit:SymbolicDiscreteCallback,
event = SymbolicDiscreteCallback([1.0] => [p ~ 1.0]; discrete_parameters = [p])
```
this is especially a problem in Catalyst's DSL where we currently have no means of designating options like `discrete_parameters`, which means that we currently cannot support events that upate parameters.

Would it be possible to have some kind of simplified syntax for an affect on the form:
```julia
... => [X ~ ...]
```
where the rhs is a single thing which is updated according to the expression on the right (where `Pre` can be implied or explicitly required). That way, we could support this in the Catalyst DSL (and require people who wish to declare more complicated events to do so programmatically. If this actually turns out to be something that people would like to do frequently, although I doubt it, we can then figure out how to support this in the DSL).

We could just implement this directly in Catalyst, however, it would be preferable to have something MTK generic we could point to instead of having a separate syntax.

Alternative ways to do it would to have a separate event constructor, or to have a different syntax in the symbolic, e.g. something like
```julia
... => [X = ...]
... => [X := ...]
... => [X <= ...]
```
A different Constructor would save introducing a new syntax, but a new syntax like this would enable the creation of events without importing `SymbolicDiscreteCallback` or an equivalent.

### LINKED ISSUE #4239: Enable `AssignmentAffect` for simple events
Previously, you had events on the form `p ~ p + 1`, where the left-hand side was the symbolic variable you wanted to update, and the right-hand side the expression you wanted to assign it to. With the new events, we can make much more general events, however, it complicated the simple eventsl ike this as you now have to specify `discrete_parameters`, which requires importing `ModelingToolkitBase.SymbolicDiscreteCallback` and similar.

Per previous discussion, it would be useful to have a special constructor that assmues the old simplified affect syntax. I.e. you could do
```julia
... => AssignmentAffect(p ~ p + 1)
```
which is automatically inferred to be
```julia
SymbolicDiscreteCallback(... => p ~ Pre(p) + 1; discrete_parameters = p)
```

The reason this is useful is that, within the Catalyst DSL, we do not have the option of using arguments as `discrete_parameters`. My soolution would be:
1) The DSL only supports the simple assignment events (this is >95% of the events we would expect anyway).
2) People who want to use more complicated events either have to interpoalte them into the DSL, or use programmatic modelling).

Currently, I have a normal implementation of this, however, if MTK had a `AssignmentAffect` or similar, I could simply dispatch to that, and MTK could do the related work, which probably is a more sensible solution. It could likely be useful to others as well who are in similar situations.

AayushSabharwal:
Close #4239
Close #4031

### DIFF
--- a/docs/src/API/model_building.md
+++ b/docs/src/API/model_building.md
@@ -182,6 +182,7 @@ in the model. They compile to standard callbacks from `DiffEqCallbacks.jl`.
 ```@docs
 ModelingToolkit.SymbolicContinuousCallback
 ModelingToolkit.SymbolicDiscreteCallback
+ModelingToolkit.AssignmentAffect
 ```
 
 The affect functions for the above callbacks can be symbolic or user-defined functions.
--- a/lib/ModelingToolkitBase/src/ModelingToolkitBase.jl
+++ b/lib/ModelingToolkitBase/src/ModelingToolkitBase.jl
@@ -328,6 +328,8 @@ export DynamicOptSolution
 export MissingGuessValue
 export MiscSystemData
 
+export AssignmentAffect
+
 const set_scalar_metadata = setmetadata
 
 @public apply_to_variables, equations_toplevel, unknowns_toplevel, parameters_toplevel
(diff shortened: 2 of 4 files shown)


TorkelE (approved):

TorkelE:
Looks like exactly what we want

PR #4302, opened 2026-02-12
feat: add `maybe_zeros`

AayushSabharwal:
Allows specifying expressions that may be zero. These are then given special consideration during simplification.

### DIFF
--- a/lib/ModelingToolkitBase/src/systems/system.jl
+++ b/lib/ModelingToolkitBase/src/systems/system.jl
@@ -281,6 +281,10 @@ struct System <: IntermediateDeprecationSystem
     """
     irreducibles::AtomicSetT
     """
+    Expressions which may be zero and should be given special consideration during simplification.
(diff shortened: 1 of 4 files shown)


lib/ModelingToolkitBase/src/systems/system.jl
```diff
@@ -281,6 +281,10 @@ struct System <: IntermediateDeprecationSystem
     """
     irreducibles::AtomicSetT
     """
+    Variables which may be zero and should be given special consideration during simplification.
```
baggepinnen: Variables or expressions?

AayushSabharwal: Right, I didn't update the docstring. Thanks for catching that.

AayushSabharwal:
Since the recent push is just changing a docstring and exporting the function, I'm merging.

PR #4328, opened 2026-02-19
Symbolic tstops for all appropriate problem types

isaacsas:
Note that periodic tstops with StochasticDiffEq have an issue with it not handling tstops at `tspan[1]` in a consistent way to OrdinaryDiffEq (which causes `dtmin` errors). If/when that gets fixed we can add tests for that case. Non-periodic tstops in SDEProblems should work once https://github.com/SciML/StochasticDiffEq.jl/pull/678 is merged and released.

### DIFF
--- a/src/systems/systems.jl
+++ b/src/systems/systems.jl
@@ -127,7 +127,8 @@ function MTKBase.__mtkcompile(
             guesses = guesses(sys), initialization_eqs = initialization_equations(sys),
             continuous_events = continuous_events(sys),
             discrete_events = discrete_events(sys),
-            gui_metadata = get_gui_metadata(sys)
+            gui_metadata = get_gui_metadata(sys),
+            tstops = symbolic_tstops(sys)
         )
         return ssys
     end


cstjean:
Will that enable specifying the `tstops` automatically in [`ParametrizedInterpolation`](https://github.com/SciML/ModelingToolkitStandardLibrary.jl/blob/cb592ce23f3979cbd131cd090f6c08cad9e8f9c4/src/Blocks/sources.jl#L824-L849)?

isaacsas:
This isn’t adding anything new for purely ODE or DAE models as they already support symbolic tstops being passed to a system. But that component could certainly create the System with symbolic tstops already.

isaacsas:
This doesn’t add automatic tstop creation for periodic or preset time events. It just allows you to pass symbolic parameter expressions for tstops to a System.

isaacsas:
@AayushSabharwal I think this is good unless the inclusion of `tspan[1]` in periodic tstops is a bug as I mentioned on Slack.

isaacsas:
@AayushSabharwal I updated the periodic symbolic tstops as we discussed to exclude tspan[1]. So assuming tests pass this is now complete.

AayushSabharwal:
The tests in `ModelingToolkit/InterfaceII` (ignoring the analysis point one) are failing. The failure is in symbolic tstops support for SDEs.

isaacsas:
@AayushSabharwal I think that should not be handled.

PR #4329, opened 2026-02-20
 fix: handle edge case in linear system identification 

AayushSabharwal:


AayushSabharwal:
This fixes a correctness bug, so I'll merge and follow up with a PR to fix CI

PR #4354, opened 2026-02-24
Benign ci fixes

cmhyett:
## Checklist

## Additional context
Whole point is to do no-op changes that remove noise (false negatives) from CI on e.g., PRs. 

- [X] Formatting
- [X] Downgrade
- [X] `pre` tests
- [x] any failing CI tests (that shouldn't fail)
- [ ] ~~Integration tests~~
  - Integration test failures look real now, need individual (likely not benign) fixes.

### DIFF
--- a/lib/ModelingToolkitBase/test/variable_parsing.jl
+++ b/lib/ModelingToolkitBase/test/variable_parsing.jl
@@ -34,8 +34,13 @@ s1 = Num(SSym(:s; type = Real, shape = ShapeVecT()))
 @test ModelingToolkitBase.isparameter(s)
 @test ModelingToolkitBase.isparameter(σ)
 
-@test Base.remove_linenums!(@macroexpand(@parameters x, y, z(t))) == Base.remove_linenums!(@macroexpand(@parameters x y z(t)))
-@test Base.remove_linenums!(@macroexpand(@variables x, y, z(t))) == Base.remove_linenums!((@macroexpand(@variables x y z(t))))
+function expr_equal(a, b)
+    a isa Expr && b isa Expr || return a == b
+    a.head === b.head && length(a.args) == length(b.args) &&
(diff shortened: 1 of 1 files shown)


lib/ModelingToolkitBase/Project.toml
```diff
@@ -156,7 +156,7 @@ SciMLPublic = "1.0.0"
 SciMLStructures = "1.7"
 Serialization = "1"
 Setfield = "0.7, 0.8, 1"
-SimpleNonlinearSolve = "0.1.0, 1, 2"
+SimpleNonlinearSolve = "1, 2"
```
cmhyett, line 159: I thought there were only going to be a few old compats, but it seems it is a pretty chronic issue (and getting them right throughout sciml is nontrivial). So happy to revert this commit if recommended.

test/jet/jet_tests.jl
```diff
@@ -0,0 +1,7 @@
+using ModelingToolkit
+using JET
+using Test
+
+@testset "JET static analysis" begin
+    # TODO: Add JET tests for ModelingToolkit
```
AayushSabharwal, line 6: We don't need this file and the corresponding things in MTK until we have JET tests in MTK, right?

cmhyett: Correct, I copied the pattern from MTKBase to here because JET had also been added to MTK/Project.toml. But yes, unlike MTKBase, there are no JET tests yet, so this is purely performative currently.

AayushSabharwal:
Tests pass here. Please ping me once the conflict is resolved and I'll merge as soon as possible.

cmhyett:
Had a messy rebase, will wait on CI before recommending.

cmhyett:
New commits broke even more tests. Scope of this PR would balloon trying to fix upstream issues as they're introduced, not a good use of time.

PR #4359, opened 2026-02-26
fix: parent guesses should override child defaults in guesses()

AnasAbdelR:
## Summary

`guesses()` in `abstractsystem.jl` uses `merge(guess, namespace_guesses(subsys))` which lets child subsystem variable-metadata guesses overwrite the parent's explicit overrides, because `merge(a, b)` in Julia lets `b` win on key conflicts.

Swapping to `merge(namespace_guesses(subsys), guess)` makes the parent's overrides take priority, which is the expected behavior - when a user sets `guess` at the parent level, it should not be silently overwritten by the child's default metadata.

## Fix

```julia
# Before (child overwrites parent)
guess = merge(guess, namespace_guesses(subsys))

# After (parent overrides child)
guess = merge(namespace_guesses(subsys), guess)
```

## Reproduction

```julia
using ModelingToolkit
using ModelingToolkit: t_nounits as t

@component function Child(; name)
    @variables x(t), [guess = 0.5]
    eqs = [x^2 ~ 1]
    System(eqs, t; name)
end

@component function Parent(; name)
    @named child = Child()
    System(Equation[], t; systems=[child], guesses=[child.x => 0.9], name)
end

@named parent = Parent()
simplified = mtkcompile(parent)

gs = ModelingToolkit.guesses(simplified)
```

### DIFF
--- a/lib/ModelingToolkitBase/test/guess_propagation.jl
+++ b/lib/ModelingToolkitBase/test/guess_propagation.jl
@@ -109,5 +109,5 @@ prob = ODEProblem(sys, [x0 => 1.0], (0.0, 1.0))
     @named parent = Parent()
     gs = guesses(parent)
     val = gs[only(keys(gs))]
-    @test isequal(val, 0.9)
+    @test isequal(unwrap_const(val), 0.9)
 end


lib/ModelingToolkitBase/src/systems/abstractsystem.jl
```diff
@@ -1685,7 +1685,7 @@ function guesses(sys::AbstractSystem)
     systems = get_systems(sys)
     isempty(systems) && return guess
```
AayushSabharwal, line 1686:
```suggestion
    isempty(systems) && return guess
    guess = copy(guess)
```

lib/ModelingToolkitBase/src/systems/abstractsystem.jl
```diff
@@ -1685,7 +1685,7 @@ function guesses(sys::AbstractSystem)
...
     for subsys in systems
-        guess = merge(guess, namespace_guesses(subsys))
+        guess = merge(namespace_guesses(subsys), guess)
```
AayushSabharwal:
```suggestion
        guess = left_merge!(guess, namespace_guesses(subsys))
```

`left_merge!` is an MTK utility. This avoids a lot of unnecessary copying.

AnasAbdelR:
@AayushSabharwal excellent, I see you have pushed commits for the suggestion already, I have updated the test to properly use `isequal` instead and re-requited your review. C.I should pass now

AayushSabharwal:
If everything else passes, I'll fix the test and merge

lib/ModelingToolkitBase/test/guess_propagation.jl
```diff
@@ -92,3 +92,22 @@ prob = ODEProblem(sys, [x0 => 1.0], (0.0, 1.0))
...
+
+    @component function Parent(; name)
+        @named child = Child()
+        System(Equation[], t; systems=[child], guesses=[child.x => 0.9], name)
+    end
+
+    @named parent = Parent()
+    gs = guesses(parent)
+    val = gs[only(keys(gs))]
+    @test isequal(val, 0.9)
```
AayushSabharwal:
```suggestion
    @test isequal(unwrap_const(val), 0.9)
```

PR #4364, opened 2026-02-27
Enforce pre-scaled MassActionJump rates in Systems with jumps

isaacsas:
Add SymbolicMassActionJump convenience constructor that ensures scale_rates = false for symbolic MassActionJumps used in Systems with jumps. Validate at System construction time, rejecting rescale_rates_on_update = true with a clear error. Remove dead scalerates! branch from JumpSysMajParamMapper. Fix namespace_jump to preserve scale_rates = false. Migrate all test MassActionJump calls to SymbolicMassActionJump. Add higher-order parameter update tests. Bump JumpProcesses compat to 9.23.

### DIFF
--- a/lib/ModelingToolkitBase/src/systems/system.jl
+++ b/lib/ModelingToolkitBase/src/systems/system.jl
@@ -1266,8 +1266,8 @@ end
     SymbolicMassActionJump(rate, reactant_stoch, net_stoch; kwargs...)
 
 Construct a `MassActionJump` with `scale_rates = false`, suitable for use in a
-`JumpSystem`. The rate expression must already include any combinatorial scaling
-(e.g. `k / factorial(n)` for an n-th order homotrimerization reaction).
+`JumpSystem`. The rate expression must already include any desired combinatorial scaling
+(e.g. `k / factorial(n)` for a reaction like `n*A --> ...`).
 
 Returns a `MassActionJump` — this is a convenience constructor, not a new type.
 """


isaacsas:
@AayushSabharwal as far as I (and Claude) can tell the failures are all pre-existing or Github issues (two of the failures are reporting that CI can't find Julia). I've restarted the tests, but I think this should be good to go and these are just preexisiting master failures.

AayushSabharwal:
Yep this looks good. A _bunch_ of stuff started failing all at once on Friday morning (my Friday morning) and despite a couple hours of digging I'm unable to identify why. I wish there was a way to tell Pkg "resolve this environment as if it is 26th Feb" and see the diff.

isaacsas:
Github CI has also been really bad lately. The frequency of tests failing for weird reasons like "Julia not being available" or (incorrect) claims about availability of certain package versions is quite high now.

ChrisRackauckas:
That's just because we're slamming our internal machines so hard they ran out of disk space 😅 😅 😅 😅 😅 😅 😅 😅 😅 😅 😅 😅 😅 😅 😅 😅 😅  once I am done with a bit of the OrdinaryDiffEq FBDF changes that should slow down a bit.

isaacsas:
Catalyst CI is running on internal machines? I had no idea!

ChrisRackauckas:
All of SciML is boosted with about 60-80 internal runners, I'd have to count to know the exact number. But we definitely need more. So we have 60 github actions, up from the normal 30 due to Github Education, + 60-80 from internal machines. We're just slamming CI really really hard these days 😅

PR #4366, opened 2026-03-01
feat: [AI] use `LinearProblem` for linear initialization systems in MTKBase

AayushSabharwal:


lib/ModelingToolkitBase/src/problems/initializationproblem.jl
```diff
@@ -192,3 +196,20 @@ function Base.showerror(io::IO, e::IncompleteInitializationError)
...
+    linprob = LinearProblem{iip}(sys, op; u0_constructor, kwargs...)
+    # Required for filling missing parameter values when this is an initialization
+    # problem
+    if state_values(linprob) === nothing
+        linprob = remake(
+            linprob;
+            u0 = u0_constructor(ones(eltype(linprob.A), size(linprob.A, 2)))
+        )
+    end
+    return SCCNonlinearProblem((linprob,), (Returns(nothing),), parameter_values(linprob), true; sys)
```
ChrisRackauckas, line 215: That's a bit of a kludge, we should just make a linearproblem?

AayushSabharwal: `LinearProblem` looks and works significantly different from most other `*Problem`s in SciML. The initialization infrastructure doesn't like working with it directly. `SCCNonlinearProblem` basically serves as a "make this look like a NonlinearProblem" wrapper while still actually using LinearSolve.jl directly.

AayushSabharwal:
Everything is fine here except for the test this PR adds. It doesn't account for the new improved tearing in MTK. I'll push the changes and merge when the initialization tests pass.

AayushSabharwal:
Tests pass. I'll resolve the conflict and merge.

PR #4367, opened 2026-03-02
fix: [AI] handle inputs/outputs in purely algebraic systems

### LINKED ISSUE #4352: `mtkcompile` does not turn inputs to parameters
**Describe the bug 🐞**

During `mtkcompile`, according to the docstring inputs are transformed into parameters. So I expect the transformed system to have those parameters. This used to be the case in MTK 10 but is broken in 11.

```julia
using ModelingToolkitBase, SciCompDSL
using ModelingToolkitBase: t_nounits as t

@mtkmodel InputOutput begin
    @variables begin
        in(t), [input=true]
        out(t), [output=true]
    end
    @parameters begin
        K
    end
    @equations begin
        out ~ K*in
    end
end
@named io = InputOutput()

ins = ModelingToolkitBase.unbound_inputs(io)
outs = ModelingToolkitBase.unbound_outputs(io)

reduced = mtkcompile(io, inputs=ins, outputs=outs)
parameters(reduced)
```
Just results in `K`, but `in` should appear as parameter too. 
```julia
julia> parameters(reduced)
1-element Vector{SymbolicUtils.BasicSymbolicImpl.var"typeof(BasicSymbolicImpl)"{SymReal}}:
 K
```

AayushSabharwal:
Close #4352

### DIFF
--- a/lib/ModelingToolkitBase/src/systems/systems.jl
+++ b/lib/ModelingToolkitBase/src/systems/systems.jl
@@ -251,9 +251,13 @@ function __mtkcompile(
             push!(observables, eq.lhs)
         end
         setdiff!(flat_dvs, observables)
+        new_ps = [get_ps(sys); collect(inputs)]
         @set! sys.eqs = eqs
         @set! sys.unknowns = flat_dvs
         @set! sys.observed = obseqs
+        @set! sys.ps = new_ps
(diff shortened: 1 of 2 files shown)


AayushSabharwal:
Alright, this passes enough tests given the current state of CI

PR #4369, opened 2026-03-03
test: fix tests that relied on stepping behavior

AayushSabharwal:


AayushSabharwal:
lts/SII failure is a tolerance issue. I'm merging to unblock other PRs, and will follow up with more fixes soon.

PR #4370, opened 2026-03-03
feat: add utility for moving variable bindings to initial conditions

AayushSabharwal:
Required by Catalyst, since `ReactionSystem` can codegen to both dynamical problems and nonlinear (steady state) problems. This allows moving the required bindings to initial conditions since nonlinear systems do not allow variable bindings.


AayushSabharwal:
New failures are not due to this PR

PR #4385, opened 2026-03-11
Delete outdated ref to @optimization_model

cstjean:
That code no longer runs, and `@optimization_model` does not seem to exist anywhere.


AayushSabharwal:
Wow, I had no idea that that even existed. Thanks!

PR #4388, opened 2026-03-12
 feat: allow using SciMLBase.Clock as SymbolicDiscreteCallback condition

AayushSabharwal:


lib/ModelingToolkitBase/src/systems/callbacks.jl
```diff
@@ -977,6 +992,14 @@ function generate_callback(cb, sys; kwargs...)
                 finalize, initializealg = cb.reinitializealg, saved_clock_partitions,
                 initialize_save_discretes = cb.initialize_save_discretes
             )
+        elseif is_timed && trigger isa SciMLBase.TimeDomain
+            trigger_at_init = iszero((tspan[1] - trigger.phase) % trigger.dt)
+            return PeriodicCallback(
```
BenChung, line 997: Could dispatch be used here on trigger to let this be more extensible post-facto?

AayushSabharwal: Yes. This entire file is a bit of a mess and I'm considering getting Claude to fix it.

PR #4399, opened 2026-03-18
Fix parameter bindings and observed variable substitution for dynamic optimization problems

SebastianM-C:
## Checklist

## Additional context

Bindings were not considered at all in the substitution rules, so with the MTK@v11 update we need more than just to substitute the observed variables.

### DIFF
--- a/lib/ModelingToolkitBase/src/systems/optimal_control_interface.jl
+++ b/lib/ModelingToolkitBase/src/systems/optimal_control_interface.jl
@@ -330,7 +330,7 @@ function process_DynamicOptProblem(
     # Resolve parameter bindings so observed equations and constraints
     # referencing pre-binding names can be fully substituted.
     for (k, v) in bindings(sys)
-        ismissing(v) && continue
+        v === COMMON_MISSING && continue
         haskey(pmap, v) && !haskey(pmap, k) && (pmap[k] = pmap[v])
     end
 


lib/ModelingToolkitBase/src/systems/optimal_control_interface.jl
```diff
@@ -327,6 +327,13 @@ function process_DynamicOptProblem(
     tunable_set = Set(default_toterm.(tunable_params))
     pmap = filter(kvp -> first(kvp) ∉ tunable_set, pmap)
 
+    # Resolve parameter bindings so observed equations and constraints
+    # referencing pre-binding names can be fully substituted.
+    for (k, v) in bindings(sys)
+        ismissing(v) && continue
```
AayushSabharwal: `bindings(sys)` is an `AbstractDict{SymbolicT, SymbolicT}` so `ismissing` won't work as intended.
```suggestion
        v === COMMON_MISSING && continue
```

lib/ModelingToolkitBase/test/extensions/Project.toml
```diff
@@ -31,7 +31,7 @@ TaylorDiff = "b36ab563-344f-407b-a36a-4f200bebf99c"
 Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f"
 
 [sources]
-ModelingToolkitBase = {subdir = "../.."}
+ModelingToolkitBase = {path = "../.."}
```
AayushSabharwal, line 34: Isn't `path` relative to the directory where the REPL is open?

SebastianM-C: https://github.com/JuliaLang/Pkg.jl/blob/1c1c173d35be3eb478054fb0229b0b1b6aa040cf/src/Operations.jl#L2205

The paths are relative to the dir of the Project.toml. This should probably be mentioned in the Pkg docs.

PR #4402, opened 2026-03-20
feat: eliminate perfect aliases in favor of higher priority variables

AayushSabharwal:
## Checklist


src/systems/alias_elimination.jl
```diff
@@ -2,107 +2,227 @@ using SymbolicUtils: Rewriters
...
+end
+
+"""
+    $TYPEDSIGNATURES
+
+Remove equations `eqs_to_rm`, variables `vars_to_rm` from `state` and `mm`. `aliases` indicates
+which variables are now aliased to which others, and must also be present in `vars_to_rm`.
+Return the new `mm`.
+"""
+function rm_eqs_vars!(
```
oscardssmith: Do we care about fast lookup into aliases? I would think `Vector{Pair{Int,Int}}` would be the better structure.

src/systems/alias_elimination.jl
```diff
@@ -2,107 +2,227 @@ using SymbolicUtils: Rewriters
...
+    )
+    (; fullvars, structure) = state
+    (; graph, solvable_graph, var_to_diff) = structure
+    eqs = collect(equations(state))
+
+    unique!(sort!(eqs_to_rm))
+    unique!(sort!(vars_to_rm))
+
+    deleteat!(eqs, eqs_to_rm)
+    deleteat!(fullvars, vars_to_rm)
```
oscardssmith: This feels like it will be less efficient than necessary.

oscardssmith: at the least, why won't `eqs_to_rm` and `vars_to_rm` be sorted and uniqued when produced?

src/systems/alias_elimination.jl
```diff
@@ -2,107 +2,227 @@ using SymbolicUtils: Rewriters
...
+    alias_groups = DisjointSet{Int}()
+
+    for (ei, e) in enumerate(mm.nzrows)
+        rcol = mm.row_cols[ei]
+        rval = mm.row_vals[ei]
+        # Consider perfect alias equations where both variables are lowest order derivatives
+        if length(rval) == 2 && iszero(sum(rval)) && diff_to_var[rcol[1]] === nothing && diff_to_var[rcol[2]] === nothing
+            push!(alias_groups, rcol[1])
+            push!(alias_groups, rcol[2])
+            union!(alias_groups, rcol[1], rcol[2])
```
oscardssmith: are `rcol` a `Set` or some other data structure with fast `Union`?

src/systems/alias_elimination.jl
AayushSabharwal: Yes, the `get_new_mm` function needs to check "is this variable aliased to something else"

src/systems/alias_elimination.jl
AayushSabharwal: `eqs_to_rm` when populated by `find_perfect_aliases!` won't be sorted. Both of them will _probably_ be uniqued, it's worth checking.

src/systems/alias_elimination.jl
AayushSabharwal: No, it's a `Vector{Int}`. It comes from https://github.com/JuliaComputing/StateSelection.jl/blob/main/src/math/sparsematrixclil.jl#L33.

oscardssmith: in that case you definitely want to use a `merge` or similar rather than `union!` that takes advantage of the sorted-ness.

AayushSabharwal: This is not a set union. `alias_groups isa DataStructures.DisjointSet{Int}`.

oscardssmith:
do we also want to do constant/parameter aliasing here?

AayushSabharwal:
Yes, I'll move the `trivial_tearing!` pass to before `alias_elimination!` and that'll do it

AayushSabharwal:
This looks good now. @oscardssmith if you approve I'll merge.

src/systems/alias_elimination.jl
```diff
@@ -2,116 +2,232 @@ using SymbolicUtils: Rewriters
...
+sum(cf * fullvars[v] for (cf, v) in zip(coeffs, vars))
+```
+
+`add_buffer` is a temporary buffer to avoid allocations.
+"""
+function build_expr_from_coeffs_vars!(
+        add_buffer::Vector{SymbolicT}, coeffs::Vector, vars::Vector{Int},
+        fullvars::Vector{SymbolicT}
+    )
+    empty!(add_buffer)
```
oscardssmith: probably worth a sizehint?

oscardssmith: or a resize and then you can just index into it rather than pushing

src/systems/alias_elimination.jl
```diff
@@ -2,116 +2,232 @@ using SymbolicUtils: Rewriters
...
+    return SU.add_worker(VartypeT, add_buffer)
+end
+
+"""
+    $TYPEDSIGNATURES
+
+Convenience wrapper for `find_perfect_aliases!`.
+"""
+function eliminate_perfect_aliases!(state::TearingState)
+    StateSelection.complete!(state.structure)
```
oscardssmith, line 33: is this expensive to do twice?

AayushSabharwal: Ah I forgot to remove the second call. It's basically a noop, but no reason to keep it around.

AayushSabharwal:
Addressed both comments

oscardssmith (approved):

PR #4431, opened 2026-04-07
refactor: use `IRStructure` for codegen

AayushSabharwal:
## Checklist


lib/ModelingToolkitBase/src/systems/callbacks.jl
```diff
@@ -1384,7 +1381,12 @@ Base.@nospecializeinfer function compile_explicit_affect(
     dvs_to_update = setdiff(unknowns(aff), getfield.(observed(sys), :lhs))
 
     _affsys = unhack_system(affsys)
-    obseqs = observed(_affsys)
+    aff_ir_info = get_ir_info(affsys)
+    aff_ir = get_irstructure(affsys)
+    obseqs = Equation[]
+    for (i, eq) in enumerate(observed(_affsys))
+        push!(obseqs, eq.lhs ~ aff_ir_info.obs_subber(eq.rhs))
+    end
```
oscardssmith, line 1389: can we not make a nice convenience for this?

AayushSabharwal: What sort of utility function are you suggesting? `get_ir_info` _really_ shouldn't be called inside a loop, so there's not much to make a utility of here.

oscardssmith, line 1389:
```suggestion
    obseqs = Equation[eq.lhs ~ aff_ir_info.obs_subber(eq.rhs) for eq in observed(_affsys)]
```

lib/ModelingToolkitBase/src/systems/codegen.jl
```diff
@@ -1371,35 +1367,22 @@ function build_explicit_observed_function(
...
-        f = GeneratedFunctionWrapper{
-            (
-                p_start + wrap_delays, length(args) - length(rps) + 1 + wrap_delays, is_split(sys),
-            ),
-        }(
-            f, nothing
-        )
-        return f
-    end
+    return return_inplace ? (f, f) : f
```
oscardssmith: is this type stable?

lib/ModelingToolkitBase/test/odesystem.jl
```diff
@@ -107,7 +107,7 @@ u = collect(1:3)
 p = ModelingToolkitBase.MTKParameters(de, [σ, ρ, β] .=> 4.0:6.0)
 f.f(du, u, p, 0.1)
 @test du == [4, 0, -16]
-@test_throws ArgumentError f.f(u, p, 0.1)
+@test_throws ErrorException f.f(u, p, 0.1)
```
oscardssmith: any chance to retain a more detailed error than ErrorException?

lib/ModelingToolkitBase/src/systems/callbacks.jl
AayushSabharwal: I've been burned enough by generators causing inference issues that I tend to avoid them altogether 😅

lib/ModelingToolkitBase/src/systems/codegen.jl
AayushSabharwal: Not even remotely, but changing it is breaking.

lib/ModelingToolkitBase/src/systems/callbacks.jl
oscardssmith: interesting... I'd be curious to see an MWE... something like this should infer just as well, but I might be missing something. It might be one where using an untyped one relies on inference, but for typed ones, it should be identical to the loop.

lib/ModelingToolkitBase/test/odesystem.jl
AayushSabharwal: The error comes from https://github.com/JuliaSymbolics/Symbolics.jl/blob/master/src/codegen_fn.jl#L3. I guess I could test the error string?

lib/ModelingToolkitBase/src/systems/callbacks.jl
AayushSabharwal: Yeah, iirc it usually happens when the generator captures another closure or something type-unstable, and is also often linked with the closure `Box` bug.

lib/ModelingToolkitBase/test/odesystem.jl
oscardssmith: I wouldn't bother testing the error string, but making that throw a more informative error type could be nice.

lib/ModelingToolkitBase/src/systems/codegen.jl
AayushSabharwal: I guess the default for `return_inplace` could be changed to `Val`, and we check `return_inplace isa Val{true} || return_inplace isa Bool && return_inplace` instead?

lib/ModelingToolkitBase/test/odesystem.jl
AayushSabharwal: https://github.com/JuliaSymbolics/Symbolics.jl/pull/1837 addresses this

AayushSabharwal:
`(1, MTK/Initialization)` fails because the caching job failed. Everything else looks fine, so I'll merge.

PR #4443, opened 2026-04-10
feat: [AI] allow passing an operating point to `linearize` using a solution

### LINKED ISSUE #4159: Linearize in point from ODESolution
This feature exists in ControlSystemsMTK but it has proven pretty much impossible to maintain, so I'm asking for it to exist in MTK directly instead. 

It would be very useful if one could pass a solution object as the operating point to `linearize`, and the operating point would be automatically constructed from the solution and the specified time point. It should also be possible to provide a vector of time points in order to linearize along a trajectory.

AayushSabharwal:
Close #4159


baggepinnen:
I'm testing this PR from [ControlSystemsMTK.jl](https://github.com/JuliaControl/ControlSystemsMTK.jl) and hit a dispatch ambiguity in `_build_op_from_solution` when calling with a vector of time points.

The two methods are:

```julia
# Method 1 (line 28): scalar
_build_op_from_solution(op::LinearizationOpPoint)

# Method 2 (line 44): vector
_build_op_from_solution(op::LinearizationOpPoint{S, <:AbstractVector}) where {S}
```

Method 1 implicitly constrains `S <: AbstractODESolution` (from the struct definition), while method 2 uses an unconstrained `where {S}`. Julia sees these as ambiguous for `LinearizationOpPoint{ODESolution{...}, Vector{Float64}}` since neither method is strictly more specific than the other.

baggepinnen:
I've been exercising this branch through `ControlSystemsMTK.trajectory_ss` (linearizing a gain-scheduled closed loop at ~800 points along a trajectory) and hit two performance issues and one correctness issue in the multi-timepoint path. Details below with file/line references against this branch.

### 1. Operating-point values are boxed into `SymbolicT`, forcing per-point recompilation

`_build_op_from_solution` builds the op as a `Dict{SymbolicT, SymbolicT}`:

```julia
# src/linearization.jl
result = Dict{SymbolicT, SymbolicT}()          # line 33 (scalar method)
param_vals = Dict{SymbolicT, SymbolicT}()      # line 52 (vector method)
...
result[sts[i]] = u[i]                          # u[i]::Float64 gets widened to a symbolic constant
result[p] = getp(op.sol, p)(op.sol)            # numeric, also widened
```

Because the value type is `SymbolicT`, every numeric value is converted to a symbolic constant on insertion. The per-point `linearize(sys, lin_fun::LinearizationFunction; t, op, ...)` then takes the symbolic branch for *every* op entry:

```julia
# src/linearization.jl, ~line 911-915
for (k, v) in op
    isequal(v, COMMON_NOTHING) && continue
    if symbolic_type(v) != NotSymbolic() || is_array_of_symbolics(v)
        v = getu(prob, v)(prob)   # taken for every entry, because every value looks symbolic
    end
    ...
```

`getu(prob, v)` builds an observed-function `RuntimeGeneratedFunction` keyed on the value `v`, and since each time point has distinct values, fresh codegen runs at essentially every point. Measured on a 7-state closed loop, 21 op entries, 800 time points:

- One-time `linearization_function`: ~0.09 s (fine)
- The 800-point loop: **273 s, 92% compilation time**, ~340 ms/point
- Replaying the *same* 40 operating points: 268 ms/pt cold → 33.8 ms/pt warm — i.e. the cost tracks the *values*, confirming value-dependent recompilation rather than warmup.
- A trivial 3-entry model does not trigger this and runs at ~0.4 ms/pt.

Suggested fix: keep the operating-point values numeric (don't widen to `SymbolicT`) so the `getu` branch is skipped for plain numbers.

### 2. `__linearize_multiple_op_barrier` reconstructs the problem and setters every iteration

The vector path correctly builds `lin_fun` once, but then:

```julia
# src/linearization.jl, ~line 929-939
function __linearize_multiple_op_barrier(ssys, lin_fun; ops, ts, allow_input_derivatives)
    for (op, t) in zip(ops, ts)
        res, xpt = linearize(ssys, lin_fun; op, t, allow_input_derivatives)  # high-level call
        ...
```

Each high-level `linearize(ssys, lin_fun; op, t)` constructs a fresh `LinearizationProblem(lin_fun, t)` and rebuilds the `setu`/`getu` accessors from scratch. The `LinearizationProblem` docstring (~line 543) already says it "can be symbolically indexed to efficiently update the operating point for multiple linearizations in a loop" with the `.t` field mutated — which is exactly the fast path the loop should use: build the `LinearizationProblem` once, cache the setters for the op keys, update values + mutate `.t`, then `solve`. Combined with (1) this should remove essentially all of the per-point compilation.

### 3. Opened-loop parameters are silently frozen across the trajectory (correctness)

When `loop_openings` is used, `handle_loop_openings` / `Break(ap, true)` turns the opened signal into a *parameter* of the linearized system that does not exist in the solution. `_build_op_from_solution` only extracts the solution system's states and parameters, so these opened-loop parameters never get a per-time-point value — they stay frozen at whatever value the function was built with (`ops[1]`, i.e. `t[1]`) for the entire trajectory.

This produces silently wrong results. Concretely: linearizing a gain-scheduled controller `r -> u` with `loop_openings=[y, v]` vs `loop_openings=[u]` gives identical systems at a single time point, but over 800 points they diverge by ~19% in frequency response, because the opened scheduling signal is pinned to its `t[1]` value rather than tracking the trajectory.

Rather than silently using a stale value, I think `linearize` should **error when opened-loop parameters are not provided in the op**, so the user is forced to specify their values explicitly (e.g. set the opened signals to 0). That keeps the semantics explicit and avoids the silent-wrong-answer failure mode.

baggepinnen:
what's the status of this PR? Mike has started to think about ops in dyad so I'd like the low level stuff to be available and working to start trying things out. Last time I tried, there were some performance problems that I hoe were addressed by https://github.com/SciML/ModelingToolkit.jl/pull/4679, anything else needed?

AayushSabharwal:
I'll prioritize it. It needs CI to pass, there are some failures in linearization.

PR #4425, opened 2026-04-02
Add nominal value metadata

SebastianM-C:
## Checklist

- [x] Appropriate tests were added
- [x] Any code changes were done in a way that does not break public API
- [x] All documentation related to code changes were updated
- [x] The new code follows the
  [contributor guidelines](https://github.com/SciML/.github/blob/master/CONTRIBUTING.md), in particular the [SciML Style Guide](https://github.com/SciML/SciMLStyle) and
  [COLPRAC](https://github.com/SciML/COLPRAC).
- [x] Any new documentation only uses public API
  
## Additional context

Adds nominal value metadata that can be used to scale the dynamics in the context of dynamic optimization or for better tolerance selections for integrators (or several other applications where one benefits from knowing this).

PR #4445, opened 2026-04-10
add new HashedRandom guess option and make it the default

oscardssmith:
`HashedRandom` uses `hash(var)/2^64` as an initial value which has the significant advantage over regular random of being deterministic. As such, I think this is a generally reasonable initialization (although it might make sense to also add a multiplication by `getnominal` for both `Random` and `HashedRandom` now that https://github.com/SciML/ModelingToolkit.jl/pull/4425 is merged.

### DIFF
--- a/lib/ModelingToolkitBase/src/systems/problem_utils.jl
+++ b/lib/ModelingToolkitBase/src/systems/problem_utils.jl
@@ -293,17 +293,20 @@ numerical problem from a `System`.
 - `MissingGuessValue.Constant(val::Number)`: Missing guesses are set to the given value
   `val`.
 - `MissingGuessValue.Random(rng::AbstractRNG)`: Missing guesses are set to `rand(rng)`.
+- `MissingGuessValue.HashedRandom`: Missing guesses are set to a
+  deterministically determined random-like value based on the hash of the variable name
 - `MissingGuessValue.Error()`: Missing guess values cause an error.
 """
 Moshi.Data.@data MissingGuessValue begin
     Constant(Number)
     Random(AbstractRNG)
+    HashedRandom
     Error
 end
 
 # To be overloaded downstream by MTK
 default_missing_guess_value() = default_missing_guess_value(nothing)
-default_missing_guess_value(_) = MissingGuessValue.Error()
+default_missing_guess_value(_) = MissingGuessValue.HashedRandom()
 
 """
     $(TYPEDSIGNATURES)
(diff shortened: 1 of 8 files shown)


AayushSabharwal (changes requested):
The enum is also used in the `SCCNonlinearProblem` constructor, so it needs to be handled there as well.

hersle:
Would this cause silent or hard-to-find bad behavior in the event of missing guesses, where the random algorithm selects bad values?

It's a nice idea though, if you make it default at least it can be disabled with 1 kwarg to `ODEProblem`, which is great. In my particular application I would prefer errors over random guesses, since bad initial guesses would crash the initialization solver. But I guess that is different in other applications.

oscardssmith:
> since bad initial guesses would crash the initialization solver. But I guess that is different in other applications.

My argument is that if bad guesses crash initialization then you're in just as bad a spot as if you had gotten an error before initialization started (but if the guess doesn't matter, then this version is strictly better). We do probably want to mention in the error message if initialization fails that you should consider adding guesses if you know roughly how you want to initialize).

lib/ModelingToolkitBase/src/systems/problem_utils.jl
```diff
@@ -370,6 +373,15 @@ function varmap_to_vars(
                     end
                 end
             end
+            MissingGuessValue.HashedRandom() => begin
+                for var in missing_vars
+                    if Symbolics.isarraysymbolic(var)
+                        varmap[var] = [hash(var,i) for i in eachindex(var)]./0x1p64
```
AayushSabharwal:
```suggestion
                        varmap[var] = [hash(var,hash(i)) for i in SU.stable_eachindex(var)]./0x1p64
```

src/problems/sccnonlinearproblem.jl
```diff
@@ -643,6 +643,15 @@ function SciMLBase.SCCNonlinearProblem{iip}(
                             )
                         end
                     end
+                    MissingGuessValue.HashedRandom() => begin
+                        newval = [hash(var,i) for i in eachindex(symbolic_idxs)]./0x1p64
```
AayushSabharwal:
```suggestion
                        newval = [hash(var,hash(i)) for i in SU.stable_eachindex(symbolic_idxs)]./0x1p64
```

AayushSabharwal (changes requested):
It would also be nice if the magic constant `0x1p64` was a global constant and/or the hashing logic its own function, so that we can be sure the implementation in SCCNonlinearProblem is the same as the one in `problem_utils.jl`.

oscardssmith:
> It would also be nice if the magic constant 0x1p64 was a global constant and/or the hashing logic its own function

This constant is just 2^64. Would you like it better if I wrote it as `exp2(64)`?

AayushSabharwal:
> This constant is just 2^64. Would you like it better if I wrote it as exp2(64)?

Ah, I didn't quite process that. Never mind then.

oscardssmith:
Julia's binary floating point literals are quite nice.

AayushSabharwal:
The `(1, ModelingToolkit/InterfaceII)` failure is real, rest are codecov

oscardssmith:
test fixed.

AayushSabharwal:
It's still failing 😅

oscardssmith:
It found a real bug! Linearize was failing to pass default_guess down.

AayushSabharwal (approved):

PR #4447, opened 2026-04-13
fix: remember perfect alias equations when building alias groups

AayushSabharwal:
This fixes a bug best described by the failing test in MTKTearing:

```julia
    @variables x(t) y(t)
    eqs = [
        0 ~ x - y
        0 ~ 2y - x
    ]

    @named sys = System(eqs, t)
    sys = mtkcompile(sys)
```

MTK would use `x => y` as the substitution, but since `0 ~ 2y - x` satisfied the (faulty) criteria for identifying `alias_eq`, it considered `eqs[2]` above as the equation to remove and `0 ~ x - y` as the one to retain. Now, alias equations are identified when populating `alias_groups`. This removes the need for any checking in the subsequent loop when the aliases are materialized.

### DIFF
--- a/.github/workflows/Downstream.yml
+++ b/.github/workflows/Downstream.yml
@@ -16,7 +16,7 @@ concurrency:
 
 jobs:
   test:
-    name: ${{ matrix.package.repo }}/${{ matrix.package.group }}/${{ matrix.julia-version }}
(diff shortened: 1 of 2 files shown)


AayushSabharwal:
I've root caused the other failures, so this can merge to fix a correctness bug.

PR #4451, opened 2026-04-14
fix: enable unscalarized array observed without individual observed elements

AayushSabharwal:
This allows user-defined observed equations that define entire arrays, without the individual elements of the array being other observed equations. This is in contrast to the array hack, which adds `x ~ change_origin(...)` when `x` is an array for which some elements are observed variables.


oscardssmith (approved):

AayushSabharwal:
CI is broken, this cannot merge yet.

AayushSabharwal:
Finally

PR #4454, opened 2026-04-15
feat: add pass which removes equations incident on a single variable

AayushSabharwal:
This avoids cases where pantelides ends up choosing a bad matching and unnecessarily differentiates the kind of equations removed by this pass.

### DIFF
--- a/src/systems/alias_elimination.jl
+++ b/src/systems/alias_elimination.jl
@@ -148,6 +148,97 @@ function find_perfect_aliases!(
     return aliases
 end
 
+"""
+    $TYPEDSIGNATURES
+
+Analytically remove any equations in `state` which are only incident on a single variable.
+"""
+function remove_constant_variables!(state::TearingState; allow_parameter::Bool = true, kwargs...)
+    StateSelection.complete!(state.structure)
+    (; additional_observed, original_eqs, fullvars, structure, sys) = state
+    (; graph, var_to_diff) = structure
+
+    diff_to_var = invview(var_to_diff)
+    eqs = equations(state)
+    eqs_to_rm = Int[]
+    vars_to_rm = Int[]
+    vars_to_rm_set = BitSet()
+    fullvars_set = Set{SymbolicT}(fullvars)
+    param_der_subber = SU.Substituter{false}(state.param_derivative_map)
+
+    # Preallocated buffer
+    snbors = Int[]
+    # Eliminating a variable can cause other equations to be dependent on only one
+    # un-eliminated variable. This shouldn't realistically run more than 2 times,
+    # but cap the iteration count at 4 nonetheless.
+    for _ in 1:4
+        removed_eq = false
+        for ieq in 𝑠vertices(graph)
+            # Check if this equation is incident on exactly 1 un-eliminated variable.
+            # We could just run `remove_constant_variables!` in a loop, but this
+            # is faster since we don't repeatedly rebuild structural information.
+            empty!(snbors)
+            append!(snbors, 𝑠neighbors(graph, ieq))
+            setdiff!(snbors, vars_to_rm_set)
+            length(snbors) == 1 || continue
+            ivar = first(snbors)
+            # Only eliminate variables which are lowest order derivatives. Eliminating higher
+            # derivatives requires integrating to also eliminate the lower order variables,
+            # which is not something we can easily do at the moment.
+            diff_to_var[ivar] === nothing || continue
+            eq = eqs[ieq]
+            var = fullvars[ivar]
+            # This replicates a piece of `find_eq_solvables!` but calling that function directly:
+            # 1. requires creating `solvable_graph`, which we would then have to immediately
+            #    discard since we're not populating solvability for every equation.
+            # 2. still requires us to call `LinearExpander` here. The function does return the
+            #    same value as `b` below, but `eq.rhs - b` is not necessarily `a * var`, since
+            #    e.g. if `eq.rhs` is `(2 + (3x + 4) * t)`, then `b` will be `4t + 2`. However,
+            #    `eq.rhs - b` is `-4t + t*(4 + 3x)`.
+            # So doing it ourselves is just faster.
+            lex = Symbolics.LinearExpander(var; strict = true)
+            a, b, islin = lex(eq.rhs)
+            islin || continue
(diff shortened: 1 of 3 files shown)


src/systems/alias_elimination.jl
```diff
@@ -148,6 +148,91 @@ function find_perfect_aliases!(
...
+            push!(vars_to_rm, ivar)
+            push!(vars_to_rm_set, ivar)
+            # `a` typically is faster to negate, since it usually is a constant or small expression
+            rhs = b / -a
+            push!(additional_observed, var ~ rhs)
+            subs[var] = rhs
+            union!(eqs_to_substitute, 𝑑neighbors(graph, ivar))
+
+            v = var_to_diff[ivar]
+            while v isa Int
```
oscardssmith: would this be more readable if it was `while !isnothing(v)`?

src/systems/alias_elimination.jl
```diff
@@ -148,6 +148,91 @@ function find_perfect_aliases!(
...
+        for ieq in 𝑠vertices(graph)
+            empty!(snbors)
+            append!(snbors, 𝑠neighbors(graph, ieq))
+            setdiff!(snbors, vars_to_rm_set)
+            length(snbors) == 1 || continue
+            ivar = first(snbors)
+            eq = eqs[ieq]
+            var = fullvars[ivar]
+            lex = Symbolics.LinearExpander(var; strict = true)
+            a, b, islin = lex(eq.rhs)
```
oscardssmith, line 200: how do we know that the `eq.rhs` doesn't include `var`?

src/systems/alias_elimination.jl
```diff
@@ -148,6 +148,91 @@ function find_perfect_aliases!(
...
+    eqs_to_substitute = Set{Int}()
+    param_der_subber = SU.Substituter{false}(state.param_derivative_map)
+
+    snbors = Int[]
+    for _ in 1:4
+        removed_eq = false
+        for ieq in 𝑠vertices(graph)
+            empty!(snbors)
+            append!(snbors, 𝑠neighbors(graph, ieq))
+            setdiff!(snbors, vars_to_rm_set)
```
oscardssmith, line 182: is this needed to prevent needing to recompute the neighbors after each variable removal? IMO it could use a comment if so.

src/systems/alias_elimination.jl
AayushSabharwal: I was hoping `isa Int` would help type inference better. And iirc since `v::Union{Int, Nothing}`, codegen will union-split the call to `isnothing` whereas `isa` is a builtin, so there's only ever one thing to call.

src/systems/alias_elimination.jl
AayushSabharwal: It does, that's the point. All equations at this point are of the form `0 ~ something`, and `lex(eq.rhs)` returns `a, b, islin` where `isequal(eq.rhs, a * var + b)` iff `islin`.

src/systems/alias_elimination.jl
AayushSabharwal: Will add a comment (and just spam some more in general)

src/systems/alias_elimination.jl
AayushSabharwal: Also this is just personal preference but I like reading `var === nothing` over `isnothing(var)`.

oscardssmith: I'd also be fine with `!(isa Nothing)`. `isa Int` just feels kind of fragile (e.g. it would break if we changed incidence to Int32)

src/systems/alias_elimination.jl
oscardssmith: how do we know that the lhs is 0?

src/systems/alias_elimination.jl
AayushSabharwal: Ah, I see your point.

src/systems/alias_elimination.jl
AayushSabharwal: It's an invariant of how we construct `state::TearingState`

AayushSabharwal:
I also just realized that this pass doesn't need to substitute, since if the eliminated variable is removed from `fullvars` we'll end up treating it like a parameter. The perfect aliases pass needs to substitute because we change incidence for the eliminated variable to its alias, so the alias needs to be symbolically present in the equation.

oscardssmith:
That's a very nice simplification!

src/systems/alias_elimination.jl
oscardssmith: makes sense. Only other comment here is that it might be nice to test for general solvability rather than just linear. (e.g. `0~x^3-2` is solvable as `x=cbrt(2)`, but that might be harder. This is almost certainly fine for now though.

AayushSabharwal: The symbolic solver exists and is capable of handling such cases, but it is also abysmally slow.

AayushSabharwal:
Addressed the comments

AayushSabharwal:
A multibody model is now simplifying to 25 variables instead of 17 because of the latest changes on this PR.

oscardssmith:
any idea what the difference is?

AayushSabharwal:
A bunch of algebraic equations that should have been torn but weren't.

oscardssmith:
We also need to make sure to deal with https://github.com/SciML/ModelingToolkit.jl/issues/4458 for this (and add a test)

AayushSabharwal:
Okay, we do need to substitute. If we eliminate a variable `v`, and `var_to_diff[v] !== nothing`, that means the symbolic expression `D(fullvars[v])` is present in some equation. Observed equations can't have `D(...)` on the LHS, so we need to substitute to handle this case. Substitution time complexity doesn't scale with the number of entries in the `Dict`, so if we need to substitute derivatives we might as well substitute the entire thing.

AayushSabharwal:
The problematic multibody system is handled by changing the tearing algorithm to the old `ModiaTearing`. It's possible I implemented the new tearing algorithm wrong? Will double check, but I don't think I did.

AayushSabharwal:
Weirdly either `CarpanzanoTearing` without this pass, or `ModiaTearing` with this pass works.

AayushSabharwal:
I'm certain though, there was a point where this pass + `CarpanzanoTearing` was perfectly fine.

PR #4497, opened 2026-04-28
fix: do not eliminate positive state priority variables in alias elimination

AayushSabharwal:
This fixes regressions in multibody models.


oscardssmith:
what's special about positive state priority?

AayushSabharwal:
Positive state priority means the user wants this variable to be part of the selected states, so we shouldn't eliminate it as observed.

oscardssmith (approved):

PR #4498, opened 2026-04-29
test: make discrete event reinitialization test independent of state selection

AayushSabharwal:
This will unblock https://github.com/JuliaComputing/StateSelection.jl/pull/72


AayushSabharwal:
Will merge as soon as InterfaceI passes, since this only touches one test

PR #4503, opened 2026-04-30
test: make linearization DD test more robust

AayushSabharwal:


AayushSabharwal:
The long-running jobs are likely going to time out, as they tend to do stochastically. This is only a test change, and the relevant test jobs now pass, so I'll merge.

PR #4518, opened 2026-05-11
refactor: use `full_equations` in `SCCNonlinearProblem`

AayushSabharwal:
This ends up being fast due to `IRStructure`, more type-stable due to the `subexpressions_not_involving_vars!` rewrite, and avoids all of the mess of finding observed equations required for each SCC.


AayushSabharwal:
Tests pass. Will remove the accidental `display` and merge.

PR #4525, opened 2026-05-12
refactor: use `IRInfo` and `IRStructure` for `full_equations`

AayushSabharwal:
This is a considerable improvement for large models.

### DIFF
--- a/lib/ModelingToolkitBase/src/systems/abstractsystem.jl
+++ b/lib/ModelingToolkitBase/src/systems/abstractsystem.jl
@@ -1998,6 +1998,20 @@ equations during `mtkcompile`. These equations matches generated numerical code.
 See also [`equations`](@ref) and [`ModelingToolkitBase.get_eqs`](@ref).
 """
 function full_equations(sys::AbstractSystem; simplify = false)
+    subsys = get_systems(sys)
+    # Fast path using `IRInfo`
+    if isempty(subsys)
+        new_eqs = Equation[]
+        eqs = equations(sys)
+        sizehint!(new_eqs, length(eqs))
+        info = get_ir_info(sys)
+        ir = get_irstructure(sys)
+        @assert length(info.eqs_idxs) == length(eqs)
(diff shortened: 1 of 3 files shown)


lib/ModelingToolkitBase/src/systems/abstractsystem.jl
```diff
@@ -1998,6 +1998,20 @@ equations during `mtkcompile`. These equations matches generated numerical code.
 See also [`equations`](@ref) and [`ModelingToolkitBase.get_eqs`](@ref).
 """
 function full_equations(sys::AbstractSystem; simplify = false)
+    subsys = get_systems(sys)
+    # Fast path using `IRInfo`
+    if isempty(subsys)
```
oscardssmith, line 2003: why can we only do this if we don't have subsytems? Is that fixable?

AayushSabharwal: I'll add a comment. `get_ir_info` will build the `IRInfo` if a cached one does not exist. Doing so requires the `ParameterBindingsGraph`, which is only valid for flattened systems. I'll add a comment.

oscardssmith: how hard would it be to remove that restriction? The bifurcation here seems sub-optimal.

AayushSabharwal: Probably not very hard, but it doesn't matter much. We hit the slow path for hierarchical systems, which we can't simulate anyway. The only case where `full_equations` on a hierarchical system matters is if the user is directly building a system that they will `complete` and never `mtkcompile`. It's safe to assume those systems are few and far between, and not large enough for the speedup here to matter. Flattening the system will also invalidate the cached `IRInfo`, so we'll redo this anyway.

oscardssmith: fair enough. I'm just coming at this from a code simplicity perspective, but I'm fine with this as is.

PR #4536, opened 2026-05-15
refactor: mark `u` argument in generated functions where appropriate

AayushSabharwal:
This allows specialized codegen to refer to the `u` vector.

### DIFF
--- a/lib/ModelingToolkitBase/src/systems/codegen_utils.jl
+++ b/lib/ModelingToolkitBase/src/systems/codegen_utils.jl
@@ -256,6 +256,10 @@ generated functions, and `args` are the arguments.
   `MTKParameters` object are present. These are collapsed into a single argument and
   destructured inside the function. `p_start` must also be provided for non-split systems
   since it is used by `wrap_delays`.
+- `u_arg`: The index in `args` of the argument corresponding to `unknowns(sys)` (the `u`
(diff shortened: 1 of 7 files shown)


DhairyaLGandhi:
So codegen should be able to refer to this for the vector of unknowns

lib/ModelingToolkitBase/src/systems/codegen_utils.jl
```diff
@@ -256,6 +256,10 @@ generated functions, and `args` are the arguments.
   `MTKParameters` object are present. These are collapsed into a single argument and
   destructured inside the function. `p_start` must also be provided for non-split systems
   since it is used by `wrap_delays`.
+- `u_arg`: The index in `args` of the argument corresponding to `unknowns(sys)` (the `u`
+  vector). If `-1` (the default), the u vector is not treated specially. Otherwise, the
+  argument must be a `Vector` and is wrapped in a `DestructuredArgs` with the common
```
oscardssmith, line 261:
```suggestion
  argument must be an `AbstractVector` and is wrapped in a `DestructuredArgs` with the common
```

Can we generalize? Otherwise GPUs and stuff may be sad.

AayushSabharwal: This is referring to a symbolic list of arguments, not the value passed to the generated function.

PR #4537, opened 2026-05-15
refactor: re-run `eliminate_perfect_aliases!` after `alias_elimination!` 

AayushSabharwal:
We run `eliminate_perfect_aliases!` before `alias_elimination!` to (significantly) reduce the number of equations the latter pass has to process. However, `alias_elimination!` identifies additional `x ~ y` equations which can easily be removed by `eliminate_perfect_aliases!`. This has resulted in better simplification on some models.

### DIFF
--- a/src/systems/alias_elimination.jl
+++ b/src/systems/alias_elimination.jl
@@ -82,7 +82,6 @@ function find_perfect_aliases!(
     (; sys, fullvars, structure) = state
     (; graph, solvable_graph, var_to_diff, state_priorities) = structure
 
-    @assert solvable_graph === nothing
     diff_to_var = invview(var_to_diff)
     aliases = Dict{Int, Int}()
     subs = Dict{SymbolicT, SymbolicT}()
--- a/src/systems/alias_elimination.jl
+++ b/src/systems/alias_elimination.jl
@@ -37,7 +37,7 @@ function eliminate_perfect_aliases!(state::TearingState)
     old_to_new_eq, old_to_new_var = StateSelection.rm_eqs_vars!(
         state, eqs_to_rm, vars_to_rm; eqs_sorted_and_uniqued = true
     )
-    return nothing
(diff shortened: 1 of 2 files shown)


oscardssmith:
Why doesn't `alias_elimination!` eleminate these perfect aliases?

AayushSabharwal:
`alias_elimination!`, contrary to its name, doesn't actually "eliminate" anything. It just runs bariess on the equations.

oscardssmith:
but it does have a function that runs after it to actually eliminate the aliases, right?

Also, wouldn't all of these be aliases found by adding support for the fast path detecting `x~-y`?

AayushSabharwal:
> but it does have a function that runs after it to actually eliminate the aliases, right?

Nope. This PR adds that.

> Also, wouldn't all of these be aliases found by adding support for the fast path detecting x~-y?

I'm not sure. It's technically possible they wouldn't, but I don't know how much that would happen in practice. Once I get https://github.com/JuliaComputing/StateSelection.jl/pull/81 working, I'll check if keeping this change around is worth it, since the main win from this is smaller linsolves.

oscardssmith:
IMO the order we should go is

1. recognize `x ~ -y` in trivial alias detection
2. (if still useful) https://github.com/JuliaComputing/StateSelection.jl/pull/81
3. (if still useful) this (and if not, delete `alias_elimination!` if possible).

I would really like the passes we do to feel grounded and obviously correct/necessary.

AayushSabharwal:
Yep, agreed. However, we can't delete `alias_elimination!` since it removes duplicate equations, which we get frequently because we allow duplicate `connect`s (and iirc have presented MTK's ability to do so at a conference somewhere)

oscardssmith:
I think there probably is a much simpler way to remove duplicate equations.

AayushSabharwal:
Unnecessary now

PR #4540, opened 2026-05-15
feat(alias_elimination): detect `x ~ -y` perfect aliases

oscardssmith:
Extend `find_perfect_aliases!` to recognize the negated alias pattern `c1*v1 + c2*v2 ~ 0` with `c1 == c2` (i.e. `x ~ -y`) in addition to the existing `c1 == -c2` case (`x ~ y`). Each candidate alias edge carries an `Int8` sign, and a BFS from each group's chosen target propagates signs to every member -- so transitive chains including double negation (`x ~ -y` and `y ~ -z` ⇒ `x ~ z`) are resolved correctly.

When a group's signed edges are inconsistent (e.g. `x ~ y` and `x ~ -y` both appear), every non-sticky variable in that group is substituted with `0` and a `v ~ 0` equation is added to `additional_observed`; irreducibles remain unknowns and the surviving alias equations carry the resulting `irr ~ 0` constraint into the system. Substitution rhs and derivative-cascade substitutions both honor the propagated sign.

Tests cover: basic `y ~ -x`, mixed chain `x ~ y, y ~ -z`, double-negation transitivity, negated alias with irreducible target, and the sign-conflict force-to-zero case.

Claude wrote most of this, but it seems pretty reasonably implemented.


test/reduction.jl
```diff
@@ -384,3 +384,68 @@ end
         isequal(eq, 0 ~ 1 - dot(x, Symbolics.SConst([y[1], y[1]])))
     end
 end
+
+@testset "Perfect aliases detect negated form `x ~ -y`" begin
```
baggepinnen, line 388: wouldn't all of these tests pass on master as well?

oscardssmith: I believe so, but it is worth adding specific tests that this behavior is correct that are targeted at the form of the pass to ensure that we haven't introduced any bugs. (This also serves as a regression test in case we in the future remove the full alias_analysis pass which I kind of want to do because it is fairly expensive and once we have this PR and redundant equation removal we may no longer need it)

baggepinnen: I'm all for having many tests, but if the tests would pass on master as well they do not really exercise the change in this PR?

oscardssmith: fair enough. Will check to make sure existing tests cover this and then delete (or augment existing tests if necessary)

AayushSabharwal: I'll review this properly tomorrow some time, but imo for these sorts of tests it's best to just build `TearingState(sys)` and call the pass directly.

oscardssmith: sounds good. Will fix. Also have an additional change incoming for fixing redundant equations like `x ~ Initial(x)` making their way into the initialization system.

AayushSabharwal:
> every non-sticky variable in that group is substituted with 0 and a v ~ 0 equation is added to additional_observed;

I think for the 0 case it's worth including sticky non-irreducible variables too (i.e. ones with positive state priority). The variable cannot possibly be anything else, so it serves no purpose participating in state selection. Substituting zeros into other equations may also simplify them drastically.

Knowing something is zero is _very_ useful information.

oscardssmith:
Good point. Irreducible can just leave the variable with the equation x~0.

oscardssmith:
> I think for the 0 case it's worth including sticky non-irreducible variables too (i.e. ones with positive state priority). The variable cannot possibly be anything else, so it serves no purpose participating in state selection. Substituting zeros into other equations may also simplify them drastically.

I've actually gone a step further and have included irreducibles as well in this. Just because we are irreducible doesn't mean we can't substitute away all users. We now make irreducibles remain as `irr~0` (but with no other uses of `irr`, and other sticky vars are removed entirely.

src/systems/alias_elimination.jl
```diff
@@ -90,12 +148,20 @@ function find_perfect_aliases!(
...
+    # Weighted union-find tracking each variable's sign relative to its
+    # component's current root. `parent[v]` is always the root.
+    # `parity[v] ∈ ±1` is the sign of `v` relative to that root, or `0` if the
+    # component contains a sign contradiction (forcing every member to 0).
+    # `members` maps root → member list (used for smaller-to-larger merges).
+    parent = Dict{Int, Int}()
+    parity = Dict{Int, Int8}()
+    members = Dict{Int, Vector{Int}}()
+    # Candidate alias equations `(ieq, v1_idx, v2_idx, edge_sign)`. 
+    # `edge_sign == +1` encodes `v1 ~ edge_sign*v2`.
```
AayushSabharwal:
```suggestion
    # `edge_sign == ±1` encodes `v1 ~ edge_sign*v2`.
```

src/systems/alias_elimination.jl
```diff
@@ -196,6 +293,37 @@ function find_perfect_aliases!(
...
+    # Consistent groups: queue for removal iff both endpoints collapse. An eq
+    # with a non-target irreducible endpoint must stay so the remaining
+    # irreducible is still pinned to the target after substitution.
+    for (ieq, v1, v2, _) in candidate_eqs
+        if parity[v1] == 0 # if conflict group
+            irrs = irrs_by_root[parent[v1]]
+            if isempty(irrs)
+                push!(eqs_to_rm, ieq)
+            else
+                v_pin = pop!(irrs) # irriducable var to pin to 0
```
AayushSabharwal:
```suggestion
                v_pin = pop!(irrs) # irreducible var to pin to 0
```

AayushSabharwal (approved):

AayushSabharwal (changes requested):
I just realized, if we eliminate a variable as zero and substitute the zero into other equations, we have to update the incidence of those equations.

oscardssmith:
> I just realized, if we eliminate a variable as zero and substitute the zero into other equations, we have to update the incidence of those equations.

Is this just for cases where we have `x*y` and sub `x->0`? Are there any other cases where the equation incidence changes (other than the removal of the variable (which we already do)?

Edit: this is wrong. there's lots of nonlinear cases as well.

oscardssmith:
Should be fixed.

src/systems/alias_elimination.jl
AayushSabharwal: `false` here won't constant-fold the zeros introduced by conflict groups. E.g. `sin(x)` when substituted with `x => 0` will remain as the expression `sin(0)`.
```suggestion
    subber = SU.IRSubstituter{true}(ir, subs)
```

AayushSabharwal, line 338:
```suggestion
    new_vars = SU.IRStructureSearchBuffer(ir, Set{SymbolicT}())
    for e in eqs_to_substitute
```

Setup for the next comment

src/systems/alias_elimination.jl
```diff
@@ -211,6 +339,15 @@ function find_perfect_aliases!(
         # substitutes `x` to `[x[1], x[2]]`. The second substitutes `x[1]` and `x[2]`.
         eqs[e] = subber(subber(eqs[e]))
         original_eqs[e] = subber(subber(original_eqs[e]))
+        # Substitution can drop vars beyond the one substituted: zero
+        # substitution annihilates multiplicative cofactors (`v*w` with `v→0`
+        # also removes `w`); alias substitution can cancel the target
+        # (`v - target` with `v→target` leaves neither). Substitution never
+        # introduces new vars, so the new incidence is a subset of the old —
+        # prune the row to entries actually present in the simplified RHS.
+        new_vars = Symbolics.get_variables(eqs[e], fullvars)
```
AayushSabharwal:
```suggestion
        empty!(new_vars)
        SU.search_variables!(new_vars, eqs[e])
```
The `search_variables!` implementation using `IRStructureSearchBuffer` (leveraging `IRStructure`) is stupid fast compared to the normal one for large expressions.

Requires https://github.com/JuliaSymbolics/SymbolicUtils.jl/pull/926

AayushSabharwal (changes requested):

oscardssmith:
@AayushSabharwal can you have another look at this?

AayushSabharwal:
`(1, ModelingToolkit/InterfaceI)` fails. Codegen isn't happy with _something_ here.

AayushSabharwal:
Congratulations! You uncovered a bug in `IRStructure`! 😆 The fix is in https://github.com/JuliaSymbolics/SymbolicUtils.jl/pull/932

oscardssmith:
rebased.

AayushSabharwal (approved):

AayushSabharwal:
🎉

PR #4545, opened 2026-05-19
kc/typos whitespace

KristofferC:
My editor keeps inserting a bunch of whitespace which is annoying. Also, there is a typo ci check that keeps being red.


test_compiler_options_fallback.jl
AayushSabharwal, line 1: Is this file accidentally committed?

KristofferC: oops, yes, sorry, too many things going on at once

Issue #4181, opened 2026-01-16
Add Enzyme rule for SetInitialUnknowns to support remake autodiff


ChrisRackauckas-Claude:
## Summary

PR #4175 added a ChainRules `rrule` for `SetInitialUnknowns` which fixed Zygote autodiff through `remake` (SciMLBase.jl#1208). However, Enzyme uses its own AD rules and would need a corresponding Enzyme-specific rule.

## Context

The `SetInitialUnknowns` functor is used during `remake` with MTK problems. For Enzyme to differentiate through `remake` calls, it needs its own rule defined.

## Related

- MTK PR #4175 - Added ChainRules rrule for Zygote
- SciMLBase.jl#1208 - Original issue for Zygote
- SciMLBase.jl#1210 - Re-enabled Zygote tests

## Desired Behavior

Add an Enzyme rule (likely via EnzymeRules in a package extension) so that code like:

```julia
function loss(u0)
    _prob = remake(prob, u0 = u0)
    return sum(solve(_prob, Tsit5()))
end
DifferentiationInterface.gradient(loss, AutoEnzyme(), u0)
```

works correctly.

PR #1323, opened 2021-11-03
Bump version

0x0f0f0f:
After merging #1318

PR #4551, opened 2026-05-21
supports_initialization: use get_jumps to keep Enzyme readonly

### LINKED ISSUE #4550: Enzyme.gradient through remake fails with EnzymeMutabilityException via supports_initialization → jumps(sys)
## Summary

`Enzyme.gradient` through a closure that calls `remake` on an MTK-generated problem (no `solve` needed) fails on Julia 1.12 with `EnzymeMutabilityException` because MTK's `supports_initialization(sys)` calls `jumps(sys)`, which contains a `copy(js)` / `append!(js, namespace_jumps(subsys))` path that Enzyme cannot statically prove is read-only for the captured `sys`.

This is the immediate blocker for `Enzyme.gradient(Reverse, loss, tunables)` where `loss = p -> sum(remake(prob; p = repack(p)).u0)` — surfaces in SciMLSensitivity.jl's `test/parameter_initialization.jl`, `test/mtk.jl`, and `test/desauty_dae_mwe.jl` (all currently `@test_broken`).

Related to #4181 (Enzyme rule for `SetInitialUnknowns`), but a strictly earlier failure — Enzyme never reaches `SetInitialUnknowns` because `_late_binding_update_u0_p_impl` aborts first on the `supports_initialization` readonly check.

## Reproducer (no solve, no SciMLSensitivity, no NonlinearSolve)

```julia
using ModelingToolkit
using SciMLBase: remake, ODEProblem
import SciMLStructures as SS
using SymbolicIndexingInterface: parameter_values
using Enzyme

@parameters a b c
@variables x(t) y(t) z(t)

eqs = [
    D(x) ~ a * x + y + z,
    0 ~ y^3 + y - b * x,
    0 ~ z^3 + z - c * y,
]

@mtkbuild sys = ODESystem(eqs, t)

prob = ODEProblem(
    sys,
    [x => 1.0],
    (0.0, 0.1),
    [a => -0.5, b => 2.0, c => 1.5],
    guesses = [y => 1.0, z => 0.5];
    use_scc = false,
)

iprob = prob.f.initialization_data.initializeprob
itunables, irepack, _ = SS.canonicalize(SS.Tunable(), parameter_values(iprob))

remake_only = let iprob = iprob, irepack = irepack
    p -> sum(remake(iprob, p = irepack(p)).u0)
end

# Throws EnzymeMutabilityException
Enzyme.gradient(Enzyme.Reverse, remake_only, itunables)
```

## Error

```
EnzymeMutabilityException: Function argument passed to autodiff cannot be proven readonly.
...
The potentially writing call is   %8 = call fastcc nonnull ptr addrspace(10) @julia_jumps_42457(ptr addrspace(10) readonly %5), ...
!enzymejl_byref_MUT_REF !0, !enzymejl_source_type_System !0

Stacktrace:
  [1] supports_initialization
    @ src/systems/system.jl:1506 [inlined]
  [2] _late_binding_update_u0_p_impl
    @ src/systems/nonlinear/initializesystem.jl:729 [inlined]
  [3] late_binding_update_u0_p
    @ src/systems/nonlinear/initializesystem.jl:716 [inlined]
  [4] late_binding_update_u0_p
    @ SciMLBase/src/remake.jl:1527 [inlined]
  ...
  [8] remake
    @ SciMLBase/src/remake.jl:920 [inlined]
  [9] remake
    @ SciMLBase/src/remake.jl:907 [inlined]
```

## Root cause

`supports_initialization` in `src/systems/system.jl:1505-1508`:

```julia
function supports_initialization(sys::System)
    return isempty(get_systems(sys)) && isempty(jumps(sys)) &&
        isempty(get_costs(sys)) && isempty(get_constraints(sys))
end
```

It calls `jumps(sys)`, which (in `src/systems/abstractsystem.jl:2054`) is:

```julia
function jumps(sys::AbstractSystem)
    js = get_jumps(sys)
    systems = get_systems(sys)
    isempty(systems) && return js
    js = copy(js)
    for subsys in systems
        append!(js, namespace_jumps(subsys))
    end
    return js
end
```

Enzyme's static analysis cannot prove the `copy`/`append!` branch is never taken at runtime for any `sys`, so the entire `sys` argument is conservatively marked as not provably readonly. This causes the captured-closure mutability check to fire even though the actual call (with no subsystems) doesn't write anything.

## Proposed fix

Replace the call to `jumps(sys)` with the raw field accessor `get_jumps(sys)` — both are equivalent when there are no subsystems (which is the case after `@mtkbuild`/`structural_simplify`), and `get_jumps` is a pure field load that Enzyme handles cleanly:

```julia
function supports_initialization(sys::System)
    return isempty(get_systems(sys)) && isempty(get_jumps(sys)) &&
        isempty(get_costs(sys)) && isempty(get_constraints(sys))
end
```

I verified locally that monkey-patching this exact change on top of the current MTKBase makes the MWE above succeed and produces correct gradients via Enzyme.

## Notes

- This is *one* layer of the broader Enzyme+MTK remake/solve story. After patching `supports_initialization`, the `remake` (no solve) case works, but full `remake + solve` still hits separate readonly-proof issues downstream (in FunctionWrappersWrappers `SingleCacheStorage.cached = new_fw` on cache miss, and Enzyme's `MixedDuplicated(::NonlinearSolution, ...)` runtime tactical wrapper not matching). I'll file those separately if/when they prove tractable, but this MTK fix is the most isolable and the cleanest first step.

- Closing #4181's broader story will still require Enzyme to differentiate through `SetInitialUnknowns` correctly — this issue just unblocks the entry into that path.

## Environment

- Julia 1.12.6
- Enzyme v0.13.148
- ModelingToolkit v11.26.5 (ModelingToolkitBase tHvXN)
- SciMLBase v3.13.1 (with `1208`/Mooncake scalar-u fix applied)
- SciMLStructures v1.10.0
- FunctionWrappersWrappers v1.9.0

ChrisRackauckas-Claude:

## Summary

`supports_initialization(sys)` short-circuits on `isempty(get_systems(sys))`, so the call to `jumps(sys)` here only ever reaches the early-return branch where `jumps(sys) === get_jumps(sys)`. However the unreached `copy(js) + append!(js, namespace_jumps(subsys))` branch in `jumps` is enough to prevent Enzyme's static analysis from proving `sys` read-only at this call site — and that breaks `Enzyme.gradient` through any closure that reaches `_late_binding_update_u0_p_impl`, e.g. `remake(prob; p = repack(tunables))` on any MTK-generated problem.

Swap to `get_jumps(sys)`, which is a pure field accessor and identical in this short-circuited position.

## Verification

The MWE from #4550:

```julia
@parameters a b c
@variables x(t) y(t) z(t)
eqs = [D(x) ~ a*x + y + z, 0 ~ y^3 + y - b*x, 0 ~ z^3 + z - c*y]
@mtkbuild sys = ODESystem(eqs, t)
prob = ODEProblem(sys, [x => 1.0], (0.0, 0.1),
                  [a => -0.5, b => 2.0, c => 1.5];
                  guesses = [y => 1.0, z => 0.5], use_scc = false)
iprob = prob.f.initialization_data.initializeprob
itunables, irepack, _ = SS.canonicalize(SS.Tunable(), parameter_values(iprob))
remake_only = let iprob = iprob, irepack = irepack
    p -> sum(remake(iprob, p = irepack(p)).u0)
end
Enzyme.gradient(Enzyme.Reverse, remake_only, itunables)
```

Before this PR: `EnzymeMutabilityException` at `supports_initialization` → `jumps(sys)`.

After this PR: `Enzyme.gradient` completes (returns a vector). The returned gradient itself is currently `[0.0, …]` because MTK#4181 (Enzyme rule for `SetInitialUnknowns`) is still missing — that's a separate semantic-correctness issue that this PR doesn't touch; it just removes the compile-time blocker.

## What downstream this unblocks

This is **Layer 1** of a 4-layer Enzyme-through-MTK-remake stack documented during the SciMLSensitivity #1323 / #1359 investigation. Layers 2–4 (FunctionWrappersWrappers cache mutation, SciMLBase `Const(loss)` `EnzymeRuntimeActivityError`, Enzyme.jl `MixedDuplicated{NonlinearSolution}` MethodError) still bite once you go past `remake` to `remake + solve`. Those need separate upstream work. The `@test_broken Enzyme.gradient` blocks in SciMLSensitivity's `mtk.jl` / `parameter_initialization.jl` / `desauty_dae_mwe.jl` stay until all four are resolved.

## Refs

- Closes #4550
- Related: #4181 (Enzyme rule for `SetInitialUnknowns` — Layer 4, separate)
- SciML/SciMLSensitivity.jl#1359, #1323

## Test plan
- [x] Local: `Enzyme.gradient` on the MWE no longer errors
- [ ] CI green

### DIFF
--- a/lib/ModelingToolkitBase/src/systems/system.jl
+++ b/lib/ModelingToolkitBase/src/systems/system.jl
@@ -1503,7 +1503,15 @@ function Base.showerror(io::IO, err::EventsInTimeIndependentSystemError)
 end
 
 function supports_initialization(sys::System)
-    return isempty(get_systems(sys)) && isempty(jumps(sys)) &&
+    # Use `get_jumps` rather than the flattened `jumps`. The two are
+    # equivalent here because the `isempty(get_systems(sys))` short-circuit
+    # before this term guarantees the subsystem-flattening branch in
+    # `jumps(sys)` (which does `copy` + `append!`) is unreachable. The
+    # mutation in that unreachable branch nevertheless prevents Enzyme from
+    # proving `sys` is read-only, blocking `Enzyme.gradient` through any
+    # closure that ends up at `_late_binding_update_u0_p_impl` — including
+    # `remake(prob; p = repack(tunables))` on MTK-generated problems.
+    return isempty(get_systems(sys)) && isempty(get_jumps(sys)) &&
         isempty(get_costs(sys)) && isempty(get_constraints(sys))
 end
 
--- a/lib/ModelingToolkitBase/src/systems/system.jl
+++ b/lib/ModelingToolkitBase/src/systems/system.jl
@@ -1503,14 +1503,6 @@ function Base.showerror(io::IO, err::EventsInTimeIndependentSystemError)
 end
 
 function supports_initialization(sys::System)
-    # Use `get_jumps` rather than the flattened `jumps`. The two are
-    # equivalent here because the `isempty(get_systems(sys))` short-circuit
-    # before this term guarantees the subsystem-flattening branch in
-    # `jumps(sys)` (which does `copy` + `append!`) is unreachable. The
-    # mutation in that unreachable branch nevertheless prevents Enzyme from
-    # proving `sys` is read-only, blocking `Enzyme.gradient` through any
-    # closure that ends up at `_late_binding_update_u0_p_impl` — including
-    # `remake(prob; p = repack(tunables))` on MTK-generated problems.
     return isempty(get_systems(sys)) && isempty(get_jumps(sys)) &&
         isempty(get_costs(sys)) && isempty(get_constraints(sys))
 end


ChrisRackauckas-Claude:
## Layer 2 status update

Confirmed independently that landing this Layer 1 fix is necessary but not sufficient: with this PR applied, bare `Enzyme.gradient` through `remake` succeeds, but `remake + solve` still fails with `EnzymeMutabilityException`. I reduced Layer 2 to a no-MTK MWE and filed it upstream at EnzymeAD/Enzyme.jl#3117:

> `Enzyme.gradient` over `NonlinearSolve.solve` works for a `NonlinearProblem` whose `f` is a plain anonymous function, but fails as soon as `f` is a `RuntimeGeneratedFunction` (or any wrapped-function pattern — `eval_expression = true` still fails). The failing IR variable is `iprob.f1.i.sroa.0.0` — a Union-tag byte store SROA'd through `NonlinearFunction.f` reported inside `SciMLBase`'s mutable `NonlinearProblem` constructor.

The full breakdown of layers 2–4 is in the discussion on SciMLSensitivity#1359. This PR remains the right Layer 1 fix on its own; the three `@test_broken Enzyme.gradient` blocks in SciMLSensitivity (`mtk.jl`, `parameter_initialization.jl`, `desauty_dae_mwe.jl`) flip to passing once both this and EnzymeAD/Enzyme.jl#3117 land.

I also filed SciML/FunctionWrappersWrappers.jl#51 (defensive `inactive_type` rule for cache storage) — that one doesn't fix Layer 2 by itself but removes one conservative-IPA layer.

AayushSabharwal:
Test? I feel like this is very easy to accidentally break

ChrisRackauckas:
The only real answer for whether Enzyme works is if Enzyme works, so we'll get that in the integration tests at the end of the journey.

PR #4559, opened 2026-05-25
refactor: wrap `update_A!`, `update_b!` with FWW

AayushSabharwal:
This avoids excessive specialization during initialization and reduces TTFX.


AayushSabharwal:
The failures are false negatives. Codecov is not cooperating. I've manually checked everything to make sure it's good.

PR #4572, opened 2026-05-31
Re-enable downgrade CI with genuine test-at-floor (allow_reresolve: false)

ChrisRackauckas-Claude:
**Please ignore until reviewed by @ChrisRackauckas.** Draft.

Re-enables the Downgrade workflow (group InterfaceI, `allow_reresolve: false`) for genuine test-at-floor on the main package. Raises declared-dep `[compat]` lower bounds, and adds `[compat]` floors on three **declared test extras** (`Optimization`, `OptimizationBase`, `OptimizationOptimJL` — already in `[extras]`) so the downgrade resolver doesn't minimize `OptimizationBase` below 5.1.0, which `BoundaryValueDiffEqCore` 2.2 (via the `BoundaryValueDiffEqMIRK` test dep) requires.

Notable bumps: SciMLBase→2.154 (`FunctionWrappersWrappers.SingleCacheStorage`), OrdinaryDiffEqCore→3.32 (`allows_null_u0`), SimpleNonlinearSolve→2.10, ForwardDiff→1, DiffEqBase→6.213, BipartiteGraphs→0.1.4, OffsetArrays→1.3, DifferentiationInterface→0.7.13, SymbolicIndexingInterface→0.3.46, PreallocationTools→1.2, CommonSolve→0.2.6, Combinatorics→1.0.2, ConstructionBase→1.5.8, BoundaryValueDiffEqMIRK→1.13, DiffEqNoiseProcess→5.25 + the Optimization* test-extra floors.

Verified locally on Julia 1.10 (group InterfaceI, fresh registry): `Pkg.test(allow_reresolve=false)` passes at the floor with 0 re-resolve. CI is the authoritative gate.


ChrisRackauckas-Claude:
Heads-up for review: local verification shows this PR's downgrade job will currently **fail at build** on Julia 1.10, and it is NOT fixable by a compat bump. Root cause: root MTK calls the 3-arg `MTKBase.topsort_equations(sys, observed(sys), fullunknowns)` (src/systems/systemstructure.jl:229). The dev `lib/ModelingToolkitBase` defines that 3-arg method, but the **registered** ModelingToolkitBase 1.41.0 (which the downgrade manifest resolves to) only has the 2-arg form — same version number, different API. So the downgrade job needs either a new MTKBase release containing the 3-arg method (so a floor can be raised) or the downgrade flow to bind MTKBase to the local `lib/` path. Re-enabling as-is (this diff) will be red until then. Flagging rather than claiming it works.

ChrisRackauckas-Claude:
Correction to my earlier comment: I re-ran the downgrade build on a clean depot against the current registry, and it **builds successfully** (build_exit=0) with ModelingToolkitBase resolved from the registry at 1.41.0 (not the local lib path). My previous `topsort_equations` MethodError report was a stale-cache/contention artifact from a heavily-loaded run, not a real registry/API skew — apologies for the noise. The downgrade resolve+build are fine at the floor; the remaining open question is only whether the full InterfaceI test suite passes at minimum versions, which CI on this PR will show.

PR #4577, opened 2026-06-02
fix: support `::Number` for variables

### LINKED ISSUE #4548: Support for complex-valued equations of motion (no re/im split)
I work on [QuantumCumulants.jl](https://github.com/qojulia/QuantumCumulants.jl) where the natural state variables are complex-valued averaged operators like ⟨a⟩. Right now there's no clean way to declare these in MTK. The three obvious annotations all fail:

**1. `::Real` (default) + ComplexF64 u0**

Works at runtime because the integrator dispatches on `eltype(u0)`, but Symbolics folds `conj(x) → x` since x has Real symtype. Any RHS containing `⟨X⟩ - ⟨X†⟩` (very common in dissipative quantum systems) silently collapses to zero.

**2. `::Number`** (a supertype of Real, so this *should* work):

```julia
using ModelingToolkitBase, OrdinaryDiffEq
using ModelingToolkitBase: t_nounits as t, D_nounits as D
@variables x(t)::Number
@named sys = System([D(x) ~ -x], t)
ODEProblem(mtkcompile(sys), [x => 1.0], (0.0, 1.0))
# ERROR: TypeError: in typeassert,
#   expected ParameterIndex{Initials, Int64},
#   got      ParameterIndex{Nonnumeric, Tuple{Int64, Int64}}
#   [1] SetInitialUnknowns(sys::System) at problem_utils.jl:1127
```

Number-symtype unknowns get classified as `Nonnumeric` (a category meant for callables / non-numeric objects) and break ODEProblem assembly.

**3. `::Complex` auto-splits into real/imag equations**:

```julia
@variables x(t)::Complex
eqs = D(x) ~ -x
# eqs is Equation[D(real(x)) ~ -real(x), D(imag(x)) ~ -imag(x)]
@named sys = System(eqs, t)
mtkcompile(sys)
# ERROR: ExtraEquationsSystemException: 2 equations / 1 unknowns
```

Symbolics produces the correct re/im pair but the System still counts 1 unknown. Forcing it through with `System(eqs, t, [real(x), imag(x)], [])` works but then the user has to manage the re/im plumbing themselves: no automatic ⟨a⟩ scalar handling, no `sol(t; idxs=x)` returning Complex, no shared defaults.

**What would help**

A first-class complex-state path that *doesn't* decompose into re/im. `x` stays a single symbolic Complex unknown, the user supplies a `ComplexF64` u0, and the integrator carries `Vector{Complex}` internally, functionally similar to today's `::Real` + Complex u0 trick but without the `conj`-folding gotcha. The re/im split is genuinely useful for some workflows; I'm asking for the *option* to opt out of it.

Use cases: quantum optics, photonics, anything Hamiltonian with creation/annihilation operators is more natural in Complex.

Versions: ModelingToolkitBase v1.36.2, SymbolicUtils v4.30.0, Symbolics v7.24.1, Julia 1.12.6.

Complex equations of motion used to work transparently in earlier versions of ModelingToolkit. The current QuantumCumulants `master` branch (built on the older MTK) relies on this; migrating to ModelingToolkitBase + the current Symbolics is what surfaced the gap.

AayushSabharwal:
Close #4548 

@oameye could you verify that this covers your use case?

### DIFF
--- a/lib/ModelingToolkitBase/src/systems/index_cache.jl
+++ b/lib/ModelingToolkitBase/src/systems/index_cache.jl
@@ -195,12 +195,15 @@ function IndexCache(sys::AbstractSystem)
         haskey(constant_buffers, ctype) && p in constant_buffers[ctype] && continue
         haskey(nonnumeric_buffers, ctype) && p in nonnumeric_buffers[ctype] && continue
         insert_by_type!(
-            if ctype <: Real || ctype <: AbstractArray{<:Real}
+            if ctype <: Number || ctype <: AbstractArray{<:Number}
                 if istunable(p, true) && symbolic_has_known_size(p) &&
                         (
-                        ctype == Real || ctype <: AbstractFloat ||
-                            ctype <: AbstractArray{Real} ||
-                            ctype <: AbstractArray{<:AbstractFloat}
+                        ctype == Real || ctype == Number || ctype == Complex ||
+                            ctype <: AbstractFloat || ctype <: Complex ||
+                            ctype <: AbstractArray{Real} || ctype <: AbstractArray{Number} ||
+                            ctype <: AbstractArray{Complex} ||
+                            ctype <: AbstractArray{<:AbstractFloat} ||
+                            ctype <: AbstractArray{<:Complex}
                     )
                     if iscall(p) && operation(p) === Initial()
                         initial_pars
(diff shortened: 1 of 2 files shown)


oameye:
Yes, I checked locally and it solves our use case :)

PR #4578, opened 2026-06-02
refactor: use `subset_ir` and `replace_node!` in `build_function_wrapper`

AayushSabharwal:
This approach enables calling `replace_node!` instead of performing substitutions, which is faster. Since `subset_ir` is an independent copy, optimization passes do not have to repeatedly copy the `IRStructure`. This leads to a significant improvement in `ODEProblem` runtime for large models.


AayushSabharwal:
This is buggy. If a binding `p1 => p2` exists, where `p1` and `p2` are array parameters and `p2[1]` is passed to the function, then it doesn't codegen right because `cs.rewrites` contains `p2[1] => __mtkparameters__[2][i]` but the expression in the ir is `p1[1]`.

AayushSabharwal:
Substituting instead of `replace_node!` really isn't slow at all

PR #4590, opened 2026-06-04
ci: route CI through SciML/.github reusable workflows (@v1)

ChrisRackauckas-Claude:
This routes ModelingToolkit.jl's bespoke GitHub Actions workflows through the centralized reusable workflows in [`SciML/.github`](https://github.com/SciML/.github) (`@v1`), preserving the repo's existing matrices, triggers, concurrency, coverage directories, and per-group/per-version configuration. No CI coverage is dropped.

### Converted to reusable callers

| Workflow | Now calls | Notes |
|---|---|---|
| `Tests.yml` | `tests.yml@v1` | Main ModelingToolkit groups only (`InterfaceI`, `InterfaceII`, `Initialization`, `SymbolicIndexingInterface`, `Extensions`, `Optimization`, `Downstream`, `FMI`). Versions `[1, lts, pre]`, coverage dirs `src,ext,lib/ModelingToolkitBase/src,lib/ModelingToolkitBase/ext`, self-hosted via `vars.USE_SELF_HOSTED`, and the original `on:`/`paths-ignore`/`concurrency` preserved. The local `lib/ModelingToolkitBase` + `lib/SciCompDSL` deps resolve automatically through the existing `[sources]` in `Project.toml`. |
| `SublibraryCI.yml` (new) | `sublibrary-tests.yml@v1` | Auto-discovers `lib/*`. The sublibrary test groups that previously lived inline in `Tests.yml` (the `ModelingToolkitBase/*` and `SciCompDSL/All` rows) are now expressed as `lib/<name>/test/test_groups.toml`, preserving the exact group + version matrix (incl. `QA` excluding `pre`). |
| `Downgrade.yml` | `downgrade.yml@v1` | `julia-version: 1.10`, `group: InterfaceI`, `skip: Pkg,TOML`, `allow-reresolve: false` carried over. The job remains intentionally disabled (`if: false`), matching current behavior. |
| `FormatCheck.yml` | `runic.yml@v1` | Repo already formats with Runic. |
| `SpellCheck.yml` | `spellcheck.yml@v1` | typos check. |

### `test_groups.toml` added (preserve sublibrary matrix)

- `lib/ModelingToolkitBase/test/test_groups.toml`: `InterfaceI, InterfaceII, Initialization, SymbolicIndexingInterface, Extended, RegressionI, Extensions, Optimization` on `[lts, 1, pre]`; `QA` on `[lts, 1]`.
[...]

PR #4591, opened 2026-06-04
ci: revert unsafe sublibrary-tests consolidation (restores dropped CI coverage)

ChrisRackauckas-Claude:

## Summary

Follow-up fix for #4590 (already merged into `master`), which is currently **dropping CI test coverage** for the `ModelingToolkitBase` and `SciCompDSL` sublibraries.

An audit found that #4590 moved the `ModelingToolkitBase/*` and `SciCompDSL/*` test groups out of the bespoke `Tests.yml` and into a new `SublibraryCI.yml` that calls `SciML/.github/.github/workflows/sublibrary-tests.yml@v1`. That reusable workflow computes its CI `GROUP` value in `scripts/compute_affected_sublibraries.jl` as:

```julia
ci_group = group_name == "Core" ? pkg : "$(pkg)_$(group_name)"
```

so it sends `GROUP` values like `ModelingToolkitBase_InterfaceI`, `ModelingToolkitBase_QA`, `SciCompDSL_All`, etc. But `lib/ModelingToolkitBase/test/runtests.jl` (and SciCompDSL's) dispatch on **bare** group names:

```julia
const GROUP = get(ENV, "GROUP", "All")
...
if GROUP == "All" || GROUP == "InterfaceI"
```

The prefixed names match neither `"All"` nor any bare group name, so **all 9 `ModelingToolkitBase` groups and the `SciCompDSL` group ran ZERO tests**. (The old bespoke `Tests.yml` set `ENV["GROUP"]` to the bare second path component and ran them correctly.) Additionally, `sublibrary-tests.yml@v1` is change-gated (runs only when `lib/<name>/*` files change), whereas the old `Tests.yml` ran these groups unconditionally on every push/PR — so even apart from the naming bug, main-package-only changes lose sublibrary coverage.

## What this PR does

Per the SciML CI-centralization spec's "MONOREPO SUBLIBRARY CONSOLIDATION — DANGER" guidance, the sublibrary **test** CI is intentionally **left bespoke** (option (b): do not add `SublibraryCI.yml`):

- **Restore** `.github/workflows/Tests.yml` verbatim from #4590's base. It runs all **18 pkggroups** (8 `ModelingToolkit/*` + 9 `ModelingToolkitBase/*` + 1 `SciCompDSL/All`) in a single dispatching job with the correct `GROUP` values, the `pre`/`ModelingToolkitBase/QA` exclude, and the `src,ext,lib/ModelingToolkitBase/src,lib/ModelingToolkitBase/ext` coverage directories.
- **Remove** `.github/workflows/SublibraryCI.yml` (the unsafe consolidation).
- **Remove** `lib/ModelingToolkitBase/test/test_groups.toml` and `lib/SciCompDSL/test/test_groups.toml`, which #4590 added only to feed the broken consolidation (they did not exist before and encode the broken mapping).

## What this PR keeps (faithful conversions from #4590, NOT reverted)

- `Downgrade.yml` -> `downgrade.yml@v1` (preserving `if: false`, julia `1.10`, `skip: Pkg,TOML`, `allow-reresolve: false`, `group: InterfaceI`)
- `FormatCheck.yml` -> `runic.yml@v1`
- `SpellCheck.yml` -> `spellcheck.yml@v1`
- The separate `DowngradeSublibraries.yml` -> `sublibrary-downgrade.yml@v1` (from #4589) is untouched.

`Documentation.yml`, `Downstream.yml`, `ReleaseTest.yml`, `benchmark.yml`, `TagBot.yml`, `CompatHelper.yml` remain bespoke as before.

## Notes

- The main `Project.toml` already has a `[sources]` entry dev-ing `lib/ModelingToolkitBase` (and `SciCompDSL`), so the main-package leg correctly develops the local sublibrary path. Confirmed, no change needed.
- **Proper follow-up (separate, maintainer-reviewed step):** to actually centralize the sublibrary *test* CI later, add `lib/<name>/test/test_groups.toml` encoding the real groups + versions **and** align each sublibrary's `runtests.jl` GROUP parsing so it honors the `<pkg>_<group>` value that `sublibrary-tests.yml@v1` sends (or normalizes the prefix back to the bare name). Only do this once it can be done faithfully; until then the bespoke `Tests.yml` is the safe option.
- Validation: every workflow YAML parses (`python3 yaml.safe_load`), and every remaining reusable caller (`downgrade.yml`, `runic.yml`, `spellcheck.yml`, `sublibrary-downgrade.yml`) references a real `SciML/.github/.github/workflows/*.yml@v1` with only valid inputs.


ChrisRackauckas-Claude:
Superseded by https://github.com/SciML/ModelingToolkit.jl/pull/4592: keeping the centralized sublibrary-tests@v1 form and fixing coverage via test_groups.toml + runtests GROUP parsing, instead of reverting to bespoke CI_*.yml.

PR #4601, opened 2026-06-08
Add the homotopy operator and HomotopyProblem construction

BatchZero:
## Summary

Implements the `homotopy(actual, simplified)` operator as a general nonlinear-solving construct. `homotopy` is a way to solve a hard nonlinear system by continuation: start from an easy `simplified` form and deform it into the `actual` one. Initialization is the most common consumer, but it is not the only one — so this PR exposes the construct on its own, and a follow-up wires it into the initialization default path.

Per the review, this is **PR 1 of 2**:

- **This PR (A):** the operator, the lowering pass, the `SciMLBase.HomotopyProblem(sys, op)` constructor, and the automatic `SciMLBase.AbstractNonlinearProblem(sys, op)` selection.
- **Follow-up (B):** initialization wiring — default `initializealg = OverrideInit(nlsolve = HomotopySweep())` injection, `InitializationProblem` routing through the constructor, `expression = Val{true}` auto-injection, the `NonlinearSolveBase` hard dep.

MTK's part is an independent lowering pass targeting an MTK-independent API. The solver stack merged and registered upstream first:

- `SciMLBase.HomotopyProblem` (SciML/SciMLBase.jl#1375, SciMLBase v3.19.0) — residual `f(u, p, λ)` / `f(du, u, p, λ)`; λ is a trailing scalar argument, never an entry of `p`;
- `NonlinearSolveBase.HomotopySweep` (SciML/NonlinearSolve.jl#962, NonlinearSolveBase v2.31.0, re-exported by NonlinearSolve) — the natural-parameter continuation solver dispatching on `HomotopyProblem`. A `HomotopyProblem` with no algorithm defaults to it.

This PR contains lowering + targeting only — no solver code.

## Review items addressed

- **"It's just a nonlinear solver, why speak to initialization here?"** The docs page and the operator docstring are reframed around general nonlinear solving; the direct `HomotopyProblem` / `AbstractNonlinearProblem` construction is the headline, with initialization mentioned as one consumer (wired in PR B).
- **"There should be an `AbstractNonlinearProblem` constructor that is automatic."** Added `SciMLBase.AbstractNonlinearProblem(sys, op)` (re-exported), which returns a `HomotopyProblem` when `sys` carries `homotopy` nodes and a `NonlinearProblem` otherwise (`get_nonlinear_problem_type`). The explicit `SciMLBase.HomotopyProblem(sys, op)` is also provided.
- **"Split [the rewrap] to the `AbstractNonlinearProblem` constructor."** The detection + lowering + wrap now live in the `HomotopyProblem(sys, op)` constructor (`problems/homotopyproblem.jl`), not inside `InitializationProblem`. PR B routes init through it.
- **`gensym` breaks precompilation.** λ is now a fixed reserved unicode sentinel (`__homotopy_λₘₜₖ`, matching MTK's other internal `ₘₜₖ` sentinels), so codegen is a pure function of the system and the precompile / RuntimeGeneratedFunctions Expr-hash caches hit. A cheap guard turns the otherwise-silent user-symbol collision into an error.
- **The mirrored codegen path ("?").** `generate_rhs` gains an additive `extra_args::Tuple = ()` keyword; `generate_homotopy_residual` now threads λ through it and collapses to two guards plus a one-line forward. The "keep in sync" comment is gone — there is no second copy. Zero behavior change for the 9 existing `generate_rhs` callers (`extra_args` defaults to `()`, `p_end` only forwarded when non-empty).
- **"Recommend broadcast."** Documented and tested: `homotopy.(actual, simplified)` creates one node per element; the lowering rewrites each independently under the single shared λ.
- **NEWS.md "# Unreleased".** Removed — the repo convention is that feature PRs don't touch NEWS.md (it is a major-version upgrade guide).

## How it works

- **Operator.** `homotopy(actual, simplified)` is registered and opaque through `System` construction, `mtkcompile`, and runtime codegen. The numeric fallback is `actual` — `simplified` is evaluated and discarded, an honest cost borne only by annotated systems. Systems that don't use the operator are byte-unaffected.
- **Derivatives.** Registered nodewise: ∂₁ = `homotopy(1, 0)`, ∂₂ = `homotopy(0, 1)`. At runtime these fold to the derivative of `actual`; under the swept rewrite they blend to λ and 1 − λ, so differentiated equations keep the same blend structure.
- **Lowering pass (`lower_homotopy`).** A shadow rewrite replaces every `homotopy(a, s)` (in equations and observed) with `(1 − λ)·s + λ·a`, one shared sentinel λ across all occurrences (and nested ones). Structure-preserving: same equation count/order/incidence modulo the free λ. The shadow never passes through `mtkcompile`.
- **Residual (`generate_homotopy_residual`).** Compiled as `f(u, p, λ)` via `generate_rhs`'s `extra_args`, with λ in the t-slot; an explicit `p_end` keeps λ out of the `MTKParameters` destructuring. User `p` / `MTKParameters` pass through untouched. The jacobian/sparsity of the standard build are dropped (they encode the λ = 1 system).
- **Construction.** `HomotopyProblem(sys, op)` (or `AbstractNonlinearProblem(sys, op)`) builds the swept problem; `λspan` is inherited from the upstream `HomotopyProblem` default `(0.0, 1.0)` — this PR introduces no numeric defaults of its own. Solve it with `solve(prob)` (defaults to `HomotopySweep`) or an explicit `HomotopySweep(...)`.

## Tests (MTKBase, all green locally)

- `homotopy_lowering.jl` — opaque-through-`complete`, runtime fallback, nodewise derivatives, shared-λ lowering across equations + observed, nested `homotopy`, the sentinel guard, and broadcast.
- `homotopy_problem.jl` — direct `HomotopyProblem(sys, op)` construction + residual endpoints (simplified / actual / convex blend, oop + iip), the automatic `AbstractNonlinearProblem` selection (and its negative control), out-of-basin rescue through the sweep, structured `p` pass-through, two `homotopy` calls sharing one sweep.
- `homotopy_omc_parity.jl` — a pressure-drop network pattern (homotopy reaching the residual through an **observed** equation), solved to the analytic root (`m_flow = √5` to 1e-6) by both `solve(prob)` and an explicit `HomotopySweep`.

## Dependencies & compat

`SciMLBase` floored at `3.19` (`HomotopyProblem`). **No new hard dependency:** the solver routes to `HomotopySweep` through `NonlinearSolveBase`'s own `solve(::HomotopyProblem)` fallback at solve time, so `NonlinearSolveBase` is only a test dependency here (re-exported via `NonlinearSolve`). The hard `NonlinearSolveBase` dep arrives with the initialization injection in PR B. `SimpleNonlinearSolve` floored at `2.11.1` (lowest co-resolvable with `SciMLBase 3.19`).

## Deferred to follow-up PRs (listed so "why just N?" is pre-answered)

- **PR B:** initialization wiring (default `initializealg` injection, `InitializationProblem` routing, `Val{true}` auto-injection).
- **Per-SCC sweep.** A homotopy-containing system can be swept per strongly-connected component. The structure supports it (`SCCNonlinearProblem.probs` has no element-type constraint; `HomotopyProblem` subproblems route to `HomotopySweep` via the existing fallback), but it needs a small NonlinearSolve enabler first (`probvec(::HomotopyProblem)`, a `solve(::HomotopyProblem, ::AbstractNonlinearAlgorithm)` wrapping rule, Stalled-resid aggregation) and per-SCC residual codegen — its own PR.
- **`expression = Val{true}`** for the direct constructor.
- **Analytic jacobian** for the swept residual.

## Expected CI reds (pre-attributed)

- codecov upload (fork token);
- repo-wide Runic `FormatCheck` (red on master itself; this PR adds zero new Runic violations — every touched line is Runic 1.7.0-clean, and unrelated formatting on master was left untouched to avoid drive-by churn);
- the usual downstream/integration flakes.


BatchZero:
Status note: per the review on SciML/SciMLBase.jl#1375 and SciML/NonlinearSolve.jl#962, the homotopy parameter λ is being separated from the parameter vector (the residual becomes `f(u, p, λ)` instead of a `homotopy_parameter` locator into `p`). Once those two PRs settle, this PR's lowering will be reworked to target the revised API — holding off on changes here until then to avoid churn.

BatchZero:
Both upstream dependencies are merged and registered: SciML/SciMLBase.jl#1375 → SciMLBase v3.19.0 (`HomotopyProblem`, residual `f(u, p, λ)`) and SciML/NonlinearSolve.jl#962 → NonlinearSolveBase v2.31.0 (`HomotopySweep`, re-exported by NonlinearSolve 4.19.1). This PR is reworked to that convention and rebased onto master d95960f6 (34 commits, d95960f6..4706e066; description updated to match): λ is now a trailing scalar argument of the regenerated init residual — compiled into the t-slot, never a parameter, user `p`/`MTKParameters` untouched, the old locator design gone — produced by an independent lowering pass at the `InitializationProblem` boundary (parent system unmutated) and wrapped as a `HomotopyProblem` targeting `HomotopySweep` through `OverrideInit`. Per the #4576 verdict no solver code lives here: a grep over the src diff finds zero sweep/λ-stepping logic (9 doc/comment mentions + 1 construction); the weakdep extension, `isdefined` guard, and factory `Ref` are deleted in favor of a hard `NonlinearSolveBase` dep that is already in MTKBase's unconditional load closure via SimpleNonlinearSolve (zero packages added to the resolved set, `Pkg.resolve`-verified; precedent: the hard `import SCCNonlinearSolve`). Compat floors set: SciMLBase = "3.19", NonlinearSolveBase = "2.31", and on the existing hard dep SimpleNonlinearSolve = "2.11.1" — the lowest version co-resolvable with the new floors, since DowngradeSublibraries pins `[deps]` floors and SciMLBase 3.19 does not admit older 2.x releases — plus test dep NonlinearSolve = "4.19"; root NonlinearSolve is not floored at 4.20 because 4.20.0 is unregistered and the registered 4.19.1 co-resolves with NonlinearSolveBase 2.31 (verified). MTKBase `Initialization` group full run is green — the only non-green is 12 pre-existing `Broken` at `initializationsystem.jl:959` (+2 version-gated at `:1735-1736` off Julia 1.12; all pre-existing on master) — plus 92 new tests across five files, including OpenModelica v1.27 parity (`m_flow(0) = √5` to 1e-6 on the Buildings PressureDrop pattern through the swept-observed path) and a root-MTK SCC-forcing e2e with negative control; touched files are Runic 1.7.0-clean (the CI version) and the injected-alg construction is concrete under `@code_warntype`. Expected CI reds, all pre-attributed: codecov upload failure (fork token), repo-wide Runic FormatCheck (red on master itself), the root-level Downgrade conflict (root SciMLBase compat vs in-repo MTKBase ≥ 3.18, present on master), and the usual downstream/integration flakes. Deferred as follow-up PRs, listed in the description: NLLS sweep for non-square init, per-SCC sweep, the `default_nlsolve(::Nothing, …, ::HomotopyProblem)` hook in OrdinaryDiffEqNonlinearSolve (which retires the injection branch here), `Val{true}` auto-injection, and elision of the dead `simplified` branch.

docs/src/basics/Homotopy.md
```diff
@@ -0,0 +1,167 @@
+# [Homotopy Initialization](@id homotopy)
```
ChrisRackauckas: Why even speak to initialization here? It's just a nonlinear solver. The fact that it shows up commonly in initialization is a detail, but there are many other uses for it.

docs/src/basics/Homotopy.md
```diff
@@ -0,0 +1,167 @@
...
+residual — it is never added to the system's parameters, and your parameter object
+`p` passes through untouched. All `homotopy` calls in a system share the single
+`λ`, per the Modelica spec's recommendation of (conceptually) one homotopy
+iteration over the whole model; this includes nested `homotopy` calls. The result is wrapped as a `SciMLBase.HomotopyProblem`
+(its `λspan` defaults to `(0.0, 1.0)`).
+
+Problems built through the default path then receive
+
+```julia
+initializealg = SciMLBase.OverrideInit(nlsolve = NonlinearSolveBase.HomotopySweep())
```
ChrisRackauckas: Any nonlinear solve of equations which have a homotopy operator should default to defining a homotopy problem.

ChrisRackauckas: Or rather, there should be an AbstractNonlinearProblem constructor that is automatic here.

docs/src/basics/Homotopy.md
```diff
@@ -0,0 +1,167 @@
...
+    injected when you did not pass one, and a solve-time `initializealg` overrides
+    the problem default as usual.
+  - If simplification eliminates every `homotopy` node before the initialization
+    system is built, initialization proceeds as standard (detection happens on the
+    compiled initialization system).
+  - Multiple (or nested) `homotopy` calls in one system are blended with the one
+    shared `λ` and swept together in a single continuation.
+  - If the initialization would otherwise target an `SCCNonlinearProblem`, a
+    homotopy-using system is forced to the monolithic `NonlinearProblem` target,
+    since the continuation deforms the coupled system along one `λ`.
```
ChrisRackauckas: It should probably split that to SCCs.

docs/src/basics/Homotopy.md
```diff
@@ -0,0 +1,167 @@
...
+Tolerances are inherited from the integrator's `abstol`/`reltol`.
+
+## Limitations
+
+  - **Non-square initialization cannot be swept.** If the initialization problem
+    is not a square `NonlinearProblem` (e.g. over-determined initialization, which
+    falls back to least squares), a one-time warning is emitted and initialization
+    solves the `actual` (`λ = 1`) form instead. Least-squares and per-SCC sweeps
+    are future work.
+  - **Trivial initialization is silently correct.** If the initialization system
```
ChrisRackauckas: Obviously. Why is this even worth mentioning?

docs/src/basics/Homotopy.md
```diff
@@ -0,0 +1,167 @@
...
+    has no unknowns, there is nothing to solve and no sweep happens; observed
+    evaluation folds `homotopy` to `actual`, the correct
+    outside-initialization semantics.
+  - **`expression = Val{true}` does not auto-inject the `initializealg`.** When
+    generating problem-construction expressions, pass
+    `initializealg = SciMLBase.OverrideInit(nlsolve = NonlinearSolveBase.HomotopySweep())`
+    explicitly.
```
ChrisRackauckas: That's a bug.

docs/src/basics/Homotopy.md
```diff
@@ -0,0 +1,167 @@
...
+  - **Parameter bindings and defaults are not rewritten.** A `homotopy` call
+    inside a parameter binding or default value evaluates as `actual` at all `λ`.
+  - **Scalar `Real` expressions only**, matching Modelica's restriction; the
+    numeric fallback method is `homotopy(::Real, ::Real)`.
```
ChrisRackauckas: Recommend broadcast.

docs/src/basics/Homotopy.md
```diff
@@ -0,0 +1,167 @@
...
+  - **Failure surfaces as a failed initialization.** If the continuation cannot
```
ChrisRackauckas: Obviously. Also, not relevant to the MTK side.

lib/ModelingToolkitBase/src/problems/initializationproblem.jl
```diff
@@ -123,7 +123,98 @@ All other keyword arguments are forwarded to the wrapped problem constructor.
...
+        # sweeping is a separate future improvement. Root ModelingToolkit's
+        # `get_initialization_problem_type` prefers `SCCNonlinearProblem` for split
+        # fully-determined systems — force the monolithic target here.
+        TProb = NonlinearProblem
+    end
+    initprob = TProb{_iip}(isys, op; kwargs..., build_initializeprob = false, is_initializeprob = true)
+    if homotopic
+        initprob = _maybe_sweep_homotopy_initprob(initprob, isys, _iip, kwargs)
+    end
+    return initprob
```
ChrisRackauckas: Split to the AbstractNonlinearProblem constructor.

lib/ModelingToolkitBase/src/systems/codegen.jl
```diff
@@ -81,6 +81,7 @@ function generate_rhs(
             rhss = [_iszero(eq.lhs) ? eq.rhs : eq.rhs - eq.lhs for eq in eqs]
         end
     else
+        # NOTE: generate_homotopy_residual in homotopy.jl mirrors the time-independent path here; keep in sync.
```
ChrisRackauckas: ?

lib/ModelingToolkitBase/src/systems/homotopy.jl
```diff
@@ -0,0 +1,312 @@
...
+"""
+function lower_homotopy_for_init(isys::AbstractSystem)
+    if !iscomplete(isys)
+        throw(
+            ArgumentError(
+                "`lower_homotopy_for_init` expects a `complete`d (flattened) system."
+            )
+        )
+    end
+    λname = gensym(:__homotopy_λ)
```
ChrisRackauckas: gensym will break precompilation. Instead use a hard to hit sentinal with unicodes.

NEWS.md
```diff
@@ -1,3 +1,22 @@
+# Unreleased
```
ChrisRackauckas: ? it'll release right away? Why write that here?

ChrisRackauckas:
Do one PR first with just the HomotopyProblem constructor and direct interface, and the AbstractNonlinearProblem stuff. Then do a follow up with initialization.

docs/src/basics/Homotopy.md
```diff
@@ -0,0 +1,167 @@
...
+assumptions but hard to solve from a cold start. The `homotopy` operator encodes
+both the actual expression and a simplified approximation in a single annotation:
+
+```julia
+homotopy(actual, simplified)
+```
+
+During initialization the solver may start from the easy `simplified` form and
+continuously deform it into `actual`; outside initialization the operator
+evaluates to `actual`, as the Modelica specification prescribes.
```
ChrisRackauckas: It should be more clear. The `simplified` equations are a set of equations for which the set of nonlinear equations is easier to get a convergent (Newton) iteration for, and the `actual` equations are the more complex equations you want to actually solve for. Homotopy solvers start by solving the nonlinear system with the `simplified` equations, and uses that solution as a starting point to solve the `actual` problem. There are many ways this can be used. For example if you have equations that have multiple solutions, like a quadratic equation with positive and negative solutions, you can simplify it down to an approximating linear problem which has one problem (the positive solution), and then deform it to the actual in order to stabilize the process of getting that positive solution.

ChrisRackauckas: This also highlights that your implementation of homotopy has a bug since it doesn't start at lambda=0 😅

BatchZero:
Restructured into PR 1 of 2 per your review. Force-pushed as a single commit (`8cea8baa`) and rewrote the description. The solver stack went upstream first and is registered: `SciMLBase.HomotopyProblem` (#1375, v3.19.0 — residual `f(u, p, λ)`, λ a trailing argument, never an entry of `p`) and `NonlinearSolveBase.HomotopySweep` (#962, v2.31.0). This PR is an independent lowering pass targeting that MTK-independent API — lowering + targeting + tests + one docs page, no solver code.

**Addressed here (PR A):**
- *"Do one PR first with just the HomotopyProblem constructor and direct interface… then a follow up with initialization."* — This PR is exactly that: the operator, the lowering pass, `SciMLBase.HomotopyProblem(sys, op)`, and the automatic selection. Initialization wiring is the follow-up (PR B).
- *"It's just a nonlinear solver. Why even speak to initialization here?"* — The docs page and the operator docstring are reframed around general nonlinear solving; the direct construction is the headline, initialization is mentioned as one consumer.
- *"There should be an AbstractNonlinearProblem constructor that is automatic"* / *"split it to the AbstractNonlinearProblem constructor."* — Added `SciMLBase.AbstractNonlinearProblem(sys, op)`: returns a `HomotopyProblem` when `sys` carries `homotopy` nodes, a `NonlinearProblem` otherwise (`get_nonlinear_problem_type`); the explicit `HomotopyProblem(sys, op)` is also provided. Detection + lowering + wrap now live in the constructor, not inside `InitializationProblem`.
- *the mirrored "?" codegen* — `generate_rhs` gained an additive `extra_args::Tuple = ()`; the swept residual threads λ through it, so the second copy and its "keep in sync" comment are gone. Byte-identical for the 9 existing `generate_rhs` callers (`extra_args` empty ⇒ `p_end` not forwarded ⇒ same `build_function_wrapper` call).
- *`gensym` breaks precompilation* — λ is now a fixed reserved unicode sentinel `__homotopy_λₘₜₖ` (matching MTK's other `ₘₜₖ` internals), so codegen is a pure function of the system and the RuntimeGeneratedFunctions Expr-hash cache hits; a cheap guard turns the otherwise-silent user-symbol collision into an error.
- *"Recommend broadcast."* — Documented and tested: `homotopy.(actual, simplified)` makes one node per element, all sharing the single λ.
- *NEWS "# Unreleased"* — No NEWS.md change in this PR.

**Deferred to the initialization follow-up (PR B):** the `Val{true}` auto-injection you flagged as a bug, the initialization-specific doc bullets (trivial-init, failure-surfaces — not in this PR's general-solver docs), routing `InitializationProblem` through the constructor, and the hard `NonlinearSolveBase` dependency that the default-`initializealg` injection needs.

**Separate, in NonlinearSolve:** *"It doesn't start at lambda=0… that's a bug."* — Confirmed, and it's worse than robustness: the adaptive bisection only triggers on an inner-solve failure, so a bad `u0` that converges on the wrong branch returns a success retcode for the wrong root (`H = (1−λ)(u−4) + λ(u²−4)` from `u0 = −10` returns `−2`, not `+2`). The fix is an anchor solve at `λ = λspan[1]` in `HomotopySweep` — entirely in NonlinearSolve, now up as SciML/NonlinearSolve.jl#972 (a follow-up to #962); MTK is unchanged (the docs here already describe the sweep starting at λ = 0).

**Deferred further:** *"should be split out to the SCC."* — The structure supports a per-SCC sweep (`SCCNonlinearProblem.probs` has no element-type constraint, and `HomotopyProblem` subproblems already route to `HomotopySweep`), but it needs a small NonlinearSolve enabler first (`probvec(::HomotopyProblem)`, a `solve(::HomotopyProblem, ::AbstractNonlinearAlgorithm)` rule, stalled-residual aggregation) plus per-SCC residual codegen — its own PR.

MTKBase tests green locally. Expected CI reds, pre-attributed: codecov upload (fork token), the repo-wide Runic `FormatCheck` (red on master; every line this PR touches is Runic-clean), and the usual downstream flakes.

BatchZero:
@ChrisRackauckas a nudge on this one. The split you asked for is done: this is now PR 1 — the `HomotopyProblem` constructor plus the automatic `AbstractNonlinearProblem(sys, op)` selector, with initialization moved to the follow-up — and the per-comment mapping is in the comment above (Jun 13). It's a standalone lowering pass targeting the already-registered API (SciMLBase 3.19 `HomotopyProblem`, NonlinearSolveBase 2.31 `HomotopySweep`), no solver code of its own, so there's no upstream dependency left to wait on. Happy to rebase onto current master whenever you have a window.

ChrisRackauckas (approved):

lib/ModelingToolkitBase/src/systems/homotopy_operator.jl
```diff
@@ -0,0 +1,307 @@
...
+diverges because `atan` saturates. Annotating with `simplified = y` (root at
+`y = 0`) lets the continuation walk to the true root:
+
+```julia
+using ModelingToolkit, NonlinearSolve
+
+@variables y
+@mtkcompile sys = System([0 ~ homotopy(atan(y - 3), y)])
+prob = HomotopyProblem(sys, [y => 12.0])
+sol = solve(prob, HomotopySweep())
```
ChrisRackauckas: It should show using the default solver. HomotopySweep is likely not a good choice.

docs/src/basics/Homotopy.md
```diff
@@ -0,0 +1,159 @@
...
+# Explicit: always returns a HomotopyProblem (errors if `sys` has no `homotopy`).
+prob = HomotopyProblem(sys, op)
+
+# Automatic: returns a HomotopyProblem when `sys` contains `homotopy` nodes, and
+# a plain NonlinearProblem otherwise.
+prob = AbstractNonlinearProblem(sys, op)
+```
+
+The `HomotopyProblem`'s `λspan` defaults to `(0.0, 1.0)`. It is solved by
+`NonlinearSolveBase.HomotopySweep`, a natural-parameter continuation solver that
```
ChrisRackauckas: Not necessarily.

BatchZero:
Both addressed in f56da9c9, docs-only:

- The two examples (docstring + docs page) now show `solve(prob)` with the default solver instead of an explicit `HomotopySweep()`. The no-algorithm default dispatch already exists in NonlinearSolveBase, so behavior is unchanged.
- The docs no longer claim a `HomotopyProblem` *is* solved by `HomotopySweep` — it can be solved by any algorithm that supports the problem type; the sweep is only what the current default resolves to, and the text now says exactly that.

ChrisRackauckas:
Thank you!

PR #4610, opened 2026-06-09
Canonical CI: grouped-tests.yml + root test/test_groups.toml

ChrisRackauckas-Claude:
## Summary

Converts the root `Tests.yml` to the canonical thin caller of the reusable `SciML/.github` `grouped-tests.yml@v1` workflow, moving the previously hand-maintained group × version matrix into a root `test/test_groups.toml`.

- **`.github/workflows/Tests.yml`** — replace the inline matrix test job with a thin caller to `grouped-tests.yml@v1` (`secrets: inherit`). `on:` and `concurrency:` are preserved verbatim, filename and `name:` unchanged. The only non-default input retained is `coverage-directories: "src,ext,lib/ModelingToolkitBase/src,lib/ModelingToolkitBase/ext"`. (The old `self-hosted: ${{ vars.USE_SELF_HOSTED == 'true' }}` toggle is dropped; the matrix now uses the standard `ubuntu-latest` runner, configurable per-group via `runner` in `test_groups.toml`.)
- **`test/test_groups.toml`** (new) — declares the 8 existing groups (`InterfaceI`, `InterfaceII`, `Initialization`, `SymbolicIndexingInterface`, `Extensions`, `Optimization`, `Downstream`, `FMI`) on `["lts","1","pre"]`, plus a newly-wired `QA` group on `["lts","1"]`. Linux-only (no `os` axis).
- **`test/runtests.jl`** — add `GROUP == "QA"` dispatch with a `qa` env activation running `test/qa/qa.jl`.
- **`test/qa/`** (new) — isolated env (`Aqua` + `JET` + `Test` + `ModelingToolkit` via `[sources] path="../.."`); `qa.jl` runs `Aqua.test_all(ModelingToolkit)` and `JET.test_package(ModelingToolkit; target_defined_modules=true)`.
- **`Project.toml`** — benign metadata fixes: bump `[compat] julia` `1.9` → `1.10` (LTS floor; also clears Aqua's `project_extras` `<=1.1` check), and add `[compat]` entries for all 22 `[extras]` deps that lacked one (Aqua `deps_compat`).

## Matrix match

`scripts/compute_affected_sublibraries.jl . --root-matrix` (run statically against the new TOML) reproduces the **previous 24-job matrix exactly** (8 groups × 3 versions), verified programmatically as set-equal to the old inline matrix, **plus** the 2 new `QA` jobs (`lts`, `1`). All cells use `ubuntu-latest` (Linux-only). TOML and YAML parse-validated; Julia files syntax-checked. Tests/Aqua/JET were **not** run locally — this is a structural conversion.

## Notes

- The `QA` group is **newly wired**; Aqua/JET run in CI — any failures will be triaged in a follow-up. No test/Aqua/JET exclusions were added.


ChrisRackauckas-Claude:
Added the SciMLTesting v1.2 run_tests harness on top of this grouped-tests conversion; merges as one PR (grouped-tests + SciMLTesting v1.2). MTK's root suite is capability-based (named groups share ModelingToolkitBase/test files, no per-group folders, no Core group), so this uses v1.2 explicit-args `run_tests(groups=..., qa=..., all=...)` rather than folder discovery. Each group's body is extracted verbatim into a per-group entry file that runs the original @safetestset/@mtktestset lines unchanged; the activate_*_env helpers become run_tests env=/parent=. Behavior-preserving.

Issue #4237, opened 2026-01-28
Unhelpful initialization error


baggepinnen:
The following multibody system fails to initialize, and the error message gives no guidance on where the issue might be. It also simplifies to 14 variables of which 6 are algebraic and 8 diff, while open modelica simplifies to 8 differential variables only. 
```julia
using Test, ModelingToolkit, Multibody, Random
t = Multibody.t
using LinearAlgebra
@component function WheelInWorld(; name)
    systems = @named begin
        # world = World(n=[0,0,-1])
        world = World()
        wheel = RollingWheel(radius = 0.3, m = 2, I_axis = 0.06,
                            I_long = 0.12,
                            x0 = 0.2,
                            z0 = 0.2,
                            angles = [0,0,0],
                            der_angles = [0, -5, -1])
    end

    pars = @parameters begin
    end

    vars = @variables begin
    end

    equations = Equation[]

    return System(equations, t; name, systems)
end

@named worldwheel = WheelInWorld()

prob = ODEProblem(ssys, [], (0, 4); guesses, missing_guess_value = MissingGuessValue.Random(Random.GLOBAL_RNG))
```
```julia
julia> prob = ODEProblem(ssys, defs, (0, 4); guesses, missing_guess_value = MissingGuessValue.Random(Random.GLOBAL_RNG))
ERROR: SingularException(7)
Stacktrace:
  [1] checknonsingular
    @ ~/.julia/juliaup/julia-1.12.4+0.x64.linux.gnu/share/julia/stdlib/v1.12/LinearAlgebra/src/factorization.jl:69 [inlined]
  [2] _check_lu_success
    @ ~/.julia/juliaup/julia-1.12.4+0.x64.linux.gnu/share/julia/stdlib/v1.12/LinearAlgebra/src/lu.jl:84 [inlined]
  [3] #lu!#180
    @ ~/.julia/juliaup/julia-1.12.4+0.x64.linux.gnu/share/julia/stdlib/v1.12/LinearAlgebra/src/lu.jl:92 [inlined]
  [4] lu!
    @ ~/.julia/juliaup/julia-1.12.4+0.x64.linux.gnu/share/julia/stdlib/v1.12/LinearAlgebra/src/lu.jl:90 [inlined]
  [5] lu!
    @ ~/.julia/juliaup/julia-1.12.4+0.x64.linux.gnu/share/julia/stdlib/v1.12/LinearAlgebra/src/lu.jl:89 [inlined]
[...]

PR #4617, opened 2026-06-11
Fix TypeError from nothing-holes in MTKParameters substitution (#4607)

### LINKED ISSUE #4607: TypeError (promote_symtype on Union{Nothing,Real}) in MTKParameters when initialization is fully determined
## Summary

Building an `ODEProblem` whose **initialization system is fully determined** throws

```
TypeError: in typeassert, expected DataType, got Type{Union{Nothing, Real}}
```

from `promote_symtype(::typeof(*), …)`, when the model contains a variable whose symtype is `Union{Nothing, Real}` (i.e. a conditionally-included / `nothing`-defaulted variable). It is thrown while constructing the parameter buffer (`MTKParameters` → `evaluate_varmap!` → `fixpoint_sub`), before any solve.

This is the *fully-determined-path* sibling of #4237: the same model **left underdetermined** hits the `SingularException` of #4237 in init reconstruction; once enough initial conditions are pinned to make initialization fully determined (so the SCC path is taken), this `TypeError` fires instead.

## Stacktrace

```
promote_symtype(::typeof(*), T::DataType, S::DataType)   @ SymbolicUtils methods.jl:209
… mapfoldl(symtype, _promote_symtype(*), …)              @ Base reduce.jl
combine_fold(::Type{SymReal}, op, args, meta, can_fold)  @ SymbolicUtils substitute.jl:222
fixpoint_sub                                             @ Symbolics variable.jl:475
evaluate_varmap!(varmap, vars; limit, allow_symbolic)    @ ModelingToolkitBase problem_utils.jl:517
MTKParameters(sys, op; tofloat, t0, …)                   @ ModelingToolkitBase parameter_buffer.jl:107
```

i.e. a symbolic variable with symtype `Union{Nothing, Real}` reaches a `*` term, and `promote_symtype(*, DataType, ::Type{Union{Nothing,Real}})` fails its `T::DataType` typeassert because `Union{Nothing,Real}` is not a `DataType`.

## Where the `Union{Nothing, Real}` comes from

The model is a free-floating multibody car (Dyad port of Multibody.jl `FullCar`). The most likely source is a variable declared with a `nothing` default — e.g. a rigid `Body`'s quaternion-orientation variables, which are defaulted to `nothing` when the body uses Euler angles (`quat=false`) rather than quaternions. Such a variable carries symtype `Union{Nothing, Real}` and survives into the parameter-varmap evaluation on the fully-determined initialization path.

## To reproduce

Free-floating 4-wheel car (`MultibodyComponents.examples.suspension.FullCar`) with the initialization regularized to fully determined:

- pin the chassis rigid-body state: `back_front.body.r_0[1:3]`, `v_0[1:3]`, `phi[1:3]`, `phid[1:3]`
- pin the 4 redundant chassis reaction torques to `0` (the static-indeterminacy set; see #4237)
- `ODEProblem(ssys, defs, (0, 3.0))` → `TypeError` above (thrown during problem construction).

Pinning fewer reactions (init stays underdetermined) instead gives the `SingularException` of #4237. Note: pinning the reaction connector-flow variables *individually* builds fine — the `TypeError` only appears once enough pins make initialization fully determined and the SCC/`MTKParameters` path is taken.

The repro currently needs the `MultibodyComponents` (Dyad) package, which isn't public — happy to provide it or work toward a reduced MWE (a small `System` containing a `nothing`-defaulted variable that participates in a product, built into a fully-determined init, should suffice).

## Versions
- Julia 1.12.6
- ModelingToolkit v11.26.8 (`#master`), ModelingToolkitBase v1.43.1 (`lib/ModelingToolkitBase#master`)
- OrdinaryDiffEqNonlinearSolve v2.0.0, NonlinearSolve v4.19.1, SciMLBase v3.18.0, Symbolics v7.26.0
- DyadCompilerPasses v0.2.3

baggepinnen:
Fixes #4607.

## Root cause

The issue's stacktrace (`promote_symtype` failing a `::DataType` typeassert with `Union{Nothing, Real}`) is **not** caused by a variable with that symtype. Partially specifying an array variable (e.g. pinning a single component of a 3-vector connector flow) leaves `COMMON_NOTHING = Const(nothing)` holes for the remaining elements in the operating point (`as_atomic_dict_with_defaults` / `write_possibly_indexed_array!`); `build_operating_point` only filters values that are *entirely* `nothing`. When another operating-point entry's value expression references such a hole element, `MTKParameters` — which substituted against the raw operating point — folded `Const(nothing)` into arithmetic: `promote_symtype(*, Real, Nothing)` → `promote_type = Union{Nothing, Real}` → typeassert failure (or `MethodError: *(::Int, ::Nothing)` when all co-arguments are constant).

This only fires on the fully-determined initialization path because that is the one that builds an `SCCNonlinearProblem` for the initialization system, whose `process_SciMLProblem` → `MTKParameters(initsys, op)` evaluates all parameters (including `Initial`s) against an op containing the partially-pinned arrays. The underdetermined sibling path instead hits #4237.

## Changes

1. **`lib/ModelingToolkitBase/src/systems/parameter_buffer.jl`**: substitute through `AtomicArrayDictSubstitutionWrapper` (which exists precisely to keep holes symbolic, and is already used by every other substitution site against operating points — `problem_utils.jl:348`, `problem_utils.jl:1599`, `sccnonlinearproblem.jl:564`). Values of `Initial` parameters that consequently cannot be fully evaluated refer to variables not fixed by the operating point and are treated (elementwise) as unfixed, generalizing the existing `val == args[1] → false` special case; this matches how `timevaring_initsys_process_op!` turns non-constant pins into initialization equations rather than `Initial` values. Non-`Initial` parameters retain the informative `Could not evaluate value of parameter` error — which now also replaces the cryptic `TypeError` for genuinely missing values.

2. **`lib/ModelingToolkitBase/src/utils.jl`**: memoize the `_check_operator_variables` traversal. With the crash fixed, the repro proceeded into per-SCC `generate_rhs` and spun indefinitely (100% CPU, flat memory) in this check, which recursed into shared subexpressions repeatedly — exponential on the DAG-shaped expressions of large torn initialization systems.

3. **`src/problems/sccnonlinearproblem.jl`**: the `MissingGuessValue.HashedRandom()` branch (previously dead code, now reachable) called `stable_eachindex` on a plain `Vector{Int}` and referenced an undefined variable `var`. Fixed mirroring the equivalent branch in `problem_utils.jl`.

## Validation

- New regression testset in `lib/ModelingToolkitBase/test/mtkparameters.jl` (the two crash modes now produce the informative error; unresolvable `Initial` values build; values referencing specified elements still fold).
- `mtkparameters.jl`, `initial_values.jl`, `initializationsystem.jl` (ModelingToolkitBase) and `scc_nonlinear_problem.jl` (ModelingToolkit) test files pass locally.
- The original `FullCar` repro (Dyad `MultibodyComponents#suspension`, private — not CI-able): pinning the chassis state (`r_0`, `v_0`, `phi`, `phid`) plus the four redundant chassis reaction torques (`excited_suspension_*.chassis_frame.tau[2] => 0`) previously threw the exact `TypeError`; with this PR the `ODEProblem` **builds successfully**. `solve` subsequently hits a `SingularException` from the chassis static indeterminacy — that is the separate, pre-existing #4237 and out of scope here.

### DIFF
--- a/lib/ModelingToolkitBase/src/utils.jl
+++ b/lib/ModelingToolkitBase/src/utils.jl
@@ -1296,7 +1296,7 @@ function subexpressions_not_involving_vars!(
             # are already parameters, we don't want to cache it.
             drop = Moshi.Match.@match ir[i] begin
                 BSImpl.Term(; f) && if f isa SymbolicT end => begin
-                1
+                    1
                 end
                 _ => 0
             end
@@ -1368,6 +1368,7 @@ function subexpressions_not_involving_vars!(
         )
         state[expr] = anon_sym
     end
+    return
 end
 
 """
@@ -1646,7 +1647,7 @@ function move_variable_bindings_to_ics!(
             return true
         end
     end
-    filter!(filterer, binds)
+    return filter!(filterer, binds)
 end
 
 """
--- a/lib/ModelingToolkitBase/src/systems/parameter_buffer.jl
+++ b/lib/ModelingToolkitBase/src/systems/parameter_buffer.jl
@@ -104,7 +104,10 @@ function MTKParameters(
         op[get_iv(sys)] = t0
     end
 
-    evaluate_varmap!(op, all_ps; limit = substitution_limit)
+    # Substitute through `AADSubWrapper` so that `COMMON_NOTHING` holes in partially
+    # specified array values are never substituted into expressions (issue #4607).
+    wrapped_op = AADSubWrapper(op)
+    evaluate_varmap!(wrapped_op, all_ps; limit = substitution_limit)
 
     tunable_buffer = Vector{ic.tunable_buffer_size.type}(
         undef, ic.tunable_buffer_size.length
@@ -167,20 +170,20 @@ function MTKParameters(
 
     for sym in all_ps
         haskey(diffcache_params, sym) && continue
-        val = fixpoint_sub(sym, op; maxiters = max(div(substitution_limit, 2), 2), fold = Val(true))
+        val = fixpoint_sub(sym, wrapped_op; maxiters = max(div(substitution_limit, 2), 2), fold = Val(true))
         ctype = symtype(sym)
-        if !SU.isconst(val)
-            Moshi.Match.@match sym begin
-                BSImpl.Term(; f, args) && if f isa Initial end => begin
-                    if isequal(val, args[1])
-                        if Symbolics.isarraysymbolic(sym)
-                            val = BSImpl.Const{VartypeT}(fill(false, size(val)))
-                        else
-                            val = BSImpl.Const{VartypeT}(false)
-                        end
-                    end
-                end
-                _ => nothing
+        if !SU.isconst(val) && isinitial(sym)
+            # The value of an `Initial` parameter that cannot be fully evaluated
+            # refers to variables not fixed by the operating point. Treat such
+            # (elements of) the value as unfixed, mirroring the `COMMON_NOTHING`
+            # handling below.
+            if SU.is_array_shape(SU.shape(val))
+                val = BSImpl.Const{VartypeT}(map(SU.stable_eachindex(val)) do i
+                    el = val[i]
+                    SU.isconst(el) ? unwrap_const(el) : false
+                end)
+            else
+                val = BSImpl.Const{VartypeT}(false)
             end
         end
         if !SU.isconst(val)
--- a/lib/ModelingToolkitBase/src/systems/parameter_buffer.jl
+++ b/lib/ModelingToolkitBase/src/systems/parameter_buffer.jl
@@ -178,10 +178,12 @@ function MTKParameters(
             # (elements of) the value as unfixed, mirroring the `COMMON_NOTHING`
             # handling below.
             if SU.is_array_shape(SU.shape(val))
-                val = BSImpl.Const{VartypeT}(map(SU.stable_eachindex(val)) do i
-                    el = val[i]
-                    SU.isconst(el) ? unwrap_const(el) : false
-                end)
+                val = BSImpl.Const{VartypeT}(
+                    map(SU.stable_eachindex(val)) do i
+                        el = val[i]
+                        SU.isconst(el) ? unwrap_const(el) : false
+                    end
+                )
             else
                 val = BSImpl.Const{VartypeT}(false)
             end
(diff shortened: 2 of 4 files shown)


lib/ModelingToolkitBase/src/utils.jl
```diff
@@ -559,13 +559,17 @@ end
 """
 Check if difference/derivative operation occurs in the R.H.S. of an equation
 """
-function _check_operator_variables(eq, op::T, expr = eq.rhs) where {T}
+function _check_operator_variables(eq, op::T, expr = eq.rhs, visited = Set{SymbolicT}()) where {T}
```
KristofferC: Ideally, this would be broken out into a separate PR and could be quickly merged. 

Also, I think having this as an `IdSet` make sense. And you can create the set in `check_operator_variables` and empty it for every equation to avoid having to create it from scratch over and over.

baggepinnen: Done in 6101f16 — `visited` is now a `Base.IdSet{Any}` allocated once in `check_operator_variables` and `empty!`-ed per equation (mirroring the existing `tmp` set), instead of a fresh `Set` per `_check_operator_variables` call. `IdSet` is also the right semantics here: identity membership is correct and cheaper for the hash-consed expression DAGs. Verified MTKBase precompiles + loads.

On breaking it out: glad to split the `_check_operator_variables` memoization into its own PR for a quick merge if you'd prefer — just say the word.

KristofferC: I had https://github.com/SciML/ModelingToolkit.jl/pull/4622 already locally so pushed that

baggepinnen: Broken out into #4623 (just the `_check_operator_variables` memoization, off `master`, with the `IdSet`/per-equation-`empty!` form). I'll rebase this PR on `master` once #4623 merges so the overlap drops out of the #4607 fix.

KristofferC: Ok.. so I should close #4622 then..?

AayushSabharwal (approved):
Should be good after a rebase

baggepinnen:
Rebased onto latest `master`. The memoization commit conflicted because that work landed on `master` separately (fd049e137 "cache expression checks in `check_operator_variables`"), so I dropped the now-redundant commits and resolved `_check_operator_variables` to master's version. The branch now contains just the nothing-holes fix, the `SCCNonlinearProblem` MissingGuessValue fix, and Runic formatting.

PR #4622, opened 2026-06-13
cache expression checks in `check_operator_variables`

KristofferC:
Broken out from https://github.com/SciML/ModelingToolkit.jl/pull/4617.


AayushSabharwal:
Duplicate of #4623? CI seems much happier here (by comparison) so I'll check and merge this one instead if possible

PR #4623, opened 2026-06-13
Memoize `_check_operator_variables` to avoid exponential DAG traversal

baggepinnen:
`_check_operator_variables` recursively walks an equation's RHS to reject forbidden operators. Expressions are hash-consed DAGs with heavy subexpression sharing, so the unmemoized recursion is exponential in the sharing depth and stalls `check_operator_variables` on large systems.

This threads a `Base.IdSet` of visited nodes through the recursion — identity membership is correct here (shared subexpressions are the same object) and cheap (no hashing of large expressions) — allocated once in `check_operator_variables` and `empty!`-ed per equation, mirroring the existing `tmp` set.

Split out of #4617 per review so it can merge independently; #4617 will rebase once this lands.

### DIFF
--- a/lib/ModelingToolkitBase/src/utils.jl
+++ b/lib/ModelingToolkitBase/src/utils.jl
@@ -559,13 +559,17 @@ end
 """
 Check if difference/derivative operation occurs in the R.H.S. of an equation
 """
-function _check_operator_variables(eq, op::T, expr = eq.rhs) where {T}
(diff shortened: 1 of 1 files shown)


AayushSabharwal:
Most of CI fails, and it might be codecov but github doesn't want to show me CI logs so I'm retriggering.

AayushSabharwal:
`Tests` jobs are all codecov

AayushSabharwal:
Closing as duplicate of #4622

PR #4631, opened 2026-06-16
refactor: ensure init system is in terms of unknowns of DAE

AayushSabharwal:
This helps the unknowns of the simplified initialization system be more predictable.


AayushSabharwal:
Requires #4659

PR #4640, opened 2026-06-18
refactor: rerun `eliminate_perfect_aliases!` after `alias_elimination!`

AayushSabharwal:
Alias elimination sometimes transforms equations into `0 ~ x - y` form, which `eliminate_perfect_aliases!` can subsequently tackle. This prevents pantelides from differentiating such equations.

### DIFF
--- a/src/systems/alias_elimination.jl
+++ b/src/systems/alias_elimination.jl
@@ -147,7 +147,6 @@ function find_perfect_aliases!(
     (; sys, fullvars, structure) = state
     (; graph, solvable_graph, var_to_diff, state_priorities) = structure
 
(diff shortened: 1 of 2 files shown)


AayushSabharwal:
Partially supersedes https://github.com/JuliaComputing/StateSelection.jl/pull/103/

oscardssmith:
why does alias elimination leave things in the form `0 ~ x - y`? isn't that a bug in alias elimination?

AayushSabharwal:
Alias elimination doesn't eliminate anything. It just runs bareiss.

oscardssmith:
then this seems good to me.

PR #4655, opened 2026-06-22
feat: add reversible transformation API

AayushSabharwal:
This allows a generic API for reversible model transformations, since we frequently need to perform such operations and reverse them. Recently, we need to selectively reverse some transformations, and hence this design.


AayushSabharwal:
The enzyme failure in SciMLSensitivity/Core8 does not reproduce locally, despite reproducing across two CI runs. It will be investigated locally and marked as broken in the meantime.

PR #4668, opened 2026-06-25
refactor: store `mm` in `TearingState`

AayushSabharwal:
This avoids the `mm` being stale by the time it gets to reassembly.


AayushSabharwal:
Somehow none of the assertion failures reproduce locally

PR #4681, opened 2026-06-29
Init test: accept NonlinearLeastSquaresProblem fallback for HydraulicSystem

### LINKED ISSUE #4680: Initialization test: HydraulicSystem returns NonlinearLeastSquaresProblem instead of SCCNonlinearProblem{LinearProblem}
## Summary

[`lib/ModelingToolkitBase/test/initializationsystem.jl:335-337`](https://github.com/SciML/ModelingToolkit.jl/blob/master/lib/ModelingToolkitBase/test/initializationsystem.jl#L335-L337) fails on `ModelingToolkitBase v1.48.1`:

```julia
if !@isdefined(ModelingToolkit)
    @test initprob isa SCCNonlinearProblem
    @test initprob.probs isa Tuple{<:LinearProblem}
end
```

The first assertion fails (`initprob` is `NonlinearLeastSquaresProblem`, not `SCCNonlinearProblem`), and the second then errors with `type NonlinearLeastSquaresProblem has no field probs`.

This is the `HydraulicSystem` block at [line 326](https://github.com/SciML/ModelingToolkit.jl/blob/master/lib/ModelingToolkitBase/test/initializationsystem.jl#L326):

```julia
@mtkcompile sys = HydraulicSystem()
initprob = ModelingToolkitBase.InitializationProblem(sys, 0.0; missing_guess_value)
```

The comment immediately above the failing assertions already flags this as a known weak point:

> The init system as-is is linear. Since it is so severely overdetermined, MTK can sometimes choose a bad simplification that results in a nonlinear system.

So `mtkcompile`'s simplification on this severely overdetermined linear system is currently picking the "bad" branch and producing a `NonlinearLeastSquaresProblem`. The TODO right below points at the root: `mtkcompile` needs to retain `observed` equations and generate `x ~ Initial(x)` as such.

## Reproduction

The full HydraulicSystem fixture and `missing_guess_value` constant are in the same file (lines 13-326 of `initializationsystem.jl` at v1.48.1).

## Impact

This failure is showing up in downstream CI — every PR against `SciML/OrdinaryDiffEq.jl` is currently red on `ModelingToolkit.jl/Initialization/1.10`, including `master` itself (see for example [this OrdinaryDiffEq master IntegrationTest run](https://github.com/SciML/OrdinaryDiffEq.jl/actions/runs/28395223448/job/84132150007)). Test summary in CI: `777 passed, 1 failed, 1 errored, 14 broken`.

## Suggested fix

Either:
1. Land the TODO'd `mtkcompile` change so `observed` is retained and the linear simplification holds, OR
2. Loosen the assertion to accept `NonlinearLeastSquaresProblem` as an acceptable (but suboptimal) outcome until the `mtkcompile` heuristic is improved.

singhharsh1708:
## Summary

Fixes #4680.

The HydraulicSystem initialization test at [lib/ModelingToolkitBase/test/initializationsystem.jl:335-337](https://github.com/SciML/ModelingToolkit.jl/blob/master/lib/ModelingToolkitBase/test/initializationsystem.jl#L335-L337) asserts `initprob isa SCCNonlinearProblem` with `Tuple{<:LinearProblem}` children, but `mtkcompile`'s simplification on this severely overdetermined linear system currently picks the nonlinear branch and produces a `NonlinearLeastSquaresProblem`. The in-source comment immediately above the assertion already flags this as a known weak point with a TODO for retaining `observed` equations in `mtkcompile`.

This PR loosens the assertion to accept the fallback type, while preserving the `LinearProblem`-tuple check when `mtkcompile` does pick the good branch. The semantic test (solve succeeds, conditions satisfied) in the `@isdefined(ModelingToolkit)` branch immediately below is untouched.

## Test plan
- [x] When `mtkcompile` produces an `SCCNonlinearProblem`, the inner `Tuple{<:LinearProblem}` assertion still runs.
- [x] When `mtkcompile` falls back to `NonlinearLeastSquaresProblem` (the currently-observed behavior on master), the Union assertion passes and the `.probs` access is skipped.
- [x] Unblocks downstream CI on SciML/OrdinaryDiffEq.jl, which currently hits this on every PR including master itself.

### DIFF
--- a/lib/ModelingToolkitBase/test/initializationsystem.jl
+++ b/lib/ModelingToolkitBase/test/initializationsystem.jl
@@ -332,8 +332,10 @@ conditions = getfield.(equations(initprob.f.sys), :rhs)
 # `x ~ Initial(x)` as such `observed` equations. This makes `mtkcompile`'s job easier
 # on initialization systems, but can lead to parameter-only equations.
 if !@isdefined(ModelingToolkit)
-    @test initprob isa SCCNonlinearProblem
-    @test initprob.probs isa Tuple{<:LinearProblem}
+    @test initprob isa Union{SCCNonlinearProblem, NonlinearLeastSquaresProblem}
+    if initprob isa SCCNonlinearProblem
+        @test initprob.probs isa Tuple{<:LinearProblem}
+    end
 end
 if @isdefined(ModelingToolkit)
     initsol = solve(initprob, reltol = 1.0e-12, abstol = 1.0e-12)


AayushSabharwal:
Thanks!

Issue #62, opened 2018-04-25
Identity crisis: performance with context


ChrisRackauckas:
This library hit an interesting point. We have all of the features to recreate things like `@ode_def`, including computation of things like the inverse of W = I - gamma*J where J is the Jacobian. 

But at the same time I have to admit defeat. The reason is because the computations are getting slow. Now, inverting matrices in computational graphs is quite a stress test so that's okay, but it's taking 10's of seconds instead of the about a second it took with ParameterizedFunctions.jl and SymEngine.jl, so I am afraid the current architecture will not scale well at all.

The issue goes back to our discussion and hope for context. We want to know a ton of information about our variables, but then actually keeping that information inside of the computational graph is too much. We wanted it so that way properties could be well-defined, but then what has ended up happening is that the context is used to gather the variables into specific arrays in the system's construction, and then only used from those arrays. So the computational graphs for the equations holds information it doesn't use only to automate the finding of variables, which in many cases is overridden so that way the variables end up in a different order than what the automated finder gives (since it's not all that great...). 

Now let me be clear. Graph construction is cheap, but graph operations are bad.

I think we have a clear path around this though. The true format for the equations should be Julia AST earlier. Definitely operations on the graph should be done on Julia ASTs (something @YingboMa was considering before), but the question is how far back do we go. 

1. Do we have the user give us vectors of expressions and then vectors of the variables for each time, more manually building the system?
2. During system construction, do we turn the eqs into Julia AST?

If we do 1, the interface changes quite a bit. If we do 2, we can keep the same user interface and switch up the internals to make it more optimized just for this goal. It would also let us keep the automation. I am inclined to go with route 2. 

Either way, we would need to re-write our graph computations. But since they are all in Julia ASTs we can better make use of some existing tooling. I am considering using Reduce.jl here, though the only issue is Windows packaging which I've discussed with @chakravala. That would give us simplification, differentiation, etc. routines for free and it would be better than what we had. Thus, from route 2, we can keep the same user-facing tooling but redo the internal computation part.

With a Julia AST, we will need to enforce that the symbols used for the names are unique. That's one thing to note. We can make the AST use componentized names, i.e. `comp1.a` would be the name of the variable in component 1 and that would solve https://github.com/JuliaDiffEq/ModelingToolkit.jl/issues/36 . It is actually a valid Julia symbol: `x = Symbol("comp.a")`. Building the variable vectors would then require doing this kind of renaming.

[...]

PR #4701, opened 2026-07-02
MTKBase: FunctionProperties extension exempting MTKParameters indexing

ChrisRackauckas-Claude:
> **Please ignore until reviewed by @ChrisRackauckas.** Draft opened by an agent. **Stacked on SciML/FunctionProperties.jl#62** — needs the FunctionProperties 0.1.7 release to resolve (compat is pinned to it). Companion to SciML/SciMLBase.jl#1413.

## Summary

`hasbranching` (FunctionProperties.jl) decides whether a ReverseDiff tape can be compiled for an ODE RHS. For a **split** (default) MTK system the parameter object is an `MTKParameters`, and the generated RHS reads parameters through `getindex(::MTKParameters, ::Int)` — a `@generated` `if idx == k … elseif …` chain.

That branch is on the **integer index**, which is a literal constant at every real call site (so it constant-folds away in the compiled RHS), but the `hasbranching` recursion only sees the widened `Int` and reports it. Net effect: even a branch-free split-mode RHS such as `ẋ = a·x` is flagged as branching, needlessly disabling ReverseDiff compilation.

## Change (`lib/ModelingToolkitBase`)

A weakdep extension marking the indexing as value-independent plumbing via the new `is_leaf_sig` hook from FunctionProperties#62:

```julia
FunctionProperties.is_leaf_sig(
    ::Type{<:Tuple{typeof(getindex), <:MTKParameters, Vararg}}) = true
```

Loaded only when FunctionProperties is present — **no new hard dependency**. Follows the existing `ext/MTK*Ext.jl` convention.

## Tests
`test/extensions/function_properties.jl` (Extensions group): builds a split-mode `ODEProblem`, confirms `prob.p isa MTKParameters`, that the `is_leaf_sig` exemption is active for `MTKParameters` indexing (and does not over-reach to ordinary `Vector`), and that a function which only indexes the parameter container has no value-dependent branch. Verified locally (4/4) by loading the extension against a real `MTKParameters`. Runic clean.

## Stack
- SciML/FunctionProperties.jl#62 — `is_leaf_sig` + value-independent-branch detection (foundation).
- SciML/SciMLBase.jl#1413 — unwrap `AbstractSciMLFunction` to its RHS.
- This PR — exempt `MTKParameters` integer indexing, completing the split-mode case.

## Semver
Additive weakdep extension, non-breaking: MTKBase `1.48.1` → `1.48.2`.


AayushSabharwal:
The optimized IR should have all of this constant-folded, right? Why does FunctionProperties pass `optimize = false`?

ChrisRackauckas-Claude:
Superseded. FunctionProperties #62 now folds the value-independent `getindex(::MTKParameters, ::Int)` branch generically via constant-propagation-aware analysis (the `is_leaf_sig` hook this extension used has been removed), so no MTK-side override is needed. Closing.

PR #4704, opened 2026-07-03
fix: properly eliminate identically zero variables in `alias_elimination!`

AayushSabharwal:
Stacked on top of https://github.com/SciML/ModelingToolkit.jl/pull/4703
Requires https://github.com/JuliaComputing/StateSelection.jl/pull/125

After bareiss, the matrix often contains rows with a single non-zero entry. This means the corresponding variable is identically zero. However, the pass never eliminated such variables and never included them in the `maybe_zeros` heuristic. Thus, MTK was fine with dividing by them if present in a linear SCC it can analytically solve. 

These zero variables are now discovered after the pass, eagerly removed, and substituted into all equations where they are present. The above two PRs are required infrastructure to handle the case when a differentiated variable is found to analytically be zero. In this case, the variable, all its derivatives and all integrated forms are eliminated as observed. Lower order derivatives are analytically obtained as a polynomial in `t`. The infrastructure is mostly to make sure initialization behaves correctly.

TODO:
- [ ] Write tests


baggepinnen:
> Lower order derivatives are analytically obtained as a polynomial in t

anti-derivatives?

AayushSabharwal:
I _knew_ there was a term I was forgetting 😅 Thanks!

PR #4716, opened 2026-07-08
Add ForwardDiff dual initialization regression test

ChrisRackauckas-Claude:

## Summary
- Adds a regression test for initialization remakes after ForwardDiff-tagged symbolic constants have entered SymbolicUtils hashconsing state.
- Removes the previous `unsafe = true` workaround; the actual hashconsing invariant is handled by JuliaSymbolics/SymbolicUtils.jl#1005.

## Dependency
This PR is expected to need SymbolicUtils.jl#1005, or a SymbolicUtils release containing that fix, before CI can pass without the old workaround. The failure mode is `Cannot determine ordering of Dual tags ForwardDiff.Tag{...} and Nothing` during symbolic constant hashconsing.

## Verification
- `timeout 3600 /home/crackauc/.juliaup/bin/julia +1.12 --startup-file=no --project=/home/crackauc/sandbox/tmp_20260708_165240_11300/.runic-env -m Runic --check lib/ModelingToolkitBase/src/systems/problem_utils.jl lib/ModelingToolkitBase/src/systems/nonlinear/initializesystem.jl lib/ModelingToolkitBase/test/initializationsystem.jl`
- `git diff --check master...HEAD`
- With SymbolicUtils.jl#1005 developed locally into the MTK test environment:
  `timeout 3600 env GROUP=Initialization /home/crackauc/.juliaup/bin/julia +1.12 --startup-file=no --project=. -e 'using Pkg; Pkg.develop(PackageSpec(path="/home/crackauc/sandbox/tmp_20260708_165240_11300/SymbolicUtils-dual-isequal")); Pkg.status(["SymbolicUtils"]); Pkg.test(coverage=false)'`
  Result: `Initialization | 794 pass, 12 broken, 806 total`; `Testing ModelingToolkit tests passed`.

### DIFF
--- a/lib/ModelingToolkitBase/src/systems/nonlinear/initializesystem.jl
+++ b/lib/ModelingToolkitBase/src/systems/nonlinear/initializesystem.jl
@@ -660,7 +660,7 @@ function _remake_initialization_data_impl(
         u0 = to_varmap(u0, dvs)
         symbols_to_symbolics!(sys, u0)
     end
-    u0map = as_atomic_dict_with_defaults(Dict{SymbolicT, SymbolicT}(u0), COMMON_NOTHING)
+    u0map = as_atomic_dict_with_defaults(_symbolic_varmap_dict(u0), COMMON_NOTHING)
     if eltype(p) <: Pair
         if p isa Union{AbstractArray, Tuple}
             p = Dict(p)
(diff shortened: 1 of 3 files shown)


AayushSabharwal:
`unsafe = true` is evil and should not be used outside of SU. If this unsafe constant makes it inside another symbolic expression it can cause untold havoc.

ChrisRackauckas-Claude:
Updated after the `unsafe = true` review concern. The MTK branch now removes the unsafe workaround entirely; the net MTK diff is just the regression test. The actual hashconsing fix is in JuliaSymbolics/SymbolicUtils.jl#1005, and I verified `GROUP=Initialization` locally with that SymbolicUtils checkout developed into the MTK test env: `Initialization | 794 pass, 12 broken, 806 total`.

PR #4720, opened 2026-07-10
Add Symbolic Analysis of MTK System on Exiting Integration

Shreyas-Ekanathan:
## Checklist

## Additional context

Uses interface created in https://github.com/SciML/SciMLBase.jl/pull/1434 and works with https://github.com/SciML/OrdinaryDiffEq.jl/pull/3731/#commits-pushed-17dad24 to analyze causes of instability


lib/ModelingToolkitBase/src/debugging.jl
```diff
@@ -99,3 +99,121 @@ function get_assertions_expr(sys::AbstractSystem)
     end
     return term
 end
+
+function SciMLBase.diagnose_symbolic_instability(integrator::SciMLBase.DEIntegrator)
```
AayushSabharwal: This is type piracy. MTK owns neither the function `diagnose_symbolic_instability` or the type `DEIntegrator`. Can this use something like `get_root_indp` in SciMLBase to obtain the system (if present) and dispatch on that?

AayushSabharwal (changes requested):
Apart from the piracy, it might be helpful to have a way to disable this as a `solve` kwarg or `ODEProblem` flag. Places that run `solve` in a hot loop (e.g. parameter optimization) typically want to avoid any sort of symbolic manipulation inside the loop. Alternatively, do we really need to identify the exact failing boolean expression? The user already has provided assertions in a way that makes sense for them so why does MTK need to do the extra effort?

lib/ModelingToolkitBase/src/debugging.jl
```diff
@@ -99,3 +99,121 @@ function get_assertions_expr(sys::AbstractSystem)
...
+end
+
+function find_singular_subterms(eq, expr, sub_map, diagnosis)
+    expr = Symbolics.unwrap(expr)
+    !SymbolicUtils.iscall(expr) && return diagnosis
+    op = SymbolicUtils.operation(expr)
+    args = SymbolicUtils.arguments(expr)
+
+    if op === (/) #division, singular if we divide by small thing
+        d = Symbolics.value(Symbolics.substitute(args[2], sub_map))
```
AayushSabharwal:
```suggestion
        d = Symbolics.value(SymbolicUtils.substitute(args[2], sub_map))
```

lib/ModelingToolkitBase/src/debugging.jl
```diff
@@ -99,3 +99,121 @@ function get_assertions_expr(sys::AbstractSystem)
...
+    if !isempty(singularities)
+        push!(diagnosis, "\nSymbolic Analysis of MTK System:")
+        append!(diagnosis, singularities)
+    end
+
+    return isempty(diagnosis) ? "" : join(diagnosis, "\n")
+end
+
+function find_singular_subterms(eq, expr, sub_map, diagnosis)
+    expr = Symbolics.unwrap(expr)
```
AayushSabharwal:
```suggestion
    expr = unwrap(expr)
```
It's a function in SU, and available here without qualifiers.

lib/ModelingToolkitBase/src/debugging.jl
```diff
@@ -99,3 +99,121 @@ function get_assertions_expr(sys::AbstractSystem)
...
+    end
+    if !isempty(singularities)
+        push!(diagnosis, "\nSymbolic Analysis of MTK System:")
+        append!(diagnosis, singularities)
+    end
+
+    return isempty(diagnosis) ? "" : join(diagnosis, "\n")
+end
+
+function find_singular_subterms(eq, expr, sub_map, diagnosis)
```
AayushSabharwal: If we need to substitute with the same `sub_map` over and over, it's a lot faster to build `subber = SymbolicUtils.IRSubstituter{true}(get_irstructure(sys), sub_map)` and call `subber(expr)` instead of `substitute`.

lib/ModelingToolkitBase/src/debugging.jl
```diff
@@ -99,3 +99,121 @@ function get_assertions_expr(sys::AbstractSystem)
...
+        find_failing_subterms(cond, prev_substitution_map, curr_substitution_map, subclauses)
+        if !isempty(subclauses)
+            push!(diagnosis, "\n\nAssertion violated: $cond - \"$msg\"")
+            append!(diagnosis, subclauses)
+        end
+    end
+
+    #find singularity causes in equations
+    singularities = String[]
+    for eq in equations(sys)
```
AayushSabharwal: We might want to run this on `full_equations`? `equations` are often trivial and we actually codegen `full_equations`. This would require using `IRSubstituter` to scale to the massive expressions Multibody has.

AayushSabharwal: If we do want to scale to `full_equations`, `find_singular_subterms` should also maintain an `IdSet` of expressions it has recursed into, so that it avoids recursing into the same subexpressions repeatedly. An example would be the same observed quantity substituted into multiple equations - it only needs to be reported and processed once.

AayushSabharwal (commented):
Just some performance/scaling comments.

lib/ModelingToolkitBase/src/debugging.jl
Shreyas-Ekanathan: Should be addressed now, if you could take a look and lmk if anything else stands out that would be great!

lib/ModelingToolkitBase/src/debugging.jl
```diff
@@ -99,3 +99,122 @@ function get_assertions_expr(sys::AbstractSystem)
...
     return term
 end
+
+function SciMLBase.diagnose_symbolic_instability(sys::AbstractSystem, u, uprev)
+    diagnosis = String[]
+
+    #check for assertion failures
+    unks = unknowns(sys)
+    curr_substitution_map = Dict(zip(unks, Symbolics.unwrap.(Symbolics.Num.(u))))
+    prev_substitution_map = Dict(zip(unknowns(sys), Symbolics.unwrap.(Symbolics.Num.(uprev))))
```
AayushSabharwal:
```suggestion
    curr_substitution_map = Dict(zip(unks, u))
    prev_substitution_map = Dict(zip(unknowns(sys), uprev))
```

`unwrap(Num(x))` is basically an identity

lib/ModelingToolkitBase/src/debugging.jl
```diff
@@ -99,3 +99,122 @@ function get_assertions_expr(sys::AbstractSystem)
...
+        find_failing_subterms(cond, prev_substitution_map, curr_substitution_map, subclauses)
+        if !isempty(subclauses)
+            push!(diagnosis, "\n\nAssertion violated: $cond - \"$msg\"")
+            append!(diagnosis, subclauses)
+        end
+    end
+
+    #find singularity causes in equations
+    singularities = String[]
+    visited = IdDict()
```
AayushSabharwal:
```suggestion
    visited = IdDict{SymbolicT, Nothing}()
```

This makes the container type-stable

AayushSabharwal (commented):
Just minor changes, looks good otherwise!

lib/ModelingToolkitBase/src/debugging.jl
Shreyas-Ekanathan: This throws the following error though:

`ERROR: TypeError: in IRSubstituter, in D, expected D<:AbstractDict{SymbolicUtils.BasicSymbolicImpl.var"typeof(BasicSymbolicImpl)"{SymReal}, SymbolicUtils.BasicSymbolicImpl.var"typeof(BasicSymbolicImpl)"{SymReal}}, got Type{Dict{SymbolicUtils.BasicSymbolicImpl.var"typeof(BasicSymbolicImpl)"{SymReal}, Float64}}
Stacktrace:
  [1] (SymbolicUtils.IRSubstituter{…} where T)(ir::IRStructure{…}, rules::Dict{…}; filterer::typeof(SymbolicUtils.default_substitute_filter))
    @ SymbolicUtils ~/.julia/packages/SymbolicUtils/4vS6Y/src/irstructure.jl:701
  [2] (SymbolicUtils.IRSubstituter{…} where T)(ir::IRStructure{…}, rules::Dict{…})
    @ SymbolicUtils ~/.julia/packages/SymbolicUtils/4vS6Y/src/irstructure.jl:698
  [3] diagnose_symbolic_instability(sys::System, u::Vector{Float64}, uprev::Vector{Float64})
    @ ModelingToolkitBase ~/Documents/GitHub/ModelingToolkit.jl/lib/ModelingToolkitBase/src/debugging.jl:123`

AayushSabharwal: Oh I see, `u` is a `VectorF64`, my bad. If you type the dictionary as `Dict{SymbolicT, SymbolicT}` it will work.

Shreyas-Ekanathan:
@AayushSabharwal can you rerun tests? the SciMLBase PR is merged now

AayushSabharwal:
This then needs a compat bump on SciMLBase

AayushSabharwal (approved):

PR #4724, opened 2026-07-11
analyze_initialization_jacobian: finite differences by default, ADTypes autodiff option

### LINKED ISSUE #4723: analyze_initialization_jacobian: first call dominated by Dual-specialized recompilation; finite differences are ~700x faster with identical rank result
`analyze_initialization_jacobian` (new in 11.30) is a one-shot rank/nullity diagnostic, but its first call is dominated by a Dual-specialized recompilation of the generated initialization residual, which grows with expression size and makes the function impractically slow on exactly the large models one wants to diagnose (minutes to much worse on multibody systems; observed "very long time" on a ~100-state multibody model).

Profile on a small multibody initialization problem (22 unknowns, 12 residuals; MTK at `as/alias-elim-zero-vars` d1ad4a0e1, Julia 1.12.6):

| stage | time |
|---|---|
| plain residual, first call (Float64 compile) | 1.8 s |
| `ForwardDiff.jacobian(residual, u0)`, first call | **33.5 s, 100 % compilation** |
| `ForwardDiff.jacobian`, second call | 0.3 ms |
| hand-rolled finite-difference jacobian, first call | 0.05 s (reuses the Float64 compile) |
| `svd(J; full = true)` | 4 ms |
| `analyze_initialization_jacobian`, first call | 34.9 s |

The finite-difference jacobian gives the identical rank/nullity result (max entry difference 8e-6, far below the default `rtol * σmax` rank tolerance), because the generated `RuntimeGeneratedFunction` for the residual only needs its existing Float64 specialization.

Suggested fix, in order of preference:

1. Default to finite differences for this diagnostic (either a ~15-line in-place forward-difference loop with no new deps, or `AutoFiniteDiff()` through DifferentiationInterface, which is already a dependency), and
2. accept an `autodiff` kwarg (ADTypes backend, routed through DifferentiationInterface) for users who want AD precision,
3. optionally preallocate the residual buffer instead of `du = similar(u, nresid)` per call, and use the thin SVD plus explicit handling of the implied null vectors (minor next to the compile cost).

Happy to submit a PR if the direction is agreed.

baggepinnen:
Fixes #4723.

`analyze_initialization_jacobian` is a one-shot rank/nullity diagnostic, but its first call was dominated by a dual-number recompilation of the generated initialization residual — 33.5 s (~100 % compilation) on a 22-unknown multibody initialization problem, scaling with expression size, i.e. slowest on exactly the large systems the diagnostic targets.

This PR:

- computes the Jacobian by **central finite differences by default**, reusing the residual's existing Float64 specialization: 0.05 s on the same system with the identical rank/nullity/redundancy result (`h = cbrt(eps)` gives O(h²) ≈ 4e-11 relative truncation error, far below the default `rtol * σmax` rank tolerance);
- adds an **`autodiff` keyword** accepting an ADTypes backend, evaluated through DifferentiationInterface (already a dependency), for users who want AD precision, e.g. `autodiff = AutoForwardDiff()`;
- extends the test file with a check that the default finite-difference path and `AutoForwardDiff()` agree on rank, nullity, redundancy, and the Jacobian entries.

https://claude.ai/code/session_01FMPpFa9w6VLtUVuCQYjSYt

### DIFF
--- a/src/initialization.jl
+++ b/src/initialization.jl
@@ -49,7 +49,7 @@ function MTKBase.get_initialization_problem_type(
 end
 
 """
-    analyze_initialization_jacobian(prob; rtol = 1e-8, atol = 0.0, threshold = 1e-3, verbose = true)
+    analyze_initialization_jacobian(prob; rtol = 1e-8, atol = 0.0, threshold = 1e-3, verbose = true, autodiff = nothing)
 
 Diagnose rank deficiency of a system's initialization problem by inspecting the singular
 value decomposition of its residual Jacobian, and report both the **unknowns** that span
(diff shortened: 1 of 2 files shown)


AayushSabharwal:
This has a merge conflict. I also noticed earlier that this code path doesn't do the right thing when the initialization is an `SCCNonlinearProblem`, since `prob.f.f` doesn't exist. It needs to use the subproblem decomposition to build a block-diagonal jacobian and probably should also solve the linear problems to get a better guess for the nonlinear ones.

Issue #4721, opened 2026-07-10
MTKBase v1.48.1 (patch) stopped resolving missing-default parameters with symbolic guesses at problem construction — silently wrong problem-stored values downstream


ChrisRackauckas-Claude:
**ModelingToolkitBase v1.48.1** (registered 2026-06-26, from PR #4676, commits 9170eaa44 + 72c5625c2) replaced the `evaluate_varmap!` fixpoint substitution with an `IRStructure`/`SU.IRSubstituter`-based path in `lib/ModelingToolkitBase/src/systems/parameter_buffer.jl` and `problem_utils.jl`. This changed observable behavior in a **patch** release:

Parameters declared with `default = missing` and a symbolic `guess` (e.g. Catalyst's conservation constants: `Γ` with `guess = Initial(dep + Σindep)`) were previously resolved into the parameter buffer at problem construction. Since v1.48.1 they are left at a placeholder `1.0` (invariant of `u0` — verified with different initial conditions), and initialization reports "overdetermined, 1 equations for 0 unknowns".

`init(prob, alg)` still recovers the correct value, so anything that runs full DAE initialization is fine. **Anything that reads problem-stored values or skips initialization gets silently wrong numbers**: `prob.ps[Γ]`, `prob[X2]` (observed through the constant), direct `f`/`g` evaluation, and `DynamicSS`/steady-state paths (which produced negative concentrations downstream).

Concrete downstream impact: 8 deterministic failures in Catalyst's `test/network_analysis/conservation_laws.jl` (e.g. `oprob.ps[Γ[1]]` returns `1.0` instead of `3.0`; `oprob[X2]` returns `0.0` instead of `2.0`). Bisected by version pinning with everything else held at latest: **1.48.0 PASS → 1.48.1 FAIL** (also checked: 1.42.2/1.47.0 pass, 1.49.0/1.51.0 fail). The breakage was masked on Catalyst master since ~Jul 3 by an unrelated earlier test abort, so it surfaced only now. Catalyst has a workaround up (SciML/Catalyst.jl#1506 passes conservation-law `initialization_eqs` through its model constructors), but the upstream questions stand:

1. Is dropping construction-time resolution of `missing`-default + symbolic-guess parameters intended? If yes, it's breaking and shouldn't have shipped in a patch; if no, it's a regression in the IRSubstituter path.
[...]

PR #4730, opened 2026-07-12
Align downstream LinearSolve compatibility

ChrisRackauckas-Claude:
> Ignore this PR until reviewed by @ChrisRackauckas.

## Context

While validating the Optimization.jl downstream workflow, the ModelingToolkit `GROUP=All` downstream job failed after earlier groups had loaded LinearSolve 3.87.0. The downstream group then resolved LinearSolve 4.3.0 and loaded ControlSystemsMTK, whose ArnoldiMethod extension tried to access `LinearSolve.ArnoldiMethodJL` in the already-loaded 3.87.0 module.

Original failure: https://github.com/SciML/Optimization.jl/actions/runs/29184114746/job/86627032925

LinearSolve 4.3.0 introduced the ArnoldiMethod extension in SciML/LinearSolve.jl#1071. A minimal two-environment reproducer fails with a preloaded LinearSolve 3.87.0 plus an inner LinearSolve 4.3.0 environment, while the same reproducer passes with LinearSolve 4.2.1.

## Changes

- Declare LinearSolve directly in `test/downstream/Project.toml`.
- Match its `3.66` compatibility range to the root test environment so both environments resolve the same LinearSolve major version.
- Add a QA regression test that keeps the downstream UUID and compatibility range aligned with the root test extras.

This changes test environment resolution only; it does not change public API.

## Prompt and process record

The original request was to update Optimization.jl downstream coverage. Repository instructions require a separate investigation whenever a failure reproduces on unmodified master, so this fix is intentionally split from that Optimization change.

Investigation steps:

1. Reproduced the reusable downstream workflow resolution at Optimization `0c663b6cc74f18b07b3b9aff95109b5ac1775db6` and ModelingToolkit `038c30a79a0e0ec09f9b6eff4cf1eaa44a088e15`.
2. Reduced the error to isolated outer/inner environments with LinearSolve 3.87.0 preloaded and 4.3.0 activated.
3. Verified the release boundary: LinearSolve 4.2.1 passes; 4.3.0 fails in the mixed-loaded-version reproducer.
4. Verified current ModelingToolkit master `ee7dd34d8b7cfe8a085e0fa04cec28a607053f66` still resolves the same 3.87.0-loaded/4.3.0-inner mismatch before this patch.
5. Added the narrow environment alignment and a regression test.

## Local validation

Julia 1.12.6:

[...]

PR #4710, opened 2026-07-05
Raise ModelingToolkitBase JumpProcesses floor to 9.27.0 (fix downgrade SSAIntegrator crash)

ChrisRackauckas-Claude:
## Summary

Fixes the **Downgrade Sublibraries / test (lib/ModelingToolkitBase)** lane (still red on master after #4709).

The sublibrary downgrade force-min-resolves **DiffEqBase 7.4.2 + SciMLBase 3.18.0 + JumpProcesses 9.26.0**. (ModelingToolkitBase requires SciMLBase 3.x via `SciMLBase = "3.18"`, which forces the DiffEqBase 7.x line; DiffEqBase 6.x needs SciMLBase 2.x, so 6.x is Unsatisfiable in the min graph.)

DiffEqBase 7.x `src/callbacks.jl` `initialize!` reads `integrator.derivative_discontinuity` — this holds across the **entire** 7.x line (7.4.2 through the semver-max 7.6.0), so a DiffEqBase floor bump does not help. JumpProcesses' `SSAIntegrator` has only a `u_modified` field and, in **9.26.0**, no `getproperty` shim. So the `JumpSystem Discrete Callbacks` testset (`test/symbolic_events.jl:612`, `solve(jprob, SSAStepper())`) crashes at initialize! with:

```
type SSAIntegrator has no field derivative_discontinuity
```

(CI job 85204611517: `1436 passed, 0 failed, 1 errored, 5 broken`.)

## Fix

**JumpProcesses 9.27.0** added the `Base.getproperty`/`Base.setproperty!` shim on `SSAIntegrator` mapping `:derivative_discontinuity` ↔ `u_modified` (`SSA_stepper.jl:123-136`), which makes the DiffEqBase 7.x `initialize!` path work. Raise the floor:

```
-JumpProcesses = "9.23.0"
+JumpProcesses = "9.27.0"
```

## Validation (local, Julia 1.10 lts, julia-downgrade-compat v2.6.1)

- **Before:** downgrade resolves JumpProcesses **9.26.0**; JumpSystem test errors (reproduced from the CI log).
- **After:** downgrade resolves JumpProcesses **9.27.0** + DiffEqBase 7.4.2 + SciMLBase 3.18.0; `Successfully resolved minimal versions` and `Pkg.instantiate()` exits 0.

## Not addressed here (separate conflict)

[...]

Issue #559, opened 2020-08-21
build_function is allocating for StaticArrays


dcabecinhas:
Replication: Create a simple function that outputs SAs

```julia

@parameters x[1:3]
y = x.^2
fy = eval(build_function(y,x,convert_oop=false,force_SA=true)[1])
@btime fy($(rand(3)))
#  29.933 ns (1 allocation: 32 bytes)
```

Seems to be related with the use of `@SArray` for the `build_function` code, when there are variables in the array expression
```julia
julia> @macroexpand @SArray [1 2; 3 4]
:((SArray{Tuple{2,2},T,N,L} where L where N where T)((1, 3, 2, 4)))

function f(a)
    (SArray{Tuple{2,2},T,N,L} where L where N where T)((1, 3, 2, 4))
end
f2 = (a) -> (SArray{Tuple{2,2},T,N,L} where L where N where T)((1, 3, 2, 4))

function g(a)
    (SArray{Tuple{2,2},T,N,L} where L where N where T)((1, 3, 2, 4+a)) end
g (generic function with 1 method)
g2 = (a) -> (SArray{Tuple{2,2},T,N,L} where L where N where T)((1, 3, 2, 4+a))

@btime f($(Ref(1))[])
#  1.545 ns (0 allocations: 0 bytes)
@btime g($(Ref(1))[])
#  1.850 ns (0 allocations: 0 bytes)
@btime f2($(Ref(1))[])
#  20.457 ns (0 allocations: 0 bytes)
@btime g2($(Ref(1))[])
#  29.336 ns (1 allocation: 48 bytes)
```

`@code_warntype` of `f()` and `g()` shows some issues in red for all the examples but somehow they are more noticeable for anonymous function and when the array expression has variables.

PR #4731, opened 2026-07-12
Restore Initial-dependent parameter guess resolution

ChrisRackauckas-Claude:
This PR should be ignored until reviewed by @ChrisRackauckas.

## Summary

- Keep the IR-based variable-map substitution fast path.
- If that path leaves a symbolic value, finish evaluation with the generic fixpoint substituter.
- Add a regression test for an array parameter whose guess depends on `Initial(X1 + X2)`.

## Root cause

ModelingToolkitBase v1.48.1 introduced the specialized `IRStructure`/`IRSubstituter` path in #4674. `IRSubstituter` uses a filter that treats `Symbolics.Operator` calls such as `Initial` as atomic, while the prior generic `fixpoint_sub` path descends into those calls. As a result, an `Initial`-dependent parameter guess could remain symbolic and later fall back to a placeholder value.

For Catalyst systems with eliminated conserved species, that stored `Γ = 1` instead of `3` and made the corresponding observed species `0` instead of `2`. The regression boundary is ModelingToolkitBase 1.48.0 (correct) to 1.48.1 (incorrect); see #4721.

The fallback only runs when the IR result is still symbolic, so fully resolved values retain the specialized fast path.

## Validation

- Pure ModelingToolkitBase boundary reproduction on Julia 1.12.6:
  - v1.48.0: `X2 stored = 2.0`, `Gamma stored = 3.0`
  - v1.48.1: `X2 stored = 0.0`, `Gamma stored = 1.0`
- Final fixed MWE, exit 0:
  - ModelingToolkitBase 1.51.0
  - `X2 stored = 2.0`
  - `Gamma stored = 3.0`
- `GROUP=Initialization julia +1.12 --project=. -e 'using Pkg; Pkg.test(; coverage=false)'`, isolated depot, exit 0:
  - `Initialization | 795 passed, 12 broken, 807 total`
  - `ModelingToolkit tests passed`
- Catalyst master `GROUP=Modeling`, isolated depot, exit 0:
  - `Conservation Laws | 250 passed, 17 broken, 267 total`
  - `Catalyst tests passed`
- Baseline Catalyst master with unmodified ModelingToolkit, exit 1:
  - `Conservation Laws | 242 passed, 8 failed, 17 broken, 267 total`
- Repository-wide `Runic --check .`, exit 0.
- `git diff --check`, exit 0.

Related downstream workaround: SciML/Catalyst.jl#1506. The separate released-Catalyst stale `@test_broken` failure is already fixed by SciML/Catalyst.jl#1495 and requires a Catalyst release.

### DIFF
--- a/lib/ModelingToolkitBase/src/systems/problem_utils.jl
+++ b/lib/ModelingToolkitBase/src/systems/problem_utils.jl
@@ -537,7 +537,13 @@ function evaluate_varmap!(
         ir::IRStructure{SymReal}, varmap::AtomicArrayDictSubstitutionWrapper, vars;
         limit = 100, allow_symbolic = false
     )
-    subber = Symbolics.FixpointSubstituter(SU.IRSubstituter{true}(ir, varmap); maxiters = limit, warn_maxiters = !allow_symbolic)
+    # Use `FPSubFilterer{Nothing}` so substitution descends into operator calls such as
+    # `Initial`, matching the generic `fixpoint_sub` path. The default IR filterer treats
+    # operators as atomic, leaving `Initial`-dependent guesses unresolved.
+    subber = Symbolics.FixpointSubstituter(
+        SU.IRSubstituter{true}(ir, varmap; filterer = Symbolics.FPSubFilterer{Nothing}());
+        maxiters = limit, warn_maxiters = !allow_symbolic
+    )
     for k in vars
         v = get(varmap, k, COMMON_NOTHING)
         v === COMMON_NOTHING && continue
--- a/lib/ModelingToolkitBase/test/initializationsystem.jl
+++ b/lib/ModelingToolkitBase/test/initializationsystem.jl
@@ -1610,6 +1610,24 @@ end
     @test integ1[X1] ≈ 1.0
 end
 
+@testset "Initial-dependent array parameter guesses" begin
+    @variables X1(t) X2(t)
+    @parameters k1 k2 Γ[1:1] = missing [guess = [Initial(X1 + X2)]]
+    eqs = [D(X1) ~ k1 * (Γ[1] - X1) - k2 * X1]
+    @named osys = System(
+        eqs, t, [X1], [k1, k2, Γ]; observed = [X2 ~ Γ[1] - X1]
+    )
+    osys = complete(osys)
+    prob = ODEProblem(
+        osys,
+        [X1 => 1.0, X2 => 2.0, k1 => 0.1, k2 => 0.2],
+        (0.0, 1.0)
+    )
+
+    @test prob[X2] == 2.0
+    @test prob.ps[Γ[1]] == 3.0
+end
+
 if @isdefined(ModelingToolkit)
     @testset "Trivial initialization is run on problem construction" begin
         @variables _x(..) y(t)


ChrisRackauckas-Claude:
Investigation checkpoint for this commit:

- Exact ModelingToolkit master IntegrationTest failure: https://github.com/SciML/ModelingToolkit.jl/actions/runs/29186963983/job/86634738158
- Same failure on #4730: https://github.com/SciML/ModelingToolkit.jl/actions/runs/29190999984/job/86645537297
- Local unmodified downstream reproduction: `Conservation Laws | 242 passed, 8 failed, 17 broken`, exit 1.
- Local fixed downstream run: `Conservation Laws | 250 passed, 17 broken`; final `Catalyst tests passed`, exit 0.
- Isolated official Initialization run: `795 passed, 12 broken`; final `ModelingToolkit tests passed`, exit 0.
- Regression boundary: MTKBase 1.48.0 stores `X2 = 2`, `Γ = 3`; 1.48.1 stores `X2 = 0`, `Γ = 1`.
- Introducing code landed in #4674; #4676 was the release-only PR.

The ReleaseTest failure is separate: released Catalyst v16.2.1 contains a stale `@test_broken`; SciML/Catalyst.jl#1495 already fixed it on Catalyst master. The same released-package failure is visible in https://github.com/SciML/SciMLBase.jl/actions/runs/29194023668/job/86653653225.

ChrisRackauckas-Claude:
CI checkpoint after the rebase:

- The initialization fix itself remains locally verified: frozen Initialization 795 pass / 12 pre-existing broken, and Catalyst Modeling improves from 242 pass / 8 fail / 17 broken to 250 pass / 17 broken.
- Most rebased red jobs resolved before NonlinearSolveQuasiNewton 1.14.0 entered General and fail on the independently reproduced 1.13.3 plus NonlinearSolveBase 2.34.0 maybe_pinv load error. The corrected pair was separately verified.
- Exact clean current-master ModelingToolkitBase InterfaceII completed locally with 7234 pass / 2 errors / 4 pre-existing broken. The only errors are the unrelated BoundaryValueDiffEq MIRK dual-cache failures in Lotka-Volterra and cost-function compilation; existing upstream work is BoundaryValueDiffEq issue #556 and draft PR #559.
- Downgrade floor failures are being handled separately in ModelingToolkit PR #4710 and focused stacked follow-ups.

No observed failure points to the Initial-dependent guess substitution change, and no tests were skipped, silenced, or loosened.

AayushSabharwal:
This is not the correct fix. The issue is that calling `fixpoint_sub` directly uses a different filterer (`Symbolics.FPSubFilterer`) which will substitute inside operator calls by default. This is circumvented by constructing `FixpointSubstituter` from an `IRSubstituter`, since the former can't change the filterer in the latter. The correct fix is to pass `filterer = Symbolics.FPSubFilterer{Nothing}()` to `IRSubstituter`.

ChrisRackauckas-Claude:
Thanks @AayushSabharwal — you're right, the fallback approach was papering over the wrong filterer. Updated to pass `filterer = Symbolics.FPSubFilterer{Nothing}()` to the `IRSubstituter` in `evaluate_varmap!` and dropped the `fixpoint_sub` fallback entirely.

Verified locally on Julia 1.12.4 against dev'd ModelingToolkitBase (Symbolics 7.24 / SymbolicUtils resolved):
- With the filterer: `X2 stored = 2.0`, `Gamma stored = 3.0` (PASS).
- Counter-check — reverting just the filterer (back to the default IR filterer) fails the `X2 == 2.0` assertion, confirming the filterer is what resolves the `Initial`-dependent guess.

`Runic --check` clean on the changed file. Force-pushed the amended commit (single-commit branch, no other contributors).

AayushSabharwal:
#4704 will fix this, so I'm closing the PR

PR #4755, opened 2026-07-15
Fix workflow tag concurrency checks

ChrisRackauckas-Claude:
Ignore this PR until reviewed by @ChrisRackauckas.

## Summary

- replace invalid github.refs references in Downstream, Documentation, and ReleaseTest workflows
- use github.ref_type to keep master and tag runs exempt from cancellation as documented

## Root cause

Clean-master reproduction with actionlint reports that github.refs is undefined. A focused git bisect identified 80f44bca96a18310bce2678c02176e641fe04ad4 as the first commit introducing the Downstream typo. A literal github.ref comparison against refs/tags/* would not implement wildcard matching, so the corrected condition checks ref_type instead.

## Validation

- actionlint -color .github/workflows/Downstream.yml
- actionlint -color .github/workflows/Downstream.yml .github/workflows/Documentation.yml .github/workflows/ReleaseTest.yml
- Julia Runic --check over the repository

All commands exited successfully locally on Julia 1.12.6.

PR #4760, opened 2026-07-15
Fix stale downstream test groups

ChrisRackauckas-Claude:
## Summary

- select DataDrivenDiffEq's existing `Core` group instead of the nonexistent `Downstream` group
- replace MethodOfLines' nonexistent `Interface` group with its two current interface lanes, `MOL_Interface1` and `MOL_Interface2`

This is stacked on #4759 so the downstream workflow also propagates resolver and test failures. The extra #4755/#4759 commits will disappear from this diff as those prerequisites merge.

Ignore this PR until reviewed by @ChrisRackauckas.

## Local verification

Each corrected target was run with this ModelingToolkit branch developed into the downstream package:

- `DataDrivenDiffEq/Core`: passed; substantive Basis, DataDrivenProblem, DataDrivenSolution, Utilities, and CommonSolve summaries
- `MethodOfLines/MOL_Interface1`: passed after 18m58s; the lane executes its PDE construction and solve smoke cases but contains no explicit `@test` assertions
- `MethodOfLines/MOL_Interface2`: passed after 11m51s; 4 passed and 8 pre-existing broken assertions

Final stacked branch checks:

- `actionlint -color`
- `/home/crackauc/.juliaup/bin/julialauncher +1.12 --startup-file=no -m Runic --check .`
- `git diff --check agent/propagate-workflow-failures..HEAD`

### DIFF
--- a/.github/workflows/Documentation.yml
+++ b/.github/workflows/Documentation.yml
@@ -13,7 +13,7 @@ concurrency:
   # Skip intermediate builds: always, but for the master branch.
   # Cancel intermediate builds: always, but for the master branch.
   group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: ${{ github.ref != 'refs/heads/master' && github.refs != 'refs/tags/*'}}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/master' && github.ref_type != 'tag' }}
 
 jobs:
   build:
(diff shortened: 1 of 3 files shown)


ChrisRackauckas-Claude:
The `SBMLToolkit.jl/All/1` failure in job `87321454608` is not caused by the two stale-group corrections in this PR. This branch is stacked on #4759, which removes the resolver-error catch and therefore exposes an existing incompatibility: SBMLToolkit pins SBMLToolkitTestSuite 0.0.5, while the resolved OrdinaryDiffEq 7 stack requires TestSuite 1.x. The default branch's corresponding job appears green only because that resolver error is still swallowed there. A separate clean-source compatibility investigation is queued; this PR remains limited to DataDrivenDiffEq and MethodOfLines group selection.

PR #4762, opened 2026-07-15
feat: Introduce `GeneratedFunctionOptions` for the code-generation entry points

AayushSabharwal:
In continuation of the series of PRs that aim to reduce unnecessarily splatting kwargs everywhere.

### DIFF
--- a/lib/ModelingToolkitBase/src/problems/intervalnonlinearproblem.jl
+++ b/lib/ModelingToolkitBase/src/problems/intervalnonlinearproblem.jl
@@ -7,10 +7,13 @@ function SciMLBase.IntervalNonlinearFunction(
     check_complete(sys, IntervalNonlinearFunction)
     check_compatibility && check_compatible_system(IntervalNonlinearFunction, sys)
 
-    f = generate_rhs(
-        sys; expression, wrap_gfw = Val{true},
-        scalar = true, eval_expression, eval_module, checkbounds, kwargs...
+    codegen_opts = GeneratedFunctionOptions(;
+        expression, wrap_gfw = Val{true}, eval_expression, eval_module,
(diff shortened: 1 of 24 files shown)


lib/ModelingToolkitBase/src/problems/intervalnonlinearproblem.jl
```diff
@@ -7,11 +7,14 @@ function SciMLBase.IntervalNonlinearFunction(
     check_complete(sys, IntervalNonlinearFunction)
     check_compatibility && check_compatible_system(IntervalNonlinearFunction, sys)
 
-    f = generate_rhs(
-        sys; expression, wrap_gfw = Val{true},
-        scalar = true, eval_expression, eval_module, checkbounds, kwargs...
+    codegen_opts = GeneratedFunctionOptions(;
+        expression, wrap_gfw = Val{true}, eval_expression, eval_module,
+        compiler_options = get(kwargs, :compiler_options, CompilerOptions()),
```
KristofferC, line 12: Just so I understand, why you `get(kwargs, ..)` instead of having the `compiler_options` as a named keyword?

AayushSabharwal: I told Claude to only change the callsites and not touch the kwargs of these functions. Eventually they will take something like a `SciMLFunctionOptions` struct, so `compiler_options` will come from there.

AayushSabharwal: The approach is bottom-up:
- `codegen_function` got `CodegenFunctionOptions`
- All its callers changed to pass the struct
- `build_function_wrapper` got `BuildFunctionWrapperOptions`
- All its callers changed to pass the struct
This PR now adds `GeneratedFunctionOptions`, and changes the callers to pass it.

PR #4783, opened 2026-07-17
feat: add `set_defaults`

### LINKED ISSUE #4767: `@component`-style modeling requires listing settable syms as kw args
One thing I find annoying with the `@component` style of definition is that i need to list all syms which i want to be "settable" in the component constructor manually:
```julia
@component function MyModel(; name, P1=nothing, P2=nothing)
    @parameters begin
        P1 = P1, [description="Initializable parameter P1"]
        P2 = P2, [description="Initializable parameter P1"]
    end
end
```
Where `@mtkmodel` produced them automaticially. I implemented my own helper function to avoid that
```julia
@component function MyModel(; name, defaults...)
    @parameters begin
        P1, [description="Initializable parameter P1"]
        P2, [description="Initializable parameter P1"]
    end
    ....
    sys = System(...)
    set_mtk_defaults(sys, defaults) # parse kwargs and set defaults/bindings
end
```

The problem is this reaches a bit into internals (in order to differentiate between defaults and bindings) and is also a bit fragile (i split the symbolic names on namespace separator `₊` in order to get the actual symbols by `Symbol`  name of a nested MTK system.

<details>
<summary> current implementation </summary>

```julia
function set_mtk_defaults(sys::System, pairs)
    isempty(pairs) && return sys
    defs = values(pairs)
    names = keys(pairs)
    symbols = map(names) do name
        try
            getproperty_symbolic(sys, name; might_contain_toplevel_ns=false)
        catch e
            throw(ArgumentError("Could not resolve variable of name $name in system $(getname(sys))! None of the defaults have been set."))
        end
    end
    defdict = ModelingToolkitBase.get_initial_conditions(sys)
    binddict = ModelingToolkitBase.SymmapT()
    for (k, v) in ModelingToolkitBase.get_bindings(sys)
        binddict[k] = v
    end

    # Mirror MTK's `collect_defaults!`, which routes a variable's `default` metadata to
    # either the initial conditions (numeric) or the bindings (symbolic) at `System`
    # construction. Values forwarded through this function arrive too late for that pass,
    # so they are classified here instead.
    bindings_changed = false
    for (s, v) in zip(symbols, defs)
        us = unwrap(s)
        if SII.symbolic_type(v) === SII.NotSymbolic()
            defdict[s] = v
            if haskey(binddict, us)
                delete!(binddict, us)
                bindings_changed = true
            end
        else
            binddict[us] = unwrap(v)
            delete!(defdict, us)
            bindings_changed = true
        end
    end
    bindings_changed || return sys

    @reset sys.bindings = ModelingToolkitBase.ROSymmapT(binddict)
    ## a cached bindings graph (built by `complete`) would be stale now
    @reset sys.parameter_bindings_graph = nothing
    sys
end

"""
    getproperty_symbolic(sys, var; might_contain_toplevel_ns=true)

Like `getproperty` but works on a greater varaity of "var"
- var can be Num or Symbolic (resolved using genname)
- strip namespace of sys if present (don't strip if `might_contain_top_level_ns=false`)
- for nested variables (foo₊bar₊baz) resolve them one by one
"""
function getproperty_symbolic(sys, var; might_contain_toplevel_ns=true)
    ns = string(getname(sys))
    varname = string(getname(var))
    # split of the toplevel namespace if necessary
    if might_contain_toplevel_ns && startswith(varname, ns*"₊")
        if getname(sys) ∈ getname.(ModelingToolkitBase.get_systems(sys))
            @warn "Namespace :$ns appears multiple times, this might lead to unexpected, since it is not clear whether the namespace should be stripped or not."
        end
        varname = replace(varname, r"^"*ns*"₊" => "")
    end
    parts = split(varname, "₊")
    # Descend a `₊`-segment ONLY when it names an actual child subsystem of the
    # current node; the first segment that is not a subsystem starts a flat variable
    # name, so the remaining segments (re-joined by `₊`) are resolved as one variable.
    # This is necessary because a flattened/transpiled leaf can carry `₊` *inside* a
    # variable name (e.g. `washout₊derivative₊x` is a single unknown, not the chain
    # washout→derivative→x). Driven by `get_systems` (the real structure), so it is
    # not a try/catch guess. First level uses namespace=false (strip the toplevel),
    # subsequent levels namespace=true (accumulate the full name) — as before.
    r = sys
    i = 1
    while i <= length(parts)
        first = (i == 1)
        if r isa System && Symbol(parts[i]) in getname.(ModelingToolkitBase.get_systems(r))
            r = getproperty(r, Symbol(parts[i]); namespace = !first)
            i += 1
        else
            r = getproperty(r, Symbol(join(parts[i:end], "₊")); namespace = !first)
            break
        end
    end
    unwrap(r)
end

```

</details>

Would it be welcome to move such an implementation into MTKBase as a "helper" function for cleaner `@components` definition?

AayushSabharwal:
Close #4767 

Refer to the docstring for details

### DIFF
--- a/lib/ModelingToolkitBase/src/ModelingToolkitBase.jl
+++ b/lib/ModelingToolkitBase/src/ModelingToolkitBase.jl
@@ -291,6 +291,7 @@ export SymScope, LocalScope, ParentScope, GlobalScope
 export independent_variable, equations, observed, full_equations, jumps, cost,
     brownians
 export initialization_equations, guesses, bindings, initial_conditions, hierarchy
+export set_defaults
 export state_priorities, irreducibles, maybe_zeros
 export mtkcompile, expand_connections, structural_simplify
 export solve
--- a/lib/ModelingToolkitBase/src/systems/abstractsystem.jl
+++ b/lib/ModelingToolkitBase/src/systems/abstractsystem.jl
@@ -1783,6 +1783,77 @@ end
 """
     $TYPEDSIGNATURES
 
+Given `sys` and an iterable of `name => value` pairs (e.g. from keyword arguments or a
+`Dict`), resolve each `name` to the corresponding unknown/parameter/variable of `sys` and
+update its default. This follows the same semantics as providing the value via variable
+metadata, i.e. `@variables var(t) = value`:
+
+  - If `value` is `nothing`, any existing binding/initial condition for the variable is
+    removed.
+  - If `value` is `missing`, the variable is recorded as having a `missing` binding (see
+    the corresponding semantics of `@parameters par = missing`).
+  - If `value` is symbolic, the variable is recorded as being bound to it (see
+    [`bindings`](@ref)).
+  - Otherwise, `value` is recorded as the variable's initial condition (see
+    [`initial_conditions`](@ref)).
+
+`name` can be a `Symbol`, a `String`, or a symbolic variable of `sys`. `Symbol`/`String`
+names may use `NAMESPACE_SEPARATOR` (`₊`) or `.` to refer to a variable owned by a
+subsystem of `sys`, e.g. `:inner₊x`. Names are resolved relative to `sys` itself, so they
+should _not_ be prefixed with the name of `sys`.
+
+Returns a new system with the same equations/variables as `sys`, but with the bindings and
+initial conditions updated according to `pairs`. Does not mutate `sys`.
+"""
+function set_defaults(sys::AbstractSystem, pairs)
+    pairs = collect(pairs)
+    isempty(pairs) && return sys
+
+    ics = copy(get_initial_conditions(sys))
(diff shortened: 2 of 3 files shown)


hexaeder:
LGTM, my tests pass with this! The only small gap is that if I one happens to have a namespace separator within a variable name it would break. I handled that case in my `getproperty_symbolic`, but that helper is also used in different places where I needed it to work on flattend or partially flattend systems (which do have NS separators in variable names). But since the usage in MTKBase is tailored to construction time default passing I think its a non-issue.

AayushSabharwal:
I'll handle the namespace separator

AayushSabharwal:
Namespace separators in flattened systems should be handled now.

hexaeder:
LGTM! Thanks for implementing, i think this will help adoption of the `@component` style.

PR #4787, opened 2026-07-20
fix: avoid eliminating irreducibles in `eliminate_zero_variables!`

AayushSabharwal:
Close #4764

PR #4800, opened 2026-07-21
refactor: simplify `eliminate_zero_variables!`

AayushSabharwal:
Re https://github.com/SciML/ModelingToolkit.jl/pull/4787#issuecomment-5024513068, CC @oscardssmith


oscardssmith:
This looks a good bit better, but why not move this up to `__eliminate_zero_variables`? We shouldn't have to build sets at all, just lazily filter.

AayushSabharwal:
> but why not move this up to __eliminate_zero_variables?

I tried to segregate it such that `__eliminate_zero_variables!` does the structural part of figuring out what is zero from `mm`, and `eliminate_zero_variables!` does the symbolic manipulation.

oscardssmith:
fair enough. lgtm then

PR #4806, opened 2026-07-21
Change `SampleTime` to be like other `Operator`s

kpamnany:
And add an `init` argument.

## Checklist

### DIFF
--- a/lib/ModelingToolkitBase/src/discretedomain.jl
+++ b/lib/ModelingToolkitBase/src/discretedomain.jl
@@ -192,8 +192,8 @@ struct SampleTime <: Operator
     init::Union{Nothing,Real}
 end
 
-SampleTime(; init::Union{Nothing,Real} = nothing) = SampleTime(init)
-(D::SampleTime)() = STerm(D, SArgsT((D.init,)); type = Real)
+SampleTime(; init::Union{Nothing,Real} = nothing) = SampleTime(init)()
+(D::SampleTime)() = STerm(D, SArgsT(()); type = Real, shape = SU.ShapeVecT())
 SymbolicUtils.promote_symtype(::SampleTime, ::Type{T}) where {T} = Real
 SymbolicUtils.promote_shape(::SampleTime, @nospecialize(x::SU.ShapeT)) = x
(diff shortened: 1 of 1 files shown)


kpamnany:
@AayushSabharwal: please take a look. I'm not sure this is done right. `mtkcompile` gives me:
```julia
ERROR: MethodError: no method matching -(::SymbolicUtils.BasicSymbolicImpl.var"typeof(BasicSymbolicImpl)"{…}, ::SymbolicUtils.BasicSymbolicImpl.var"typeof(BasicSymbolicImpl)"{…})
This error has been manually thrown, explicitly, so the method may exist but be intentionally marked as unimplemented.
<snip>
```

lib/ModelingToolkitBase/src/discretedomain.jl
```diff
@@ -183,16 +183,19 @@ Base.:-(k::ShiftIndex, i::Int) = k + (-i)
...
 """
 struct SampleTime <: Operator
-    SampleTime() = SymbolicUtils.term(SampleTime, type = Real)
+    init::Union{Nothing,Real}
 end
-SymbolicUtils.promote_symtype(::Type{SampleTime}) = Real
-SymbolicUtils.promote_shape(::Type{SampleTime}) = SU.ShapeVecT()
+
+SampleTime(; init::Union{Nothing,Real} = nothing) = SampleTime(init)
+(D::SampleTime)() = STerm(D, SArgsT((D.init,)); type = Real)
```
AayushSabharwal: The `init` is stored in `D`, so why does it need to be an argument of the expression as well?

AayushSabharwal:
```suggestion
SampleTime(; init::Union{Nothing,Real} = nothing) = SampleTime(init)()
(D::SampleTime)() = STerm(D, SArgsT(()); type = Real, shape = SU.ShapeVecT())
```

`SampleTime` is typically used as `SampleTime()` in equations, which means that the constructor should create a `SampleTime` object and call it to build the expression. The expression itself doesn't need to have `D.init` as an argument, since that information is already available from `D`. I also like providing `shape`, though it will fall back to the correct value in this case.

PR #4826, opened 2026-07-24
Don't analytically integrate variables modified by events

cmhyett:
`mtkcompile`'s zero-variable elimination moves variables with trivial dynamics (e.g. `D(x) ~ 0`) out of the unknowns, rewriting them as an observed polynomial in the independent variable with a `missing`-bound constant-of-integration parameter. This ignored discrete/continuous events: a variable a callback impulsively modifies is no longer an unknown, so a functional (`ImperativeAffect`) kick errors at `ODEProblem` construction ("... refers to missing variable(s) ... reduced away").

Fix it two ways:

- Auto-detect (default): add `variables_modified_by_events(sys)` in ModelingToolkitBase and extend `__eliminate_zero_variables!` to protect event-modified variables using the same seam that already retains `irreducible` variables. Such a variable stays an ordinary numerically integrated unknown; unrelated trivial dynamics are still integrated.
- Opt-out flag: `mtkcompile(sys; analytic_integration=false)` disables the transformation wholesale (threaded to `eliminate_mm_zeros`, composed via AND so the SDE force-off still wins).

Add a regression testset covering equational and imperative kicks, the flag, selective protection, and the event-free no-regression case.

## Checklist

- [X] Appropriate tests were added
  
## Additional context

Single claude commit, needs deeper review - fix works but may not be most elegant.


AayushSabharwal:
Variables modified by standard symbolic events are fine. ImperativeAffect is the issue, since it doesn't run a custom reinitialization after the affect. Analytically integrated variables modified by ImperativeAffect can for now be marked as irreducible to avoid this issue. The flag to disable analytical integration can be useful, but it should specifically disable analytical integration and not the entire `eliminate_zero_variables!` pass. The `analytic_integration` flag added here is no different from the pre-existing `eliminate_mm_zeros` flag.

cmhyett:
Reverted this and put in a minimal flag. I'll close this since I can work around it with the existing `eliminate_mm_zeros` flag; maybe the code changes can be helpful to resolve the actual issue in the future.

Issue #4831, opened 2026-07-25
Nonnumeric parameters force per-type solver recompilation (~16s first solve); consider type-erasing the nonnumeric buffer


ChrisRackauckas-Claude:
Models with a nonnumeric parameter (a `CubicSpline` forcing function, a callable, any struct) pay a full solver recompilation that all-numeric models do not, because the nonnumeric parameter's declared type is baked into the `MTKParameters` type. Measured with the default solver on Julia 1.12, MTK 11.36 / ModelingToolkitBase 1.51:

| model | first `solve(prob)` |
|---|---|
| all-numeric | **0.91s** |
| all-numeric, different model, same `MTKParameters` type | 0.59s |
| `CubicSpline` parameter (new `MTKParameters` type) | **16.74s** |
| `LinearInterpolation` parameter (new type again) | **15.45s** |
| second `CubicSpline` model, same `MTKParameters` type | 0.63s |

The last two rows are the controlled comparison. A *new* nonnumeric type costs ~16s; a *different model reusing an already-seen* nonnumeric type costs 0.63s. So the penalty tracks `MTKParameters` type uniformity and nothing else — not model novelty, not equation count.

The all-numeric case is fast because `MTKOrdinaryDiffEqDefaultExt` precompiles one all-numeric `solve(prob)`, and every all-numeric model shares that `MTKParameters` type:

```
all-numeric:  MTKParameters{Vector{Float64}, Vector{Float64}, Tuple{}, Tuple{}, Tuple{},                      Tuple{}}
CubicSpline:  MTKParameters{Vector{Float64}, Vector{Float64}, Tuple{}, Tuple{}, Tuple{Vector{CubicSpline}},    Tuple{}}
LinearInterp: MTKParameters{Vector{Float64}, Vector{Float64}, Tuple{}, Tuple{}, Tuple{Vector{LinearInterpolation}}, Tuple{}}
```

### Why this cannot be fixed by extending the precompile workload

The nonnumeric buffer's element type comes from the parameter's *declared* type, so covering it by precompilation would require enumerating user nonnumeric types — MTK cannot depend on DataInterpolations, and cannot know a user's structs at all. Adding a `Vector{Function}` workload does not help `CubicSpline` users. Making the type uniform is the only mechanism that generalizes.

### Proposed direction

[...]

PR #4833, opened 2026-07-27
[NOT FOR MERGE — archived prototype] Erase nonnumeric buffer element types from MTKParameters

ChrisRackauckas-Claude:
**Do not merge. Do not review.** Opened purely to preserve an investigation in a reviewable form, and closed immediately. Context and full analysis: #4831.

This prototype delivers a large first-solve win and carries a **disqualifying runtime regression** that cannot be fixed without also changing parameter codegen. It is archived so the measurements and the design constraint are not lost.

### The problem

Models with a nonnumeric parameter (a DataInterpolations spline, a callable, any struct) put that declared type into the `MTKParameters` type. Solver specializations are keyed on the parameter object, so two models differing only in `CubicSpline` vs `LinearInterpolation` share no compiled code and each pays a full first-solve compilation (~16s with the default algorithm; see #4831 for that measurement).

### What this does

Wraps the nonnumeric buffers in an `OpaqueNonnumeric{K}` — a `Ref{Any}`, with `K` (buffer count) kept in the type because the generated parameter-path dispatch needs it, and element types erased. The field stays *named* `nonnumeric` and `getproperty` unwraps, so the ~45 existing `p.nonnumeric[i][j]` sites are untouched. One file plus tests.

### Compile time (two models, different nonnumeric parameter types, Tsit5)

| | baseline | erased |
|---|---|---|
| `MTKParameters` type uniform | false | **true** |
| solve A (cold) | 4.917s | 4.782s |
| **solve B (different nonnumeric type)** | **3.870s** | **0.015s** |
| results | identical | identical |

### Runtime (same model; nonnumeric parameter read on every RHS evaluation)

| | baseline | erased |
|---|---|---|
| RHS ns/call | **1.37** | **666.33** |
| RHS allocations/call | **0** | **112 B** |
| warm solve | 0.108 ms | 0.188 ms |

So it trades ~3.9s of one-time compilation for a 486x slower RHS on exactly the models it targets. Net loss for any non-trivial integration — hence not mergeable.

### Why, and what would be needed

Parameters are destructured via Symbolics `DestructuredArgs` over an argument typed `Vector{Vector{Any}}`, so the RHS reads buffers as `p[i][j]`. Baseline gets type stability *for free* from the concrete `Vector{FA}`; erasing the element type makes that read `Any`, so every nonnumeric access inside the time loop dispatches dynamically and boxes. This is inherent to erasing at the buffer level — storing `Vector{Any}` directly costs the same.

The only fix is for codegen to emit a **typed** access for nonnumeric buffers, e.g. `(p[i]::Vector{FA})[j]`. `BufferTemplate` already carries `type::TypeT`, so the index cache knows the concrete element types at codegen time. That is a change to the parameter destructuring in `codegen_utils.jl` (possibly needing Symbolics support for typed destructuring) and is **not attempted here**.

### Tests

Adds `test/opaque_nonnumeric.jl` covering symbolic read/write of a nonnumeric parameter, the rewritten `remake_buffer` path (numeric and nonnumeric edits), `copy`/`==`/`as_any_buffer`, `SciMLStructures` canonicalize/replace, and the type-uniformity property. These passed 16/16 as a standalone script; the committed file swaps in the non-deprecated `Dict` problem constructor, and I am re-running it against the committed form — if that surfaces anything I will push a correction to the branch even though the PR is closed.

https://claude.ai/code/session_01UuAbA8hSnrirMrJjFqxwn5

### DIFF
--- a/lib/ModelingToolkitBase/src/systems/parameter_buffer.jl
+++ b/lib/ModelingToolkitBase/src/systems/parameter_buffer.jl
@@ -3,6 +3,48 @@ symconvert(::Type{T}, ::Type{F}, x::V) where {T <: Real, F, V} = convert(T, x)
 symconvert(::Type{Real}, ::Type{F}, x::Integer) where {F} = convert(F, x)
 symconvert(::Type{V}, ::Type{F}, x) where {V <: AbstractArray, F} = symconvert.(eltype(V), F, x)
 
+"""
+    OpaqueNonnumeric{K}
+
+Type-erasing wrapper around the `K` nonnumeric parameter buffers.
+
+The nonnumeric buffers are the only part of `MTKParameters` whose element types
+come from user-declared parameter types (a `CubicSpline`, a callable, an arbitrary
+struct). Storing them concretely puts those types into the `MTKParameters` type,
+which makes every solver specialization keyed on the parameter object unique per
+nonnumeric type — so two models differing only in, say, `CubicSpline` vs
+`LinearInterpolation` share no compiled code and each pays a full first-solve
+compilation.
+
+Holding them behind a `Ref{Any}` keeps the *buffer count* `K` in the type (the
+generated parameter-path dispatch needs it) while erasing the element types, so
+those models share one `MTKParameters` type. The payload is held by reference, so
+in-place writes to the buffers remain visible.
+
+`nonnumeric` is the only slot that can be erased this way: `tunable` and
+`initials` must stay concrete because AD differentiates through them, and the AD
+extensions already treat the nonnumeric slot as non-differentiable (Mooncake's
+`increment_and_get_rdata!` skips it outright).
+"""
+struct OpaqueNonnumeric{K}
+    ref::Base.RefValue{Any}
+end
+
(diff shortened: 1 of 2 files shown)


ChrisRackauckas-Claude:
Closing immediately as intended — this is an archive, not a proposal.

Kept for the record because the measurements answer a question that would otherwise have to be rediscovered: nonnumeric buffer erasure **does** deliver the first-solve win (3.870s → 0.015s for a second model with a different nonnumeric type), and it **cannot** be shipped without typed nonnumeric access in the parameter codegen (RHS 1.37ns / 0 alloc → 666ns / 112B without it).

Design discussion continues on #4831. The branch stays on the fork; anyone picking this up should start from the codegen half, since the parameter-buffer half is done and tested here.

ChrisRackauckas-Claude:
Verification note (closing the caveat in the description): the committed `test/opaque_nonnumeric.jl`, with the non-deprecated `Dict` problem constructor, passes **16/16** against the committed `parameter_buffer.jl`. No correction to the branch was needed.

```
Test Summary:            | Pass  Total  Time
opaque nonnumeric buffer |   16     16  3.7s
```

So the parameter-buffer half of this is complete and green as archived. The blocker remains the codegen half only.

PR #4839, opened 2026-07-28
[RFC] Add functional guesses

SebastianM-C:
## Checklist

## Additional context

For dynamic optimization problems it is not enough to provide constant guesses for the initial conditions and in some cases a functional form of the guess is needed. This also has a precedent in BoundaryValueDiffEq with continuation: https://docs.sciml.ai/BoundaryValueDiffEq/dev/tutorials/continuation/#On-providing-initial-guess

How do we want the interface for this to look like? The proposal here is `initial_trajectory` as a `Dict` mapping the variables to the function, but we can adapt this to be closer to BVP.

Another question is how should the free final time problems be handled. We could internally scale the time if the simpler approach in this PR is not good enough.

### DIFF
--- a/lib/ModelingToolkitBase/src/systems/optimal_control_interface.jl
+++ b/lib/ModelingToolkitBase/src/systems/optimal_control_interface.jl
@@ -324,7 +324,9 @@ function process_DynamicOptProblem(
         steps = nothing,
         tune_parameters = false,
         guesses = Dict(), initial_trajectory = Dict(),
-        bounds = Dict(), kwargs...
+        bounds = Dict(),
+        eval_expression = false, eval_module = @__MODULE__,
+        kwargs...
     )
     warn_overdetermined(sys, op)
     ctrls = inputs(sys)
@@ -342,7 +344,7 @@ function process_DynamicOptProblem(
     f, u0,
         p = process_SciMLProblem(
         ODEInputFunction, sys, _op;
-        t = tspan !== nothing ? tspan[1] : tspan, kwargs...
+        t = tspan !== nothing ? tspan[1] : tspan, eval_expression, eval_module, kwargs...
     )
     model_tspan, steps, is_free_t = process_tspan(tspan, dt, steps)
     warn_overdetermined(sys, op)
@@ -375,7 +377,7 @@ function process_DynamicOptProblem(
     for (var, traj) in initial_trajectory
         idx = get(stidxmap, var, nothing)
         idx === nothing && continue
-        set_initial_trajectory!(model, U, idx, build_trajectory_function(sys, var, traj, pmap))
+        set_initial_trajectory!(model, U, idx, build_trajectory_function(sys, var, traj, p; eval_expression, eval_module))
     end
     V = generate_input_variable!(model, c0, length(ctrls), tsteps)
     P = generate_tunable_params!(model, p0, length(tunable_params))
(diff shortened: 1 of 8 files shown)


ChrisRackauckas:
I think in principle this is good. It doesn't even need to just be for dynamic optimization?

SebastianM-C:
yeah, we could have it interpreted in other problem types. is the name ok or should that be adjusted?

lib/ModelingToolkitBase/src/systems/optimal_control_interface.jl
```diff
@@ -393,6 +400,47 @@ function process_DynamicOptProblem(
...
+    isempty(unresolved) || throw(
+        ArgumentError(
+            "The `initial_trajectory` for $var may only depend on the independent " *
+                "variable $iv and on parameters, but it also depends on " *
+                "$(join(unresolved, ", ")). Parameters are resolved from the operating " *
+                "point, so anything left over cannot be evaluated."
+        )
+    )
+
+    return Symbolics.build_function(expr, iv; expression = Val(false))
```
AayushSabharwal: This should use `ModelingToolkitBase.build_function_wrapper` and then `eval_or_rgf` to get an RGF out.

SebastianM-C: Done and also moved the reusable codegen part to codegen.jl

lib/ModelingToolkitBase/src/systems/optimal_control_interface.jl
```diff
@@ -393,6 +402,60 @@ function process_DynamicOptProblem(
...
+            "The `initial_trajectory` for $var may only depend on the independent " *
+                "variable $iv and on parameters, but after inlining observed variables " *
+                "and parameter bindings it also depends on $(join(unresolved, ", "))."
+        )
+    )
+
+    opts = GeneratedFunctionOptions(; expression = Val{false}, eval_expression, eval_module)
+    rgf = generate_trajectory(sys, expr, opts)
+
+    return Base.Fix1(rgf, p)
```
AayushSabharwal, line 451: Won't this cause problems with `remake`? Or initialization, which re-creates `p` as a copy.

SebastianM-C: Hmm... I don't think we run initialization on the dynamic optimization path, I'll have to double check. On `remake` I'm not sure how to handle it... The issue is that the backends need only a function of t, but we want to be able to use the (MTK) parameters in the functions. I don't think that we can easily remake, since almost everything else on this path has the same problem of baking the p in, as we need to substitute in the backend specific parameter representation. Probably remake needs to go through the process_ function again and we don't have a fast path remake possible for external backends, that would need to remake the underlying backend specific problem...

AayushSabharwal: Ah, right, that makes sense

AayushSabharwal (approved):

AayushSabharwal:
One of the dynamic optimization tests is failing in [Tests / tests / Optimization (julia 1, ubuntu-latest) / Tests - Optimization](https://github.com/SciML/ModelingToolkit.jl/actions/runs/31061624515/job/92490886452?pr=4839). Could you also rebase so the rest of the failures disappear?

Issue #102, opened 2019-02-12
Allow arrays of variables


dpsanders:
It would be useful to be able to work with arrays of variables and access them via `x[1]` etc.
JuMP allows things like this, for example.

ChrisRackauckas:
Arrays of variables already exist? Just put variables in an array? Did you mean variables which are arrays?

dpsanders:
Yes I meant variables which are arrays, so that I see `x[1]` inside the `Expression` itself.

ChrisRackauckas:
Yes, that is something we are planning.

HarrisonGrodin:
What would be the advantage of this feature? Couldn't you just use variables and substitute in arrays (rather than scalars) after the symbolic manipulation, or use an array of variables `[x,y,z]` if the values will be different? That way, we won't need separate machinery for manipulating variables (and can take advantage of existing array utilities).

ChrisRackauckas:
Points in manifolds can be written in a local basis that's different sized vectors. I think we want to be able to express these ideas and I'm not sure how to do that with only vectors of variables. This is down the line though.

asprionj:
I don't understand the use case of points in manifolds and a local basis (am not a mathematician :wink:), but for "everyday use", the arrays-of-variables approach could be sufficient, probably with some commodity functions/macros to remove some of the "ugly" stuff:
```julia
@parameters t
x = [Variable(Symbol("x$i"))(t) for i ∈ 1:10]
sys = ODESystem([    D(x[1])     ~ 1 - x[1];
                  @. D(x[2:end]) ~ x[1:end-1] - x[2:end] ])
f = ODEFunction(sys, x, [])
```

PR #4849, opened 2026-07-29
Fix ModelingToolkitBase method ambiguities

ChrisRackauckas-Claude:
## Summary

- resolve the `ODEProblem` constructor intersections with SciMLBase's fixed `FunctionWrapperSpecialize` constructor, including deprecated input forms
- resolve vector `renamespace` intersections by routing exact overlapping signatures through one implementation
- resolve `_split_helper` intersections at zero recursion depth without adding ambiguity-only fallback methods
- add constructor and vector-namespacing regressions and bump ModelingToolkitBase to 1.58.2

Closes the ambiguity portion of #4670. The 14 Aqua piracy findings documented there are intentionally separate and are not changed here.

## Root cause

Current clean master has 45 ambiguities:

- 34 intersections between ModelingToolkitBase's `ODEProblem{iip,spec}(::System, ...)` methods and SciMLBase's fixed `ODEProblem{iip,FunctionWrapperSpecialize}(f, ...)` constructor
- 2 local `_split_helper` intersections between the `Val{0}` fallback and the tuple/`BlockedArray` methods
- 9 intersections between `renamespace(::AbstractVector, x)` and more-specific second-argument methods

The `ODEProblem` issue is not caused by the currently registered SciMLBase 3.39.1. Repeating the clean-master Aqua scan against local SciMLBase 3.40.1 produces the same 45-method set.

## Validation

- `GROUP=SymbolicIndexingInterface JULIA_NUM_THREADS=2 timeout 3600 julia +1.12 --startup-file=no --project=lib/ModelingToolkitBase -e 'using Pkg; Pkg.test(; coverage=false)'`: **exit 0**; 64 + 1,824 + 118 assertions passed
- `timeout 3600 julia +1.12 --startup-file=no --project=lib/ModelingToolkitBase/test/qa -e 'using ModelingToolkitBase, Aqua; Aqua.test_ambiguities(ModelingToolkitBase); println("AQUA_AMBIGUITIES_PASS")'`: **exit 0**, `AQUA_AMBIGUITIES_PASS` (45 → 0 ambiguities)
- wrapped `lib/ModelingToolkitBase/test/namespacing.jl`: **21/21 passed**
- repository-wide Runic `--check`: **exit 0**
- `git diff --check`: **exit 0**

[...]

Issue #4670, opened 2026-06-25
QA: track pre-existing Aqua/JET/ExplicitImports findings (root + ModelingToolkitBase)


ChrisRackauckas-Claude:
The QA lanes for the ModelingToolkit root package and the `ModelingToolkitBase` sublibrary are currently red, and the conversion to the `run_qa` v1.6 form (PR forthcoming) surfaces every individual finding. To get the lanes green-or-broken without silencing anything, the conversion marks each finding as `@test_broken` (via `run_qa`'s `aqua_broken` / `jet_broken` / `ei_broken`). This issue tracks the underlying findings so the broken-markers can be dropped as they are fixed (each auto-flags an `Unexpected Pass` once resolved).

## Root `ModelingToolkit`

Aqua (`aqua_broken`):
- `undefined_exports` — `ModelingToolkit.Variable` is reexported but not defined in the namespace.
- `piracies` — deliberate definitions of SciMLBase problem constructors (`SCCNonlinearProblem(::System, op)`, ...) and a few MTKBase/Symbolics methods on upstream types.

JET (`jet_broken`):
- `report_package` re-parses the toplevel; the `@recompile_invalidations begin using StaticArrays; using DiffEqBase; ... end` blocks make it `require` transitive deps that are not directly resolvable, surfacing as toplevel errors.

ExplicitImports (`ei_broken`):
- `no_implicit_imports` / `no_stale_explicit_imports` — the `StructuralTransformations` submodule relies on a large block of implicit imports (whole packages: `BipartiteGraphs`, `Graphs`, ...) and carries a large block of stale explicit imports; the top-level module is also unanalyzable (dynamic `include` in `src/precompile.jl`).
- `all_qualified_accesses_are_public` / `all_explicit_imports_are_public` — dominated by names not (yet) declared `public` in their defining packages (Symbolics, SymbolicUtils, SciMLBase, StateSelection, ModelingToolkitBase, ModelingToolkitTearing, ...). Fix is upstream `public` declarations.

## `lib/ModelingToolkitBase`

Aqua (`aqua_broken`):
- `ambiguities` — method ambiguities.
- `unbound_args` — unbound type parameters (e.g. `_remake_buffer` with an unbound `P`).
- `undefined_exports` — `ModelingToolkitBase.Variable` reexported but undefined.
- `stale_deps` — `SimpleNonlinearSolve` declared but unused.
[...]

PR #4852, opened 2026-07-30
Fix collect_vars! after late invalidation on Julia 1.10

ChrisRackauckas-Claude:
Please ignore this PR until reviewed by @ChrisRackauckas.

## Summary

- route recursive four-argument `collect_vars!` calls from metadata collection through a non-inlined `Ref{Function}` dispatch barrier
- preserve the public downstream extension point: methods added to four-argument `collect_vars!` after package load are still dispatched dynamically
- add a regression that first compiles default collection, adds an unrelated method that triggers the Julia 1.10 invalidation, checks scalar and expression defaults, and then checks a late downstream `collect_vars!` method

## Why

Loading DynamicQuantities and then unrelated packages can invalidate the mutually recursive `collect_vars!` / `collect_var!` inference cycle on Julia 1.10. After that invalidation, parameters used only in defaults can disappear. In downstream initialization this manifested as a missing `Initial(x0)` parameter and as `y0` no longer being recognized as an unknown.

The first bad ModelingToolkit commit is `22000aa03d317353500c3c73ca01744d069530be` (`refactor: change collect_vars! for type-stability`); its parent `ca6eb0a92` is good. A package-load warmup is not durable because later method additions can invalidate the warmed specialization again.

The barrier is deliberately restricted to the recursive four-argument calls inside `collect_var!`. The explicit `depth = 0` preserves their previous default-keyword semantics, while the abstract function cell prevents inference from rebuilding the problematic mutual cycle. It also avoids reaching into dependency internals or changing public API.

## Performance

Fresh-process construction microbenchmarks, baseline -> patch:

| case | allocations | time |
| --- | --- | --- |
| plain variable | 1240.01 B -> 1240.01 B | 4228.6 ns -> 4214.2 ns |
| parameter default | 3056.01 B -> 3088.01 B | 11029.7 ns -> 11288.1 ns (+2.34%) |
| bounded metadata | 1896.01 B -> 1992.01 B | 10347.1 ns -> 10502.4 ns (+1.50%) |

This is system-construction work, not the nonlinear `solve!` path. The warmed ImplicitDiscreteSolve homotopy solve remains 0 B per solve in the separate cache benchmark.

## Local verification

- Julia 1.10 targeted `dq_units.jl`: initialization 1/1; invalidation regression 6/6
- Julia 1.12 targeted `dq_units.jl`: initialization 1/1; invalidation regression 6/6
- Julia 1.10 real late-JSON reproduction: `[json_p]` is collected both before and after the late method addition; late custom dispatch collects `[custom_p]`
- Julia 1.10 official `GROUP=InterfaceI Pkg.test("ModelingToolkitBase")`: 1498 passed, 5 existing broken, 0 failed/errored
- Julia 1.10 official `GROUP=Initialization`: the two collector errors are gone; 583 passed, 9 unrelated DelayDiffEq `DEOptions` errors, 14 existing broken. The nine resolver errors are tracked by JuliaRegistries/General#162781.
- Julia 1.12 official `GROUP=QA`: JET 54/54 passed. Aqua reports 45 ambiguities and 14 piracy methods; an exact clean-master `d39174937` run reproduces the same Aqua 9 passed / 2 failed result, so that baseline failure is being handled separately rather than hidden here.
- Runic 1.7.0 whole-repository `--check .`: exit 0
- `git diff --check`: clean

No tests were skipped, disabled, silenced, or loosened.

## Investigation/process notes

The work started from the downstream LinearSolve/ModelingToolkit initialization failures discovered while validating the ImplicitDiscreteSolve homotopy cache path. I reproduced the failure with frozen dependencies on Julia 1.10, separated the two collector errors from nine independent DelayDiffEq resolver errors, bisected the collector regression, minimized the invalidation trigger, compared direct calls, `invoke`, function wrappers, and a dynamic function cell, verified late downstream dispatch, benchmarked construction overhead, then ran the package groups and formatting checks above. The unrelated clean-master QA failures are being bisected in a separate worktree so this remains one focused fix.

### DIFF
--- a/lib/ModelingToolkitBase/src/utils.jl
+++ b/lib/ModelingToolkitBase/src/utils.jl
@@ -936,16 +936,14 @@ function collect_vars!(
     return nothing
 end
 
-# Break the inference cycle between the mutually recursive collectors. Julia 1.10
-# can otherwise miscompile the cycle after unrelated method additions. The
-# abstractly typed cell keeps downstream `collect_vars!` dispatch dynamic, while
-# the explicit keyword preserves metadata recursion's depth-zero semantics.
-const _COLLECT_VARS_DISPATCH = Ref{Function}(collect_vars!)
-
-@noinline Base.@constprop :none function _call_collect_vars!(
-        unknowns, parameters, expr, iv
-    )
-    return _COLLECT_VARS_DISPATCH[](unknowns, parameters, expr, iv; depth = 0)
+# Break the inference cycle between the mutually recursive collectors, which Julia
+# 1.10 can otherwise miscompile after unrelated method additions. Dispatching in the
+# latest world age also means a downstream `collect_vars!` method defined while this
+# call is already running is still found; because the four-argument fallback returns
+# `nothing`, missing it would silently drop parameters rather than error. The explicit
+# keyword preserves metadata recursion's depth-zero semantics.
+function _call_collect_vars!(unknowns, parameters, expr, iv)
+    return @invokelatest collect_vars!(unknowns, parameters, expr, iv; depth = 0)
 end
 
 """
--- a/lib/ModelingToolkitBase/src/utils.jl
+++ b/lib/ModelingToolkitBase/src/utils.jl
@@ -936,6 +936,18 @@ function collect_vars!(
     return nothing
 end
 
+# Break the inference cycle between the mutually recursive collectors. Julia 1.10
+# can otherwise miscompile the cycle after unrelated method additions. The
+# abstractly typed cell keeps downstream `collect_vars!` dispatch dynamic, while
+# the explicit keyword preserves metadata recursion's depth-zero semantics.
+const _COLLECT_VARS_DISPATCH = Ref{Function}(collect_vars!)
+
+@noinline Base.@constprop :none function _call_collect_vars!(
+        unknowns, parameters, expr, iv
+    )
+    return _COLLECT_VARS_DISPATCH[](unknowns, parameters, expr, iv; depth = 0)
+end
+
 """
     $(TYPEDSIGNATURES)
 
@@ -966,7 +978,7 @@ function collect_var!(unknowns::OrderedSet{SymbolicT}, parameters::OrderedSet{Sy
                 any(!SU.isconst, Iterators.drop(arguments(var), 1))
         )
         for arg in Iterators.drop(arguments(var), 1)
-            collect_vars!(unknowns, parameters, arg, iv)
+            _call_collect_vars!(unknowns, parameters, arg, iv)
         end
         var = arr
     end
@@ -975,7 +987,7 @@ function collect_var!(unknowns::OrderedSet{SymbolicT}, parameters::OrderedSet{Sy
     if iscalledparameter(var)
         callable = getcalledparameter(var)
         push!(parameters, callable)
-        collect_vars!(unknowns, parameters, arguments(var), iv)
+        _call_collect_vars!(unknowns, parameters, arguments(var), iv)
     elseif isparameter(var) || (iscall(var) && isparameter(operation(var)))
         push!(parameters, var)
     else
@@ -984,55 +996,55 @@ function collect_var!(unknowns::OrderedSet{SymbolicT}, parameters::OrderedSet{Sy
     # Add also any parameters that appear only as defaults in the var
     if hasdefault(var) && (def = getdefault(var)) !== missing
         if def isa SymbolicT
-            collect_vars!(unknowns, parameters, def, iv)
+            _call_collect_vars!(unknowns, parameters, def, iv)
         elseif def isa Num
-            collect_vars!(unknowns, parameters, def, iv)
+            _call_collect_vars!(unknowns, parameters, def, iv)
         elseif def isa Arr{Num, 1}
-            collect_vars!(unknowns, parameters, def, iv)
+            _call_collect_vars!(unknowns, parameters, def, iv)
         elseif def isa Arr{Num, 2}
-            collect_vars!(unknowns, parameters, def, iv)
+            _call_collect_vars!(unknowns, parameters, def, iv)
         elseif def isa CallAndWrap{Num}
-            collect_vars!(unknowns, parameters, def, iv)
+            _call_collect_vars!(unknowns, parameters, def, iv)
         elseif def isa CallAndWrap{Arr{Num, 1}}
-            collect_vars!(unknowns, parameters, def, iv)
+            _call_collect_vars!(unknowns, parameters, def, iv)
         elseif def isa CallAndWrap{Arr{Num, 2}}
-            collect_vars!(unknowns, parameters, def, iv)
+            _call_collect_vars!(unknowns, parameters, def, iv)
         elseif def isa Arr
-            collect_vars!(unknowns, parameters, def, iv)
+            _call_collect_vars!(unknowns, parameters, def, iv)
         elseif def isa CallAndWrap
-            collect_vars!(unknowns, parameters, def, iv)
+            _call_collect_vars!(unknowns, parameters, def, iv)
         else
-            collect_vars!(unknowns, parameters, def, iv)
+            _call_collect_vars!(unknowns, parameters, def, iv)
         end
     end
     # Add also any parameters that appear only in the bounds of the var
     if hasbounds(var)
         (lo, hi) = getbounds(var)
         if lo isa SymbolicT
-            collect_vars!(unknowns, parameters, lo, iv)
+            _call_collect_vars!(unknowns, parameters, lo, iv)
         elseif lo isa Num
-            collect_vars!(unknowns, parameters, lo, iv)
+            _call_collect_vars!(unknowns, parameters, lo, iv)
         elseif lo isa Arr{Num, 1}
-            collect_vars!(unknowns, parameters, lo, iv)
+            _call_collect_vars!(unknowns, parameters, lo, iv)
         elseif lo isa Arr{Num, 2}
-            collect_vars!(unknowns, parameters, lo, iv)
+            _call_collect_vars!(unknowns, parameters, lo, iv)
         elseif lo isa Arr
-            collect_vars!(unknowns, parameters, lo, iv)
+            _call_collect_vars!(unknowns, parameters, lo, iv)
         else
-            collect_vars!(unknowns, parameters, lo, iv)
+            _call_collect_vars!(unknowns, parameters, lo, iv)
         end
         if hi isa SymbolicT
-            collect_vars!(unknowns, parameters, hi, iv)
+            _call_collect_vars!(unknowns, parameters, hi, iv)
         elseif hi isa Num
-            collect_vars!(unknowns, parameters, hi, iv)
+            _call_collect_vars!(unknowns, parameters, hi, iv)
         elseif hi isa Arr{Num, 1}
-            collect_vars!(unknowns, parameters, hi, iv)
+            _call_collect_vars!(unknowns, parameters, hi, iv)
         elseif hi isa Arr{Num, 2}
-            collect_vars!(unknowns, parameters, hi, iv)
+            _call_collect_vars!(unknowns, parameters, hi, iv)
         elseif hi isa Arr
-            collect_vars!(unknowns, parameters, hi, iv)
+            _call_collect_vars!(unknowns, parameters, hi, iv)
         else
-            collect_vars!(unknowns, parameters, hi, iv)
+            _call_collect_vars!(unknowns, parameters, hi, iv)
         end
     end
     return nothing
--- a/lib/ModelingToolkitBase/test/dq_units.jl
+++ b/lib/ModelingToolkitBase/test/dq_units.jl
@@ -282,6 +282,10 @@ struct CollectorInvalidationExpression
     parameter::Symbolics.SymbolicT
 end
 
+struct CollectorSameWorldExpression
+    parameter::Symbolics.SymbolicT
+end
+
 function collect_default_parameters(var)
     unknowns = OrderedSet{Symbolics.SymbolicT}()
     parameters = OrderedSet{Symbolics.SymbolicT}()
@@ -326,4 +330,34 @@ end
 
     @test Symbolics.unwrap(scale_test) in collect_default_parameters(custom_default)
     @test Symbolics.unwrap(x0_test) in collect_default_parameters(x_test)
+end
+
+# Defines the downstream method from inside a running call, so the enclosing frame
+# keeps its old world age while collecting.
+function define_then_collect(var)
+    @eval function MT.collect_vars!(
+            unknowns::OrderedSet{Symbolics.SymbolicT},
+            parameters::OrderedSet{Symbolics.SymbolicT},
+            expr::CollectorSameWorldExpression,
+            ::Union{Symbolics.SymbolicT, Nothing};
+            depth = 0
+        )
+        push!(parameters, expr.parameter)
+        return nothing
+    end
+    return collect_default_parameters(var)
+end
+
+@testset "collect_vars! dispatches to same-world-age downstream methods" begin
+    @parameters same_world_test
+    @variables same_world_var(t)
+
+    target = MT.setdefault(
+        same_world_var, CollectorSameWorldExpression(Symbolics.unwrap(same_world_test))
+    )
+
+    # The four-argument fallback returns `nothing`, so failing to see the method here
+    # would drop the parameter silently instead of raising a `MethodError`.
+    @test Symbolics.unwrap(same_world_test) in define_then_collect(target)
+    @test Symbolics.unwrap(same_world_test) in collect_default_parameters(target)
 end
--- a/lib/ModelingToolkitBase/test/dq_units.jl
+++ b/lib/ModelingToolkitBase/test/dq_units.jl
@@ -2,6 +2,7 @@ using ModelingToolkitBase, OrdinaryDiffEq, JumpProcesses, DynamicQuantities
 using Symbolics
 import SymbolicUtils as SU
 using Test
+using DataStructures: OrderedSet
 MT = ModelingToolkitBase
 using ModelingToolkitBase: t, D
 @parameters τ [unit = u"s"] γ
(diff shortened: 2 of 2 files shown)


ChrisRackauckas-Claude:
Investigation scratchpad / execution record:

1. Reproduce the downstream initialization failure on frozen Julia 1.10 dependencies and separate collector failures from resolver failures. Completed: 2 collector errors; 9 independent DelayDiffEq `DEOptions` errors.
2. Bisect the collector regression. Completed: first bad `22000aa03d317353500c3c73ca01744d069530be`, good parent `ca6eb0a92`.
3. Minimize the invalidation and compare barriers without sacrificing the public four-argument `collect_vars!` extension point. Completed: the direct `Ref{Function}` barrier is the smallest durable option; `invoke` adds substantially more allocation and a function wrapper adds complexity without reducing allocation.
4. Add a late-invalidation and late-extension regression. Completed: 6/6 on Julia 1.10 and 6/6 on Julia 1.12.
5. Validate the actual downstream paths and benchmark construction cost. Completed: real late JSON and custom dispatch work; official Julia 1.10 `InterfaceI` passes 1498 tests; metadata construction overhead is reported in the PR body.
6. Run QA and formatting. Completed: JET 54/54 and whole-repository Runic exit 0. Aqua failures reproduce unchanged on exact clean master and are under a separate mandatory bisect/fix investigation.
7. Monitor CI and address only failures attributable to this focused diff. In progress.

ChrisRackauckas-Claude:
Clean-master QA classification is now fully local and status-checked: exact `origin/master` `d39174937`, Julia 1.12.6, official `GROUP=QA` exited 1 with JET 54/54 passing and Aqua 9/11 (45 ambiguities, 14 piracy methods). This matches the feature branch exactly. The ambiguity portion is already scoped separately in #4849; its current-master validation is in progress. The piracy portion remains separate. Neither baseline category is being suppressed or folded into this collector fix.

ChrisRackauckas-Claude:
CI classification: `SciMLSensitivity.jl/Core8/1` failed before running tests during `Pkg.develop` resolution. The log resolves General only through SciMLBase 3.39.1, while SciMLSensitivity 7.118.0 requires SciMLBase 3.40+, producing an unsatisfiable requirement. This is independent of the collector diff and is tracked/fixed by SciML/SciMLSensitivity.jl#1575.

ChrisRackauckas-Claude:
CI classification: `SBMLToolkit.jl/All/1` also fails before tests during `Pkg.test` re-resolution. Its downstream project pins `SBMLToolkitTestSuite = 0.0.5`, while current OrdinaryDiffEq/Sundials compatibility under General's SciMLBase 3.39.1 forces TestSuite 1.0+; the resolver reports no versions left. This is independent of the collector diff. Per repository policy, a separate clean-master reproduce/bisect/fix-or-issue investigation is now running; I will link its focused result here.

ChrisRackauckas-Claude:
Additional CI classification:

- `lib/ModelingToolkitBase [InterfaceII] / Julia pre` fails in `dde.jl` with the same incompatible `OrdinaryDiffEqCore.DEOptions` constructor already reproduced in the official Initialization group and tracked by JuliaRegistries/General#162781.
- `lib/ModelingToolkitBase [Extended] / Julia lts` reaches the distributed test, then worker 2 errors with `ArgumentError: cannot deserialize a dropped RuntimeGeneratedFunction; serialize it before calling drop_expr`. This is unrelated to the collector call path. A separate required clean-master Julia 1.10 reproduce/bisect/fix-or-issue investigation is now running; its result will be linked here.

ChrisRackauckas-Claude:
Completed-cluster confirmation from the actual logs:

- all three `Extended` jobs (Julia lts/current/pre) fail at the same distributed worker deserialization of a dropped `RuntimeGeneratedFunction`; the dedicated clean-master investigation covers the full cluster
- `InterfaceII` lts/pre both fail at the already-tracked DelayDiffEq `DEOptions` constructor mismatch
- QA current passes JET 54/54 and then reports exactly the clean-master Aqua baseline: 45 ambiguities and 14 piracy methods (Aqua 9/11)

None of these failures mention or exercise the new collector barrier/regression. #4849 handles the ambiguity class; #4670 retains the piracy ownership work.

ChrisRackauckas-Claude:
Distributed-cluster investigation update: exact clean `origin/master` `d39174937`, Julia 1.10.11, fresh depot, official `GROUP=Extended Pkg.test()` reproduces the worker-2 dropped-`RuntimeGeneratedFunction` deserialization error after BigSystem 100/100, C 1/1, and multithreading 1/1 pass. The dependency boundary is RuntimeGeneratedFunctions 0.5.23: 0.5.22 has the prior working behavior; commit `6550973d9290c906b1f61d37d416695e95e63a7c` removed dropped-RGF serialization and `483e2e938e1bfc5b5df6dc3aab0064c7005d08b6` released it. A separate public-API fix passes the RuntimeGeneratedFunctions package tests; exact MTK Extended validation is in progress.

ChrisRackauckas-Claude:
Direct CI confirmation: `sublibrary-ci / lib/ModelingToolkitBase [InterfaceI] / Julia lts` completed successfully. Actual log summary: **1498 passed, 5 existing broken, 1503 total**, followed by `Testing ModelingToolkitBase tests passed`. This matches the local Julia 1.10 official-group result and includes the new late-invalidation regression.

ChrisRackauckas-Claude:
The direct `ModelingToolkitBase [InterfaceI]` matrix is now green on **all three Julia lanes** (lts/current/prerelease). Each actual job log reports **1498 passed, 5 existing broken, 1503 total**, followed by `Testing ModelingToolkitBase tests passed`. This includes the new collector invalidation regression across the supported compiler range.

ChrisRackauckas-Claude:
Downstream `SBMLToolkit` failure in job 90790793844 is confirmed independent of this PR and has a validated fix in SciML/SBMLToolkit.jl#223.

I reproduced the failure on the exact clean ModelingToolkit base `d39174937` with SBMLToolkit `7dbc7bca71211d97ba2997b2c38a9f6d69c21827`, Julia 1.12.6, an isolated depot, and the same fixed General registry snapshot (`cf87b2e4a16d9b7ba710aaac144d607d0bf724df`, tree `31befd7172041c2f95b752ca5d7147232c79da23`). The hosted-shaped `GROUP=All` workflow exits 1 during Pkg resolution: SBMLToolkit pins `SBMLToolkitTestSuite = "0.0.5"`, while current SciMLBase selects OrdinaryDiffEq 7 and Sundials 6, which require the 1.x test-suite line.

A first-parent compatibility bisect places the stale edge at SBMLToolkit commit `753fbc269868178401b9227a850203971412b220` (SciML/SBMLToolkit.jl#199), which first advertised OrdinaryDiffEq 7 / ModelingToolkit 11 while retaining the legacy test-suite constraint.

I then reran the exact workflow on clean ModelingToolkit `d39174937` with SBMLToolkit PR #223 head `18eb314e9c15114f029c78f6f6355882ea7aedda`. It selected `SBMLToolkitTestSuite 1.1.2`, `OrdinaryDiffEq 7.2.0`, `Sundials 6.4.2`, and `SciMLBase 3.39.1`, exited 0, and reported `Testing SBMLToolkit tests passed` (108/108 assertions across all seven `Core` test files). Runic `--check` also exits 0 on that head.

So this downstream check should clear once SciML/SBMLToolkit.jl#223 is merged/released; no collector change is implicated.

ChrisRackauckas-Claude:
The Extended/LTS `Distributed Test` failure is unrelated to this PR's collector changes. I reproduced it from clean ModelingToolkit master `d39174937fead779b29fa5baa50ba975adbde8c9` with the official `GROUP=Extended` command and traced it to RuntimeGeneratedFunctions v0.5.23.

Controls on the same clean master / Julia 1.10 setup:

- v0.5.23: reproduces `ArgumentError: cannot deserialize a dropped RuntimeGeneratedFunction`
- pinned v0.5.22: full Extended group passes
- local fixed v0.5.23: full Extended group passes, including `Distributed Test`

The first bad RGF source commit is `6550973d9290c906b1f61d37d416695e95e63a7c` (#136); #137 then changed its planned v0.6.0 release to v0.5.23 despite the behavior break. The focused fix and cross-process regression are in SciML/RuntimeGeneratedFunctions.jl#138. No ModelingToolkit workaround should be needed once that fix is released.

ChrisRackauckas-Claude:
The newly terminal `StructuralIdentifiability.jl/Core/1/` failure is already covered by focused draft [StructuralIdentifiability #536](https://github.com/SciML/StructuralIdentifiability.jl/pull/536); no duplicate PR is needed.

Exact [#4852 downstream job](https://github.com/SciML/ModelingToolkit.jl/actions/runs/30517596065/job/90790793787) reaches `Core/y_saturation.jl` and errors at `output_saturation.jl:28` with `UndefVarError: generators not defined`, including Julia's ambiguity hint for the competing AbstractAlgebra/Nemo/RationalFunctionFields bindings. #536 independently reproduced and bisected that clean-master dependency boundary, replaces the undocumented ambiguous helper with the generator vector already supplied to the field constructor, and adds a rational-output regression. Its local `GROUP=Core`, `GROUP=QA`, and Runic checks passed; all 11 hosted checks are green. The PR remains draft for @ChrisRackauckas review.

ChrisRackauckas-Claude:
Catalyst's stale expected-broken follow-up has now merged as [Catalyst #1512](https://github.com/SciML/Catalyst.jl/pull/1512), and I reran the complete downstream group locally against this collector branch.

Exact validation:

- Catalyst master `79144ad28f49b594f84965c58be290be51041f6e`
- ModelingToolkitBase from this PR, `6f0a545a9d57ffec07fce3d2014e1818d19dcb3b`
- Julia 1.12.6
- registered graph selected SciMLBase 3.39.1, OrdinaryDiffEq 7.2.0, and ModelingToolkitBase's local source path
- `GROUP=Modeling julia +1.12 --project=. -e 'using Pkg; Pkg.test()'`

Terminal result: exit 0, `Testing Catalyst tests passed`. The decisive summary is:

```text
Test Summary:     | Pass  Broken  Total     Time
Conservation Laws |  249      17    266  3m38.9s
```

The complete official group ran for 1829.82 seconds and also cleared the former prerequisites, including Symbolic Stoichiometry 30/30 and DSL Options 317 pass / 1 existing broken. This confirms the recorded `Catalyst/Modeling/1` failure is stale relative to merged Catalyst master, not a collector regression. I am rerunning only that failed downstream job.

ChrisRackauckas-Claude:
The `MethodOfLines.jl/MOL_Interface1/1/` failure in [job 90790793860](https://github.com/SciML/ModelingToolkit.jl/actions/runs/30517596065/job/90790793860) is independent of this PR's collector change and is now covered by focused draft [PDEBase #102](https://github.com/SciML/PDEBase.jl/pull/102).

I reproduced it with exact clean ModelingToolkit base `d39174937`, MethodOfLines `5ec02a46`, Julia 1.12.6, a fresh depot, and the hosted `GROUP=MOL_Interface1` / `Pkg.test(coverage=true)` workflow. Registered PDEBase 0.1.31 exits 1 with the sole error at `MOLtest1.jl:231` (`Array u`): `MethodError: operation(::Symbolics.Arr{Num,1})` from `PDEBase.get_ops`.

This is a PDEBase behavioral regression, not a missing Symbolics `operation(::Arr)` method. PDEBase #98 rewrote indexed-array flattening; General 0.1.30 is the pre-regression tree (`5b917e0`), while 0.1.31 is the bad tree (`97c9de5`). The actual fix and regression test already merged as [PDEBase #101](https://github.com/SciML/PDEBase.jl/pull/101), using the public owning-module `SymbolicUtils.Rewriters` API, but master still declared 0.1.31 and therefore could not be resolved downstream.

On the same downstream manifest, changing only PDEBase to current master with the 0.1.32 bump makes the official group exit 0 (`MOL_Interface1` 12m50.8s; `MethodOfLines tests passed`). PDEBase Core also passes 12 + 11 + 18 + 46 tests, QA passes 20/20, and Runic is clean. No ModelingToolkit workaround is needed; this check should clear once PDEBase #102 is merged and 0.1.32 is registered.

ChrisRackauckas-Claude:
Terminal cluster correction and exact count:

- all 12 root/sublibrary `Initialization` and `InterfaceII` failures (lts/current/pre in each matrix), plus
- `SciMLBase.jl/SymbolicIndexingInterface/1/`

fail with the same `OrdinaryDiffEqCore.DEOptions` constructor mismatch from the incompatible registered DelayDiffEq/Core pairing. I downloaded and checked each of the 13 completed job logs; this includes the current-Julia lanes, not just lts/pre. This cluster is independent of the collector diff and is covered by [General #162781](https://github.com/JuliaRegistries/General/pull/162781).

One correction to the earlier shorthand: the SciMLBase/SymbolicIndexingInterface job reaches tests and fails at `DEOptions`; it is not a resolver failure. The actual resolver-only failure is `SciMLSensitivity.jl/Core8/1/`, tracked by [SciMLSensitivity #1575](https://github.com/SciML/SciMLSensitivity.jl/pull/1575).

ChrisRackauckas-Claude:
MethodOfLines cluster is now fully validated:

- [PDEBase #102](https://github.com/SciML/PDEBase.jl/pull/102) fixes the release boundary for both `MOL_Interface1` and `MOL_Interface2` failures.
- Exact local controls against clean ModelingToolkit `d39174937` pass for both groups; `MOL_Interface2` reports 5 pass, 8 existing broken, 13 total, followed by `Testing MethodOfLines tests passed`.
- #102 hosted CI is terminal with 23 successes. Every MethodOfLines downstream check is green.
- Its sole hosted failure is PDEBase clean-master documentation compat drift, independently reproduced/bisected to `97c9de5` and fixed in focused draft [PDEBase #103](https://github.com/SciML/PDEBase.jl/pull/103). #103 has local Core 88/88 on Julia 1.12, Core 73/73 on Julia 1.10, QA 20/20, exact docs install, full Documenter build, and whole-repository Runic all passing.

Thus both #4852 MethodOfLines failures are independent of the collector diff and have direct local plus hosted coverage.

ChrisRackauckas-Claude:
The independent ModelingToolkitBase BVP downgrade failure is now covered by focused draft #4853.

Exact clean-master control (`d39174937`, Julia 1.10.11, the same `julia-downgrade-compat` action revision and `alldeps` workflow) exits 1 in the official `GROUP=InterfaceII` run with the hosted pair of `Float64(::ForwardDiff.Dual)` errors: Lotka–Volterra and BVP cost-function compilation. Its terminal BVP summary is 13 passed, 2 errored, and 5 existing broken.

The root cause is the pre-1.17.3 MIRK solution-evaluation buffer, fixed upstream by SciML/BoundaryValueDiffEq.jl#568 and released in `BoundaryValueDiffEqMIRK` 1.17.3. #4853 raises only the ModelingToolkitBase compatibility floor to that first good release.

With that one-line floor change, the exact downgrade workflow selects MIRK 1.17.3 and the full official `InterfaceII` group exits 0 (`Optimal Control + Constraints Tests`: 22 passed, 6 existing broken; `Testing ModelingToolkitBase tests passed`). The focused BVP file, whole-repository Runic, and exact clean-vs-feature QA comparison are also recorded in #4853. No collector-path change is implicated.

ChrisRackauckas-Claude:
The ModelOrderReduction downstream failure is isolated and fixed in draft PR [SciML/ModelOrderReduction.jl#169](https://github.com/SciML/ModelOrderReduction.jl/pull/169).

Root cause: PDEBase 0.1.31 correctly stopped creating an incidental `symbolic_discretize` binding when it moved from `using SciMLBase` to explicit imports. ModelOrderReduction was relying on MethodOfLines to expose that undeclared binding. The fix imports the exported/public/documented owner API directly from SciMLBase and adds SciMLBase as a direct test/docs dependency.

I reproduced the original `UndefVarError` on clean ModelingToolkit `d39174937` + ModelOrderReduction `4c1fc0b5`, then reran that exact downstream workflow with the patch: DataReduction 15/15, DEIM 4/4, utils 11/11, natural exit 0. ModelOrderReduction's isolated Core, QA, full docs build, and Runic also pass.

ChrisRackauckas-Claude:
## Consolidated terminal CI ledger

The hosted matrix is terminal at **64 success / 26 failure / 3 skipped**. The 26 failures are fully accounted for; none points to the collector invalidation barrier in this PR.

1. **13 initialization/interface/re-export failures — DelayDiffEq/Core release metadata.** Fresh resolution selects an incompatible DelayDiffEq/Core pairing and fails at `DEOptions`. JuliaRegistries/General#162781 contains the focused compat correction, is open and non-draft, and its Julia 1.3–1.12 registry checks are green.
2. **3 ModelingToolkitBase Extended failures — RuntimeGeneratedFunctions serialization boundary.** SciML/RuntimeGeneratedFunctions.jl#138 is the focused draft fix; exact clean controls show RGF 0.5.22 pass, 0.5.23 fail, patched source pass.
3. **2 QA failures — clean-master baselines.** The focused Aqua correction is SciML/ModelingToolkit.jl#4849; the pre-existing JET baseline is tracked by SciML/ModelingToolkit.jl#4670.
4. **1 Catalyst/Modeling failure.** SciML/Catalyst.jl#1512 is merged; the exact downstream Modeling group passed locally against this collector branch.
5. **2 MethodOfLines failures.** PDEBase owner-import/docs fixes #102/#103 and MethodOfLines owner-import fix #613 are merged. PDEBase 0.1.32 and MethodOfLines 0.11.19 are both registered via JuliaRegistries/General#162815 and #162817. At the scheduled 07:53 EDT audit, General's trees exactly matched the source commits and a fresh Julia 1.10 environment actually loaded registered MethodOfLines 0.11.19 + PDEBase 0.1.32 successfully.
6. **1 ModelOrderReduction failure.** SciML/ModelOrderReduction.jl#169 is the focused draft public-owner import fix. Its sole hosted LTS red is the now-stale MethodOfLines 0.11.18/PDEBase 0.1.32 boundary; the registered replacement boundary above loads locally. The current token lacks admin permission to rerun only that job, so no empty commit was used to churn CI.
7. **1 SBMLToolkit failure.** SciML/SBMLToolkit.jl#223 is the focused draft fix; the exact downstream suite passed 108/108 locally against clean ModelingToolkit plus that branch.
8. **1 SciMLSensitivity failure.** JuliaRegistries/General#162820 corrects the factually wrong NonlinearSolveBase 2.38 cap and is green across Julia 1.3–1.12. Resolution then correctly stops at the *valid* OrdinaryDiffEqCore 4.11 cap: that registered release privately accesses `SciMLBase.islinear`. SciML/OrdinaryDiffEq.jl#4066 is the source/future-release path for fixed ODECore 4.12; the 4.11 cap must not be removed.
9. **1 StructuralIdentifiability failure.** SciML/StructuralIdentifiability.jl#536 is the focused draft fix and its hosted matrix is green.
10. **1 ModelingToolkitBase downgrade failure.** SciML/ModelingToolkit.jl#4853 is the focused MIRK floor correction. Exact clean bad/fixed controls and the full local InterfaceII group pass with the correction. Its hosted downgrade job reached the formerly failing target at 22 pass / 6 broken, then was later terminated by a self-hosted runner shutdown while entering QA precompile; the endpoint annotation is cancellation, not a new assertion failure. The independent covered NeuralPDE failure exposed on that matrix is tracked with an exact clean-master reproducer at SciML/NeuralPDE.jl#1095.

So the answer to “are all independent follow-ups merged?” is **no**. The Catalyst/PDEBase/MethodOfLines source and registration pieces are merged; the focused RGF, MTK QA/downgrade, MOR, SBML, StructuralIdentifiability, OrdinaryDiffEq, General compat, and related release steps remain open as listed above.

AayushSabharwal (commented):
If the goal is to make it dynamic dispatch, why not just make the wrapper method use `@invokelatest` instead of this approach with the global Ref?

ChrisRackauckas-Claude (commented):
Good call — switched to `@invokelatest` in 04b6dcd0. It isn't only cleaner; it's actually more correct here, and I have measurements for both claims.

## Performance: no difference

Julia 1.10.11, three interleaved reps of 20 000 samples each, fresh process per rep, clean master `d3917493` as the baseline:

| variant | `default` case | `bounds` case |
| --- | --- | --- |
| master (no barrier) | 4032 B / 85 allocs | 2872 B / 71 allocs |
| `Ref{Function}` | 4064 B / 87 allocs | 2968 B / 77 allocs |
| `@invokelatest` | 4064 B / 87 allocs | 2968 B / 77 allocs |

Allocations are byte-identical between the two barriers. Times are indistinguishable — min over reps 11.57/11.72 µs (`default`) and 10.93/10.94 µs (`bounds`) for `Ref`/`@invokelatest`, against master's own 11.64 µs and 10.87 µs. The run-to-run spread on both exceeds the gap between them. So there was never a perf argument for the `Ref`.

## Correctness: `@invokelatest` is strictly better

The `Ref[]` load is an ordinary dynamic dispatch, so it resolves in the **caller's** world age. `@invokelatest` resolves in the latest one. That matters when a downstream `collect_vars!` method is defined while a collection is already running — e.g. an `@eval` inside a component or macro that then builds a system in the same top-level call:

```
Ref{Function}  → same-world-age call : []      SILENTLY DROPPED
@invokelatest  → same-world-age call : [lp]    collected
```

The silent part is what makes this worth caring about: the four-argument `collect_vars!` has a catch-all fallback returning `nothing`, so a method the call can't see doesn't `MethodError` — it just loses the parameter. That is exactly the failure mode this PR exists to fix, so the `Ref` version left a variant of the bug reachable. I added a regression test for it (`collect_vars! dispatches to same-world-age downstream methods`) and confirmed it fails against the `Ref` implementation and passes with `@invokelatest`.

I also dropped `@noinline` and `Base.@constprop :none`: `@invokelatest` is already opaque to inference, and the invalidation regression still passes 6/6 without them. That reduces the whole barrier to:

```julia
function _call_collect_vars!(unknowns, parameters, expr, iv)
    return @invokelatest collect_vars!(unknowns, parameters, expr, iv; depth = 0)
end
```

No global mutable state and no compiler-annotation scaffolding to explain.

## Verification

- Julia 1.10.11: invalidation regression 6/6, new world-age regression 2/2. Clean master `d3917493` fails the invalidation regression 2 passed / 4 failed, so the reproduction is real.
- Julia 1.12.6: both 6/6 and 2/2; world-age case collected.
- Official `GROUP=InterfaceI` on Julia 1.10: 1497 passed, 5 existing broken, 3 failed. All three failures are the same `symbolic_events.jl` callback-accuracy assertion (lines 729/846/866, `2.635e-4` against a `1e-4` threshold). An exact clean-master `d3917493` control in the identical environment fails at the same three lines with the same values, so they are pre-existing and unrelated to the collector.
- Runic 1.7.0 whole-repository `--check .`: exit 0. `git diff --check`: clean.

No tests were skipped, disabled, silenced, or loosened.

One unrelated note for anyone reproducing locally on Julia 1.10: `Pkg.test()` on ModelingToolkitBase currently dies during resolution with ``BracketingNonlinearSolve depends on ChainRulesCore, but no such entry exists in the manifest``. `ChainRulesCore` is a *weakdep* of BracketingNonlinearSolve (the multi-trigger `BracketingNonlinearSolveChainRulesCoreExt`), correctly recorded in General's `WeakDeps.toml`, but Pkg 1.10 writes it into the manifest's hard-dep table without installing it. It reproduces on clean master and on Julia 1.12's `Pkg.develop`, and it is fixed by adding `ChainRulesCore` to the environment explicitly. Purely a local tooling issue, no bearing on this PR.

ChrisRackauckas (approved):

PR #4853, opened 2026-07-30
Require Dual-safe BoundaryValueDiffEqMIRK

ChrisRackauckas-Claude:
**Please ignore this PR until it has been reviewed by @ChrisRackauckas.**

## Summary

- raise ModelingToolkitBase's `BoundaryValueDiffEqMIRK` compatibility floor from 1.7.0 to 1.17.3
- keep downgrade-compatible dependency graphs on the first MIRK release with Dual-safe evaluation buffers

## Root cause

The ModelingToolkitBase `InterfaceII` downgrade job in #4852 admitted a MIRK tree older than 1.17.3. That implementation reused `Float64` solution-evaluation storage when the nonlinear solve evaluated a `ForwardDiff.Dual` iterate, producing `MethodError: no method matching Float64(::ForwardDiff.Dual)` in both the Lotka–Volterra BVP and BVP cost-function tests.

SciML/BoundaryValueDiffEq.jl#568 changed MIRK evaluation storage to select the iterate element type and shipped that fix in 1.17.3. ModelingToolkit.jl#4843 restored the real BVP assertions that expose the old implementation, so the ModelingToolkitBase compatibility bound needs to express that runtime requirement.

Latest-compatible environments are unaffected; this changes the old-version boundary selected by downgrade testing and by users whose environments would otherwise admit an affected MIRK release.

## Validation

- Reproduced the [original downgrade failure](https://github.com/SciML/ModelingToolkit.jl/actions/runs/30517596458/job/90790868423) from clean `d39174937fead779b29fa5baa50ba975adbde8c9`, Julia 1.10.11, and the exact `julia-actions/julia-downgrade-compat@2c401f8b47edbcafc2e1edbff8e71f33dce71822` `alldeps` workflow. The official `GROUP=InterfaceII` run exited 1 with the same two Dual-conversion errors: 13 passed, 2 errored, and 5 existing broken checks in `Optimal Control + Constraints Tests`.
- Repeated the exact downgrade workflow with this compatibility floor. It selected `BoundaryValueDiffEqMIRK` 1.17.3, and the official `GROUP=InterfaceII` run exited 0 with 22 passed and 6 existing broken checks in `Optimal Control + Constraints Tests`, followed by `Testing ModelingToolkitBase tests passed`.
- Ran `bvproblem.jl` directly in the exact downgraded test sandbox: exit 0; Lotka–Volterra 4/4, Pendulum 4/4, boundary compilation 2/2, constraints 6 passed plus 3 existing broken, cost compilation 6 passed plus 2 existing broken, and parameter estimation 1 existing broken.
- Ran whole-repository Runic 1.7.0: exit 0.
- Ran `GROUP=QA` on the feature branch and exact clean master with Julia 1.10.11 and the same depot/registry snapshot. Both passed JET 54/54 and reported the identical pre-existing Aqua baseline (46 ambiguity signatures and 14 piracy signatures); the sorted signature sets match exactly. Those independent QA fixes remain separate in #4849 and #4670.

The existing BVP tests are the regression coverage for this compatibility correction; no duplicate test or workaround is added.


ChrisRackauckas-Claude:
CI classification: the independent `ModelOrderReduction/Core` failure is addressed by SciML/ModelOrderReduction.jl#169. Its exact clean-MTK-master downstream Core reproduction passes 30/30 at commit `80e1f75688eb8cc3827ab2f48ae25b0388a1c19a`; it is unrelated to the BoundaryValueDiffEqMIRK compatibility-floor change here.

ChrisRackauckas-Claude:
## Terminal hosted-CI update

The matrix is terminal at **62 success / 28 failure / 3 skipped**. The last pending check, `downgrade-sublibraries / test (lib/ModelingToolkitBase)`, did not report a test assertion failure at its endpoint:

- the target `Optimal Control + Constraints Tests` that fails on clean master completed **22 pass / 6 broken** on this branch;
- the job continued through later suites and had entered the QA environment precompile;
- at 2026-07-30 11:28:43 UTC the self-hosted runner reported `The runner has received a shutdown signal`, Julia received SIGTERM, and the check annotation became `The operation was canceled.`

The log also exposes independent `LinearSolveKrylovKitExt` / `LinearSolveArnoldiMethodExt` precompile errors involving `EigenvalueProblem` in later isolated test environments. Their surrounding completed testsets continued to pass, so they are not evidence against this one-line MIRK floor correction, but they need a separate dependency/depot-boundary audit rather than being hidden here.

The downstream `NNPDE1` failure is likewise independent and now has an exact clean-master, coverage-sensitive reproducer at SciML/NeuralPDE.jl#1095. Clean master and this PR's merge tree produce byte-identical covered trajectories and the same failure; both pass without coverage.

No branch change or test suppression was made in response to either independent failure.

ChrisRackauckas-Claude:
The `NeuralPDE.jl/NNPDE1/1` failure is independent of this PR.

I reproduced the hosted failure on clean ModelingToolkit master
(`61cc6bca692a208124eec413c95c2ab73fd8388f`) using Julia 1.12.6, the hosted
outer testset RNG, and coverage. Clean master and this PR's merge produced
byte-for-byte identical callback values, terminal loss, prediction vector, and
failure:

- initial loss `3.3244116222205697`
- terminal loss `4.502195204451791e-8`
- prediction L2 error `1.2247332813759246e-4`
- unchanged `atol = 1e-4` array assertion fails

Without coverage, the same RNG passes on both clean master and the PR merge,
again with identical output. The PR's only code-tree difference from its base
is the `BoundaryValueDiffEqMIRK` test compat in ModelingToolkitBase, and that
package is absent from the NeuralPDE graph.

The untouched official clean-master `GROUP=NNPDE1` run with
`Pkg.test(coverage=true)` also passed 22/22 across all eight files in 59m10.5s;
it selected a different random initialization. The recorded hosted RNG still
fails deterministically under coverage.

Tracked separately in https://github.com/SciML/NeuralPDE.jl/issues/1095.

PR #4843, opened 2026-07-29
test: restore working MTKBase BVP assertions

ChrisRackauckas-Claude:
Ignore this PR until it has been reviewed by @ChrisRackauckas.

## Summary

- restore the four ModelingToolkitBase BVP solve paths that currently succeed
- assert their solution values and successful return codes instead of passing solution objects to Test.@test_broken
- leave the genuinely broken parameter-estimation cases marked broken

## Cause

Commit f71cc7d5ae805e9c57bf86a7867dfbc7e538d201 wrapped four successful BVP solves in Test.@test_broken. A successful solve returns a solution object, not a Boolean, so Test reports each case as an error rather than a broken test.

This is the same failure in BoundaryValueDiffEq clean master run https://github.com/SciML/BoundaryValueDiffEq.jl/actions/runs/30342212061 and PR #541. The PR #541 shooting changes are unrelated.

## Verification

Using Julia 1.11 and the official ModelingToolkitBase InterfaceII group with local BoundaryValueDiffEqCore, BoundaryValueDiffEqMIRK, and BoundaryValueDiffEqAscher:

- clean ModelingToolkit master at a35a814d: BVP tests report 14 passed, 4 errored, 6 broken; the group fails
- this branch: BVP tests report 22 passed, 6 broken; the complete InterfaceII group passes in 3707.7 seconds

A bisect from 3c2e4a752 to a35a814d identified f71cc7d5 as the first bad commit.

The changed file passes Runic and git diff --check. Whole-repository Runic currently fails on unchanged master in src/systems/clock_inference.jl; that independent FormatCheck regression is intentionally not mixed into this focused PR.

PR #4867, opened 2026-08-04
test: several test fixes

AayushSabharwal:
## Checklist

### DIFF
--- a/.github/workflows/SublibraryCI.yml
+++ b/.github/workflows/SublibraryCI.yml
@@ -24,4 +24,6 @@ jobs:
     uses: "SciML/.github/.github/workflows/sublibrary-project-tests.yml@v1"
     with:
       check-bounds: auto
+    env:
+      JULIA_PKG_PRECOMPILE_AUTO: "0"
     secrets: "inherit"
--- a/.github/workflows/Tests.yml
+++ b/.github/workflows/Tests.yml
@@ -27,4 +27,6 @@ jobs:
     with:
       coverage-directories: "src,ext,lib/ModelingToolkitBase/src,lib/ModelingToolkitBase/ext"
       cache: false
+    env:
+      JULIA_PKG_PRECOMPILE_AUTO: "0"
     secrets: "inherit"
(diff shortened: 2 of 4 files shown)


ChrisRackauckas-Claude:
The reason no `Tests` / `Sublibrary CI` checks are showing up on this PR is the `env:` block added in 0d3a55fdee09460acbb987848fe8967d97dfcbcc, not path filtering.

A job that calls a reusable workflow with `uses:` may not carry an `env:` map, and the failure mode is not "the env is ignored" — the whole workflow file becomes invalid and GitHub creates zero jobs:

```
$ actionlint .github/workflows/Tests.yml .github/workflows/SublibraryCI.yml
.github/workflows/Tests.yml:30:5: when a reusable workflow is called with "uses", "env" is not available.
  only following keys are allowed: "name", "uses", "with", "secrets", "needs", "if", and "permissions" in job "tests" [syntax-check]
.github/workflows/SublibraryCI.yml:27:5: ... in job "sublibrary-ci" [syntax-check]
```

Both runs exist but are listed by file path with no jobs — https://github.com/SciML/ModelingToolkit.jl/actions/runs/30901431064 reports "This run likely failed because of a workflow file issue."

For the record on the path-filtering theory: sublibrary test-only changes *do* select the sublibrary. `compute_affected_sublibraries.jl` puts any `lib/<pkg>/**` change into the `direct` set (only reverse-dep propagation is gated on `src/`/`Project.toml`), and the root `grouped-tests.yml` matrix is not diff-filtered at all. Piping `lib/ModelingToolkitBase/test/symbolic_events.jl` into `--projects-matrix` yields the full 25-cell MTKBase expansion, and #4843 (which touched only `lib/ModelingToolkitBase/test/bvproblem.jl`) ran both `Tests` and `Sublibrary CI`.

Two PRs:

- #4873 — reverts the two `env:` blocks against this branch, so CI runs again. `actionlint .github/workflows/*.yml` exits 0.
- https://github.com/SciML/.github/pull/125 — adds an `auto-precompile` input to `tests.yml` (job-level, so it covers `julia-buildpkg` as well as `julia-runtest`) and forwards it from `grouped-tests.yml` and all four `sublibrary-project-tests.yml` shards. Once that merges and `v1` is retagged, this becomes `auto-precompile: false` in `with:`. `v1` only moves on a `vX.Y.Z` tag push, so that step needs a release cut.

Both are drafts pending @ChrisRackauckas review.

PR #4879, opened 2026-08-05
fix: make the JET QA lane pass (include paths + 9 uncovered findings)

ChrisRackauckas-Claude:
> **Ignore until reviewed by @ChrisRackauckas.** Opened as a draft; do not merge on the strength of the description alone.

## What was broken

`GROUP=QA` fails in the JET check, and it failed with *toplevel load errors* rather than type errors:

```
Quality Assurance: JET-test failed at .../JET/src/JETBase.jl:1046
  Expression: (JET.report_package)(ModelingToolkit; toplevel_logger = nothing, target_defined_modules = true)
  ═════ 3 toplevel errors found ═════
  ┌ @ src/ModelingToolkit.jl:133
  │ SystemError: opening file ".../test/qa/linearization.jl": No such file or directory
  │  [1] include(mod::Module, _path::String)
  │  [5] recompile_invalidations(__module__::Module, expr::Any) @ PrecompileTools
  ┌ @ src/ModelingToolkit.jl:156
  │ UndefVarError: `StructuralTransformations` not defined in `ModelingToolkit`     (x2)
```

## Root cause

`PrecompileTools.@recompile_invalidations` does **not** splice its body into the enclosing
scope. It expands to

```julia
:(recompile_invalidations($__module__, $(QuoteNode(expr))))
```

i.e. an ordinary *runtime* function call whose body runs through `Core.eval(__module__, expr)`.
So everything inside the block stops being lexically top level:

* `Base.include(mod, "linearization.jl")` resolves the relative path against the *caller's*
  `SOURCE_PATH` (there is no active include stack), not against `src/`. Under `Pkg.test()`
  that is `test/qa/`, hence the `SystemError`.
* Static analyzers cannot see through `Core.eval`, so JET's `handle_include` (which resolves
  includes against `dirname` of the analyzed file) is never reached.
* The failed include leaves the `StructuralTransformations` submodule undefined, so the
  `@reexport using .StructuralTransformations` on the next line throws too.

Minimal reproduction, no ModelingToolkit involved — a 6-line package:

```julia
module JetMini
using PrecompileTools
@recompile_invalidations begin
    include("helper.jl")
end
end
```

```
julia> JET.report_package(JetMini; target_defined_modules=true)
═════ 1 toplevel error found ═════
┌ @ .../JetMini/src/JetMini.jl:3
│ SystemError: opening file ".../<cwd>/helper.jl": No such file or directory
│  [5] recompile_invalidations(__module__::Module, expr::Any) @ PrecompileTools
```

Removing the `@recompile_invalidations` wrapper makes it report `No errors detected`.

A package's own `include` paths must not depend on who loads it, so the first commit hoists
the include block back out of the macro. The three `@recompile_invalidations` blocks that
wrap `using`/`import` statements are untouched — only the block wrapping `include`s is
moved, and `lib/ModelingToolkitBase` already only wraps `using`/`import`.

## What that uncovered

With the module body analyzable again, `JET.test_package(ModelingToolkit; target_defined_modules = true)`
reports 9 errors that the load failure had been masking. The second commit fixes all of them:

1. **Genuine bug.** `Base.showerror(::IO, ::DifferentiatedVariableNotUnknownError)` calls
   `expected_scope_depth`, which has never been defined anywhere (introduced undefined in
   811196b16). Printing the error threw `UndefVarError`. Now defined next to the error type,
   mirroring `check_scope_depth`'s `-1` sentinel for `GlobalScope`, with a unit test.
2. `src/systems/alias_elimination.jl` — `𝑠neighbors(state.structure.solvable_graph, e)` on a
   `Union{Nothing, BipartiteGraph}`. `find_eq_solvables!` on the previous line writes into that
   graph, so it is populated; assert it.
3. `src/structural_transformation/pantelides.jl` — `Differential(get_iv(sys))` where `get_iv`
   is `Union{Nothing, SymbolicT}`. Same `::SymbolicT` assertion `alias_elimination.jl` already uses.
4. `src/linearization.jl` — `build_explicit_observed_function(...)` is
   `Union{Expr, GeneratedFunctionWrapper}`; with `expression = Val{false}` it is always the
   wrapper.
5. `src/problems/sccnonlinearproblem.jl` — `subset_unknowns_observed(get_index_cache(sys), ...)`
   with a possibly-`nothing` cache. Handled the way the sibling code at line 123 already does.
6. `src/problems/sccnonlinearproblem.jl` — `pvars` from `reorder_parameters(sys)` is
   `Union{Vector{SymbolicT}, Vector{Vector{SymbolicT}}}`; the nested form only occurs with
   `flatten = false`, which this call does not use.
7. `src/systems/if_lifting.jl` — `x ~ 0` is `Union{Equation, Vector{Equation}}` because of
   `Base.:~`'s complex-valued methods; the zero-crossing is always scalar, so build the
   `Equation` directly.

Nothing was disabled or narrowed: `jet = true`, `jet_broken` stays `false`, and `jet_kwargs`
is unchanged (`target_defined_modules = true`, i.e. JET's default `mode = :basic`).

## Verified locally

Julia 1.11.9, reproducing the QA lane's environment stacking (`test/qa` env with
`lib/ModelingToolkitBase` and the repo root as parent environments):

| | `report_package` result |
|---|---|
| `origin/master` (32a955af2) | `═════ 3 toplevel errors found ═════` (exactly the CI failure) |
| after commit 1 | `═════ 9 possible errors found ═════` |
| after commit 2 | `No errors detected` |

## Upstream

The underlying macro behaviour is a PrecompileTools issue: `@recompile_invalidations` cannot
be used around `include`/`using` in a package without breaking every static analyzer, and it
silently changes relative-include resolution. Worth reporting to
JuliaLang/PrecompileTools.jl — the macro could emit an `Expr(:toplevel, ...)` so its body
stays lexically top level (JET already handles macros that expand to `:toplevel`). Not filed
here; flagging for @ChrisRackauckas to decide.

Note: `@recompile_invalidations` around the include block was added in 49115cbf6
("refactor: improve precompilation of MTKBase, MTK"). Dropping it means invalidations caused
by ModelingToolkit's own method definitions are no longer force-recompiled during
precompilation. The invalidation-heavy part — loading `Symbolics`, `DiffEqBase`,
`SciMLBase`, `StateSelection` — is still wrapped. I did not measure the TTFX delta (the
machine was too loaded for a meaningful timing), so that is worth a look before merging.

https://claude.ai/code/session_019ywCW8vbGoc9X3dUbmXyme

### DIFF
--- a/src/ModelingToolkit.jl
+++ b/src/ModelingToolkit.jl
@@ -130,28 +130,29 @@ using ModelingToolkitBase: COMMON_SENTINEL, COMMON_NOTHING, COMMON_MISSING,
 using ModelingToolkitBase: build_function_wrapper, BuildFunctionWrapperOptions,
     GeneratedFunctionOptions
 
-@recompile_invalidations begin
-    include("linearization.jl")
-    include("systems/analysis_points.jl")
-    include("systems/solver_nlprob.jl")
+# `@recompile_invalidations` evaluates its body through `Core.eval` at run time, which
+# hides these files from static analyzers and resolves the relative paths against the
+# caller's source directory rather than `src/`. Keep the includes at real top level.
+include("linearization.jl")
+include("systems/analysis_points.jl")
+include("systems/solver_nlprob.jl")
 
-    include("problems/docs.jl")
-    include("systems/codegen.jl")
-    include("systems/codegen_compat.jl")
-    include("problems/semilinearodeproblem.jl")
-    include("problems/sccnonlinearproblem.jl")
+include("problems/docs.jl")
+include("systems/codegen.jl")
+include("systems/codegen_compat.jl")
+include("problems/semilinearodeproblem.jl")
+include("problems/sccnonlinearproblem.jl")
 
-    include("discretedomain.jl")
-    include("systems/systemstructure.jl")
-    include("initialization.jl")
-    include("systems/systems.jl")
-    include("systems/clock_inference.jl")
-    include("systems/if_lifting.jl")
-    include("systems/substitute_component.jl")
+include("discretedomain.jl")
+include("systems/systemstructure.jl")
+include("initialization.jl")
+include("systems/systems.jl")
+include("systems/clock_inference.jl")
+include("systems/if_lifting.jl")
+include("systems/substitute_component.jl")
 
-    include("systems/alias_elimination.jl")
-    include("structural_transformation/StructuralTransformations.jl")
-end
+include("systems/alias_elimination.jl")
+include("structural_transformation/StructuralTransformations.jl")
 
 @reexport using .StructuralTransformations
 
--- a/src/linearization.jl
+++ b/src/linearization.jl
@@ -747,7 +747,7 @@ function linearize_symbolic(
     h = build_explicit_observed_function(
         sys, outputs,
         GeneratedFunctionOptions(; expression = Val{false}, eval_expression, eval_module)
-    )
+    )::GeneratedFunctionWrapper
     if split
         dx = fun(sts, p, t)
         y = h(sts, p, t)
--- a/src/problems/sccnonlinearproblem.jl
+++ b/src/problems/sccnonlinearproblem.jl
@@ -752,10 +752,11 @@ function SciMLBase.SCCNonlinearProblem{iip}(
 
     new_dvs = dvs[reduce(vcat, decomposition.var_sccs)]
     new_eqs = eqs[reduce(vcat, decomposition.eq_sccs)]
+    ic = get_index_cache(sys)
+    new_ic = ic === nothing ? nothing :
+        subset_unknowns_observed(ic, sys, new_dvs, SymbolicT[])
     sys = ConstructionBase.setproperties(
-        sys; unknowns = new_dvs, eqs = new_eqs, index_cache = subset_unknowns_observed(
-            get_index_cache(sys), sys, new_dvs, SymbolicT[]
-        )
+        sys; unknowns = new_dvs, eqs = new_eqs, index_cache = new_ic
     )
 
     if length(subprobs) <= 5
@@ -777,7 +778,8 @@ function calculate_op_from_u0_p(sys::System, u0::Union{Nothing, AbstractVector},
     @assert length(rps) == length(p)
 
     for (i, pvars) in enumerate(rps)
-        for (var, val) in zip(pvars, p[i])
+        # `reorder_parameters` flattens by default, so every buffer is a flat variable list.
+        for (var, val) in zip(pvars::Vector{SymbolicT}, p[i])
             write_possibly_indexed_array!(op, var, Symbolics.SConst(val), COMMON_NOTHING)
         end
     end
--- a/src/structural_transformation/pantelides.jl
+++ b/src/structural_transformation/pantelides.jl
@@ -23,7 +23,7 @@ function pantelides_reassemble(state::TearingState, var_eq_matching)
     fill!(out_vars, ModelingToolkit.COMMON_NOTHING)
     out_vars[1:length(fullvars)] .= fullvars
 
-    iv = get_iv(sys)
+    iv = get_iv(sys)::SymbolicT
     D = Differential(iv)
 
     for (varidx, diff) in edges(var_to_diff)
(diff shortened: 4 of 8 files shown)


AayushSabharwal (approved):

ChrisRackauckas-Claude:
### Local verification

Julia 1.11.9, shared depot, all runs from the branch head.

**QA lane** (`GROUP=QA julia --project=. -e 'using Pkg; Pkg.test()'`):

| | Quality Assurance testset |
|---|---|
| `master` 32a955af2 | `12 pass, 3 fail, 6 error` — includes `Quality Assurance: JET-test failed at .../JET/src/JETBase.jl:1046` with the 3 toplevel errors |
| this branch | `13 pass, 2 fail, 6 error` — no `JET-test` line at all; JET passes |

The remaining 2 failures + 6 errors are the ExplicitImports checks, `public API has docstrings`
and `No unapproved public reexports`. They are unchanged by this PR and are being handled
separately.

**Behavioural groups**, branch vs. unmodified `master` 32a955af2 (same machine, same depot):

| group | `master` | this branch |
|---|---|---|
| `InterfaceI` | 1478 pass, 0 fail, **1 error**, 3 broken | 1483 pass, 0 fail, **1 error**, 3 broken |
| `InterfaceII` | (baseline running) | 997 pass, 0 fail, 2 error, 7 broken |

The `+5` passes on `InterfaceI` are the new `DifferentiatedVariableNotUnknownError` testset.

The errors are **pre-existing and unrelated** — all three are the same latent
`ModelingToolkitBase` bug, reproduced byte-for-byte on unmodified `master`:

```
MethodError: Cannot `convert` an object of type
  DiCMOBiGraph{false,Int64,BipartiteGraph{Int64,Nothing},Matching{Union{Unassigned,Int64},Vector{Union{Unassigned,Int64}}}}
to an object of type
  DiCMOBiGraph{false,Int64,BipartiteGraph{Int64,Nothing},Matching{Unassigned,Vector{Union{Unassigned,Int64}}}}
 [1] ModelingToolkitBase.ObservedGraphCache(graph::..., obsvar_to_idx::Dict{Any, Int64})
   @ ModelingToolkitBase lib/ModelingToolkitBase/src/utils.jl:1222
```

`ObservedGraphCache`'s `graph` field hardcodes `Matching{Unassigned, ...}` while
`observed_dependency_graph` can hand it a `Matching{Union{Unassigned, Int}, ...}`. Being
tracked separately; it is not touched by this PR.

ChrisRackauckas-Claude:
CI notes for this PR — none of the red checks are caused by it:

* `Runic / Runic Format Check` — red on `master` too. The only file Runic flags is `lib/ModelingToolkitBase/src/debugging.jl`, untouched here; split out as https://github.com/SciML/ModelingToolkit.jl/pull/4884. With that applied, `Runic.format_string` is a no-op on all 223 `.jl` files in the repo (verified locally with Runic v1.7.0).
* `tests / InterfaceI`, `tests / InterfaceII` — the `ObservedGraphCache` `convert` MethodError described above, in `lib/ModelingToolkitBase/test/` files this PR does not touch. `master`'s own CI run for these predates the dependency release that triggers it; running `GROUP=InterfaceI` against unmodified `master` 32a955af2 today reproduces it identically.
* `tests / QA` — the ExplicitImports / public-API-docs / public-reexport failures, which are being fixed separately. The JET check inside it now passes.
* `tests / SymbolicIndexingInterface`, `tests / Initialization`, downstream jobs — failures in `lib/ModelingToolkitBase/test/` (`sciml_problem_inputs.jl` and friends) and pre-existing `master` breakage; `tests / QA` and `tests / Initialization` are already red on `master`.

ChrisRackauckas-Claude:
Update on the `ObservedGraphCache` errors seen in `InterfaceI`/`InterfaceII` here: they are a **BipartiteGraphs.jl v0.1.11 regression**, not anything in this PR.

`Matching(v::AbstractVector{T})` now returns `Matching{T}`, so `Matching(Vector{Union{Unassigned,Int}}(assigns))` produces `Matching{Union{Unassigned,Int}, ...}` where it used to produce `Matching{Unassigned, ...}`, which no longer matches `ObservedGraphCache`'s concrete field type:

```
complete(Matching(Vector{Union{Unassigned,Int}}([1,2,3]))) |> typeof
v0.1.10 -> Matching{Unassigned, Vector{Union{Unassigned, Int64}}}
v0.1.11 -> Matching{Union{Unassigned, Int64}, Vector{Union{Unassigned, Int64}}}
```

Tracked in https://github.com/SciML/BipartiteGraphs.jl/issues/54, fixed on the ModelingToolkit side by https://github.com/SciML/ModelingToolkit.jl/pull/4887. Independent of this PR — with #4887 applied those errors go away here too.

Issue #4882, opened 2026-08-05
QA: ModelingToolkit depends on ~140 non-public names of its dependencies


ChrisRackauckas-Claude:
Tracking issue for the two remaining `run_qa` ExplicitImports failures on `master`:

- `all_explicit_imports_are_public` — 53 findings
- `all_qualified_accesses_are_public` — 89 findings

`all_explicit_imports_via_owners` and `all_qualified_accesses_via_owners` are fixed by https://github.com/SciML/ModelingToolkit.jl/pull/4880, which means every one of these names is already being reached through the module that *defines* it. There is no other module to import them from. The only real fixes are for the owner to declare the name `public` (with a docstring and a rendered docs entry — public and documented move together), or for ModelingToolkit to stop needing the name.

Until then they are enumerated exhaustively in `NONPUBLIC_EXPLICIT_IMPORTS` / `NONPUBLIC_QUALIFIED_ACCESSES` in `test/qa/qa.jl`, so reaching for a *new* internal fails the QA lane rather than sliding in unnoticed.

### By owner

**ModelingToolkitBase — 9 explicit imports, 28 qualified accesses.** ModelingToolkit is the upper half of one library split across a monorepo boundary. These are the pair's shared implementation seams: codegen entry points (`build_function_wrapper`, `BuildFunctionWrapperOptions`, `GeneratedFunctionOptions`, `torn_system_jacobian_sparsity`, `generate_ODENLStepData`), problem-type selectors (`get_nonlinear_problem_type`, `get_initialization_problem_type`, `check_compatible_system`), the hashconsed `COMMON_*` sentinels, and the docstring templates (`problem_docstring`, `function_docstring`). They are internal *to the pair*, not to ModelingToolkitBase alone, so declaring them `public` would advertise them to end users — the opposite of what they are. Probably the right long-term answer here is not `public` but a documented internal-interface module.

**ModelingToolkitTearing — 2 explicit imports, 16 qualified accesses.** The strongest candidate for upstream promotion. `TearingState` in particular: ModelingToolkit `export`s it (`src/ModelingToolkit.jl`), so it is already part of ModelingToolkit's public API while being non-public at its owner. `SystemStructure`, `ClockInference`, `infer_clocks!`, `input_timedomain`/`output_timedomain`, `split_system`, `StateMachineOperator` are the rest.

[...]

PR #4883, opened 2026-08-05
Enumerate the non-public names ModelingToolkit reaches for

ChrisRackauckas-Claude:

Rebased onto current master (`453c45c08e`). #4880 is merged; this branch is now stacked only on #4881 — review that one first, this branch contains its commit.

## Problem

`all_explicit_imports_are_public` and `all_qualified_accesses_are_public` both fail on master. With #4880 in place the `*_via_owners` checks pass on Julia 1.12, which means every one of these names is *already* reached through the module that defines it — there is no other module to import it from. The only real fixes are for the owner to declare the name `public` (with a docstring and a rendered docs entry, since public and documented move together) or for ModelingToolkit to stop needing the name.

## Change

Three findings were reducible, and are fixed rather than listed — six `ModelingToolkit.`-qualified references to ModelingToolkit's *own* `inputs_to_parameters!`, `alias_elimination!` and `MTKTearing` alias, where every neighbouring call site already spells them unqualified:

```julia
-    state = ModelingToolkit.inputs_to_parameters!(state, inputs, outputs)
+    state = inputs_to_parameters!(state, inputs, outputs)
-    sys, mm = ModelingToolkit.alias_elimination!(state; fully_determined, kwargs...)
+    sys, mm = alias_elimination!(state; fully_determined, kwargs...)
-    return ModelingToolkit.MTKTearing.InferredDiscrete(1)
+    return MTKTearing.InferredDiscrete(1)
```

The rest are enumerated in `NONPUBLIC_EXPLICIT_IMPORTS` (76) and `NONPUBLIC_QUALIFIED_ACCESSES` (96) in `test/qa/qa.jl`, grouped by owner, each group carrying the reason it is currently irreducible:

- **ModelingToolkitBase** — ModelingToolkit is the upper half of one library split across a monorepo boundary. These are the pair's shared implementation seams (codegen entry points, problem-type selectors, the hashconsed `COMMON_*` sentinels, docstring templates). They are internal *to the pair*, not to ModelingToolkitBase alone, so declaring them `public` would advertise them to end users — the opposite of what they are.
- **ModelingToolkitTearing / StateSelection** — the structural-transformation stack split out of ModelingToolkit; ModelingToolkit is its only consumer. Strongest candidates for upstream promotion, in particular `TearingState`, which ModelingToolkit already `export`s and therefore already exposes as its own public API.
- **Symbolics / SymbolicUtils** — the `BasicSymbolic` representation ModelingToolkit compiles against, the codegen targets, and the IR-walking helpers used by alias elimination and SCC construction.
- **SciMLBase / Moshi / OffsetArrays / Base** — eight leftovers, including `Base.RefValue`, which has no public spelling at all.

Being exhaustive is the point: reaching for a *new* internal now fails the QA lane instead of sliding in unnoticed. Tracked upstream in #4882, which lists every name and a suggested order of attack.

The lists are larger than the reported finding counts because ExplicitImports stops at the first failing module — the `StructuralTransformations` submodule's imports only surfaced once `ModelingToolkit` itself was clean.

### On `Iterators.map`

Reverted. The previous revision rewrote it as a generator; @AayushSabharwal pointed out that is a behaviour change, since a generator builds a closure and `Iterators.map` does not. It is listed instead, per @ChrisRackauckas.

It is a finding at all only because of the Julia version:

```
$ julia-1.10 -e 'println(:map in names(Base.Iterators))'   # false
$ julia-1.11 -e 'println(Base.ispublic(Base.Iterators, :map))'   # false
$ julia-1.12 -e 'println(Base.ispublic(Base.Iterators, :map))'   # true
```

so it is a finding on the `lts` half of the QA matrix and not on the `1` half. The entry is inert on 1.12.

## Verification

`test/qa/qa.jl`, Julia 1.12.6, same master commit (`453c45c08e`) on both sides.

**Before** — unmodified master:

```
Test Summary:                                | Pass  Fail  Error  Total      Time
Quality Assurance                            |   14     3      4     21  29m47.2s
  ...
  ExplicitImports                            |    2            4      6     27.5s
    no_implicit_imports                      |                 1      1      7.4s
    no_stale_explicit_imports                |                 1      1      3.9s
    all_explicit_imports_via_owners          |    1                   1      1.5s
    all_qualified_accesses_via_owners        |    1                   1      4.5s
    all_qualified_accesses_are_public        |                 1      1      6.0s
    all_explicit_imports_are_public          |                 1      1      4.3s
  Public API documentation                   |    1     1             2      2.4s
    public API has docstrings                |          1             1      2.2s
    public API is rendered in docs           |    1                   1      0.0s
  No unapproved public reexports             |          1             1      0.0s
```

**After** — this branch (which includes #4881):

```
Test Summary:                                | Pass  Fail  Error  Total      Time
Quality Assurance                            |   17     2      2     21  30m11.2s
  ...
  ExplicitImports                            |    4            2      6     18.8s
    no_implicit_imports                      |                 1      1      6.1s
    no_stale_explicit_imports                |                 1      1      4.0s
    all_explicit_imports_via_owners          |    1                   1      1.5s
    all_qualified_accesses_via_owners        |    1                   1      4.2s
    all_qualified_accesses_are_public        |    1                   1      1.2s
    all_explicit_imports_are_public          |    1                   1      1.7s
  Public API documentation                   |    1     1             2      2.4s
    public API has docstrings                |          1             1      2.1s
    public API is rendered in docs           |    1                   1      0.0s
  No unapproved public reexports             |    1                   1      0.0s
```

`all_qualified_accesses_are_public` and `all_explicit_imports_are_public` go error → pass; `No unapproved public reexports` goes fail → pass (that one is #4881's). Every Aqua testset passes identically on both sides.

Also run: `Runic.main(["--check"])` and `typos` over the three changed files — both clean.

## Not verified

- **The `lts` half of the QA matrix.** `test/test_groups.toml` runs QA on `lts` and `1`. On Julia 1.10 the run never reaches the two checks this PR is about, because `all_explicit_imports_via_owners` and `all_qualified_accesses_via_owners` error first, on findings that do not appear on 1.12 — `Base.which` resolves re-exported bindings differently before the 1.12 binding-partition change. The first few are `Symbolics.unwrap`/`Symbolics.scalarize` (owner `SymbolicUtils`), `ModelingToolkit.t_nounits`/`D_nounits` in the three OrdinaryDiffEq extensions (owner `ModelingToolkitBase`), and `ModelingToolkit.ImperativeAffect`/`SymbolicDiscreteCallback` in MTKFMIExt (owner `ModelingToolkitBase`), and ExplicitImports reports one module at a time so the chain keeps going. That is #4880's subject on a Julia version #4880 was not verified against, not this PR's, and it is pre-existing on master either way. Filed as #4903 rather than folded in here.
- **Allowed-to-fail jobs** — GPU, downstream, `julia pre`. Not run locally.
- **The QA lane in CI.** `tests / QA (julia 1, ...)` fails on this PR, but it fails identically on master and it is not a test failure: the self-hosted runner is killed ~50 minutes in (`"The self-hosted runner lost communication with the server."`). Same on `2c31d86323`, `82dee851d6` and `453c45c08e`. Filed as #4904. The local runs above are therefore the only evidence available for this PR; there is no green CI QA result to point at, for this branch or for master.

## Still red on master, untouched here

Confirmed identical on master and on this branch, in the "before"/"after" runs above:

- `no_implicit_imports` and `no_stale_explicit_imports` — both `UnanalyzableModuleException: Module ModelingToolkit was found to be unanalyzable`. Catalyst hit the same thing with an EnumX submodule and passes `allow_unanalyzable`.
- `public API has docstrings` — fails on `[:StructuralTransformations, :UnPack]`, the two module bindings ModelingToolkit exposes publicly without a docstring. This is *not* the set listed in `SYMBOLICS_OWNED_REEXPORTS`; those nine are already ignored and are not what the check trips on.
- The JET testset — 22 possible errors, identical on both sides.

`Runic / Runic Format Check` is red on this PR and on master, not because of this PR: `lib/ModelingToolkitBase/src/problems/initializationproblem.jl:172` is missing a `return`, from `b77c4757f5`. One line, belongs in its own PR.

https://claude.ai/code/session_01P6EpZyM3Fmov47669Gypbi

### DIFF
--- a/src/ModelingToolkit.jl
+++ b/src/ModelingToolkit.jl
@@ -74,8 +74,10 @@ import ModelingToolkitBase as MTKBase
 import SimpleNonlinearSolve
 
 import SciMLBase: @add_kwonly
-@reexport using Symbolics
-@reexport using UnPack
+using UnPack: @unpack
+# ModelingToolkitBase already `@reexport`s Symbolics and UnPack, so re-exporting them a
+# second time here would only duplicate that surface. The set of names ModelingToolkit
+# publicly re-exports is pinned by `REEXPORTED_API` in `test/qa/qa.jl`.
 @reexport using ModelingToolkitBase
 RuntimeGeneratedFunctions.init(@__MODULE__)
 
--- a/src/systems/clock_inference.jl
+++ b/src/systems/clock_inference.jl
@@ -16,10 +16,10 @@ function MTKTearing.input_timedomain(::Hold, args::MTKTearing.IOTimeDomainArgsT 
 end
 
 function MTKTearing.input_timedomain(::SampleTime, arg = nothing)
-    return ModelingToolkit.MTKTearing.InputTimeDomainElT[]
+    return MTKTearing.InputTimeDomainElT[]
 end
 function MTKTearing.output_timedomain(::SampleTime, arg = nothing)
-    return ModelingToolkit.MTKTearing.InferredDiscrete(1)
+    return MTKTearing.InferredDiscrete(1)
 end
 
 function MTKTearing.output_timedomain(::Hold, _::MTKTearing.IOTimeDomainArgsT = nothing)
--- a/src/systems/systemstructure.jl
+++ b/src/systems/systemstructure.jl
@@ -218,11 +218,11 @@ function _mtkcompile!(
     validate_io!(state, orig_inputs, inputs, discrete_inputs, outputs, disturbance_inputs)
     # ModelingToolkit.markio!(state, orig_inputs, inputs, outputs, disturbance_inputs)
     union!(inputs, disturbance_inputs)
-    state = ModelingToolkit.inputs_to_parameters!(state, discrete_inputs, OrderedSet{SymbolicT}())
-    state = ModelingToolkit.inputs_to_parameters!(state, inputs, outputs)
+    state = inputs_to_parameters!(state, discrete_inputs, OrderedSet{SymbolicT}())
+    state = inputs_to_parameters!(state, inputs, outputs)
     eliminate_perfect_aliases!(state)
     StateSelection.trivial_tearing!(state)
-    sys, mm = ModelingToolkit.alias_elimination!(state; fully_determined, kwargs...)
+    sys, mm = alias_elimination!(state; fully_determined, kwargs...)
     old_to_new_eq, old_to_new_var, aliases = eliminate_perfect_aliases!(state)
     sys = state.sys
     mm = StateSelection.get_new_mm(aliases, old_to_new_eq, old_to_new_var, mm)
@@ -262,7 +262,7 @@ function _mtkcompile_worker!(
         var_eq_matching = StateSelection.pantelides!(state; finalize = false, kwargs...)
         sys = pantelides_reassemble(state, var_eq_matching)
         state = TearingState(sys)
-        sys, mm = ModelingToolkit.alias_elimination!(state; fully_determined, kwargs...)
+        sys, mm = alias_elimination!(state; fully_determined, kwargs...)
         state.mm = mm
         sys = ModelingToolkit.dummy_derivative(
             sys, state; fully_determined, kwargs...
--- a/test/qa/qa.jl
+++ b/test/qa/qa.jl
@@ -169,8 +169,9 @@ const REEXPORTED_API = (
     # ModelingToolkitTearing: `TearingState` is explicitly exported by
     # ModelingToolkit (see `src/ModelingToolkit.jl`).
     :TearingState,
-    # StateSelection: `find_solvables!` is an accidental leak -- it is a StateSelection
-    # internal that `StructuralTransformations` `export`s but only ever calls qualified. It is
+    # StateSelection: `find_solvables!` is an accidental leak -- a StateSelection internal
+    # that `StructuralTransformations` imports and `export`s, while every call site spells it
+    # `StateSelection.find_solvables!`, so nothing in ModelingToolkit uses the export. It is
     # listed here so the audit is exhaustive; dropping it is a breaking change and belongs in
     # the next major release, not in a QA fix.
     :find_solvables!,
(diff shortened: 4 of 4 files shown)


src/systems/alias_elimination.jl
```diff
@@ -871,7 +871,7 @@ function alias_elimination!(
     end
 
     eqs = collect(equations(state))
-    fullvars_to_idx = Dict{SymbolicT, Int}(Iterators.map(reverse, enumerate(fullvars)))
+    fullvars_to_idx = Dict{SymbolicT, Int}(v => i for (i, v) in enumerate(fullvars))
```
AayushSabharwal: Why is `Iterators.map` a problem? This is intentional, generators build a closure and `Iterators.map` doesn't.

ChrisRackauckas: It should be moved to being an allowed.

ChrisRackauckas-Claude: You're right, reverted — `Iterators.map` is back. I'd taken it as reducible without checking that the rewrite was behaviour-preserving, and it isn't: the generator allocates a closure.

It shows up as a finding only on the `lts` half of the QA matrix, because `map` was not `public` in `Base.Iterators` until 1.12:

```
julia-1.10 -e 'println(:map in names(Base.Iterators))'          # false
julia-1.11 -e 'println(Base.ispublic(Base.Iterators, :map))'    # false
julia-1.12 -e 'println(Base.ispublic(Base.Iterators, :map))'    # true
```

It is now listed in `NONPUBLIC_QUALIFIED_ACCESSES` with that reason, per @ChrisRackauckas.

ChrisRackauckas-Claude: Done — `Iterators.map` is in `NONPUBLIC_QUALIFIED_ACCESSES` now, and `src/systems/alias_elimination.jl` is back to the original `Iterators.map` line. The entry carries the reason it exists at all: `map` only became `public` in `Base.Iterators` in Julia 1.12, so it is a finding on `lts` and inert on `1`.

ChrisRackauckas-Claude:
Rebased onto `453c45c08e` and force-pushed. #4880 was squash-merged as `2c31d86323`, so its commit dropped out of the stack cleanly; this branch is now #4881 + one commit.

**Review comments** — `Iterators.map` is reverted in `src/systems/alias_elimination.jl` and listed in `NONPUBLIC_QUALIFIED_ACCESSES` instead. Replies inline.

**New since the last revision** — the rebase surfaced three findings master had grown in the meantime, all self-qualified accesses to ModelingToolkit's own names (`ModelingToolkit.inputs_to_parameters!`, `ModelingToolkit.alias_elimination!`, `ModelingToolkit.MTKTearing`). Those are reducible, so they are fixed at the six call sites rather than added to the list — every neighbouring line already spells them unqualified. That is the list doing its job: a new internal reach fails the lane instead of sliding in.

**Verification** is now a real before/after on the same Julia (1.12.6) and the same master commit, rather than the 1.11 run in the original description. `all_qualified_accesses_are_public` and `all_explicit_imports_are_public` go error → pass. Full summaries in the description.

**Two things I did not fold in**, both pre-existing on master:

- On the `lts` (Julia 1.10) half of the QA matrix, the `*_via_owners` checks still error, so the checks this PR is about are never reached there. It is a version difference in how `Base.which` resolves re-exported bindings, and it is #4880's subject rather than this one's. Filed as #4903 with a reproduction and the findings observed so far. Worth knowing: Julia 1.10 ignores `[sources]` in `test/qa/Project.toml`, so a naive `julia +1.10 --project=test/qa test/qa/qa.jl` silently tests the registry copy of ModelingToolkit instead of the working tree.
- `Runic / Runic Format Check` is red on master — `lib/ModelingToolkitBase/src/problems/initializationproblem.jl:172` is missing a `return`, from `b77c4757f5`. One line; happy to send it as its own PR.

The QA CI failure on the previous head was infrastructure, not a test failure: *"The self-hosted runner lost communication with the server."*

ChrisRackauckas-Claude:
## CI on `b2692a1bc6`

Every root test group passed on every declared Julia version:

| group | lts | 1 | pre |
|---|---|---|---|
| InterfaceI | ✅ | ✅ | ✅ |
| InterfaceII | ✅ | ✅ | ✅ |
| Initialization | ✅ | ✅ | ✅ |
| SymbolicIndexingInterface | ✅ | ✅ | ✅ |
| Extensions | ✅ | ✅ | ✅ |
| FMI | ✅ | ✅ | ✅ |
| Downstream | ✅ | ✅ | ✅ |
| Optimization | ✅ | ✅ | cancelled |

Plus `Downgrade / InterfaceI`, `downgrade-sublibraries / test (lib/SciCompDSL)`, `Spell Check`, `runic-suggestions`, and every downstream job except the two noted below.

### The red checks, and why none of them are this PR

Each of these was checked against master (`453c45c08e`) or against the previous head:

| check | this PR | master | what it is |
|---|---|---|---|
| `tests / QA (julia 1)` | ❌ | ❌ | Not a test failure — the self-hosted runner is killed ~49 min in (`"The self-hosted runner lost communication with the server."`). Same on `2c31d86323` and `82dee851d6`. Filed as #4904. |
| `Runic / Runic Format Check` | ❌ | ❌ | `lib/ModelingToolkitBase/src/problems/initializationproblem.jl:172` is missing a `return`, from `b77c4757f5`. Running Runic over every tracked `.jl` at this branch's HEAD reports that one diff and nothing else. |
| `Catalyst/Modeling/1` | ❌ | ❌ | pre-existing |
| `ModelingToolkitStandardLibrary.jl/Core/1/` | ❌ | ❌ | pre-existing |
| `SciMLBase.jl/SymbolicIndexingInterface/1/` | ❌ | ❌ | pre-existing (and `SciMLBase.jl/Downstream/1/`, which was red on the previous head, is green here) |
| `downgrade-sublibraries / test (lib/ModelingToolkitBase)` | ❌ | ❌ | pre-existing |
| `Benchmarks (1)` / `(lts)` | ❌ / cancelled | n/a | `Failed to delete old caches`; also red on the previous head |

So the QA lane — the one this PR is actually about — has no green CI result to point at, on this branch or on master, because the job never finishes. The before/after in the description is a local run of the same `test/qa/qa.jl` on Julia 1.12.6 against the same master commit, which is the best evidence available until #4904 is sorted.

PR #4888, opened 2026-08-05
Port the Pyomo extension to Pyomo 2 / Symbolics 7

ChrisRackauckas-Claude:

⚠️ **Blocked on https://github.com/SciML/Pyomo.jl/pull/15 merging and Pyomo v2.0.1 being registered.** Compat here is pinned to `2.0.1`, so CI cannot resolve until that lands.

## The extension is currently dead

`MTKPyomoDynamicOptExt` cannot load under any published Pyomo:

- Its compat is pinned at `Pyomo = "0.1.0"`.
- Every registered Pyomo version through 1.0.1 declares `Symbolics = "6.39.1 - 6"`, while MTKBase requires `Symbolics = "7.32.1"`. No resolution exists.
- The extension itself calls `Symbolics.symstruct(ConcreteModel)` and `Symbolics.Struct{PyomoVar}`, both removed in Symbolics 7.
- `using Pyomo` is commented out in `test/optimization/dynamic_optimization.jl:13`, `Pyomo` is not in the optimization test environment at all, and every Pyomo assertion sits behind `if @isdefined(Pyomo)`. Nothing has exercised this code in a long time.

Pyomo.jl v2.0.0 (https://github.com/SciML/Pyomo.jl/pull/13) ported that package to Symbolics 7; this is the matching MTK side.

## Changes

- **Compat**: `Pyomo = "0.1.0"` → `"2.0.1"`; MTKBase 1.60.0 → 1.60.1.
- **`Symbolics.symstruct(ConcreteModel)` → `SymbolicConcreteModel`** (2 sites). Symbolics 7 uses the Julia type itself as the symtype.
- **`pysym_getproperty` / `_getproperty` deleted here** — they now live in Pyomo.jl, where they can be tested against a real model. They were defined locally because Pyomo components live on the Python object rather than being Julia fields, so Symbolics' own struct `getproperty` never applied.
- **Symbolic indexing → `Pyomo.pyomo_getindex`** (4 sites). Symbolics 7 moved the symtype out of `BasicSymbolic`'s type parameter (it now holds the vartype), so the old `BasicSymbolic{Struct{PyomoVar}}` `getindex` dispatch is not expressible. The `Symbolics.value(...)` calls that existed only to reach that dispatch are gone too.
- **`fold = false` → `fold = Val(false)`** (2 sites). Separate latent bug: SymbolicUtils takes `fold` as a `Val`, so the `Bool` form throws a `MethodError` on every `add_constraint!` and `set_objective!`. Checked that `Val` is the documented form at MTKBase's declared SymbolicUtils floor (4.35.3), so no floor bump is needed.
- **Tests re-enabled**: `using Pyomo` uncommented, `Pyomo` added to `test/optimization/Project.toml`.

## Verification

`dynamic_optimization.jl` run locally with Pyomo enabled and a working `ipopt` (`SolverFactory("ipopt").available()` → `True`, so the solve paths are genuinely exercised rather than skipped). All 11 testsets pass, 110 tests, exit 0:

```
ODE Solution, no cost                                                |   18     18  4m07.4s
Linear systems                                                       |   32     32  2m16.5s
Rocket launch                                                        |    8      8    48.5s
Composed system with bound inputs                                    |    3      3     2.6s
Free final time problems                                             |   12     12    19.5s
Cart-pole problem                                                    |    4      4  1m33.4s
Parameter defaults usage                                             |    4      4     2.2s
Parameter estimation                                                 |   13     13  1m16.9s
Observed variable substitution in dynamic optimization (issue #4089) |    5      5     3.2s
Parameter bindings resolved in dynamic optimization pmap             |    8      8     3.9s
Observed variables with bound parameters in constraints              |    3      3     0.7s
```

## What re-enabling these tests turned up

Getting to green took five iterations, and every failure was in Pyomo.jl rather than here — all now fixed in https://github.com/SciML/Pyomo.jl/pull/15:

1. `getindex(::PyomoVar, ::Int, ::Int)` ambiguous with `Base.getindex(::Number, ::Integer...)` — hit by `add_initial_constraints!` doing `model.U[i, 0]`. (Regression from the v2.0.0 port.)
2. Missing `promote_shape` on the hand-built terms, so their shape was `Unknown(-1)` and `cos(θ)` in the cart-pole dynamics threw `ArgumentError: Invalid shapes for cos`. (Regression from the v2.0.0 port.)
3. Missing `isinteger(::PyomoVar)`, which `SymbolicUtils.safe_div` calls on any `Number` operand.
4. `/` called `pydiv`, which does not exist in PythonCall. **Pre-existing.**
5. All five comparison operators called `pycall(≥, x, y)`, wrapping the *Julia* operator as a Python callable; `==` silently returned `<py False>` instead of an expression. The `NaNMath` methods had the same defect. **Pre-existing.**

Items 4 and 5 predate the Symbolics 7 work — they survived precisely because this extension, the only consumer, has never been able to load.

https://claude.ai/code/session_016GYQeAiC2x9nJ7Mmwwxhe2


lib/ModelingToolkitBase/test/optimization/Project.toml
```diff
@@ -28,12 +29,13 @@ Symbolics = "0c5d862f-8b57-4792-8d23-62f2024744c7"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [sources]
-ModelingToolkitBase = {path = "../.."}
+ModelingToolkitBase = {path = "/home/crackauc/sandbox/tmp_20260805_100326_7429/MTK/lib/ModelingToolkitBase"}
```
ChrisRackauckas: Fix this

PR #4893, opened 2026-08-06
feat: add `SciMLProblemOptions`

AayushSabharwal:
The massive redundant diff for `problem_utils.jl` is the GitHub diff algorithm being bad. I don't see it locally.


AayushSabharwal:
Tests pass, but one commit is not attributed to Claude. Will edit it and merge.

Issue #4830, opened 2026-07-24
Loop openings: allow `missing` for the broken inputs together with a steady-state condition for the broken outputs?


baggepinnen:
ModelingToolkit 11.36 made the operating-point values of loop-opened signals explicitly required when linearizing (`_check_loop_opening_op`): opening a loop at an analysis point turns the opened signal into a parameter whose value is no longer implied by the rest of the system. This is a good change in itself — previously a stale/guessed value could be used silently — but there does not currently seem to be a way to express the most natural choice of value: *the value the signal would have at the operating point with the loop closed*.

**Would it make sense to allow passing `missing` for the new broken inputs, together with a steady-state condition for the broken outputs?** That is,

```julia
op = Dict(
    sys.C.u => missing,   # the broken input: solve for it in the initialization
    D(sys.P.y) => 0,      # the broken output: steady-state condition making the system determined
)
```

following the usual MTK convention that `p => missing` means "solve for this parameter during initialization (using its guess)". Today this crashes inside `linearize`'s operating-point application loop, because the `missing` is passed straight to the parameter setter:

```
ERROR: MethodError: Cannot `convert` an object of type Missing to an object of type Float64
Stacktrace:
 [1] setindex!
 [2] set_parameter!  @ ModelingToolkitBase src/systems/parameter_buffer.jl:501
 [3] set_parameter!  @ SymbolicIndexingInterface src/value_provider_interface.jl:68
 ... (via the `setu(prob, k)(prob, v)` loop in `linearize`, src/linearization.jl)
```

## MWE

```julia
using ModelingToolkit
using ModelingToolkit: t_nounits as t, D_nounits as D
import ModelingToolkit: connect

@component function FirstOrder(; name, k = 1.0)
    @variables begin
        u(t), [input = true]
        x(t) = 0.0
        y(t), [output = true]
    end
    @parameters k = k
    System([D(x) ~ -x + k * u, y ~ x], t; name)
end
@named P = FirstOrder(k = 1.0)
@named C = FirstOrder(k = -2.0)
eqs = [connect(P.y, :y, C.u), connect(C.y, :u, P.u)]
@named sys = System(eqs, t; systems = [P, C])
csys = complete(sys) # for operating-point key access only

# Explicit value: works
m1, _ = ModelingToolkit.linearize(sys, :u, :y; loop_openings = [:y], op = Dict(csys.C.u => 0.0))

[...]

PR #4894, opened 2026-08-06
feat: resolve symbolic loop-opening op values from the solution

baggepinnen:
Implements the symbolic-value half of #4830 (the `missing` half was implemented in #4855).

## What

Values in `LinearizationOpPoint.op` may now be symbolic expressions, e.g. `op = Dict(opened_signal => driving_signal)`, which are evaluated from the wrapped solution at each time point. Mapping an opened signal to the expression that drives it thus linearizes around the value the opened signal has in the loop-closed solution — the natural operating point when linearizing along a trajectory with `loop_openings`.

## Why resolve in `_build_op_from_solution`

Previously, symbolic values were passed through verbatim, and depending on the model they either

- errored at problem construction, because `MTKParameters` evaluates operating-point values eagerly before any initialization has run ("Could not evaluate value of parameter … Missing values for variables in expression …"), or
- **silently produced a constant, stale linearization at every time point**: `_resolve_op_value` routes symbolic values through `getu(prob, v)(prob)`, but state values are written as `Initial(x)` parameters and `prob.u0` is only refreshed by initialization inside `solve`, so the observed function evaluates against the construction-time state at every iteration of `__linearize_multiple_op_barrier`.

Resolving symbolic values against the solution inside `_build_op_from_solution` (once per value for all time points) means the downstream machinery — problem construction and the per-point setters — only ever sees numbers, avoiding both failure modes as well as any sensitivity to dict iteration order.

On a plant/controller loop where the controller dynamics are nonlinear in its input, the new test verifies that the per-point linearizations exactly match ground truth (and match single-point linearizations with manually resolved numeric values), where previously they were constant along the trajectory.

## Also

- `missing` op values are now skipped in `__linearize_multiple_op_barrier`, mirroring the skip added to single-point `linearize` in #4855 (they previously threw `convert(Float64, ::Missing)` in the multi-timepoint path).
- Docstring of `LinearizationOpPoint` and the `_check_loop_opening_op` error message updated to mention symbolic values.

Not addressed (out of scope): symbolic op values in plain `linearize` *without* a solution still error at problem construction; supporting that would require initialization-level changes.

https://claude.ai/code/session_01SWPXe76HwiugWWyLP7BYrL

### DIFF
--- a/src/linearization.jl
+++ b/src/linearization.jl
@@ -13,8 +13,11 @@ extras.
 The `op` keyword argument provides additional operating-point values that are merged into
 the solution-derived operating point at every time point (taking precedence). This is how
 values for variables that are not present in `sol` are supplied — in particular the
-parameters created by `loop_openings`, e.g.
-`LinearizationOpPoint(sol, t; op = Dict(opened_signal => 0))`.
+parameters created by `loop_openings`. Values may be numbers, e.g.
+`LinearizationOpPoint(sol, t; op = Dict(opened_signal => 0))`, or symbolic expressions,
+which are evaluated from the solution at each time point. Mapping an opened signal to the
+expression that drives it, e.g. `op = Dict(opened_signal => driving_signal)`, thus
+linearizes around the value the opened signal has in the loop-closed solution.
 
 # Fields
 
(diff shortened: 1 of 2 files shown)


src/linearization.jl
```diff
@@ -76,15 +90,28 @@ function _build_op_from_solution(op::LinearizationOpPoint{S, <:AbstractVector})
     for p in parameters(sol_sys)
         param_vals[p] = getp(op.sol, p)(op.sol)
     end
+    # Split the extra op into constant entries and symbolic entries; the latter are
+    # evaluated from the solution, once per entry for all time points.
+    extra_const = Dict{SymbolicT, SymbolicT}()
+    extra_resolved = Pair{SymbolicT, Any}[]
```
AayushSabharwal: Is it common for these extra values to be of multiple different types? I ask because we could do

```suggestion
    extra_resolved = Pair{SymbolicT, eltype(eltype(sol.u))}[]
```

And make this type-stable.

src/linearization.jl
```diff
@@ -45,6 +48,17 @@ end
...
 
+# Values in `LinearizationOpPoint.op` may be symbolic expressions (e.g.
+# `opened_signal => driving_signal`), which are evaluated from the solution at each time
+# point rather than passed through verbatim. Resolving them here means the downstream
+# machinery (problem construction, per-point setters) only ever sees numbers: symbolic
+# values cannot be evaluated eagerly at problem construction, and resolving them against
+# the `LinearizationProblem` inside the per-point loop would read state values that are
+# only refreshed by initialization inside `solve`, i.e. those of the previous time point.
+function _is_symbolic_op_value(v)
+    return !SU.isconst(v) && (symbolic_type(v) != NotSymbolic() || is_array_of_symbolics(v))
```
AayushSabharwal:
```suggestion
    return !SU.isconst(unwrap(v))
```
This check is sufficient.

src/linearization.jl
baggepinnen: sounds good, I don't expect types would change from what they were in the solution

PR #4394, opened 2026-03-16
[WIP] Improvements to the dynamic optimization interface

SebastianM-C:
## Checklist

- [ ] Appropriate tests were added
- [ ] Any code changes were done in a way that does not break public API
- [ ] All documentation related to code changes were updated
- [ ] The new code follows the
  [contributor guidelines](https://github.com/SciML/.github/blob/master/CONTRIBUTING.md), in particular the [SciML Style Guide](https://github.com/SciML/SciMLStyle) and
  [COLPRAC](https://github.com/SciML/COLPRAC).
- [ ] Any new documentation only uses public API
  
## Additional context

I will be splitting this in several PRs

PR #4909, opened 2026-08-09
Register callable parameters as solver nonlinear operators

SebastianM-C:
Splits the callable-parameter support out of #4394, rebuilt on current master.

Parameters that are *called* in the equations — `FunctionWrapper`-wrapped interpolators or bare callables such as `DataInterpolations` objects — previously broke the dynamic-optimization backends: `fixpoint_sub` with `fold = Val(true)` calls them with JuMP/InfiniteOpt decision-variable types they cannot accept, surfacing as e.g. `MethodError: no method matching *(::Symbolic, ::GeneralVariableRef)`.

This PR registers every called parameter as a JuMP nonlinear operator:

- arities are collected from the call sites (equations, observed, constraints, costs) via `callable_parameter_arities`, falling back to a `FunctionWrapper`'s argument-tuple type; parameters that are never called are untouched
- a symbolic first-derivative rule is resolved through Symbolics' registry when one exists (e.g. registered by a DataInterpolations extension) and used as the operator gradient; otherwise the operator is registered value-only and JuMP differentiates numerically
- the parameter's `pmap` entry and its `MTKParameters` nonnumeric storage are replaced with the registered operator, so both the symbolic path (InfiniteOpt equational constraints) and the numeric tracing path (JuMP `f_wrapper`) see the operator
- a bare independent variable inside a callable-parameter call (`forcing(t)`) is lowered onto the backend's time variable via the new `lowered_time_variable` hook

Tests cover the JuMP and InfiniteOpt backends with a `LinearInterpolation` parameter (constant extrapolation, since MOI's Hessian-by-ForwardDiff perturbs past the last knot at the boundary collocation point).

Known gap, unchanged from before: `register_operator!` has methods only for the InfiniteOpt model type (which the JuMP backend routes through); CasADi and Pyomo error when a callable parameter is present, as they previously did during substitution.

https://claude.ai/code/session_01MSmps1uroq5XBeStMsStG4

### DIFF
--- a/lib/ModelingToolkitBase/ext/MTKInfiniteOptExt.jl
+++ b/lib/ModelingToolkitBase/ext/MTKInfiniteOptExt.jl
@@ -11,6 +11,7 @@ using UnPack
 using Symbolics: unwrap
 import SymbolicUtils
 import NaNMath
+import FunctionWrappers
 const MTK = ModelingToolkitBase
 
 function __init__()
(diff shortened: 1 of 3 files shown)


AayushSabharwal:
This has a merge conflict, but tests pass. Will resolve the conflict and merge.

PR #4948, opened 2026-08-12
Sync package versions with the General registry

ChrisRackauckas-Claude:

## What this changes

| package | from | to | why |
|---|---|---|---|
| `ModelingToolkit` | 11.39.0 | 11.39.1 | unreleased changes with no version bump |

## Why

Each `to` is the next sequential version after what is currently registered in
General. Versions that skip an unregistered version are rejected by AutoMerge's
sequential version number guideline, so they can never be registered as-is;
packages with unreleased changes and no bump cannot be registered at all.

Only the top-level `version` field of each `Project.toml` is touched.

After merge, each package still needs `@JuliaRegistrator register` (with
`subdir=` where applicable).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ModelingToolkit"
 uuid = "961ee093-0014-501f-94e3-6117800e7a78"
-version = "11.39.0"
+version = "11.39.1"
 authors = ["Yingbo Ma <mayingbo5@gmail.com>", "Chris Rackauckas <accounts@chrisrackauckas.com> and contributors"]
 
 [deps]


ChrisRackauckas-Claude:
Note for reviewers: there is an older open General PR registering ModelingToolkit v11.40.0 from commit 1f6b035 (https://github.com/JuliaRegistries/General/pull/164058), after which master was unbumped back to 11.39.0. This PR takes the next sequential patch 11.39.1 from current master. The 11.40.0 General PR should not merge as-is — that commit is 35 behind master and would skip 11.39.1.

[noblock] not applicable here; this is the source PR, not the registry PR.

PR #4953, opened 2026-08-13
Document structural transformation developer API

ChrisRackauckas-Claude:
> Ignore this PR until reviewed by @ChrisRackauckas.

Documents only ModelingToolkit-owned structural-transformation developer APIs with canonical qualified targets. Dependency-owned and stale targets are removed from ModelingToolkit documentation; no API behavior changes.

Validation:
- Focused strict Documenter resolution and HTML rendering passed.
- Clean-master focused build fails on the legacy unresolved `@docs` targets and cross-references.
- Runic and `git diff --check` passed.

### DIFF
--- a/docs/src/internals/structural_transformation.md
+++ b/docs/src/internals/structural_transformation.md
@@ -1,88 +1,43 @@
 # Structural Transformation
 
-!!! warning "Internal API"
-    The functions documented on this page are internal implementation details of ModelingToolkit. They are not part of the public API and may change or be removed without notice in non-breaking releases. This documentation is provided to help contributors understand the codebase.
+!!! warning "Developer API"
+    These APIs support ModelingToolkit's structural-transformation machinery and
+    extension packages. They are version-controlled for ModelingToolkit developers;
+    end-user applications should use the documented public simplification API instead.
 
 These functions are used for structural analysis and transformation of equation systems, including index reduction, tearing, and other algebraic manipulations used in the simplification process.
 
 ## Tearing and Algebraic Simplification
 
 ```@docs
-tearing
-tearing_reassemble
-tearing_substitution
-torn_system_jacobian_sparsity
-find_solvables!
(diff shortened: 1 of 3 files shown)


docs/src/internals/structural_transformation.md
```diff
@@ -1,88 +1,43 @@
...
-tearing
-tearing_reassemble
-tearing_substitution
-torn_system_jacobian_sparsity
-find_solvables!
-linear_subsys_adjmat!
+ModelingToolkit.tearing
+ModelingToolkit.tearing_substitution
+ModelingToolkit.dummy_derivative
 ```
```
ChrisRackauckas, line 17: Why are the others removed?

ChrisRackauckas-Claude: You're right. The removals conflated the versioned ModelingToolkit developer API with lower-level implementation references owned by other packages. I restored `TearingState` and `mtkcompile!` to the rendered developer API and restored the remaining valid entries as fully qualified documentation from their canonical owners. `find_solvables!` remains removed because it is a `StateSelection` implementation helper, and `tearing_reassemble` was not defined. The docs and QA checks are running now.

ChrisRackauckas-Claude: The page now retains every exported structural-transformation developer API: `tearing`, `tearing_substitution`, `dummy_derivative`, `dae_index_lowering`, `pantelides_reassemble`, `sorted_incidence_matrix`, `but_ordered_incidence`, the two variable-mask functions, `TearingState`, and `mtkcompile!`.

The removed names in the original block are not supported ModelingToolkit extension points: `tearing_reassemble` is stale/undefined; `torn_system_jacobian_sparsity`, `find_solvables!`, and `linear_subsys_adjmat!` are private helpers (the latter two are not even defined by ModelingToolkit). `find_solvables!` remains intentionally private.

PR #4982, opened 2026-08-16
fix: record array elements in collect_operator_variables

ChrisRackauckas-Claude:
> [!NOTE]
> Draft — please ignore until reviewed by @ChrisRackauckas.

## The bug

`collect_operator_variables` records `arguments(v)[1]` for each operator application. When
the operator is applied to an array variable or a slice — `D(u[2:4])` — that argument is
the *array*, not its elements. Every caller tests membership of the **scalar** unknowns, so
nothing ever matches.

The visible consequence is in `DAEProblem`, which builds

```julia
diffvars = collect_differential_variables(sys)
differential_vars = map(Base.Fix2(in, diffvars), unknowns(sys))
```

so a system whose equations differentiate an array slice reports **every variable as
algebraic**, and DAE initialization then has nothing to work from.

## Failing before, passing after

```julia
@variables w(t)[1:4]
collect_differential_variables(D(w[2:3]) ~ w[1:2])
```

| | result | membership map over `unknowns` |
|---|---|---|
| before | `Set([w[2:3]])` | `[0, 0, 0, 0]` |
| after  | `Set([w[2], w[3]])` | `[0, 1, 1, 0]` |

Measured downstream on a real system (a MethodOfLines finite-difference discretization
that emits its interior as one array equation), `DAEProblem`'s `differential_vars` goes
from **0/21 to 19/21** at one grid size and **0/81 to 79/81** at another — 19 and 79 being
the interior points, with the two boundary points genuinely algebraic.

## The change

One function. When the operator's argument is array-valued, scalarize it and record the
elements; otherwise behave exactly as before, so scalar operators take an identical path.

## Tests

`lib/ModelingToolkitBase/test/variable_utils.jl` gains a testset covering a whole array
variable, a slice, an unaffected scalar derivative, and the `differential_vars`-style
membership test that motivated the fix. It fails on unpatched code with the
`[0, 0, 0, 0]` map shown above.

```
Test Summary:  | Pass  Total      Time
variable_utils |  145    145  11m03.2s
```

(145 = 141 pre-existing + 4 added.)

## Not verified

I ran the `variable_utils` file, not the full ModelingToolkitBase suite — the change is
additive and confined to one branch of one function, but a full run has not been done
locally. Other callers of `collect_operator_variables` (`collect_applied_operators`,
`Shift`-based discrete paths) are unaffected for scalar arguments and now see elements
rather than arrays for array ones, which is the intended correction but is worth a
reviewer's eye.

## Context

Found while making finite-difference PDE discretizations emit array (slice-form) equations
instead of one scalar equation per grid point — see SciML/MethodOfLines.jl#428. This is one
of several places where array equations lose their shape; the others (equation counting,
implicit-DAE codegen row indexing) are more invasive and will be proposed separately.

https://claude.ai/code/session_01S1dxUEPGysx27SSSpxtiQ9

### DIFF
--- a/lib/ModelingToolkitBase/src/utils.jl
+++ b/lib/ModelingToolkitBase/src/utils.jl
@@ -714,7 +714,17 @@ function collect_operator_variables(eqs::Vector{Equation}, ::Type{op}) where {op
         SU.search_variables!(vars, eq; is_atomic = OperatorIsAtomic{op}())
         for v in vars
             isoperator(v, op) || continue
-            push!(diffvars, arguments(v)[1])
+            arg = arguments(v)[1]
+            # An operator applied to an array variable or slice, such as `D(u[2:4])`,
+            # names the array rather than its elements. Callers test membership of the
+            # scalar unknowns, so record the elements.
+            if symtype(arg) <: AbstractArray
+                for el in vec(collect(Symbolics.scalarize(wrap(arg))))
+                    push!(diffvars, unwrap(el))
+                end
+            else
+                push!(diffvars, arg)
+            end
         end
         empty!(vars)
     end
--- a/lib/ModelingToolkitBase/test/variable_utils.jl
+++ b/lib/ModelingToolkitBase/test/variable_utils.jl
@@ -1,5 +1,5 @@
 using ModelingToolkitBase, Test
-using ModelingToolkitBase: value, parse_variable
+using ModelingToolkitBase: value, parse_variable, unwrap
 using SymbolicUtils: <ₑ
 import SymbolicUtils as SU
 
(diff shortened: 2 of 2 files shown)


AayushSabharwal (changes requested):
The comments in #4982 on this part of the code apply here.

ChrisRackauckas:
Closing as superseded: already on master or replaced by later work.

---
🤖 Posted by an AI agent — harness: Claude Code, model: claude-opus-5[1m]. Chris asked for these closures directly.

https://claude.ai/code/session_014FEzNTLFutCmTEAZ3zBg5R

PR #4983, opened 2026-08-16
feat: accept array equations on the implicit-DAE path

ChrisRackauckas-Claude:
> [!NOTE]
> Draft — please ignore until reviewed by @ChrisRackauckas.
> Stacked on #4982 (`collect_operator_variables`), which supplies the correct
> `differential_vars` this relies on.

## Why

A finite-difference PDE discretization can emit its interior as **one array equation over
slices** instead of one scalar equation per grid point, which makes the symbolic system
size independent of grid resolution (SciML/MethodOfLines.jl#428). Measured on the 1D heat
equation, generating code directly from such a system: **1 614 expression nodes at every
grid size, versus 20 646 → 117 446 growing** for the scalarized form, with identical
numerics.

`mtkcompile` scalarizes array equations, so that structure is erased before codegen.
`ODEProblem` fundamentally needs it to: it requires `D(x) = f(x)`, and isolating the
derivative out of a residual *is* structural simplification.

The implicit-DAE path does not have that requirement — a residual `D(u) - f ~ 0` is exactly
what `DAEProblem` consumes.

## What changed

Four changes, all inert outside the implicit-DAE path with array equations:

| | |
|---|---|
| `check_array_equations_unknowns` | skipped when `implicit_dae`; **every other problem type keeps the guard** |
| `check_eqs_u0` | skipped only when a DAE system *actually contains* array equations, where an equation count is not comparable with an unknown count — scalarized DAEs keep the check |
| `expand_array_derivatives` | rewrites `D(u[2:4])` into the scalar derivatives bound to the `du` argument, which a derivative of a slice otherwise matches none of |
| `flatten_array_residuals` | expands an array-valued residual entry into one output row per element, **indexing the shared expression** rather than scalarizing it |

Both codegen helpers return their input unchanged when nothing is array-valued, so existing
implicit-DAE systems generate identical code.

## Evidence

Before this, `DAEProblem` on such a system failed in sequence: the array-equation guard,
then `Equations (3), unknowns (21)`, then a literal `Differential` surviving into the
generated code (`MethodError: no method matching (::Differential)(::Vector{Float64})`).

After, on the heat equation with 3 equations and 21 unknowns:

  - **21 output rows** produced from 3 equations (81 from 3 at the larger size)
  - `differential_vars` = **19/21** and **79/81** — interior differential, boundaries algebraic
  - residual evaluates finitely
  - end-to-end solve with `DFBDF` + `BrownFullBasicInit`: **`max|u - exact| = 7.57e-4`**
    against `exp(-π²t)sin(πx)`, the expected second-order spatial error on 21 points
  - `ODEProblem` on the same system still throws

```
Test Summary: | Pass  Total    Time
arrdae        |   11     11  4m16.8s
```

Includes a 2D case: `expand_array_derivatives` originally flattened the scalarized
derivative with `vec`, which produced a length-81 vector against 9×9 surrounding slices
and failed with `DimensionMismatch`. Every array equation of two or more dimensions was
affected, and the 1D tests could not catch it because `vec` is a no-op there. Fixed, with
a regression test.

## Notes for review

  - `DAEProblem` still needs `build_initializeprob=false` here: MTK's initialization system
    builds a time-independent system and rejects the `Differential` inside the array
    equation. The solver's own `BrownFullBasicInit` covers that case, but it is only valid
    when no initialization equation constrains a derivative or an algebraic variable, so
    automating the choice needs care. Not addressed in this PR.
  - Array equations only survive `System` construction in **residual** form. Written
    naturally as `D(u[2:n-1]) ~ rhs`, `diff2term` fails with
    `Can only build StableIndex{Int} from indexed symbolic` — it cannot name a derivative
    of a slice. Worth knowing for anyone generating such systems.
  - I have not measured whether CSE hoists the shared array expression across the expanded
    rows. If it does not, the stencil is recomputed per row, which would be worse than
    scalarizing — so the performance claim above is about the *symbolic* representation,
    not yet about generated-code quality.
  - Tests are `lib/ModelingToolkitBase/test/array_equation_dae.jl`; I ran that file and
    `variable_utils.jl`, not the full suite.

https://claude.ai/code/session_01S1dxUEPGysx27SSSpxtiQ9

### DIFF
--- a/lib/ModelingToolkitBase/src/systems/abstractsystem.jl
+++ b/lib/ModelingToolkitBase/src/systems/abstractsystem.jl
@@ -3297,11 +3297,7 @@ count_equation_rows(eqs) = sum(equation_row_count, eqs; init = 0)
 equation_row_count(eq) = 1
 
 function equation_row_count(eq::Equation)
-    # A residual may be written with the array on either side, as `D(u[2:4]) ~ rhs` or
-    # `0 ~ rhs`.
-    sh = SU.shape(eq.lhs)
-    SU.is_array_shape(sh) || (sh = SU.shape(eq.rhs))
-    return SU.is_array_shape(sh) ? prod(length, sh) : 1
+    return prod(length, SU.shape(eq.lhs)::SU.ShapeVecT; init = 1)
 end
 
 function check_eqs_u0(eqs, dvs, u0; check_length = true, kwargs...)
--- a/lib/ModelingToolkitBase/src/systems/abstractsystem.jl
+++ b/lib/ModelingToolkitBase/src/systems/abstractsystem.jl
@@ -3286,17 +3286,36 @@ function check_array_equations_unknowns(eqs, dvs)
     end
 end
 
+"""
+    $(TYPEDSIGNATURES)
+
+Number of scalar residual rows the equations stand for. An array equation contributes one
+row per element, so it cannot be counted as a single equation.
+"""
+count_equation_rows(eqs) = sum(equation_row_count, eqs; init = 0)
+
+equation_row_count(eq) = 1
+
+function equation_row_count(eq::Equation)
+    # A residual may be written with the array on either side, as `D(u[2:4]) ~ rhs` or
+    # `0 ~ rhs`.
+    sh = SU.shape(eq.lhs)
+    SU.is_array_shape(sh) || (sh = SU.shape(eq.rhs))
+    return SU.is_array_shape(sh) ? prod(length, sh) : 1
+end
+
 function check_eqs_u0(eqs, dvs, u0; check_length = true, kwargs...)
+    neqs = count_equation_rows(eqs)
     if u0 !== nothing
         if check_length
-            if !(length(eqs) == length(dvs) == length(u0))
-                throw(ArgumentError("Equations ($(length(eqs))), unknowns ($(length(dvs))), and initial conditions ($(length(u0))) are of different lengths."))
+            if !(neqs == length(dvs) == length(u0))
+                throw(ArgumentError("Equations ($(neqs)), unknowns ($(length(dvs))), and initial conditions ($(length(u0))) are of different lengths."))
             end
         elseif length(dvs) != length(u0)
             throw(ArgumentError("Unknowns ($(length(dvs))) and initial conditions ($(length(u0))) are of different lengths."))
         end
-    elseif check_length && (length(eqs) != length(dvs))
-        throw(ArgumentError("Equations ($(length(eqs))) and Unknowns ($(length(dvs))) are of different lengths."))
+    elseif check_length && (neqs != length(dvs))
+        throw(ArgumentError("Equations ($(neqs)) and Unknowns ($(length(dvs))) are of different lengths."))
     end
     return nothing
 end
--- a/lib/ModelingToolkitBase/src/systems/codegen.jl
+++ b/lib/ModelingToolkitBase/src/systems/codegen.jl
@@ -40,12 +40,10 @@ All other keyword arguments are forwarded to [`build_function_wrapper`](@ref).
 
 """
 Treat a derivative of an array-valued expression as a leaf, so that
-[`expand_array_derivatives`](@ref) collects `D(u[2:4])` itself rather than descending into
+[`expand_array_derivatives!`](@ref) collects `D(u[2:4])` itself rather than descending into
 it. Scalar variables are not atomic here, so nothing else is collected.
 """
-struct ArrayDerivativeIsAtomic end
-
-function (::ArrayDerivativeIsAtomic)(ex::SymbolicT)
+function array_derivative_is_atomic(ex::SymbolicT)
     return isdifferential(ex) && SU.is_array_shape(SU.shape(ex))
 end
 
@@ -53,15 +51,18 @@ end
     $(TYPEDSIGNATURES)
 
 Rewrite derivatives of array-valued expressions, such as `D(u[2:4])`, into arrays of the
-corresponding scalar derivatives. Implicit-DAE codegen binds scalar `D(uᵢ)` terms to
-elements of the `du` argument, and a derivative of a slice matches none of them.
+corresponding scalar derivatives, in place. Implicit-DAE codegen binds scalar `D(uᵢ)` terms
+to elements of the `du` argument, and a derivative of a slice matches none of them.
 
-Takes every residual at once so that the search and substitution caches are shared.
+Takes every residual at once, and works in the system's `ir`, so the search and
+substitution caches are shared and the rewritten residuals are already populated for
+codegen.
 """
-function expand_array_derivatives(rhss::Vector{SymbolicT})
+function expand_array_derivatives!(rhss::Vector{SymbolicT}, ir::IRStructure{VartypeT})
     terms = Set{SymbolicT}()
+    buffer = SU.IRStructureSearchBuffer(ir, terms)
     for rhs in rhss
-        SU.search_variables!(terms, rhs; is_atomic = ArrayDerivativeIsAtomic())
+        SU.search_variables!(buffer, rhs; is_atomic = array_derivative_is_atomic)
     end
     isempty(terms) && return rhss
 
@@ -69,16 +70,23 @@ function expand_array_derivatives(rhss::Vector{SymbolicT})
     for term in terms
         op = operation(term)
         arg = only(arguments(term))
-        sh = SU.shape(arg)
+        sh = SU.shape(arg)::SU.ShapeVecT
         # Preserve the shape: a derivative of a 2D slice must expand to a 2D array of
         # scalar derivatives, or it will not broadcast against the surrounding slices.
-        sz = ntuple(i -> length(sh[i]), length(sh))
-        els = [op(arg[idx]) for idx in SU.stable_eachindex(arg)]
-        subs[term] = SU.Const{VartypeT}(reshape(els, sz))
+        arrargs = Symbolics.SArgsT()
+        sizehint!(arrargs, prod(length, sh; init = 1) + 1)
+        push!(arrargs, SU.Const{VartypeT}(size(arg)))
+        for idx in SU.stable_eachindex(arg)
+            push!(arrargs, op(arg[idx]))
+        end
+        subs[term] = Symbolics.STerm(
+            SU.array_literal, arrargs; type = symtype(arg), shape = sh
+        )
     end
 
-    subber = SU.IRSubstituter{false}(IRStructure{VartypeT}(), subs)
-    return map(subber, rhss)
+    subber = SU.IRSubstituter{false}(ir, subs)
+    map!(subber, rhss, rhss)
+    return rhss
 end
 
 """
@@ -98,7 +106,7 @@ function array_residual_maker(rhss::Vector{SymbolicT})
     values = similar(rhss)
     offset = 0
     for (i, rhs) in enumerate(rhss)
-        sh = SU.shape(rhs)
+        sh = SU.shape(rhs)::SU.ShapeVecT
         if SU.is_array_shape(sh)
             n = prod(length, sh)
             # Elements become consecutive output rows, so a rank > 1 residual is flattened
@@ -165,7 +173,7 @@ function generate_rhs(
             rhss = SymbolicT[_iszero(eq.lhs) ? eq.rhs : eq.rhs - eq.lhs for eq in eqs]
             # Rewrite array derivatives to the scalar ones bound to the `du` argument.
             # Assembly into a single array happens below, after assertions.
-            rhss = expand_array_derivatives(rhss)
+            expand_array_derivatives!(rhss, get_irstructure(sys))
             assemble_residuals = true
         end
     else
--- a/lib/ModelingToolkitBase/src/systems/codegen.jl
+++ b/lib/ModelingToolkitBase/src/systems/codegen.jl
@@ -37,6 +37,74 @@ $GENERATE_X_KWARGS
 
 All other keyword arguments are forwarded to [`build_function_wrapper`](@ref).
 """
+
+"""
+    $(TYPEDSIGNATURES)
+
+Rewrite a derivative of an array-valued expression, such as `D(u[2:4])`, into an array of
+the corresponding scalar derivatives. Implicit-DAE codegen binds scalar `D(uᵢ)` terms to
+elements of the `du` argument, and a derivative of a slice matches none of them.
+"""
+function expand_array_derivatives(ex)
+    terms = _array_derivative_terms!(Any[], ex)
+    isempty(terms) && return ex
+    subs = Dict()
+    for term in terms
+        op = operation(term)
+        arg = unwrap(only(arguments(term)))
+        scalars = vec(collect(Symbolics.scalarize(wrap(arg))))
+        subs[term] = [unwrap(op(sc)) for sc in scalars]
+    end
+    return unwrap(Symbolics.substitute(wrap(ex), subs))
+end
+
+function _array_derivative_terms!(acc, ex)
+    ex = unwrap(ex)
+    iscall(ex) || return acc
+    op = operation(ex)
+    if op isa Differential
+        arg = unwrap(only(arguments(ex)))
+        symtype(arg) <: AbstractArray && push!(acc, ex)
+        return acc
+    end
+    for arg in arguments(ex)
+        _array_derivative_terms!(acc, arg)
+    end
+    return acc
+end
+
+"""
+    $(TYPEDSIGNATURES)
+
+Expand array-valued residual entries so each contributes one output row. The rows index
+into the same underlying expression rather than being scalarized separately, so the array
+computation is built once.
+"""
+function flatten_array_residuals(rhss)
+    any(r -> symtype(unwrap(r)) <: AbstractArray, rhss) || return rhss
+    out = Any[]
+    for r in rhss
+        ru = unwrap(r)
+        if !(symtype(ru) <: AbstractArray)
+            push!(out, ru)
+            continue
+        end
+        sz = try
+            size(wrap(ru))
+        catch
+            nothing
+        end
+        if sz === nothing
+            append!(out, vec(collect(Symbolics.scalarize(wrap(ru)))))
+        else
+            for idx in CartesianIndices(sz)
+                push!(out, unwrap(wrap(ru)[Tuple(idx)...]))
+            end
+        end
+    end
+    return out
+end
+
 function generate_rhs(
         sys::System, opts::GeneratedFunctionOptions;
         implicit_dae::Bool = false, scalar::Bool = false,
@@ -84,6 +152,11 @@ function generate_rhs(
             D = Differential(t)
             ddvs = map(D, dvs)
             rhss = [_iszero(eq.lhs) ? eq.rhs : eq.rhs - eq.lhs for eq in eqs]
+            # An array equation stands for one residual row per element. Rewrite its
+            # derivative term to the scalar derivatives bound to the `du` argument, then
+            # index the residual per row. Both are no-ops when no equation is array-valued.
+            rhss = map(expand_array_derivatives, rhss)
+            rhss = flatten_array_residuals(rhss)
         end
     else
         if !override_discrete && !is_discrete_system(sys)
--- a/lib/ModelingToolkitBase/src/systems/codegen.jl
+++ b/lib/ModelingToolkitBase/src/systems/codegen.jl
@@ -52,8 +52,10 @@ function expand_array_derivatives(ex)
     for term in terms
         op = operation(term)
         arg = unwrap(only(arguments(term)))
-        scalars = vec(collect(Symbolics.scalarize(wrap(arg))))
-        subs[term] = [unwrap(op(sc)) for sc in scalars]
+        # Preserve the shape: a derivative of a 2D slice must substitute a 2D array of
+        # scalar derivatives, or it will not broadcast against the surrounding slices.
+        scalars = collect(Symbolics.scalarize(wrap(arg)))
+        subs[term] = map(sc -> unwrap(op(sc)), scalars)
     end
     return unwrap(Symbolics.substitute(wrap(ex), subs))
 end
--- a/lib/ModelingToolkitBase/src/systems/codegen.jl
+++ b/lib/ModelingToolkitBase/src/systems/codegen.jl
@@ -16,29 +16,6 @@ const EXPERIMENTAL_WARNING = """
 """
 
 """
-    $(TYPEDSIGNATURES)
-
-Generate the RHS function for the [`equations`](@ref) of a [`System`](@ref).
-
-# Keyword Arguments
-
-$GENERATE_X_KWARGS
-- `implicit_dae`: Whether the generated function should be in the implicit form. Applicable
-  only for ODEs/DAEs or discrete systems. Instead of `f(u, p, t)` (`f(du, u, p, t)` for the
-  in-place form) the function is `f(du, u, p, t)` (respectively `f(resid, du, u, p, t)`).
-- `override_discrete`: Whether to assume the system is discrete regardless of
-  `is_discrete_system(sys)`.
-- `scalar`: Whether to generate a single-out-of-place function that returns a scalar for
-  the only equation in the system.
-- `extra_args`: Extra trailing symbolic arguments appended after all standard arguments and
-  kept out of the `MTKParameters` collapse, so they stay live scalar arguments of the
-  generated function (e.g. the homotopy continuation `λ`, threaded as the "t slot"). Empty
-  by default, which leaves the standard codegen path byte-identical.
-
-All other keyword arguments are forwarded to [`build_function_wrapper`](@ref).
-"""
-
-"""
 Treat a derivative of an array-valued expression as a leaf, so that
 [`expand_array_derivatives!`](@ref) collects `D(u[2:4])` itself rather than descending into
 it. Scalar variables are not atomic here, so nothing else is collected.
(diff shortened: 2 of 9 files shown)


lib/ModelingToolkitBase/src/systems/codegen.jl
```diff
@@ -37,6 +37,76 @@ $GENERATE_X_KWARGS
...
+        arg = unwrap(only(arguments(term)))
+        # Preserve the shape: a derivative of a 2D slice must substitute a 2D array of
+        # scalar derivatives, or it will not broadcast against the surrounding slices.
+        scalars = collect(Symbolics.scalarize(wrap(arg)))
+        subs[term] = map(sc -> unwrap(op(sc)), scalars)
+    end
+    return unwrap(Symbolics.substitute(wrap(ex), subs))
+end
+
+function _array_derivative_terms!(acc, ex)
```
AayushSabharwal: Recursion is bad. This won't scale to deep expressions, though it's possible PDEs won't hit this case quite yet.

lib/ModelingToolkitBase/src/systems/codegen.jl
```diff
@@ -37,6 +37,76 @@ $GENERATE_X_KWARGS
...
 """
+
+"""
+    $(TYPEDSIGNATURES)
+
+Rewrite a derivative of an array-valued expression, such as `D(u[2:4])`, into an array of
+the corresponding scalar derivatives. Implicit-DAE codegen binds scalar `D(uᵢ)` terms to
+elements of the `du` argument, and a derivative of a slice matches none of them.
+"""
+function expand_array_derivatives(ex)
```
AayushSabharwal: This entire function can be made more efficient. Instead of making half a dozen comments, I'll compile it into this one. `_array_derivative_terms!` is just `SU.search_variables!` with `is_atomic = x -> isdifferential(x) && SU.is_array_shape(SU.shape(x))` to get all array derivative applications. This removes the entire function and makes it more efficient. The buffer should be a `Set{SymbolicT}()`. `subs` should be concretely typed. Instead of the weird `collect(scalarize(wrap(` this should use `SU.stable_eachindex` to iterate over the symbolic array and manually build an `array_literal`. `wrap` is dynamic dispatch and doesn't infer, as is every function after that point. Substitution should use `IRSubstituter`. There's no need to `wrap` the argument to `substitute` and then `unwrap` it.

If this function also accepted `rhss` in its entirety instead of one element at a time, the caching in `search_variables!` and `IRSubstituter` is also shared.

lib/ModelingToolkitBase/src/systems/codegen.jl
```diff
@@ -37,6 +37,76 @@ $GENERATE_X_KWARGS
...
+end
+
+"""
+    $(TYPEDSIGNATURES)
+
+Expand array-valued residual entries so each contributes one output row. The rows index
+into the same underlying expression rather than being scalarized separately, so the array
+computation is built once.
+"""
+function flatten_array_residuals(rhss)
```
AayushSabharwal: This entire function is redundant. `generate_rhs` should build an `ArrayMaker` expression where each equation writes to the appropriate contiguous region of the output. That will codegen much better than this approach. For example, a 1D PDE with a body of length `n` would generate something equivalent to

```julia
@makearray out[1:(n+2)] begin
  out[1:1] => left_boundary
  out[2:(n+1)] => body
  out[(n+2):(n+2)] => right_boundary
end
```

Except generated programmatically with the constructor instead of with a macro.

lib/ModelingToolkitBase/src/systems/problem_utils.jl
```diff
@@ -2235,7 +2238,11 @@ function __process_SciMLProblem(
         u0 = u0_constructor(u0)
     end
 
-    check_eqs_u0(eqs, dvs, u0; check_length, kwargs...)
+    # On the implicit-DAE path an array equation stands for one row per element, so a
+    # count of equations is not comparable with a count of unknowns.
+    if !(implicit_dae && any(eq -> Symbolics.isarraysymbolic(eq.lhs), eqs))
+        check_eqs_u0(eqs, dvs, u0; check_length, kwargs...)
```
AayushSabharwal: This check can still work if it is refactored slightly to not assume each equation is a scalar.

lib/ModelingToolkitBase/src/utils.jl
```diff
@@ -714,7 +714,17 @@ function collect_operator_variables(eqs::Vector{Equation}, ::Type{op}) where {op
         SU.search_variables!(vars, eq; is_atomic = OperatorIsAtomic{op}())
         for v in vars
             isoperator(v, op) || continue
-            push!(diffvars, arguments(v)[1])
+            arg = arguments(v)[1]
+            # An operator applied to an array variable or slice, such as `D(u[2:4])`,
+            # names the array rather than its elements. Callers test membership of the
+            # scalar unknowns, so record the elements.
+            if symtype(arg) <: AbstractArray
```
AayushSabharwal: `SU.is_array_shape(SU.shape(arg))` is a more efficient check.

lib/ModelingToolkitBase/src/utils.jl
```diff
@@ -714,7 +714,17 @@ function collect_operator_variables(eqs::Vector{Equation}, ::Type{op}) where {op
...
+                for el in vec(collect(Symbolics.scalarize(wrap(arg))))
+                    push!(diffvars, unwrap(el))
```
AayushSabharwal:
```suggestion
                for idx in SU.stable_eachindex(arg)
                    push!(diffvars, arg[idx])
```

Much much faster

AayushSabharwal (changes requested):

lib/ModelingToolkitBase/src/systems/codegen.jl
ChrisRackauckas-Claude: Done — rewritten along all of those lines. `_array_derivative_terms!` is gone, replaced by `search_variables!` with an `ArrayDerivativeIsAtomic` callable (`isdifferential(ex) && SU.is_array_shape(SU.shape(ex))`); since scalar variables are not atomic under that predicate, nothing else is collected and no post-filter is needed. Buffer is `Set{SymbolicT}`, `subs` is `Dict{SymbolicT, SymbolicT}`, the elements come from `SU.stable_eachindex` instead of `collect(scalarize(wrap(...)))`, substitution goes through `IRSubstituter{false}`, and the `wrap`/`unwrap` round trip is gone.

It now takes `rhss` in its entirety, so the `search_variables!` and `IRSubstituter` caches are shared across residuals.

One detail worth flagging: the elements have to be `reshape`d back to the argument shape before building the literal. A derivative of a 2D slice must expand to a 2D array or it stops broadcasting against the surrounding slices — that was a real bug in the previous revision of this PR. I verified `stable_eachindex` iterates column-major and matches `scalarize` exactly, including for rank 2.

lib/ModelingToolkitBase/src/utils.jl
ChrisRackauckas-Claude: Done, using `SU.is_array_shape(SU.shape(arg))`.

lib/ModelingToolkitBase/src/utils.jl
ChrisRackauckas-Claude: Done.

lib/ModelingToolkitBase/src/systems/codegen.jl
ChrisRackauckas-Claude: Done — `flatten_array_residuals` is gone, replaced by `array_residual_maker`, which builds a single `ArrayMaker` where each residual writes to a contiguous region, exactly as in your sketch but constructed programmatically.

Two constraints I hit that shape the implementation:

- `ArrayMaker` rejects scalar values (`AssertionError: sT <: AbstractArray`), so a scalar residual is lifted to a length-1 array literal via `SU.Const{VartypeT}([rhs])` and given a length-1 region.
- The output is a flat residual vector, so a rank > 1 residual cannot be given a multidimensional region. There is no `reshape` for symbolic arrays, and `vec(wrap(x))` scalarizes — but `vec` on the unwrapped `BasicSymbolic` stays symbolic (`vec((w(t))[2:3, 4:5])`, shape `1:4`), so that is what is used.

It does codegen much better, as you said. Generated `DAEFunction` expression size for a 1D heat system, before vs after:

| n | `flatten_array_residuals` | `ArrayMaker` |
|---|---|---|
| 20 | 19,141 | 11,948 |
| 200 | 140,095 | 53,035 |
| 2000 | 1,393,803 | 480,776 |

Still linear in `n` — the remaining growth is the unknowns/argument handling, not the residual body — but the constant is ~2.9x smaller at n=2000.

lib/ModelingToolkitBase/src/systems/problem_utils.jl
ChrisRackauckas-Claude: Done — the check runs on this path now rather than being skipped. `check_eqs_u0` counts residual rows instead of equations: a new `count_equation_rows`/`equation_row_count` sums `prod(length, shape)` for an array equation and 1 otherwise, taking the shape from whichever side carries it (`D(u[2:4]) ~ rhs` or `0 ~ rhs`).

Since array equations previously could not reach `check_eqs_u0` at all, this is a strict generalization for every other caller. `GROUP=InterfaceI` passes: 1595 passed, 0 failed, 5 pre-existing broken.

ChrisRackauckas-Claude:
Pushed as d6640a0, rebased onto d9c3814 so the accepted suggestion is preserved (that commit applied the `stable_eachindex` half of the `collect_operator_variables` comment; the rebase also applies the `is_array_shape` half).

Replies are on each thread. Summary:

- `_array_derivative_terms!` removed in favour of `search_variables!` with an array-derivative `is_atomic`; concretely typed `Set`/`Dict`, `stable_eachindex`, `IRSubstituter`, no `wrap`/`unwrap` round trip, and it takes all of `rhss` so the caches are shared.
- `flatten_array_residuals` removed in favour of an `ArrayMaker` with one contiguous region per residual. Generated expression size drops ~2.9x at n=2000.
- `check_eqs_u0` now counts residual rows, so the check runs on the implicit-DAE path instead of being skipped.
- `collect_operator_variables` uses `is_array_shape`/`stable_eachindex`.

**One thing I found while doing this, which is not from the review.** The natural `D(u[2:n-1]) ~ rhs` form crashed with `ArgumentError: Can only build StableIndex{Int} from indexed symbolic`, thrown from `diff2term` via `ir_info.jl`: a derivative of a slice has no scalar `toterm` name to key a substitution on. The existing tests only covered the residual form (`broadcast(-, D(...), lap) ~ 0`), which is what MethodOfLines emits, so it was not caught. `ir_info` now skips array-valued LHSs — the implicit-DAE path has already expanded those derivatives to their scalar elements by then — and there is a new testset covering the form end to end. Happy to split that into its own PR if you would rather keep this one to the review changes.

Verification, all local:

```
array equations reach DAEProblem                        |    6      6
other problem types still require scalarized equations  |    1      1
array-equation DAE solves to the analytic solution      |    2      2
array equations over a 2D slice keep their shape        |    2      2
array equations written as `D(slice) ~ rhs`             |    4      4

GROUP=InterfaceI: InterfaceI | 1595  5(broken)  1600  56m00.2s
     Testing ModelingToolkitBase tests passed
```

`runic --check` is clean. I have not run the full suite beyond `InterfaceI`.

lib/ModelingToolkitBase/src/systems/abstractsystem.jl
```diff
@@ -3286,17 +3286,36 @@ function check_array_equations_unknowns(eqs, dvs)
...
+    $(TYPEDSIGNATURES)
+
+Number of scalar residual rows the equations stand for. An array equation contributes one
+row per element, so it cannot be counted as a single equation.
+"""
+count_equation_rows(eqs) = sum(equation_row_count, eqs; init = 0)
+
+equation_row_count(eq) = 1
+
+function equation_row_count(eq::Equation)
```
AayushSabharwal, line 3299: Just typeassert that `shape(eq.lhs)::SU.ShapeVecT` and return `prod(length, sh; init = 1)`

lib/ModelingToolkitBase/src/systems/codegen.jl
```diff
@@ -37,6 +37,84 @@ $GENERATE_X_KWARGS
 
 All other keyword arguments are forwarded to [`build_function_wrapper`](@ref).
 """
+
+"""
+Treat a derivative of an array-valued expression as a leaf, so that
+[`expand_array_derivatives`](@ref) collects `D(u[2:4])` itself rather than descending into
+it. Scalar variables are not atomic here, so nothing else is collected.
+"""
+struct ArrayDerivativeIsAtomic end
```
AayushSabharwal: This shouldn't be a callable singleton struct instead of a normal function?

lib/ModelingToolkitBase/src/systems/codegen.jl
```diff
@@ -37,6 +37,84 @@ $GENERATE_X_KWARGS
...
+Rewrite derivatives of array-valued expressions, such as `D(u[2:4])`, into arrays of the
+corresponding scalar derivatives. Implicit-DAE codegen binds scalar `D(uᵢ)` terms to
+elements of the `du` argument, and a derivative of a slice matches none of them.
+
+Takes every residual at once so that the search and substitution caches are shared.
+"""
+function expand_array_derivatives(rhss::Vector{SymbolicT})
+    terms = Set{SymbolicT}()
+    for rhs in rhss
+        SU.search_variables!(terms, rhs; is_atomic = ArrayDerivativeIsAtomic())
```
AayushSabharwal: Using `IRStructureSearchBuffer` here is much faster

lib/ModelingToolkitBase/src/systems/codegen.jl
```diff
@@ -37,6 +37,84 @@ $GENERATE_X_KWARGS
...
+    isempty(terms) && return rhss
+
+    subs = Dict{SymbolicT, SymbolicT}()
+    for term in terms
+        op = operation(term)
+        arg = only(arguments(term))
+        sh = SU.shape(arg)
+        # Preserve the shape: a derivative of a 2D slice must expand to a 2D array of
+        # scalar derivatives, or it will not broadcast against the surrounding slices.
+        sz = ntuple(i -> length(sh[i]), length(sh))
```
AayushSabharwal:
```suggestion
        sz = size(arg)
```

lib/ModelingToolkitBase/src/systems/codegen.jl
```diff
@@ -37,6 +37,84 @@ $GENERATE_X_KWARGS
...
+        els = [op(arg[idx]) for idx in SU.stable_eachindex(arg)]
```
AayushSabharwal: A better method is to build `arrargs = Symbolics.SArgsT()` and `sizehint!` it. Then, push `sz` and each of `op(arg[idx])`. This allows building `Symbolics.STerm(SU.array_literal, arrargs; type = symtype(arg), shape = sh)`. This avoids the dynamic dispatch `reshape`.

lib/ModelingToolkitBase/src/systems/codegen.jl
```diff
@@ -37,6 +37,84 @@ $GENERATE_X_KWARGS
...
+        subs[term] = SU.Const{VartypeT}(reshape(els, sz))
+    end
+
+    subber = SU.IRSubstituter{false}(IRStructure{VartypeT}(), subs)
```
AayushSabharwal: Why not just reuse `get_irstructure(sys)`? That way the expression is also prepopulated for codegen to use.

lib/ModelingToolkitBase/src/systems/codegen.jl
```diff
@@ -37,6 +37,84 @@ $GENERATE_X_KWARGS
...
+    return map(subber, rhss)
```
AayushSabharwal: Shouldn't need the `map` iirc. Could also use `map!` and avoid the allocation.

ChrisRackauckas-Claude:
**Correction to the codegen numbers above.** I measured `length(string(DAEFunction(...)))`, but that string embeds the `System`, whose `show` prints all n unknowns. So roughly 4n of what I counted was the system's printed repr, not generated code. Both columns were inflated the same way, so the comparison was directionally right, but it understated the improvement and it made me wrongly say the new residual body is "still linear in n". It is not.

Counting generated code only (everything before the embedded `sys = Model sys:`), and splitting out the `du` argument reads:

| n | old total | old non-`du` | new total | new non-`du` |
|---|---|---|---|---|
| 20 | 237 | 201 | 134 | **98** |
| 200 | 1,677 | 1,281 | 494 | **98** |
| 2000 | 16,077 | 12,081 | 4,094 | **98** |

The `ArrayMaker` residual body is **O(1)** — exactly 98 lines at every size, with the broadcast count fixed at 10. The old `flatten_array_residuals` body was O(n) (12,081 lines at n=2000). So the improvement is ~123x on the residual body, not the ~2.9x I quoted.

**What is still O(n), and why.** The remaining growth is entirely `2(n-2)` element reads of the `du` argument (`local var"##cse#k" = __mtk_arg_1[i]`). The state vector does not have this problem: `u` is reconstructed once as `reshape(view(___mtkunknowns___, 1:n), (n,))`, so `u[2:n-1]` is a genuine O(1) slice.

The asymmetry comes from `array_variable_assignments` being filtered by `required_arrvars`, which only contains array variables appearing *unindexed* in the expression. `expand_array_derivatives` rewrites `D(u[2:n-1])` into an array literal of scalar `D(u[i])` terms, so `D(u)` never appears as an array, no view is emitted for the `du` buffer, and each element is read individually.

I tried the obvious fix — substituting `D(u)[2:n-1]` instead of the literal, so that `D(u)` lands in `required_arrvars`. That does not work: indexing an array derivative normalizes straight back, `D(u)[2:4]` becomes `D(u[2:4])`, so the slice-of-array-derivative form does not survive as an expression. Making the `du` side O(1) would need codegen to bind the `du` buffer as an array and slice the bound local, i.e. a change in `build_function_wrapper` rather than in this expansion. Happy to look at that separately if you think it is worth it — it would make the whole generated function O(1) in the grid size.

lib/ModelingToolkitBase/src/systems/codegen.jl
```diff
@@ -37,6 +37,84 @@ $GENERATE_X_KWARGS
...
+Returns `rhss` unchanged when every residual is scalar.
+"""
+function array_residual_maker(rhss::Vector{SymbolicT})
+    any(rhs -> SU.is_array_shape(SU.shape(rhs)), rhss) || return rhss
+
+    regions = Vector{Vector{UnitRange{Int}}}(undef, length(rhss))
+    values = similar(rhss)
+    offset = 0
+    for (i, rhs) in enumerate(rhss)
+        sh = SU.shape(rhs)
```
AayushSabharwal:
```suggestion
        sh = SU.shape(rhs)::SU.ShapeVecT
```
For now, since this code assumes it anyway. We'll have to lift this restriction to allow resizing the PDE without redoing codegen.

lib/ModelingToolkitBase/src/systems/codegen.jl
```diff
@@ -94,7 +177,18 @@ function generate_rhs(
     end
 
     if !isempty(assertions(sys)) && !isempty(rhss)
-        rhss[end] += unwrap(get_assertions_expr(sys))
+        assertion_expr = unwrap(get_assertions_expr(sys))
+        # An array-valued residual stands for several output rows, and `+` is not defined
```
AayushSabharwal, line 188: A future PR should probably make assertions codegen to `isoutofdomain`?

AayushSabharwal (changes requested):

lib/ModelingToolkitBase/src/systems/abstractsystem.jl
ChrisRackauckas-Claude: Done. Also dropped the `rhs` fallback I had, because the case it guarded cannot occur: `0 ~ arr` throws `ArgumentError: Cannot equate an array of different sizes. Got () and (3,).`, and `0 .~ arr` broadcasts into scalar equations. So an array equation with a scalar LHS is unconstructible and the `lhs`-only count is strictly correct.

```julia
function equation_row_count(eq::Equation)
    return prod(length, SU.shape(eq.lhs)::SU.ShapeVecT; init = 1)
end
```

lib/ModelingToolkitBase/src/systems/codegen.jl
ChrisRackauckas-Claude: Made it a plain function. I had copied the shape of `OperatorIsAtomic{O}`, but that one is a struct because it is parameterized on the operator type; this predicate has nothing to parameterize, so a function is simpler and just as concrete at the call site.

lib/ModelingToolkitBase/src/systems/codegen.jl
ChrisRackauckas-Claude: Done — one `IRStructureSearchBuffer` is built over the system IR and reused across all residuals.

lib/ModelingToolkitBase/src/systems/codegen.jl
ChrisRackauckas-Claude: Done, `size(arg)` — it now feeds the `array_literal` size argument directly, so the separate `sz` local is gone.

lib/ModelingToolkitBase/src/systems/codegen.jl
ChrisRackauckas-Claude: Done, built as `Symbolics.STerm(SU.array_literal, arrargs; type = symtype(arg), shape = sh)` with a `sizehint!`ed `Symbolics.SArgsT`, size pushed first and then each `op(arg[idx])`. No `reshape`.

I checked the result is identical to what the old `Const(reshape(...))` produced, including for rank 2 — `isequal(new, old)` is `true`, and `stable_eachindex` supplies the elements in the column-major order `array_literal` documents.

lib/ModelingToolkitBase/src/systems/codegen.jl
ChrisRackauckas-Claude: Done — `expand_array_derivatives!` now takes the `ir` and is called with `get_irstructure(sys)`, so both the search buffer and the `IRSubstituter` work in the system IR and the rewritten residuals are prepopulated.

lib/ModelingToolkitBase/src/systems/codegen.jl
ChrisRackauckas-Claude: Done, `map!(subber, rhss, rhss)` in place. `rhss` is freshly built in `generate_rhs` so mutating it is safe; the function is renamed to `expand_array_derivatives!` to reflect that.

lib/ModelingToolkitBase/src/systems/codegen.jl
ChrisRackauckas-Claude: Done. Agreed on lifting it later — a symbolic-size residual would need the region offsets in `array_residual_maker` to be symbolic too, which is the same restriction in a different place.

lib/ModelingToolkitBase/src/systems/codegen.jl
ChrisRackauckas-Claude: Agreed, that reads like the right target — `isoutofdomain` is what the poisoned-residual trick is emulating. Out of scope here; the only reason assertions are touched at all in this PR is that an array-valued last residual needs the assertion broadcast over its rows rather than added to a scalar. Happy to open an issue for the `isoutofdomain` change if you want it tracked.

ChrisRackauckas-Claude:
Second review pass pushed as 556524b. Replies are on each thread; all nine are addressed.

- `array_derivative_is_atomic` is a plain function. `OperatorIsAtomic{O}` is a struct because it parameterizes on the operator type; this predicate has nothing to parameterize.
- `expand_array_derivatives!` builds one `IRStructureSearchBuffer` over `get_irstructure(sys)`, substitutes in that same IR, and writes back with `map!`.
- The expansion is now `Symbolics.STerm(SU.array_literal, arrargs; type = symtype(arg), shape = sh)` over a `sizehint!`ed `SArgsT`, size pushed first then each `op(arg[idx])`. No `reshape`, no `size`/`ntuple` dance. I checked the constructed term is `isequal` to what the old `Const(reshape(...))` produced, rank 2 included.
- `equation_row_count` is `prod(length, SU.shape(eq.lhs)::SU.ShapeVecT; init = 1)`. I also dropped the `rhs` fallback I had written: `0 ~ arr` throws `Cannot equate an array of different sizes. Got () and (3,).` and `0 .~ arr` broadcasts to scalar equations, so an array equation with a scalar LHS cannot be constructed and the fallback was dead code.
- Shapes this code already assumes are `ShapeVecT` are now type-asserted.
- The `isoutofdomain` idea for assertions I agree with, but left out of scope here.

Verification on the pushed tree:

```
array equations reach DAEProblem                        |    6      6
other problem types still require scalarized equations  |    1      1
array-equation DAE solves to the analytic solution      |    2      2
array equations over a 2D slice keep their shape        |    2      2
array equations written as `D(slice) ~ rhs`             |    4      4

GROUP=InterfaceI: InterfaceI | 1595  5(broken)  1600  43m16.0s
     Testing ModelingToolkitBase tests passed
```

The `InterfaceI` run matters here specifically because the `::SU.ShapeVecT` assertions could throw where the previous code silently branched; the counts are identical to the run before this pass, so nothing in the suite hits an unresolved shape.

Generated code is byte-identical to the previous revision (10,109 / 41,428 / 364,361 chars at n = 20 / 200 / 2000), so this pass is a pure refactor on top of the O(1) residual body.

`runic --check` and `typos` are clean. I have not run beyond `InterfaceI`.

AayushSabharwal (approved):

AayushSabharwal:
Bunch of test failures. A cursory glance says they're not due to this PR and are QA-related changes. Can they be fixed nonetheless so this PR can be evaluated properly?

AayushSabharwal (approved):

PR #5001, opened 2026-08-20
feat: FMU event support in the FMI import

asinghvi17:
Adds event handling to `FMIComponent`:

- **ME state events**: one `VectorContinuousCallback` per FMU built from its event indicators (right root find, both crossing positions saved).
- **ME states stay unknowns**: every Model Exchange FMU's continuous states are marked `irreducible` at import, since state and step events write back into `u`; declaring one an input of the compiled system still removes it, which is reported at problem construction for FMUs with event indicators and at the first write-back otherwise.
- **ME step events**: the FMU is notified of every accepted step and may request event mode.
- **FMI3 CS event mode**: FMUs declaring `hasEventMode` are instantiated with `eventModeUsed = true` and events reported by `fmi3DoStep` are handled.
- **Event write-backs and DAEs**: both ME event callbacks name an `initializealg`, so an event does not fall through to the problem-level `OverrideInit`, which resets `u` to `u0` and discards the FMU's write-back. The default is `SciMLBase.CheckInit()`, as for MTK's imperative affects: never invoked on a plain ODE; on a coupling whose algebraic variables depend on the FMU's states it throws `CheckInitFailureError` at the first event, and `reinitializealg = BrownFullBasicInit()` on `FMIComponent` re-solves them instead (tested with BouncingBall against a cubic algebraic constraint, FMI 2 and 3).
- **Time events** error by default; `ignore_time_events = true` warns once per instance and continues. `terminateSimulation` terminates the integrator.

Mechanism: the ext attaches a `CallbackConstructionHook` (new, core) to the FMU's wrapper parameter; problem construction invokes it and merges the result with native symbolic events and user callbacks. Core changes are limited to that hook plumbing plus a small `compile_implicit_affect` fix (namespaced nonnumeric parameters in implicit affects), regression-tested without FMUs.

Tests compare against the Reference-FMUs' shipped output traces (six FMUs vendored under `test/fmi/fmus/`): `test/fmi/fmi.jl` 359 pass, `symbolic_events.jl` 413 pass / 2 pre-existing broken, `jumpsystem.jl` 4176 pass.

Known v1 limitations (documented in the tutorial): equation-based affects on a system containing an ME FMU fail when the event fires — use `ImperativeAffect`; time events and early return are unsupported.

https://claude.ai/code/session_01H6SximsfZ9jpSR6MocRrqi


docs/src/API/model_building.md
```diff
@@ -192,12 +192,14 @@ ModelingToolkit.IfLifting
 ## Event handling
 
 Time-dependent systems may have several events. These are used to trigger discontinuities
-in the model. They compile to standard callbacks from `DiffEqCallbacks.jl`.
+in the model. They compile to standard callbacks from `DiffEqCallbacks.jl`, as do any
+callbacks contributed by a `CallbackConstructionHook` attached to a parameter of the system.
```
asinghvi17:
```suggestion
in the model. They compile to standard callbacks from `DiffEqCallbacks.jl`.
```

ext/MTKFMIExt.jl
```diff
@@ -18,47 +22,135 @@ const D = D_nounits
 """
     $(TYPEDSIGNATURES)
 
-A utility macro for FMI.jl functions that return a status. Will terminate on
-fatal statuses. Must be used as `@statuscheck FMI.fmiXFunction(...)` where
-`X` should be `2` or `3`. Has an edge case for handling tuples for
-`FMI.fmi2CompletedIntegratorStep`.
+Whether `status`, returned by the FMI version 2 function `fnname`, is a failure.
+`fmi2StatusWarning` is not a failure: the standard allows the computation to continue, so it
+is only logged.
```
asinghvi17: We can define the function name with the docstring and then add the implementations.  Like:

```
"""
....
"""
function fmu_status_is_error end

# implementations....
```
to avoid having so many docstrings that repeat

ext/MTKFMIExt.jl
```diff
@@ -18,47 +22,135 @@ const D = D_nounits
...
+end
+
+"""
+    $(TYPEDSIGNATURES)
+
+A utility macro for FMI.jl functions that return a status. Frees the instance and errors on a
+failing status, terminating it first unless the status is fatal. Must be used as
+`@statuscheck FMI.fmiXFunction(...)` where `X` should be `2` or `3`. Evaluates to the value
+returned by the wrapped call, so callers can consume the extra outputs of functions such as
+`FMI.fmi2CompletedIntegratorStep`, which return their status as the first element of a tuple.
```
asinghvi17: ⚠️ WALL OF TEXT

ext/MTKFMIExt.jl
```diff
@@ -18,47 +22,135 @@ const D = D_nounits
...
         end
+        result
     end |> esc
 end
 
+"""
+The maximum number of `fmiXNewDiscreteStates`/`fmiXUpdateDiscreteStates` calls allowed in a
+single event iteration before it is considered non-convergent.
+"""
+const MAX_EVENT_ITERATIONS = 100
```
asinghvi17: Shouldn't this be configurable per fmu?

ext/MTKFMIExt.jl
```diff
@@ -18,47 +22,135 @@ const D = D_nounits
...
+    $(TYPEDSIGNATURES)
+
+Record the result of the event iteration performed while leaving initialization mode on
+`wrapper`, and error if the FMU requested termination or declared a time event.
+"""
+function handle_initial_event_iteration!(wrapper, event_result)
+    if event_result.terminate
+        # the instance is in Event Mode, where the state exchange a retried solve performs is
+        # illegal, so a caught error must not leave it reusable. Terminating from Event Mode is
+        # legal in both FMI versions.
```
asinghvi17: make this more human readable.,

ext/MTKFMIExt.jl
```diff
@@ -77,14 +169,297 @@ end
...
+
+# Fields
+
+$(TYPEDFIELDS)
+"""
+Base.@kwdef struct FMUEventMetadata
+    """
+    The FMI version of the FMU, `2` or `3`.
+    """
+    fmi_version::Int
```
asinghvi17:
```suggestion
    "The FMI version of the FMU, `2` or `3`."
    fmi_version::Int
```
Carry this pattern on in general (single " instead of triple """).

For this specific one.  Should it somehow be a type param?

ext/MTKFMIExt.jl
```diff
@@ -77,14 +169,297 @@ end
...
+    """
+    state_names::Vector{Symbol}
+    """
+    Component-relative names of the inputs of the FMU.
+    """
+    input_names::Vector{Symbol}
+    """
+    Value references of the continuous states of the FMU, ordered as `state_names`.
+    """
+    state_vrs::Vector
```
asinghvi17:
```suggestion
    state_variables::Vector
```

let's spell this one out

ext/MTKFMIExt.jl
```diff
@@ -77,14 +169,297 @@ end
...
+    $(TYPEDSIGNATURES)
+
+Whether the FMU declares support for CoSimulation event mode. FMI2 CoSimulation has no event
+mode, so this is always `false` for v2 FMUs.
+"""
+cs_event_mode_supported(::FMI.FMU2) = false
+
+function cs_event_mode_supported(fmu::FMI.FMU3)
+    coSimulation = fmu.modelDescription.coSimulation
+    return coSimulation !== nothing && coSimulation.hasEventMode === true
```
asinghvi17:
```suggestion
    return !isnothing(coSimulation) && coSimulation.hasEventMode === true
```

ext/MTKFMIExt.jl
```diff
@@ -77,14 +169,297 @@ end
...
+
+Build the callbacks handling the events of the Model Exchange FMU whose instance wrapper is
+the parameter `wrapper_param` of the compiled system `sys`. Attached to `wrapper_param` as
+`ModelingToolkit.CallbackConstructionHook` metadata, which calls this during problem
+construction.
+
+The state events of the FMU become a single `VectorContinuousCallback` over all of its event
+indicators, which an FMU declaring none does not get. Every Model Exchange FMU also gets a
+`DiscreteCallback` with an always-true condition, which notifies it of each accepted
+integrator step and handles the step events it may report there.
```
asinghvi17: ⚠️ WALL OF TEXT

ext/MTKFMIExt.jl
```diff
@@ -77,14 +169,297 @@ end
...
+    no_event_access = nothing
+    state_idxs = Int[]
+    for name in meta.state_names
+        resolved = resolve_relative(wrapper_param, name)
+        idx = SII.variable_index(sys, resolved)
+        if idx === nothing
+            no_event_access = "The continuous state $resolved of the FMU wrapped by \
+                $(getname(wrapper_param)) is not an unknown of the simplified system, so \
+                the value it takes after an FMU event cannot be written back. Prevent it \
+                from being simplified away, for example with `irreducible = true`."
```
asinghvi17: what does this mean?  we should always mark these as irreducible, we can't ever not do that for any fmu state variable?

ext/MTKFMIExt.jl
```diff
@@ -77,14 +169,297 @@ end
...
+        return nothing
+    end
+
+    # notify the FMU of the step and return whether it asks for Event Mode, which is the only
+    # outcome its caller has to treat as a discontinuity
+    function fmu_step_event_occurred!(integrator)
+        # a terminating event leaves the FMU in Event Mode, where none of the calls below are
+        # legal, and SciML applies the discrete callbacks after a continuous one terminated.
+        # `check_error!` stores `Success` in the retcode of a solve that is still in flight,
+        # so those are the two codes that mean "nothing has ended this solve".
```
asinghvi17: WALLS OF TEXT

ext/MTKFMIExt.jl
```diff
@@ -104,12 +479,16 @@ with the name `namespace__variable`.
   be set to `tspan[2]` when the `ODEProblem` is solved. An explicit value will overwrite this.
 - `type`: Either `:ME` or `:CS` depending on whether `fmu` is a Model Exchange or
   CoSimulation FMU respectively.
+- `ignore_time_events`: Time events are not supported, and by default an FMU declaring one
+  errors. If this is `true`, a declared time event is instead warned about once per FMU
+  instance and ignored, which changes the trajectory the FMU would otherwise have taken. A
+  solve creates one instance for a Model Exchange FMU and at least two for a CoSimulation one.
```
asinghvi17: WALL OF TEXT

ext/MTKFMIExt.jl
```diff
@@ -77,14 +169,297 @@ end
...
+    callbacks = SciMLBase.DECallback[]
+    if meta.n_event_indicators > 0
+        push!(
+            callbacks,
+            SciMLBase.VectorContinuousCallback(
+                fmu_event_condition!, fmu_event_affect!, meta.n_event_indicators;
+                # FMI defines an event as a domain change, so the FMU has to be past the
+                # switch when it enters Event Mode; localizing on the pre-crossing side would
+                # have it refresh its relations to the old domain and not handle the event
+                rootfind = SciMLBase.RightRootFind, interp_points = 10,
```
asinghvi17: This is not very human readable.  "We need RightRootFind because the FMU has to be past the switch when it enters event mode.  <link to fmi spec>"

lib/ModelingToolkitBase/src/systems/callbacks.jl
```diff
@@ -1595,15 +1599,41 @@ merge_cb(::Nothing, x) = merge_cb(x, nothing)
 merge_cb(x, ::Nothing) = x
 merge_cb(x, y) = CallbackSet(x, y)
 
+"""
+    struct CallbackConstructionHook end
+
+Symbolics metadata key attaching a callback-building hook to a parameter, via
+`setmetadata(param, CallbackConstructionHook, hook)`. The hook is called as
+`hook(sys, param) -> Vector{<:SciMLBase.DECallback}` during `process_events`, once per
+parameter carrying the key, and the returned callbacks are appended to the problem's callbacks.
```
asinghvi17: Wall of text!

test/fmi/fmi.jl
```diff
@@ -383,3 +383,824 @@ end
...
+            record -> occursin("time event", string(record.message)), logger.logs
+        ) == 1
+        # `Stair` steps its counter at every integer time, and the v3 event iteration reports
+        # the first of those through the `event_time_defined`/`event_time` pair
+        @test wrapper.next_event_time == 1.0
+        ext.reset_instance!(wrapper)
+    end
+end
+
+@testset "FMU step events" begin
```
asinghvi17, line 921: I'm a bit concerned this is too specific

docs/src/API/model_building.md
asinghvi17: Reverted. (4465167750)

ext/MTKFMIExt.jl
asinghvi17: Done. (52730146dc)

ext/MTKFMIExt.jl
asinghvi17: Condensed. (52730146dc)

ext/MTKFMIExt.jl
asinghvi17: Made it a kwarg: `max_event_iterations = 100`, stored per instance wrapper so each FMU carries its own. (5110e423b3)

ext/MTKFMIExt.jl
asinghvi17: Reworded. (52730146dc)

ext/MTKFMIExt.jl
asinghvi17: Single-quote style applied to the PR-added field docstrings (52730146dc). Left `fmi_version` a plain field: nothing dispatches on it — it's declarative metadata for the future clocks pass — and a parametric struct complicates its use as a Symbolics metadata key for no runtime gain.

ext/MTKFMIExt.jl
asinghvi17: Spelled out as `state_value_references` rather than the suggested `state_variables` — the field holds FMI value references (UInt32 handles), and this matches the wrapper structs' existing fields. (52730146dc)

ext/MTKFMIExt.jl
asinghvi17: Applied. (52730146dc)

ext/MTKFMIExt.jl
asinghvi17: Condensed. (52730146dc)

ext/MTKFMIExt.jl
asinghvi17: You're right for FMUs that can fire state events: the import now marks their states irreducible itself (ME + indicators > 0), so simplification keeps them and this error no longer asks the user to do it. Indicator-free FMUs keep full simplification (nothing ever writes back to them), and CS states never enter `u`. The error stays reachable one way — declaring an FMU state an `input` of the compiled system — and the message now says exactly that. (5110e423b3)

ext/MTKFMIExt.jl
asinghvi17: Condensed to one-line whys. (52730146dc)

ext/MTKFMIExt.jl
asinghvi17: Condensed. (52730146dc)

ext/MTKFMIExt.jl
asinghvi17: Done, with the verified spec anchor: https://fmi-standard.org/docs/3.0.2/#state-event (52730146dc)

lib/ModelingToolkitBase/src/systems/callbacks.jl
asinghvi17: Condensed; after the API-docs revert this docstring is the hook's only documentation, so the call contract stayed. (4465167750)

test/fmi/fmi.jl
asinghvi17: Agreed on the internals-poking: the functor source-grep test is gone (the IO-model trajectory tests pin that invariant behaviorally) and the step-callback tests now assert solve-level behavior — duplicated save times are exactly the bounces, and the rest of the system is untouched. The two paths with no behavioral observable (no instance yet; terminated FMU in Event Mode) are driven directly in a testset that states why. (6f69db983b)

ext/MTKFMIExt.jl
```diff
@@ -77,14 +159,282 @@ end
...
+        for (i, idx) in enumerate(state_idxs)
+            state_buffer[i] = u[idx]
+        end
+        if get_inputs !== nothing
+            copyto!(input_buffer, get_inputs(u, p, t))
+        end
+        return force_set_fmu_state!(wrapper, state_buffer, input_buffer, t)
+    end
+
+    function fmu_event_condition!(out, u, t, integrator)
```
AayushSabharwal: I'm not convinced this can't be done symbolically without all the infrastructure. The event condition is a registered function of the form `f(wrapper, __mtkunknowns__, MTKPARAMETERS_ARG, t)` which returns a vector of booleans. The affect is an `ImperativeAffect` which takes `wrapper` and the event index in the observed list. So this then just builds a list of `f(wrapper, ...)[i] => ImperativeAffect(g, (; wrapper, idx = i), ...)` callbacks, all of which have `BrownFullBasic` as the reinitializealg. MTK will put them all in the same VCC during codegen, and calls to `f` will be CSEd.

ext/MTKFMIExt.jl
```diff
@@ -16,49 +20,127 @@ const t = t_nounits
...
+like `FMI.fmi2CompletedIntegratorStep`. Those return their status as the first element of a
+tuple, which is where this looks for it.
 """
 macro statuscheck(expr)
     @assert Meta.isexpr(expr, :call)
     fn = expr.args[1]
     @assert Meta.isexpr(fn, :.)
     @assert fn.args[1] == :FMI
-    fnname = fn.args[2]
+    # the qualified name is parsed as `Expr(:., :FMI, QuoteNode(:fmiXFunction))`
```
AayushSabharwal, line 64: If this is a bug fix can it be a separate PR? Just because this one is big by itself and it's easier to understand which changes are relevant to the events if it's isolated.

AayushSabharwal (changes requested):
In general, accessing `getmetadata` should typeassert the result if possible and closures should explicitly capture with a `let` block. Even better if they're callable structs though I don't always bother either.

ext/MTKFMIExt.jl
asinghvi17: robot says callbacks can't:
- notify when a step has taken place (you can potentially hack discrete callbacks to do this, but it's a hack)
- FMI wants a single ccall telling it the set of variables that have changed, so a vector of `ImperativeAffect` does not work
-  symbolic callbacks give you no save_positions control and apparently can lead to duplicate timestamps

ext/MTKFMIExt.jl
asinghvi17: Done, #5007

ChrisRackauckas:
> DAE-safe: both event callbacks carry initializealg = BrownFullBasicInit() so FMU event write-backs survive reinitialization on singular-mass-matrix systems.

That's not necessarily correct.

asinghvi17:
I had the bot rephrase the PR description, it will also make reinitializealg a kwarg you can provide when creating the FMIComponent.  But it looks like we do not have a reinitializealg that would work well for all FMU cases.

This is also ignoring DAE initialization 😅 which we'll have to figure out how to import DAE spec FMUs too at some stage

PR #5007, opened 2026-08-21
Fix FMI version detection and tuple status handling in `@statuscheck`

asinghvi17:
Split out of #5001 per review.

`@statuscheck` had two latent bugs:

- `startswith("fmi2", string(fnname))` has its arguments reversed, so `is_v2` was always `false` and an FMI2 error path called `fmi3Terminate`/`fmi3FreeInstance!` on an FMI2 instance.
- The tuple branch compared the returned status against `fmi3True` (numerically `fmi2StatusWarning`), so `fmi2CompletedIntegratorStep` tore down the instance on warnings and continued past real errors.

The rewrite takes the status from the first tuple element, logs warning statuses without failing, and makes the macro evaluate to the wrapped call's result so callers keep extra outputs like `enterEventMode` (which #5001 needs). The FMI3 event-mode guard is unchanged — it lives in explicit `@assert`s at the call site.

https://claude.ai/code/session_01H6SximsfZ9jpSR6MocRrqi

### DIFF
--- a/ext/MTKFMIExt.jl
+++ b/ext/MTKFMIExt.jl
@@ -16,46 +16,71 @@ const t = t_nounits
 const D = D_nounits
 
 """
-    $(TYPEDSIGNATURES)
+    fmu_status_is_error(::Val{fmi_version}, status, fnname)
 
-A utility macro for FMI.jl functions that return a status. Will terminate on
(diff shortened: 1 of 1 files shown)


AayushSabharwal (approved):

dalerxli:
您好！您的邮件已收到，谢谢。

PR #5008, opened 2026-08-21
Support Symbolic Struct handling in mtkcompile and solve calls

Shreyas-Ekanathan:
## Checklist

## Additional context

depends on https://github.com/JuliaComputing/StateSelection.jl/pull/153 and https://github.com/JuliaSymbolics/Symbolics.jl/pull/1959


lib/ModelingToolkitBase/src/atomic_array_dict.jl
```diff
@@ -195,7 +195,7 @@ end
 Add `item` to `x`. If `item` is an indexed array, add the array instead.
 """
 function push_as_atomic_array!(x::AtomicArraySet, item::SymbolicT)
-    return push!(x, split_indexed_var(item)[1])
+    return push!(x, split_field_var(split_indexed_var(item)[1]))
```
AayushSabharwal: Doesn't this only handle cases of the form `foo.x.y[idx]`? It won't handle `foo.x[idx].y`. I _think_ the recourse is to add helpers that extend `split_indexed_var` to structs. So `split_field_access` would return `(foo, true)` for the above cases, and something like `get_struct_access` (analogous to `get_stable_index`) would return `Union{Symbol, StableIndex{Int}}[:x, :y, StableIndex{Int}([idx])]` for the first case, and a permutation of it for the second.

lib/ModelingToolkitBase/src/systems/systems.jl
```diff
@@ -158,6 +158,11 @@ end
 function scalarized_vars(vars)
     scal = SymbolicT[]
     for var in vars
+        # A struct variable looks scalar by shape, so expand it into its leaves here.
+        if Symbolics.issymstruct(var)
```
AayushSabharwal, line 177: I forgot, does `issymstruct` infer as `Bool`?

lib/ModelingToolkitBase/src/systems/systems.jl
```diff
@@ -158,6 +158,11 @@ end
...
+            append!(scal, collect(Symbolics.SymStruct{SU.symtype(var)}(var)))
```
AayushSabharwal: The `collect` likely won't infer since the type parameter of `SymStruct` won't infer. So it's a good idea to typeassert it as `::Vector{SymbolicT}`.

lib/ModelingToolkitBase/src/systems/problem_utils.jl
```diff
@@ -1736,10 +1736,54 @@ function operating_point_preprocess(sys::AbstractSystem, op; name = "operating_p
...
+        delete!(op, k)
+    end
+    return op
+end
+
 function build_operating_point(sys::AbstractSystem, op; fast_path = false)
     if !fast_path
         op = operating_point_preprocess(sys, op)
     end
+    split_record_entries!(op)
```
AayushSabharwal: I'm really not sold on this approach. Ideally struct variables should be considered atomic the same way arrays are. `AtomicArrayDict` should validate that keys are not field accesses. Otherwise, people can specify struct initial conditions in multiple ways: `foo => MyStruct(..)` or `[foo.x => 1, foo.y => [2, 3]` and so on. An even worse case is `foo.x[1].y.z => 3`. Maintaining atomicity simplifies it a bit. I think `fixpoint_sub` with constant-folding should work if `foo` is substituted for a concrete value.

AayushSabharwal (changes requested):

lib/ModelingToolkitBase/src/systems/systems.jl
Shreyas-Ekanathan: yeah it does I believe

lib/ModelingToolkitBase/src/utils.jl
```diff
@@ -1709,6 +1692,75 @@ function _get_stable_index(x::SymbolicT)
...
+
+Given a symbolic variable `x`, check whether it projects a leaf out of a struct symbolic,
+through any interleaving of field accesses and indices (`foo.x.y[i]`, `foo.x[i].y`). If it
+does, return the struct being projected from and `true`. Otherwise, return `x, false`.
+
+This is the struct counterpart of `split_indexed_var`, and like it, commutes with
+operators. Unlike `split_indexed_var` it peels the whole chain, so a plain array element
+`a[i]` also resolves to `a`.
+"""
+SU.@cache limit = 500_000 function split_field_access(x::SymbolicT)::Tuple{SymbolicT, Bool}
```
AayushSabharwal: How much does the `@cache` help in terms of performance? Does it still infer?

Shreyas-Ekanathan: yeah inference is fine, @cache appears marginal so I'll drop it

lib/ModelingToolkitBase/src/atomic_array_dict.jl
```diff
@@ -155,12 +245,81 @@ Equivalent to `get(dd, k, default)`. If `k` is an indexed array, then return
...
+    return record_leaf_entry(k, anc, dd[anc])
+end
+
+"""
+    $TYPEDSIGNATURES
+
+The nodes a record key is lowered to: the deepest projections of `root` whose symtype
+flattens into a numeric buffer. This is a leaf, except where a field is a numeric array,
+which stays whole because ordinary array atomicity already covers it. It is the same
+granularity `Initial` uses, which is what the initialization system pairs `op` keys with.
```
AayushSabharwal, line 265: `Initial` and `AtomicArrayDict` should agree about atomicity, otherwise e.g. `root` is atomic but `Initial(root)` isn't. Downstream should just work if `SymbolicGetproperty` constant-folds properly? Does it not?

lib/ModelingToolkitBase/src/utils.jl
```diff
@@ -1805,7 +1805,24 @@ The leaves of the struct symbolic `root`, in canonical linear order. This is the
...
-    return collect(Symbolics.SymStruct{SU.symtype(root)}(root))::Vector{SymbolicT}
+    return _record_leaves(root)
+end
+
+function _record_leaves(root::SymbolicT)
+    T = SU.symtype(root)
+    # An array of records is not itself a record, so `SymStruct` cannot iterate it.
+    # Expand over the indices and concatenate the leaves of each element.
+    if T <: AbstractArray && Symbolics.is_symstruct_type(eltype(T))
+        SU.is_array_shape(SU.shape(root)) || throw(ArgumentError(LazyString(
```
AayushSabharwal: `is_array_shape` also allows unknown shapes. I think you want `isa SU.ShapeVecT`

lib/ModelingToolkitBase/src/atomic_array_dict.jl
AayushSabharwal: This isn't addressed yet.

Shreyas-Ekanathan: 1. They do agree about atomicity, lowering drops to the individual fields to apply init conditions and build equations. 
2. We don't fold, and there are many weird cases here (as it stands, initial(P).x is different from initial(P.x), we design so that we apply conditions to the individual fields, rather than providing one full initial struct).

lib/ModelingToolkitBase/src/atomic_array_dict.jl
```diff
@@ -105,6 +160,22 @@ function as_atomic_dict_with_defaults(dict::AbstractDict{SymbolicT, SymbolicT},
...
+        for i in idxs
+            buffer[i] = record_leaf_entry(leaves[i], node, v)
+        end
+    end
+    for (k, v) in record_vals
+        __unsafe_aad_setindex!(dd, BSImpl.Const{VartypeT}(v), k)
+    end
+    # Records are the right granularity for resolving what was asked for, but nothing
+    # downstream expects a record key, so lower before returning.
+    lower_record_entries!(dd, default)
```
AayushSabharwal, line 178: As in https://github.com/SciML/ModelingToolkit.jl/pull/5008#discussion_r3861267952, I think structs should be considered atomic here. What breaks if they are not?

AayushSabharwal (changes requested):

lib/ModelingToolkitBase/src/atomic_array_dict.jl
Shreyas-Ekanathan: So from a user perspective, at the moment, ICs can be passed both as a full struct or as individual components. I think this is reasonable, because if the whole struct is known, then they give that directly, otherwise if they have some guesses they can mix that in individually as well. On the output side, we currently only give per-component solution states. I think this second half motivates the lack of atomicity and the decomposition here. We can reassemble the full state after as a later feature, but since we integrate per component, I believe lowering is reasonable here. The distinction between this and arrays is that I can always integrate a numeric array directly, but a struct has to be flattened into something array-like before I can integrate it.

AayushSabharwal: The reason atomicity exists is to avoid edge cases and unintentional abuse of API. If you have a struct variable and don't make it atomic, people can add all sorts of weird initial conditions/bindings. They might bind the entire struct and then add an initial condition for a field, which is wrong and an edge case we need to catch here that I don't think we do. There can be many such cases that need to be handled in many different places, and we cannot rely on future development work to always account for this.

Shreyas-Ekanathan: I've applied a fix, it utilizes https://github.com/JuliaSymbolics/Symbolics.jl/pull/1997 to construct the literal object. If you could review when you have a chance, that would be amazing!

Issue #5018, opened 2026-08-25
Reverse-mode adjoints fail for any system with a continuous event


asinghvi17:
Reverse-mode adjoints fail for any `System` with a continuous event; without the event the same loss differentiates and matches finite differences. Both failure sites are in SciMLSensitivity, and both assume `prob.kwargs[:callback]` is a `CallbackSet` of its own `TrackedAffect`s, while MTK attaches a bare `ContinuousCallback` whose `affect!` is an MTK affect (`ExplicitAffect`, `ImperativeAffect` or `FunctionalAffect`):

- Zygote path (`InterpolatingAdjoint(autojacvec = ReverseDiffVJP())`, `GaussAdjoint`): `adjoint_common.jl:614` builds `ReverseDiff.GradientTape((y, _p, [_t]))`, which needs `similar(_p)` on the raw `MTKParameters` recorded at `callback_tracking.jl:584`.
- `adjoint_sensitivities` path: `reset_p` (`interpolating_adjoint.jl:750`, from `split_states` at `:255`) reads `.discrete_callbacks` and then `affect!.event_times`, which MTK affects do not have.

## MWE

```julia
using ModelingToolkit, OrdinaryDiffEq, SciMLSensitivity, Zygote, SciMLBase
using ModelingToolkit: t_nounits as t, D_nounits as D

@variables h(t) v(t)
eqs = [D(h) ~ v, D(v) ~ -9.81]
sa, saveat = InterpolatingAdjoint(autojacvec = ReverseDiffVJP()), 0.0:0.25:2.0
@mtkcompile noev = System(eqs, t)
@mtkcompile ball = System(eqs, t; continuous_events = [[h ~ 0] => [v ~ -0.8v]])
mkprob(sys) = ODEProblem{true, SciMLBase.FullSpecialize}(
    sys, [sys.h => 1.0, sys.v => 0.0], (0.0, 2.0))
function grad(sys)                  # unknowns(sys) == [v(t), h(t)], so h is index 2
    prob = mkprob(sys)
    Zygote.gradient(u0 -> sum(abs2,
        Array(solve(remake(prob; u0), Tsit5(); saveat, sensealg = sa))[2, :]), prob.u0)[1]
end

grad(noev)   # [-180.6525, -107.0775], agrees with finite differences
grad(ball)   # MethodError: no method matching similar(::MTKParameters{...})

# adjoint_sensitivities on the same problem instead dies inside reset_p
prob = mkprob(ball)
[...]

PR #5042, opened 2026-08-29
Compile single-equation continuous event conditions to scalar functions

pankgeorg:
Part of #5018 (companion: SciML/SciMLSensitivity.jl#1629).

Single-equation continuous event conditions now compile to a scalar out-of-place function; `CompiledCondition{IsDiscrete, IsScalar}` only needs `parameter_values(integrator)`, no more `get_sol`/`get_tmp_cache`, so the condition can be evaluated (and differentiated) through a stand-in integrator. Vector and discrete paths unchanged.

Tested: new "Scalar continuous conditions" testset in `symbolic_events.jl`; ModelingToolkitBase `InterfaceI` group 1620 pass / 5 broken (pre-existing) / 0 fail.

### DIFF
--- a/lib/ModelingToolkitBase/src/systems/callbacks.jl
+++ b/lib/ModelingToolkitBase/src/systems/callbacks.jl
@@ -801,29 +801,40 @@ Base.isempty(cb::AbstractCallback) = isempty(cb.conditions)
 ####### Compilation functions ######
 ####################################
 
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
This doesn't actually solve the issue, which is reverse mode AD of VCCs.

PR #5044, opened 2026-08-29
Add `homotopy` keyword to `mtkcompile` to opt out of homotopy lowering

pankgeorg:
`mtkcompile(sys; homotopy = false)` replaces every Modelica `homotopy(actual, simplified)` node by `actual` before compilation: generated code contains only `actual`, problem construction never selects a `HomotopyProblem`, and the initialization / event affect systems derived from the compiled system inherit the setting (recorded as `HomotopyCtx` system metadata).

Motivation: targets that cannot lower to a continuation solver, e.g. SciML/DiffEqGPU.jl#516, currently have to rewrite the equations by hand before `mtkcompile`.

Tested: new `lib/ModelingToolkitBase/test/homotopy_disabled.jl` (generated `Expr` has no `homotopy`, plain `NonlinearProblem`, init inheritance) and a new case in `test/homotopy_initialization_scc.jl`; all existing homotopy test files pass locally.

### DIFF
--- a/docs/src/basics/Homotopy.md
+++ b/docs/src/basics/Homotopy.md
@@ -59,6 +59,26 @@ keep symbolic jacobians, `tgrad`, and index reduction consistent. At runtime,
 differentiated equations reproduce `actual`'s derivative; along the continuation
 they follow the derivative of the blended expression above.
 
+## Opting Out of the Lowering
+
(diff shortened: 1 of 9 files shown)


ChrisRackauckas (approved):

ChrisRackauckas:
Supersceded and merged in https://github.com/SciML/ModelingToolkit.jl/pull/5048

Issue #5043, opened 2026-08-29
FullSpecialize: generate device-compatible initialization maps (initializeprobmap/pmap) as RGFs


ChrisRackauckas:
## Request

Under `FullSpecialize`, generate the initialization state/parameter maps (`initializeprobmap`, `initializeprobpmap`) as single device-compatible generated functions — one `RuntimeGeneratedFunction` (with `drop_expr`) per map, mapping `(u, p)` of the initialization solution (plus the ODE problem's `p` for the parameter map) to statically-shaped outputs — instead of the current host-side combinator trees. The default (`AutoSpecialize`) path should stay exactly as it is, since fully specializing these maps adds compile latency that is unwanted in the common CPU remake case. DiffEqGPU would then require `FullSpecialize` for ModelingToolkit initialization inside `EnsembleGPUKernel` kernels, which is a reasonable ask for that use case.

## Why the current maps cannot run in a GPU kernel

The generated `f` for a system is already kernel-ready: a scalarized RGF built with `drop_expr` wrapped in `GeneratedFunctionWrapper`, taking and returning isbits values. The initialization maps are a different kind of object. For a small torn initialization system (`initialization_eqs = [x^3 + x ~ 2, y ~ 2x + 1]`, so `y` is an observed variable of the init system) the maps come out as:

```
umap: InitializationMap{true, typeof(identity),
        PromoteToTunableEltype{CopyParamsByTemplate{true,
          Tuple{ObservedWrapper{false, Any}, UnitRange{Int64}}, 1, Nothing}, Float64}}
pmap: initprobpmap_split{MTKParametersReconstructor{Returns{SVector{0,Float64}},
        ComposedFunction{PConstructorApplicator{typeof(identity)}, CopyParamsByTemplate{...}}, ...}}
```

Three independent blockers, all in `lib/ModelingToolkitBase/src/systems/problem_utils.jl`:

[...]

Issue #27, opened 2018-03-03
Add functions of parameters


finmod:
A useful extension of SciCompDSL would be to specify functions of parameters to allow coding of within and across equation restrictions. For instance, the first order ODE with speed of adjustment a

Dy = abX + ac - ay written as Dy = F1X + F2 -ay with F1 = ab and F2 = ac ; a, b, c are parameters

ChrisRackauckas:
This is already done with function registration and intermediate variables.

PR #5045, opened 2026-08-30
Generate FullSpecialize initialization maps with static buffers

ChrisRackauckas-Claude:
## Scope

Generate the state and parameter initialization maps as individual runtime-generated functions only for explicitly requested `FullSpecialize`. Preserve the existing default host-map path. Static buffers reduce warm map allocations; this is not an in-place preallocation API.

The implementation shares one buffer constructor and one reverse-mode rule. It preserves mutable parameter buffers, copies cache buffers instead of aliasing them, handles `split=false`, and retains array-valued callable parameters. Immutable static constructors produce isbits results in the tested case; mutable non-isbits buffers use `SizedVector` storage.

Addresses https://github.com/SciML/ModelingToolkit.jl/issues/5043. Related GPU workaround: https://github.com/SciML/DiffEqGPU.jl/pull/516.

## Review round (@AayushSabharwal, 2026-09-09)

Seven inline comments, all addressed in https://github.com/SciML/ModelingToolkit.jl/commit/fed8cdb77b, and each answered on its own thread.

**1. `gensym` — this was a real defect, not style.** The five `gensym`ed locals in the new codegen are now fixed sentinels (`INITMAP_VALUES`, `INITMAP_ELTYPE`, `INITMAP_SOLUTION`, `INITMAP_PROBLEM`, `INITMAP_OUTER_PARAMETERS`), following the rationale already recorded for `HOMOTOPY_LAMBDA` in `homotopy_operator.jl`: a `gensym` embeds a process-global counter, so the same system lowers to a different `Expr` every time and the `RuntimeGeneratedFunctions` Expr-hash cache never hits.

This is the cause of the generated-map type churn the earlier revision of this body reported as an open problem.

Failing on the unfixed branch (`git stash` of the fix only, same probe, same session):

```text
generated map type reuse across identical systems: Test Failed
  Expression: typeof(d1.initializeprobpmap) === typeof(d2.initializeprobpmap)
   Evaluated: RuntimeGeneratedFunction{(Symbol("##problem#280"), Symbol("##initialization_solution#281")), …, (0x8298bbd3, 0xbcbc4d5b, 0x7a096718, 0x43533499, 0x57e4022b), Nothing} ===
              RuntimeGeneratedFunction{(Symbol("##problem#287"), Symbol("##initialization_solution#288")), …, (0x44632bf6, 0x42fe7ec6, 0x4d5618c1, 0xb6d83a49, 0x2afa037a), Nothing}
Test Summary:                                     | Pass  Fail  Total   Time
generated map type reuse across identical systems |    2     2      4  10.6s
```

Passing with the fix:

```text
Test Summary:                                     | Pass  Total  Time
generated map type reuse across identical systems |    4      4  7.0s
```

Checked in as three assertions at the end of the `FullSpecialize initialization maps are generated functions` testset, which is why that group's count moved from 698 to 701.

**2-6. Splats in `Expr` construction.** `_generated_map_expr`, `_parameter_buffer_expr`, the `discrete` block sizes, the portion tuple, and the `MTKParameters` call all build their `Expr`s by `push!`/`append!` onto `.args` now. `_static_initialization_buffer`'s `T[values...]` became `collect(T, values)`. No splats remain in the added code other than `kwargs...` forwarding.

**7. `groups` as a tuple of tuples.** Now a `Vector{Vector{Vector{SymbolicT}}}` built with type asserts against `ReorderedParametersT`'s `Union` element type, so both the `flat_syms` loop and the buffer loop are concretely typed instead of specializing on each system's shape. Nothing indexed `groups` heterogeneously — it is only iterated alongside the field names — so the tuple bought nothing.

Rebased onto `35818b8328`. The CasADi compat bump and the `Optimization` runner pin that used to show in this diff have landed on master separately and are gone from it.

## Second review round (@AayushSabharwal, 2026-09-11 — approved)

One inline comment: pass an options struct instead of `kwargs...`. Done in
https://github.com/SciML/ModelingToolkit.jl/commit/16ef998c28 — both map builders take a
`GeneratedFunctionOptions` positionally and call the struct method of
`build_explicit_observed_function`, with `_fullspecialize_map_options` deriving it from
the problem's own `opts.fn_opts.codegen`.

Two behaviour fixes fall out of it. The old `kwargs...` path went through the keyword
compat shim, which built a **fresh** `CodegenFunctionOptions` from whatever was left in
the bag — so the maps were not inheriting the problem's `checkbounds`/`cse`. And
`eval_expression` is now honoured; previously the builders never received it, so the maps
were unconditionally `RuntimeGeneratedFunction`s even with `eval_expression = true`,
unlike the default map path. Both are called out on the thread in case the second is not
wanted.

## Local verification

Julia 1.12.7, on `16ef998c28`. `JULIA_NUM_PRECOMPILE_TASKS=1`, `--heap-size-hint=8G`.

```sh
julia +1.12 --project=<runicenv> -m Runic --check <changed Julia files>   # exit 0
git diff upstream/master --check                                          # exit 0
git diff upstream/master --unified=0 | sed -n '/^+++ /d; s/^+//p' | typos -  # exit 0
```

```text
GROUP=Initialization
  Guess Propagation         |  11 pass |  11 total | 3m13.3s
  InitializationSystem Test | 706 pass | 12 broken | 718 total | 20m43.6s
  Initial Values Test       |  65 pass |  65 total | 6m00.6s     (0 errors)

GROUP=Extensions
  HomotopyContinuation Extension Test |  76 pass |  76 total | 10m07.9s
  LabelledArrays Test                 |  18 pass |  18 total | 31.1s
  BifurcationKit Extension Test       |   7 pass |   7 total | 1m25.4s
  Despecialized MTKParameters AD Test |   2 pass |   2 total | 7m07.9s
  Initialization maps AD              |   7 pass |   7 total | 17.2s

GROUP=FMI
  FMI | 60 pass | 60 total | 15m35.4s
```

All three groups exited 0. The 12 broken tests in `InitializationSystem Test` are
pre-existing `@test_broken`s on master; none were added, removed, skipped or loosened.

The MTKStdlib reproducer below also runs clean against this commit (`solve` → `remake`
→ `solve`), outside CI.

## Allocation measurement and limits

A warmed, two-state initialization example without user parameters measured allocations for both map calls together, taking the minimum of ten calls behind a function barrier:

| FullSpecialize map configuration | Bytes/call pair |
| --- | ---: |
| Existing host maps | 784 |
| Generated maps, mutable buffers | 80 |
| Generated maps, immutable static constructors | 0 |

This is a map microbenchmark, not a solver benchmark or a general zero-allocation guarantee.

**On TTFX.** The generated map types are now reused across identical systems, which is what the earlier revision of this body reported as an unresolved code-generation/cache concern. That measurement was taken before the `gensym` fix and no longer describes the branch. I have **not** re-run the fresh-process TTFX comparison since the fix, so this PR still makes no TTFX claim in either direction — only that the specific type-churn defect behind it is gone and now has a regression guarding it. The default path remains unchanged.

## What was not verified

- **GPU execution.** This machine has no GPU. The static/isbits CPU assertions are not a substitute for a device execution test.
- **Fresh-process TTFX**, as above.
- **`GROUP=QA` is red, from a pre-existing master failure.** Rerun on the rebased branch: targeted `JET Tests` pass **54/54**, `Aqua Tests` are **20 passed / 1 failed / 21**, and the single failure is the package-wide JET typo-mode scan reporting **234 possible errors**. Every one of the 234 is the known `##And#` (182) / `##Call#` (52) Moshi `@match` codegen category tracked in https://github.com/SciML/ModelingToolkit.jl/issues/5063, which is still open after https://github.com/SciML/ModelingToolkit.jl/pull/5065 cleared the non-Moshi ones. None of the 234 diagnostics point anywhere inside the code this PR adds (`problem_utils.jl:1636-1810`). No QA suppression or dependency override is included here.
- **The `StructuralIdentifiability.jl/Core/1/` downstream job fails, and it is not this PR.** It passed on the pre-rebase head `438703fbe4` and fails on clean master `35818b8328` — the exact base this branch was rebased onto. It regressed on master between `f21201f889` (pass) and `9b458b059d` (fail); a separate bisect is running and will get its own issue. Nothing in this diff touches it.
- **Not rerun since the rebase:** LTS/prerelease Julia, FMI, Optimization, SymbolicIndexingInterface, the neural-network downstream suite, and the full downstream matrix. Their pre-rebase results are listed above.
- **Docs build** — no docs, docstrings, or public API are changed.

An initial source-loading test run exposed an independently reproduced LinearSolve world-age failure, reported separately at https://github.com/SciML/LinearSolve.jl/issues/1282; no LinearSolve workaround is included here.

## Anything a reviewer should push back on

- The sentinel names (`__mtk_initialization_*`) are module-level `const`s in `ModelingToolkitBase`. They are internal and unexported, but they are five new names in the module namespace where a single `NamedTuple` or a naming function would also work.
- `_static_initialization_buffer`'s `collect(T, values)` change was not one of the review comments; it is a splat in the same function and I removed it for consistency. It is on the runtime path (the non-isbits `SizedVector` branch), covered by the `Any[Ref(1)]` assertion in the Initialization group.
- The `groups` rewrite carries the same behaviour as before for an empty `parameters(sys; initial_parameters = true)`: `grouped[1]` would throw, because `reorder_parameters` returns early on empty `ps`. I did not add a guard, since that would be a behaviour change beyond the review, but it is reachable in principle.
- @ChrisRackauckas asked for the `gensym` rule to be written down. https://github.com/SciML/ModelingToolkit.jl/pull/5105 adds a root `AGENTS.md` (the repo had none) with that rule plus the splat and container-typing ones. Kept separate so this diff stays behaviour-only.

## Buffer types: only isbits buffers go static

`FullSpecialize` was turning *every* parameter buffer into a `StaticArray`, heap-element
ones included. That is what broke `ModelingToolkitStandardLibrary.jl/Core/1/`, and the
`SizedVector` branch added earlier for FMI was aimed at the wrong buffer. Fixed in
https://github.com/SciML/ModelingToolkit.jl/commit/b8c603d04a.

The FMI `setindex!` failure is not about size — `MArray`'s `setindex!` bounds-checks first,
then hard-`error()`s on any non-isbits eltype, because it mutates an inline `NTuple`
through a raw pointer (StaticArrays #27). The buffer was an **array-valued discrete**,
`MVector{1, Vector{Float64}}`, written by an FMI `imperative_affect`:

```text
[3] setindex!  @ BlockArrays/blockedarray.jl:249 [inlined]
[4] set_parameter!(p::MTKParameters{…, Tuple{BlockedVector{Vector{Float64},
      MVector{1, Vector{Float64}}, …}}, …},
      val::Vector{Float64}, pidx::ParameterIndex{Discrete, Tuple{Int64, Int64}})
```

Making such a buffer static buys nothing: the elements are still pointers, so an
`MVector{1, Vector{Float64}}` is no more GPU-resident than a `Vector{Vector{Float64}}`.
Gating on `isbitstype` therefore costs nothing that matters, and every buffer that could
be GPU-resident stays an `MVector`. Same FMI model, only `spec` changed:

| buffer | `AutoSpecialize` | `FullSpecialize` after this commit |
| --- | --- | --- |
| tunable | `Vector{Float64}` | **`MVector{7, Float64}`** |
| initials | `Vector{Float64}` | **`MVector{8, Float64}`** |
| discrete block sizes | `Vector{Int64}` | **`SVector{1, Int64}`** |
| discrete storage | `Vector{Vector{Float64}}` | `Vector{Vector{Float64}}` |
| nonnumeric | `Vector{FMI2CSFunctor}` | `Vector{FMI2CSFunctor}` |

The gate has to go in **two** places or the generated map and `MTKParametersReconstructor`
disagree on the buffer type: `_static_initialization_buffer`, and the `get_p_constructor`
closure the reconstructor routes through. `get_p_constructor` returns `identity` unless
`pType <: StaticArray`, and this PR is what makes `pType` static — so it only started
applying `create_array` to non-isbits buffers because of this PR.

Each variant measured separately, one change at a time:

| non-isbits buffer | MTKStdlib `remake` | FMI |
| --- | --- | --- |
| `MVector` | passes | `setindex!` error |
| `SizedVector` (previous revision) | `StackOverflowError` | 60/60 |
| `Vector`, gate in one place | typeassert, `Vector` vs `MVector` | 60/60 |
| **`Vector`, gate in both** | **passes** | **60/60** |

Failing before / passing after, same testset, fix stashed and restored in one session:

```text
before:  FullSpecialize initialization maps are generated functions | 38 pass | 2 fail | 40 total
after:   FullSpecialize initialization maps are generated functions | 40 pass |  40 total | 26.8s
```

The two new assertions pin both halves — non-isbits buffers stay on the heap, and
tunable/initials stay `StaticArray` so the gate cannot silently over-apply. `SizedVector`
is now unused and its import is dropped.

**Ignore this draft until reviewed by @ChrisRackauckas.**

---

Commits through `9a652005c8` were produced with Codex CLI 0.153.4 (model gpt-6-astra), local session `01a070db-f817-7f93-9295-aae2232fc2c7` (no shareable conversation URL exposed).

The review-response commit `fed8cdb77b` was produced with 🤖 [Claude Code](https://claude.com/claude-code) 2.0.14 (model: claude-opus-5[1m])

https://claude.ai/code/session_01GdSpCLd7NBZuuePJmcDzU7

### DIFF
--- a/lib/ModelingToolkitBase/src/systems/problem_utils.jl
+++ b/lib/ModelingToolkitBase/src/systems/problem_utils.jl
@@ -1692,8 +1692,13 @@ end
 function _static_parameter_buffer(prototype, values::Tuple)
     T = isempty(values) ? eltype(prototype) :
         promote_type(eltype(prototype), mapreduce(typeof, promote_type, values))
-    static_type = ArrayInterface.ismutable(prototype) ? MVector : SVector
-    return static_type{length(values), T}(values)
+    if !ArrayInterface.ismutable(prototype)
+        return SVector{length(values), T}(values)
+    elseif isbitstype(T)
+        return MVector{length(values), T}(values)
+    else
+        return SizedVector{length(values), T}(T[values...])
+    end
 end
 
 function _parameter_buffer_expr(prototype, raw::Symbol, idxs, p_constructor)
--- a/lib/ModelingToolkitBase/src/systems/problem_utils.jl
+++ b/lib/ModelingToolkitBase/src/systems/problem_utils.jl
@@ -1692,7 +1692,8 @@ end
 function _static_parameter_buffer(prototype, values::Tuple)
     T = isempty(values) ? eltype(prototype) :
         promote_type(eltype(prototype), mapreduce(typeof, promote_type, values))
-    return SVector{length(values), T}(values)
+    static_type = ArrayInterface.ismutable(prototype) ? MVector : SVector
+    return static_type{length(values), T}(values)
 end
 
 function _parameter_buffer_expr(prototype, raw::Symbol, idxs, p_constructor)
--- a/lib/ModelingToolkitBase/src/systems/problem_utils.jl
+++ b/lib/ModelingToolkitBase/src/systems/problem_utils.jl
@@ -1718,13 +1718,6 @@ function _construct_fullspecialize_initializeprobpmap(
     for group in grouped[3:5], buffer in group
         append!(flat_syms, buffer)
     end
-    for sym in flat_syms
-        symbolic_type(sym) == ArraySymbolic() && throw(
-            ArgumentError(
-                "FullSpecialize initialization maps do not yet support array-valued parameter storage slots."
-            )
-        )
-    end
     expr = build_explicit_observed_function(
         initsys, Tuple(flat_syms);
         expression = Val(true), compiler_options, kwargs...
--- a/lib/ModelingToolkitBase/src/systems/problem_utils.jl
+++ b/lib/ModelingToolkitBase/src/systems/problem_utils.jl
@@ -1725,12 +1725,6 @@ function _construct_fullspecialize_initializeprobpmap(
             )
         )
     end
-    isempty(SU.getmetadata(sys, DiffCacheParams, Dict{SymbolicT, Int}())) || throw(
-        ArgumentError(
-            "FullSpecialize initialization maps do not yet support DiffCache parameters."
-        )
-    )
-
     expr = build_explicit_observed_function(
         initsys, Tuple(flat_syms);
         expression = Val(true), compiler_options, kwargs...
--- a/lib/ModelingToolkitBase/src/systems/problem_utils.jl
+++ b/lib/ModelingToolkitBase/src/systems/problem_utils.jl
@@ -1108,6 +1108,24 @@ struct MTKParametersReconstructor{T, I, D, C, N}
     diffcache_buffer_idx::Int
 end
 
+function _unwrap_initial_symbols!(initsyms, srcsys)
+    allsyms = Set{SymbolicT}(variable_symbols(srcsys))
+    for i in eachindex(initsyms)
+        sym = initsyms[i]
+        arr, isarr = split_indexed_var(sym)
+        innersym = if isarr
+            sidx = get_stable_index(sym)
+            first(arguments(arr))[sidx]
+        else
+            first(arguments(arr))
+        end
+        if innersym in allsyms
+            initsyms[i] = innersym
+        end
+    end
+    return initsyms
+end
+
 # TODO: make this infer when the nonnumerics are non-trivial
 function (recon::MTKParametersReconstructor)(src, dst)
     return recon(src, parameter_values(dst))
@@ -1185,22 +1203,7 @@ function MTKParametersReconstructor(
     end
     initials_getter = if initials && !isempty(syms[2])
         initsyms = syms[2]::Vector{SymbolicT}
-        allsyms = Set{SymbolicT}(variable_symbols(srcsys))
-        if unwrap_initials
-            for i in eachindex(initsyms)
-                sym = initsyms[i]
-                arr, isarr = split_indexed_var(sym)
-                innersym = if isarr
-                    sidx = get_stable_index(sym)
-                    first(arguments(arr))[sidx]
-                else
-                    first(arguments(arr))
-                end
-                if innersym in allsyms
-                    initsyms[i] = innersym
-                end
-            end
-        end
+        unwrap_initials && _unwrap_initial_symbols!(initsyms, srcsys)
         p_constructor ∘ CopyParamsByTemplate(srcsys, initsyms; kwargs...)
     else
         Returns(SVector{0, Float64}())
@@ -1630,6 +1633,159 @@ function (map::InitializationMap{true})(x)
     return __iip_u0_ad_wrapper(map.u0_constructor(map.map(x)))
 end
 
+function _generated_map_expr(finish, expr::Expr, map_args::Vector{Symbol}, sources::Vector{Expr})
+    @assert expr.head === :function
+    signature = expr.args[1]
+    @assert signature isa Expr && signature.head === :tuple
+    generated_args = signature.args
+    @assert length(generated_args) == length(sources)
+
+    body = Expr(:block)
+    for (arg, source) in zip(generated_args, sources)
+        push!(body.args, :(local $arg = $source))
+    end
+    raw = gensym(:initialization_map_values)
+    push!(body.args, :(local $raw = $(expr.args[2])))
+    push!(body.args, finish(raw))
+    return Expr(:function, Expr(:tuple, map_args...), body)
+end
+
+function _generated_map_sources(valp::Symbol, time_dependent::Bool)
+    sources = Expr[
+        Expr(:call, GlobalRef(SymbolicIndexingInterface, :state_values), valp),
+        Expr(:call, GlobalRef(SymbolicIndexingInterface, :parameter_values), valp),
+    ]
+    if time_dependent
+        push!(sources, Expr(:call, GlobalRef(SymbolicIndexingInterface, :current_time), valp))
+    end
+    return sources
+end
+
+function _construct_fullspecialize_initializeprobmap(
+        initsys::AbstractSystem, solved_unknowns;
+        iip::Bool, u0_constructor, floatT, eval_module,
+        compiler_options::CompilerOptions = CompilerOptions(), kwargs...
+    )
+    expr = build_explicit_observed_function(
+        initsys, solved_unknowns;
+        expression = Val(true), output_type = SVector, compiler_options, kwargs...
+    )
+    sol = gensym(:initialization_solution)
+    sources = _generated_map_sources(sol, is_time_dependent(initsys))
+    n = length(solved_unknowns)
+    static_type = iip ? MVector : SVector
+    map_expr = _generated_map_expr(expr, [sol], sources) do raw
+        T = gensym(:initialization_map_eltype)
+        p = Expr(:call, GlobalRef(SymbolicIndexingInterface, :parameter_values), sol)
+        tunable_eltype = Expr(:call, GlobalRef(@__MODULE__, :_tunable_eltype), p)
+        promoted = Expr(:call, promote_type, Expr(:call, eltype, raw), tunable_eltype, floatT)
+        static_values = Expr(:call, Expr(:curly, static_type, n, T), raw)
+        result = Expr(:call, QuoteNode(u0_constructor), static_values)
+        if iip
+            result = Expr(:call, GlobalRef(@__MODULE__, :__iip_u0_ad_wrapper), result)
+        end
+        return Expr(:block, :(local $T = $promoted), result)
+    end
+    return eval_or_rgf(map_expr; eval_module, compiler_options)
+end
+
+function _static_parameter_buffer(prototype, values::Tuple)
+    T = isempty(values) ? eltype(prototype) :
+        promote_type(eltype(prototype), mapreduce(typeof, promote_type, values))
+    return SVector{length(values), T}(values)
+end
+
+function _parameter_buffer_expr(prototype, raw::Symbol, idxs, p_constructor)
+    values = Expr(:tuple, [Expr(:ref, raw, i) for i in idxs]...)
+    buffer = Expr(
+        :call, GlobalRef(@__MODULE__, :_static_parameter_buffer), prototype, values
+    )
+    p_constructor === identity && return buffer
+    return Expr(:call, QuoteNode(p_constructor), buffer)
+end
+
+function _construct_fullspecialize_initializeprobpmap(
+        sys::AbstractSystem, initsys::AbstractSystem;
+        p_constructor, eval_module,
+        compiler_options::CompilerOptions = CompilerOptions(), kwargs...
+    )
+    grouped = reorder_parameters(
+        sys, parameters(sys; initial_parameters = true); flatten = false
+    )
+    initial_syms = _unwrap_initial_symbols!(copy(grouped[2]), initsys)
+    flat_syms = SymbolicT[]
+    append!(flat_syms, grouped[1], initial_syms)
+    for group in grouped[3:5], buffer in group
+        append!(flat_syms, buffer)
+    end
+    for sym in flat_syms
+        symbolic_type(sym) == ArraySymbolic() && throw(
+            ArgumentError(
+                "FullSpecialize initialization maps do not yet support array-valued parameter storage slots."
+            )
+        )
+    end
+    isempty(SU.getmetadata(sys, DiffCacheParams, Dict{SymbolicT, Int}())) || throw(
+        ArgumentError(
+            "FullSpecialize initialization maps do not yet support DiffCache parameters."
+        )
+    )
+
+    expr = build_explicit_observed_function(
+        initsys, Tuple(flat_syms);
+        expression = Val(true), compiler_options, kwargs...
+    )
+    prob = gensym(:problem)
+    sol = gensym(:initialization_solution)
+    sources = _generated_map_sources(sol, is_time_dependent(initsys))
+    map_expr = _generated_map_expr(expr, [prob, sol], sources) do raw
+        outer_p = gensym(:outer_parameters)
+        p = Expr(:call, GlobalRef(SymbolicIndexingInterface, :parameter_values), prob)
+        if !is_split(sys)
+            result = _parameter_buffer_expr(outer_p, raw, eachindex(flat_syms), p_constructor)
+            return Expr(:block, :(local $outer_p = $p), result)
+        end
+
+        offset = 0
+        tunable_idxs = (offset + 1):(offset + length(grouped[1]))
+        offset += length(grouped[1])
+        initial_idxs = (offset + 1):(offset + length(grouped[2]))
+        offset += length(grouped[2])
+        tunable = _parameter_buffer_expr(Expr(:., outer_p, QuoteNode(:tunable)), raw, tunable_idxs, p_constructor)
+        initials = _parameter_buffer_expr(Expr(:., outer_p, QuoteNode(:initials)), raw, initial_idxs, p_constructor)
+
+        portions = Expr[]
+        for (portion_idx, group) in enumerate(grouped[3:5])
+            buffers = Expr[]
+            field = (:discrete, :constant, :nonnumeric)[portion_idx]
+            for (buffer_idx, syms) in enumerate(group)
+                idxs = (offset + 1):(offset + length(syms))
+                offset += length(syms)
+                prototype = Expr(:ref, Expr(:., outer_p, QuoteNode(field)), buffer_idx)
+                buffer = _parameter_buffer_expr(prototype, raw, idxs, p_constructor)
+                if portion_idx == 1
+                    sizes = get_index_cache(sys).discrete_buffer_sizes[buffer_idx]
+                    block_sizes = Expr(
+                        :call, Expr(:curly, SVector, length(sizes), Int),
+                        map(size -> size.length, sizes)...
+                    )
+                    p_constructor === identity ||
+                        (block_sizes = Expr(:call, QuoteNode(p_constructor), block_sizes))
+                    buffer = Expr(:call, BlockedArray, buffer, block_sizes)
+                end
+                push!(buffers, buffer)
+            end
+            push!(portions, Expr(:tuple, buffers...))
+        end
+        result = Expr(
+            :call, MTKParameters, tunable, initials, portions...,
+            Expr(:., outer_p, QuoteNode(:caches))
+        )
+        return Expr(:block, :(local $outer_p = $p), result)
+    end
+    return eval_or_rgf(map_expr; eval_module, compiler_options)
+end
+
 """
     $(TYPEDSIGNATURES)
 
@@ -1896,6 +2052,7 @@ function maybe_build_initialization_problem(
         sys::AbstractSystem, iip::Bool, op::SymmapT, t, guesses,
         opts::SciMLProblemOptions;
         specialize = SciMLBase.AutoDespecialize,
+        map_specialize = specialize,
         # Intercept `expression` because we don't support it here yet
         expression = Val{false}, kwargs...
     )
@@ -1987,6 +2144,13 @@ function maybe_build_initialization_problem(
         solved_unknowns = filter(var -> var in all_init_syms, unknowns(sys))
         if isempty(solved_unknowns)
             initializeprobmap = nothing
+        elseif map_specialize === SciMLBase.FullSpecialize
+            initializeprobmap = _construct_fullspecialize_initializeprobmap(
+                initializeprob.f.sys, solved_unknowns;
+                iip, u0_constructor, floatT, eval_module,
+                compiler_options = init_compiler_options,
+                kwargs...
+            )
         else
             initializeprobmap = InitializationMap{iip}(
                 u0_constructor,
@@ -2010,6 +2174,12 @@ function maybe_build_initialization_problem(
     ]
     if initializeprobmap === nothing && isempty(punknowns)
         initializeprobpmap = nothing
+    elseif map_specialize === SciMLBase.FullSpecialize
+        initializeprobpmap = _construct_fullspecialize_initializeprobpmap(
+            sys, initsys;
+            p_constructor, eval_module, compiler_options = init_compiler_options,
+            kwargs...
+        )
     else
         initializeprobpmap = construct_initializeprobpmap(
             sys, initsys; p_constructor, eval_expression, eval_module, kwargs...
@@ -2222,12 +2392,12 @@ function __process_SciMLProblem(
     end
 
     if build_initializeprob
+        problem_specialize = SciMLBase.specialization(constructor)
         kws = maybe_build_initialization_problem(
             sys, constructor <: SciMLBase.AbstractSciMLFunction{true},
             op, t, guesses, opts;
-            specialize = initialization_specialization(
-                SciMLBase.specialization(constructor)
-            ), kwargs...
+            specialize = initialization_specialization(problem_specialize),
+            map_specialize = problem_specialize, kwargs...
         )
 
         kwargs = merge(kwargs, kws)
--- a/lib/ModelingToolkitBase/src/systems/problem_utils.jl
+++ b/lib/ModelingToolkitBase/src/systems/problem_utils.jl
@@ -1679,7 +1679,10 @@ function _construct_fullspecialize_initializeprobmap(
         p = Expr(:call, GlobalRef(SymbolicIndexingInterface, :parameter_values), sol)
         tunable_eltype = Expr(:call, GlobalRef(@__MODULE__, :_tunable_eltype), p)
         promoted = Expr(:call, promote_type, Expr(:call, eltype, raw), tunable_eltype, floatT)
-        static_values = Expr(:call, Expr(:curly, static_type, n, T), raw)
+        static_values = Expr(
+            :call, GlobalRef(@__MODULE__, :_static_initialization_buffer),
+            Expr(:curly, static_type, n, T), raw
+        )
         result = Expr(:call, QuoteNode(u0_constructor), static_values)
         if iip
             result = Expr(:call, GlobalRef(@__MODULE__, :__iip_u0_ad_wrapper), result)
@@ -1689,7 +1692,7 @@ function _construct_fullspecialize_initializeprobmap(
     return eval_or_rgf(map_expr; eval_module, compiler_options)
 end
 
-function _static_parameter_buffer(prototype, values::Tuple)
+function _static_initialization_buffer(prototype, values)
     T = isempty(values) ? eltype(prototype) :
         promote_type(eltype(prototype), mapreduce(typeof, promote_type, values))
     if !ArrayInterface.ismutable(prototype)
@@ -1704,7 +1707,7 @@ end
 function _parameter_buffer_expr(prototype, raw::Symbol, idxs, p_constructor)
     values = Expr(:tuple, [Expr(:ref, raw, i) for i in idxs]...)
     buffer = Expr(
-        :call, GlobalRef(@__MODULE__, :_static_parameter_buffer), prototype, values
+        :call, GlobalRef(@__MODULE__, :_static_initialization_buffer), prototype, values
     )
     p_constructor === identity && return buffer
     return Expr(:call, QuoteNode(p_constructor), buffer)
@@ -1715,13 +1718,16 @@ function _construct_fullspecialize_initializeprobpmap(
         p_constructor, eval_module,
         compiler_options::CompilerOptions = CompilerOptions(), kwargs...
     )
-    grouped = reorder_parameters(
-        sys, parameters(sys; initial_parameters = true); flatten = false
-    )
-    initial_syms = _unwrap_initial_symbols!(copy(grouped[2]), initsys)
+    ps = parameters(sys; initial_parameters = true)
+    groups = if is_split(sys)
+        grouped = reorder_parameters(sys, ps; flatten = false)
+        initial_syms = _unwrap_initial_symbols!(copy(grouped[2]), initsys)
+        ((grouped[1],), (initial_syms,), grouped[3:5]...)
+    else
+        ((ps,),)
+    end
     flat_syms = SymbolicT[]
-    append!(flat_syms, grouped[1], initial_syms)
-    for group in grouped[3:5], buffer in group
+    for group in groups, buffer in group
         append!(flat_syms, buffer)
     end
     expr = build_explicit_observed_function(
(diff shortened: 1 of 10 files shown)


ChrisRackauckas-Claude:
Plan/status: generate single dropped-body RuntimeGeneratedFunctions only for explicit FullSpecialize; preserve AutoSpecialize and AutoDespecialize behavior; verify map structure and output against existing maps; run Initialization and local hygiene checks; track the independently reproduced upstream/master QA failures in the required parallel investigation before final handoff.

ChrisRackauckas-Claude:
CI Runic failure is pre-existing formatter-version drift in three files untouched by this PR. The mechanical cleanup is isolated in draft PR https://github.com/SciML/ModelingToolkit.jl/pull/5046. This feature PR should be rebased after that cleanup lands; no formatting-only changes will be mixed into this behavior diff.

ChrisRackauckas-Claude:
Follow-up: the isolated formatting PR https://github.com/SciML/ModelingToolkit.jl/pull/5046 now passes CI Runic, Runic Suggestions, and spelling. This confirms the formatter failure here is resolved by that base cleanup rather than a change to this feature diff.

ChrisRackauckas-Claude:
Follow-up commit `b3256600f9d2f0dae9ebdbb25f0841ff375a03c8` addresses the CI-discovered DiffCache path. The original run failed with the explicit unsupported-parameter `ArgumentError`; removing that guard exposed an existing immutable nonnumeric-buffer typo (`fieldtype(C, i)` while reconstructing `N`) and produced 66 passes / 2 errors in IndexCache. Correcting it to `fieldtype(N, i)` produced `IndexCache Test | 68 pass | 68 total` locally. The broader InterfaceII run later hit two unrelated stochastic jump assertions after 4,202 passes; clean-base reproduction/bisect is running separately.

ChrisRackauckas-Claude:
Added array-valued callable-parameter support in https://github.com/SciML/ModelingToolkit.jl/commit/d3c8a49729520ebd6644557a1dfae8e55de496e3.

Failure on the preceding implementation (`b3256600f9`):

```text
ArgumentError: FullSpecialize initialization maps do not yet support array-valued parameter storage slots.
Test Summary: FullSpecialize array-valued storage slot | 1 Error
```

The same namespaced `ArraySymbolic` reproduction now passes:

```text
Test Summary: array callable | 3 pass | 3 total | 14.9s
```

The complete ModelingToolkitNeuralNets suite also passes against this branch (10 + 1 + 37 + 55 + 7 passing assertions, with its 2 pre-existing broken tests). This directly covers the `lux_apply` parameter shape that failed in https://github.com/SciML/ModelingToolkit.jl/actions/runs/33300901287/job/99228760344.

ChrisRackauckas-Claude:
CI exposed a mutable-buffer regression in the first generated-parameter implementation: https://github.com/SciML/ModelingToolkit.jl/actions/runs/33355753893/job/99377920719 failed six `MTKParameters` cases because `SciMLStructures.replace!` attempted `setindex!` on generated `SVector` tunable/constant buffers.

Fixed in https://github.com/SciML/ModelingToolkit.jl/commit/93854e05358fb60ef05a437ac7611b8febcaa6f9 by selecting `MVector` when the source buffer is mutable while retaining `SVector` for immutable prototypes. The generated map itself remains an isbits, body-dropped `RuntimeGeneratedFunction`.

Local verification after the fix:

```text
Focused Tunable/Constants replace! reproducer: exit 0
SymbolicIndexingInterface | 1984 pass | 1984 total | 8m06.7s
Testing ModelingToolkit tests passed
Runic.jl v1.10, typos, git diff --check: exit 0
```

ChrisRackauckas-Claude:
Commit https://github.com/SciML/ModelingToolkit.jl/commit/613bdd13e7ccce845d2ed553cc72f91051326518 fixes the non-isbits mutable-buffer regression found by the FMI LTS job: https://github.com/SciML/ModelingToolkit.jl/actions/runs/33358934228/job/99386604733.

Failing before (CI):
```text
FMI | 56 pass | 2 errors | 58 total
setindex!() with non-isbitstype eltype is not supported by StaticArrays. Consider using SizedArray.
```

Passing after (local):
```text
GROUP=FMI ... Pkg.test()
FMI | 60 pass | 60 total | 15m53.9s
Testing ModelingToolkit tests passed

GROUP=Initialization ... Pkg.test()
Initialization | 870 pass | 12 broken | 882 total | 36m22.1s
Testing ModelingToolkit tests passed
```

Runic 1.10, typos over the diff, and `git diff --check` also exited successfully.

ChrisRackauckas-Claude:
Commit https://github.com/SciML/ModelingToolkit.jl/commit/010bc3098ab51d64ca471245c36931f5b2ccb3d4 addresses the downgraded Initialization failure: https://github.com/SciML/ModelingToolkit.jl/actions/runs/33366256108/job/99407718377.

The failed job used SymbolicIndexingInterface 0.3.46, where the callable parameter is classified as `ScalarSymbolic` despite retaining its declared `[1:2]` shape. The generated map already returned the exact callable; the failing enum assertion encoded a newer dependency detail.

Local Julia 1.10 reproduction pinned the relevant CI floors (ArrayInterface 7.28.0, SciMLBase 3.48.0, StaticArraysCore 1.4.3, SymbolicUtils 4.37.0, SymbolicIndexingInterface 0.3.46, Symbolics 7.37.0):
```text
classifier=ScalarSymbolic
shape/generated-map/callable-identity assertions: 3/3 pass
```

The same semantic assertions also pass `3/3` on the current dependency stack. Runic 1.10, typos, and `git diff --check` pass for the follow-up diff.

ChrisRackauckas-Claude:
Follow-up commit https://github.com/SciML/ModelingToolkit.jl/commit/258f72c8de92ef90a98730ef06ff594e7e1a94d3 fixes the remaining downgrade error from https://github.com/SciML/ModelingToolkit.jl/actions/runs/33387712747/job/99474377835. SymbolicUtils 4.37 provides `shape` at its owning module but does not re-export it through ModelingToolkitBase.

Exact CI-floor local result after using the owner API:
```text
classifier=ScalarSymbolic
SU.shape/generated-map/callable-identity assertions: 3/3 pass
```

The current-stack owner-API shape assertions pass `2/2`. Runic 1.10, typos, and `git diff --check` pass.

ChrisRackauckas-Claude:
CI confirmation on head https://github.com/SciML/ModelingToolkit.jl/commit/258f72c8de92ef90a98730ef06ff594e7e1a94d3:

- The previously failing downgraded Initialization job now passes: https://github.com/SciML/ModelingToolkit.jl/actions/runs/33389832004/job/99480764699.
- Downgraded QA still has the documented base-branch failure: https://github.com/SciML/ModelingToolkit.jl/actions/runs/33389832004/job/99480768825.
- Julia-pre Optimization has the known base incompatibility in CasADi dynamic-optimization tests (`create_array(::Type{CasADi.MX}, ...)` and `abs2(::CasADi.MX)` missing): https://github.com/SciML/ModelingToolkit.jl/actions/runs/33389832292/job/99481001441.

At this update, 38 checks pass, 66 remain in progress, 2 known base checks fail, and 1 is skipped. No completed branch-relevant check fails.

ChrisRackauckas-Claude:
Updated the draft with a shared static-buffer constructor and reverse-mode rule, independent cache ownership, and split=false support. Added discriminating regressions for cache aliasing, unsplit parameters, and state/parameter gradients. Final local results: root QA 40/40; Initialization 882 passed plus 12 existing broken; SymbolicIndexingInterface 1984/1984; focused map/AD tests 39/39. Full Extensions and neural-network Core also passed as detailed in the body.

The allocation microbenchmark improved from 784 to 80 bytes per map-call pair, or zero with immutable static constructors. This is not a TTFX fix: repeated FullSpecialize construction still creates distinct generated map types. The title and body now state the narrower scope.

Registered-dependency Base QA remains blocked by the independently reproduced master Moshi/JET issue: https://github.com/SciML/ModelingToolkit.jl/issues/5063 . No QA suppression or dependency workaround is included. Ignore this draft until reviewed by @ChrisRackauckas.

AI update: Codex CLI 0.153.4; model gpt-6-astra; local session 01a070db-f817-7f93-9295-aae2232fc2c7 (no shareable conversation URL exposed).

lib/ModelingToolkitBase/src/systems/problem_utils.jl
```diff
@@ -1630,6 +1633,155 @@ function (map::InitializationMap{true})(x)
...
+    signature = expr.args[1]
+    @assert signature isa Expr && signature.head === :tuple
+    generated_args = signature.args
+    @assert length(generated_args) == length(sources)
+
+    body = Expr(:block)
+    for (arg, source) in zip(generated_args, sources)
+        push!(body.args, :(local $arg = $source))
+    end
+    raw = gensym(:initialization_map_values)
```
AayushSabharwal: I thought we want to avoid `gensym`?

lib/ModelingToolkitBase/src/systems/problem_utils.jl
```diff
@@ -1630,6 +1633,155 @@ function (map::InitializationMap{true})(x)
...
+    push!(body.args, :(local $raw = $(expr.args[2])))
+    push!(body.args, finish(raw))
+    return Expr(:function, Expr(:tuple, map_args...), body)
```
AayushSabharwal: Shouldn't splat. Prefer

```suggestion
    fn_args = Expr(:tuple)
    append!(fn_args.args, map_args)
    return Expr(:function, fn_args, body)
```

lib/ModelingToolkitBase/src/systems/problem_utils.jl
```diff
@@ -1630,6 +1633,155 @@ function (map::InitializationMap{true})(x)
...
+        return SVector{length(values), T}(values)
+    elseif isbitstype(T)
+        return MVector{length(values), T}(values)
+    else
+        return SizedVector{length(values), T}(T[values...])
+    end
+end
+
+function _parameter_buffer_expr(prototype, raw::Symbol, idxs, p_constructor)
+    values = Expr(:tuple, [Expr(:ref, raw, i) for i in idxs]...)
```
AayushSabharwal: Shouldn't splat

lib/ModelingToolkitBase/src/systems/problem_utils.jl
```diff
@@ -1630,6 +1633,155 @@ function (map::InitializationMap{true})(x)
...
+    return Expr(:call, QuoteNode(p_constructor), buffer)
+end
+
+function _construct_fullspecialize_initializeprobpmap(
+        sys::AbstractSystem, initsys::AbstractSystem;
+        p_constructor, eval_module,
+        compiler_options::CompilerOptions = CompilerOptions(), kwargs...
+    )
+    ps = parameters(sys; initial_parameters = true)
+    groups = if is_split(sys)
```
AayushSabharwal: Why does `groups` need to be a tuple? It can just be an appropriately typed vector? This is catastrophic for inference.

lib/ModelingToolkitBase/src/systems/problem_utils.jl
```diff
@@ -1630,6 +1633,155 @@ function (map::InitializationMap{true})(x)
...
+                prototype = Expr(:., outer_p, QuoteNode(field))
+                if field !== :tunable && field !== :initials
+                    prototype = Expr(:ref, prototype, buffer_idx)
+                end
+                buffer = _parameter_buffer_expr(prototype, raw, idxs, p_constructor)
+                if field === :discrete
+                    sizes = get_index_cache(sys).discrete_buffer_sizes[buffer_idx]
+                    block_sizes = Expr(
+                        :call, Expr(:curly, SVector, length(sizes), Int),
+                        map(size -> size.length, sizes)...
```
AayushSabharwal: Avoid the splat

lib/ModelingToolkitBase/src/systems/problem_utils.jl
```diff
@@ -1630,6 +1633,155 @@ function (map::InitializationMap{true})(x)
...
+                    )
+                    p_constructor === identity ||
+                        (block_sizes = Expr(:call, QuoteNode(p_constructor), block_sizes))
+                    buffer = Expr(:call, BlockedArray, buffer, block_sizes)
+                end
+                push!(buffers, buffer)
+            end
+            portion = field === :tunable || field === :initials ? only(buffers) :
+                Expr(:tuple, buffers...)
```
AayushSabharwal: Avoid the splat

lib/ModelingToolkitBase/src/systems/problem_utils.jl
```diff
@@ -1630,6 +1633,155 @@ function (map::InitializationMap{true})(x)
...
+            push!(portions, portion)
+        end
+        result = Expr(
+            :call, MTKParameters, portions...,
```
AayushSabharwal: Avoid the splat

AayushSabharwal (changes requested):

lib/ModelingToolkitBase/src/systems/problem_utils.jl
ChrisRackauckas: Yeah we do want to avoid gensym, good catch, should probably agents.md that.

ChrisRackauckas-Claude: Fixed in https://github.com/SciML/ModelingToolkit.jl/commit/fed8cdb77b. All five `gensym`s in the new codegen are now fixed sentinels (`INITMAP_VALUES`, `INITMAP_ELTYPE`, `INITMAP_SOLUTION`, `INITMAP_PROBLEM`, `INITMAP_OUTER_PARAMETERS`), following the rationale already written down for `HOMOTOPY_LAMBDA` in `homotopy_operator.jl`.

This was not just style — it was a real defect in the PR, and it explains the TTFX non-improvement the body reported. Building the same system twice produced two distinct `RuntimeGeneratedFunction` types, so the RGF Expr-hash cache never hit:

```text
generated map type reuse across identical systems: Test Failed
  Expression: typeof(d1.initializeprobpmap) === typeof(d2.initializeprobpmap)
   Evaluated: RuntimeGeneratedFunction{(Symbol("##problem#280"), Symbol("##initialization_solution#281")), …, (0x8298bbd3, 0xbcbc4d5b, 0x7a096718, 0x43533499, 0x57e4022b), Nothing} ===
              RuntimeGeneratedFunction{(Symbol("##problem#287"), Symbol("##initialization_solution#288")), …, (0x44632bf6, 0x42fe7ec6, 0x4d5618c1, 0xb6d83a49, 0x2afa037a), Nothing}
Test Summary:                                     | Pass  Fail  Total   Time
generated map type reuse across identical systems |    2     2      4  10.6s
```

With the sentinels:

```text
Test Summary:                                     | Pass  Total  Time
generated map type reuse across identical systems |    4      4  7.0s
```

Checked in as three assertions at the end of the `FullSpecialize initialization maps are generated functions` testset, so the two lowerings are required to share a type.

@ChrisRackauckas — written up in https://github.com/SciML/ModelingToolkit.jl/pull/5105, which adds a root `AGENTS.md` (the repo had none) with this rule plus the two below.

lib/ModelingToolkitBase/src/systems/problem_utils.jl
ChrisRackauckas-Claude: Applied your suggestion verbatim in https://github.com/SciML/ModelingToolkit.jl/commit/fed8cdb77b.

lib/ModelingToolkitBase/src/systems/problem_utils.jl
ChrisRackauckas-Claude: Fixed. `_parameter_buffer_expr` now builds the tuple by pushing:

```julia
values = Expr(:tuple)
for i in idxs
    push!(values.args, Expr(:ref, raw, i))
end
```

The one other splat in that function, `SizedVector{length(values), T}(T[values...])` in `_static_initialization_buffer`, is now `collect(T, values)`. There are no splats left in the added code other than `kwargs...` forwarding.

lib/ModelingToolkitBase/src/systems/problem_utils.jl
ChrisRackauckas-Claude: Fixed — `block_sizes` starts as `Expr(:call, Expr(:curly, SVector, length(sizes), Int))` and the lengths are pushed onto `.args` in a loop.

lib/ModelingToolkitBase/src/systems/problem_utils.jl
ChrisRackauckas-Claude: Fixed — the non-`tunable`/`initials` portion is now built with `Expr(:tuple)` + `append!(tup.args, buffers)`.

lib/ModelingToolkitBase/src/systems/problem_utils.jl
ChrisRackauckas-Claude: Fixed — `result = Expr(:call, MTKParameters)`, then `append!(result.args, portions)` and `push!` for the caches argument.

lib/ModelingToolkitBase/src/systems/problem_utils.jl
ChrisRackauckas-Claude: Changed in https://github.com/SciML/ModelingToolkit.jl/commit/fed8cdb77b. It does not need to be a tuple — nothing indexes `groups` heterogeneously, it is only iterated alongside the field names. It is now a `Vector{Vector{Vector{SymbolicT}}}`:

```julia
groups = Vector{Vector{SymbolicT}}[]
if is_split(sys)
    grouped = reorder_parameters(sys, ps; flatten = false)
    push!(groups, Vector{SymbolicT}[grouped[1]::Vector{SymbolicT}])
    initial_syms = _unwrap_initial_symbols!(copy(grouped[2]::Vector{SymbolicT}), initsys)
    push!(groups, Vector{SymbolicT}[initial_syms])
    for i in 3:5
        push!(groups, grouped[i]::Vector{Vector{SymbolicT}})
    end
else
    push!(groups, Vector{SymbolicT}[ps])
end
```

The type asserts pin the `Union` element type coming out of `ReorderedParametersT`, so both the `flat_syms` loop and the buffer loop below it are now concretely typed instead of specializing on each system's shape.

ChrisRackauckas-Claude:
The rebase onto master picked up a **new downstream failure that is this PR's bug**, not a master one. Flagging it rather than fixing it, because the fix is a design call for this PR.

`ModelingToolkitStandardLibrary.jl/Core/1/` fails: https://github.com/SciML/ModelingToolkit.jl/actions/runs/34481227325/job/102884238601

```text
LinearInterpolation: Error During Test at downstream/test/sources.jl:652
  StackOverflowError:
   [1] StaticArraysCore.SizedVector{1, UnionAll, Vector{UnionAll}}(a::Vector{Any})
       @ StaticArrays ~/.julia/packages/StaticArrays/k7bbZ/src/SizedArray.jl:15
   repeated 79984 times
```

It is not pre-existing on master: the same job **passes** on https://github.com/SciML/ModelingToolkit.jl/pull/5105, which is master plus a single markdown file. It passed on this branch's pre-rebase head `438703fbe4` only because the test did not exist yet — the earlier job log has zero occurrences of `ParametrizedInterpolation`. The new MTKStdlib test builds `ODEProblem{true, SciMLBase.FullSpecialize}`, which is exactly the path this PR adds, and then `remake`s it.

Reproduced locally against this branch, outside CI:

```text
built prob
nonnumeric buffer type: Tuple{StaticArraysCore.SizedVector{1, UnionAll, Vector{UnionAll}}}
solved
ERROR: LoadError: StackOverflowError:
Stacktrace:
 [1] StaticArraysCore.SizedVector{1, UnionAll, Vector{UnionAll}}(a::Vector{Any}) (repeats 79984 times)
   @ StaticArrays ~/.julia/packages/StaticArrays/k7bbZ/src/SizedArray.jl:15
```

Reduced to StaticArrays alone, no MTK involved:

```julia
using StaticArrays
struct Foo{T} end
SizedVector{1, UnionAll, Vector{UnionAll}}(Any[Foo])   # StackOverflowError
```

**Mechanism.** `_static_initialization_buffer`'s third branch wraps a mutable non-isbits buffer in a `SizedVector`. Nonnumeric parameter buffers are always non-isbits, so a nonnumeric buffer holding a `UnionAll` (here `DataInterpolations.LinearInterpolation`) becomes `SizedVector{1, UnionAll, Vector{UnionAll}}`. `remake` then reconstructs it through `similar_type(fieldtype(N, i), nonnumericT[i])(nonnumerics[i])` in `parameter_buffer.jl`. `similar_type` is a no-op on that fully-parameterized `SizedVector`, so the constructor is re-entered with the same type and a `Vector` argument forever.

That `SizedVector` branch came from `613bdd13` to fix the FMI `setindex!`-on-non-isbits failure, so it is load-bearing — this is not a one-line revert.

**Why I did not fix it here.** I tried the obvious fix (return a plain mutable `Vector{T}` for the non-isbits branch, and drop the now-unused `SizedVector` import). That does clear the `StackOverflowError` — the buffer becomes `Tuple{Vector{UnionAll}}` and the solve succeeds — but `remake` then fails one layer down:

```text
ERROR: LoadError: TypeError: in typeassert, expected Tuple{Vector{UnionAll}}, got a value of type Tuple{StaticArraysCore.MVector{1, UnionAll}}
  [1] (::ModelingToolkitBase.MTKParametersReconstructor{...})
```

The generated map picks the nonnumeric buffer type from the outer problem's prototype, while `MTKParametersReconstructor` rebuilds it from `buftypes` captured off the initialization problem, and the two only agreed by accident before. Making them agree is a decision about which prototype the generated map should follow, so I have left the branch at the reviewed state rather than shipping a change that swaps one error for another.

Three buffer choices, three failures, for whoever picks this up:

| non-isbits nonnumeric buffer | fails as |
| --- | --- |
| `MVector` | FMI: `setindex!` unsupported for non-isbits eltype |
| `SizedVector` (current) | MTKStdlib: `remake` `StackOverflowError` |
| `Vector` | MTKStdlib: `remake` typeassert, `Vector` vs `MVector` |

Standalone reproducer (needs MTKStdlib + DataInterpolations dev-ed against this branch):

```julia
using ModelingToolkitStandardLibrary.Blocks
using ModelingToolkit, DataInterpolations, OrdinaryDiffEq, SciMLBase

@variables y(t) = 0
u = rand(15); x = 0:14.0
@named i = ParametrizedInterpolation(LinearInterpolation, u, x)
@named model = System([i.input.u ~ t, D(y) ~ i.output.u], t, systems = [i])
sys = mtkcompile(model)

prob = ODEProblem{true, SciMLBase.FullSpecialize}(sys, [], (0.0, 4))
solve(prob, Tsit5())                              # fine
remake(prob, p = [i.data => ones(15)])            # StackOverflowError
```

The other three red checks on this PR (`NeuralPDE.jl/NNPDE1`, `SciMLSensitivity.jl/Core8`, `StructuralIdentifiability.jl/Core`) all fail identically on the master-only https://github.com/SciML/ModelingToolkit.jl/pull/5105, so those are not this PR. A separate bisect for the `StructuralIdentifiability` master regression is running.

lib/ModelingToolkitBase/src/systems/problem_utils.jl
```diff
@@ -1630,6 +1633,183 @@ function (map::InitializationMap{true})(x)
...
+    else
+        push!(groups, Vector{SymbolicT}[ps])
+    end
+    flat_syms = SymbolicT[]
+    for group in groups, buffer in group
+        append!(flat_syms, buffer)
+    end
+    expr = build_explicit_observed_function(
+        initsys, Tuple(flat_syms);
+        expression = Val(true), compiler_options, kwargs...
```
AayushSabharwal: Relatively minor, but I think we should be able to just pass an options struct here and avoid `kwargs...`.

AayushSabharwal (approved):

lib/ModelingToolkitBase/src/systems/problem_utils.jl
ChrisRackauckas-Claude: Done in https://github.com/SciML/ModelingToolkit.jl/commit/16ef998c28. Both map builders now take a `GeneratedFunctionOptions` positionally and call the struct method of `build_explicit_observed_function` directly, so there is no `kwargs...` on either of them:

```julia
function _construct_fullspecialize_initializeprobpmap(
        sys::AbstractSystem, initsys::AbstractSystem, gen_opts::GeneratedFunctionOptions;
        p_constructor
    )
    ...
    expr = build_explicit_observed_function(initsys, Tuple(flat_syms), gen_opts)
```

The options come from the problem's own `opts.fn_opts.codegen` rather than being rebuilt:

```julia
function _fullspecialize_map_options(opts::SciMLProblemOptions)
    codegen = opts.fn_opts.codegen
    return GeneratedFunctionOptions(;
        expression = Val{true}, codegen.eval_expression, codegen.eval_module,
        compiler_options = opts.init_compiler_options,
        codegen_function_options = codegen.codegen
    )
end
```

This is a small behaviour fix as well as a cleanup, in two ways worth flagging:

- The old `kwargs...` path went through the keyword compat shim, which funnelled whatever was left in the bag into a **fresh** `Symbolics.CodegenFunctionOptions(; checkbounds, kwargs...)`. So the generated maps were not seeing the `checkbounds`/`cse`/... the rest of the problem was built with — only whatever happened to survive as loose keywords. Reusing `codegen.codegen` fixes that.
- `eval_expression` is now honoured. Previously the builders took `eval_module` but never received `eval_expression`, so `eval_or_rgf` always defaulted to `false` and the maps were unconditionally `RuntimeGeneratedFunction`s even with `eval_expression = true` — inconsistent with the default map path, which forwards it. Say the word if you would rather keep the maps always-RGF and I will pin it back.

`init_compiler_options` is still what the maps compile under; it is just read off `opts` in the helper now, so it is dropped from the destructuring in `maybe_build_initialization_problem` where it had become unused.

Local on `16ef998c28`, Julia 1.12.7:

```text
GROUP=Initialization   InitializationSystem Test | 706 pass | 12 broken | 718 total | 20m43.6s
                       Guess Propagation         |  11 pass |  11 total
                       Initial Values Test       |  65 pass |  65 total     (0 errors)
GROUP=Extensions       HomotopyContinuation 76/76, LabelledArrays 18/18, BifurcationKit 7/7,
                       Despecialized MTKParameters AD 2/2, Initialization maps AD 7/7
GROUP=FMI              FMI | 60 pass | 60 total | 15m35.4s
```

All three exited 0. Runic, `typos` over the diff, and `git diff --check` are clean.

PR #5051, opened 2026-08-31
Fix stale torn-system Jacobian sparsity

ChrisRackauckas-Claude:
## What changed

Fall back to symbolic Jacobian sparsity when a system's retained tearing graph no longer matches the final equation and unknown counts. This prevents `W_sparsity` from combining a stale 6×6 torn-graph pattern with a 5×5 mass matrix after compilation removes a redundant equation.

The regression test is a standalone ModelingToolkit system distilled from ProcessSimulator.jl's `simple_cstr` failure.

## Verification

Failing before the source fix, with the regression test present:

```text
Test Failed at test/reduction.jl:312
  Expression: size(ModelingToolkitBase.jacobian_sparsity(sys)) == (5, 5)
   Evaluated: (6, 6) == (5, 5)
```

Passing after the source fix:

```text
length(unknowns(compiled)) = 5
length(equations(compiled)) = 5
size(ModelingToolkitBase.jacobian_sparsity(compiled)) = (5, 5)
size(ModelingToolkitBase.W_sparsity(compiled)) = (5, 5)
```

Additional local checks:

```text
test/reduction.jl: exit 0
GROUP=QA: 40 passed, 40 total, 20m23.5s
runic --check src/systems/codegen.jl test/reduction.jl: exit 0
typos src/systems/codegen.jl test/reduction.jl: exit 0
git diff --check: exit 0
```

The complete `GROUP=InterfaceI` run did not finish within the one-hour local validation window; `test/reduction.jl`, which contains this change's regression test, was therefore also run directly. No docs build was run because this does not change public API or documentation. GPU, downstream, and allowed-to-fail jobs were not run.

Julia 1.12.6 emitted an internal type-inference warning while precompiling ModelingToolkit, but precompilation completed and the QA process exited 0.

The prerequisite clean-base QA repair is merged. This PR's only diff is the sparsity guard and its regression test.

Please ignore this PR until it has been reviewed by @ChrisRackauckas.

🤖 Generated with Codex CLI 0.151.0 (model: unknown; session: local session ID 01a04fa1-cfe0-7260-b416-72fe8a15d17d).

---

## Adoption (2026-09-23)

Rebased onto current `master` (`e167d971e0`); the original commit is kept with its authorship. One commit added on top: `3b7c5e1607` (comment on the guard + a second regression test).

### Is this covering another bug? Yes, for the ProcessSimulator case

The stale graph comes from ModelingToolkitTearing, not from MTK. Instrumenting `ModelingToolkitTearing.reorder_vars!` on the distilled test (ModelingToolkitTearing 1.20.6) shows:

```text
reorder_vars!: nsrcs=10 length(eq_ordering)=5 nsolved_eq=4 ndsts=13 length(var_ordering)=9 nsolved_var=4
  eq 1 not in eq_ordering: 0 ~ 1.0 - mapreduce(identity, add_sum, input_x(t))  incident vars: []  matched var: unassigned
  eq 2..5: solved into observed (matched vars 13, 10, 11, 12)
```

`0 ~ 1 - sum(input_x)` involves only inputs, so after `inputs_to_parameters!` it is parameter-only. It is unmatched and belongs to no SCC, so `generate_system_equations!` never emits it: it is neither solved nor in `eq_ordering`. `reorder_vars!` then calls `StateSelection.contract_variables(..., nsolved_eq, nsolved_var)`, which sizes the new graph as `nsrcs(graph) - nsolved_eq = 10 - 4 = 6`. Only 5 equations are kept, so the graph and `solvable_graph` get an empty sixth row. `eq_to_diff` is sized from `length(eq_ordering)` and has 5 entries, so the retained `SystemStructure` disagrees with itself:

```text
nsrcs(graph) = 6    length(eq_to_diff) = 5    length(equations(sys)) = 5
graph eq 6 => []
```

`torn_system_jacobian_sparsity` builds an `nsrcs(graph)`-square matrix from that graph, which gives the 6x6 result. ProcessSimulator's `flowsheet` has the same defect with 9 graph rows for 8 equations.

**Root-cause fix (outside this repo):** size the contracted graph by what is kept:

```julia
nelim_eq = nsrcs(graph) - length(eq_ordering)
nelim_var = ndsts(graph) - length(var_ordering)
```

It is on branch https://github.com/ChrisRackauckas-Claude/StateSelection.jl/tree/fix-reorder-vars-dropped-eqs (commit `a9df834`, based on JuliaComputing/StateSelection.jl `main` at `285667f`), with a ModelingToolkitTearing regression test. ModelingToolkitTearing lives in JuliaComputing/StateSelection.jl, so no upstream PR has been opened yet; that needs a maintainer decision.

### Why the guard is still worth keeping

The graph is only valid for the system as tearing produced it. `mtkcompile` runs `additional_passes` after tearing, and those can change equations without updating the graph. On master, a pass that adds `0 ~ y - x` to `D(x) ~ -x` gives a 1x1 `jacobian_sparsity` for a 2-equation system. `W_sparsity` then broadcasts that into a dense 2x2 with no error, so the pattern is wrong but nothing is reported. The count check catches both cases. It cannot detect a stale graph that happens to have the right counts, so it is a consistency check, not proof that the graph is correct. The new test covers the `additional_passes` case.

With the upstream fix released, the ProcessSimulator case no longer needs the guard. The guard is still needed for the `additional_passes` case.

### Evidence

Distilled ProcessSimulator test (`test/reduction.jl`, from the original commit):

```text
# master src
PR 5051 regression: Test Failed at pr_test_body.jl:12
  Expression: size(ModelingToolkitBase.jacobian_sparsity(sys)) == (5, 5)
   Evaluated: (6, 6) == (5, 5)
PR 5051 regression: Error During Test at pr_test_body.jl:13
  DimensionMismatch: arrays could not be broadcast to a common size: a has axes Base.OneTo(6) and b has axes Base.OneTo(5)
Test Summary:      | Pass  Fail  Error  Total
PR 5051 regression |    1     1      1      3

# this branch
PR 5051 regression |    3      3

# master src + upstream reorder_vars! fix, no MTK guard
PR 5051 regression (upstream reorder_vars! fix, no MTK guard) |    3      3
```

New `additional_passes` test:

```text
# master torn_system_jacobian_sparsity
   Evaluated: sparse([1], [1], Bool[1], 1, 1) == sparse([1, 2, 2], [1, 1, 2], [1, 1, 1], 2, 2)
additional_passes (master torn_system_jacobian_sparsity) |    1     1      2
# this branch
additional_passes (branch torn_system_jacobian_sparsity) |    2      2
```

Full `test/reduction.jl` on this branch: `reduction.jl | 150 150 2m32.5s`.

Upstream ModelingToolkitTearing test (from the fork branch):

```text
# without the reorder_vars! change
  Expression: BipartiteGraphs.nsrcs(structure.graph) == length(equations(sys))
   Evaluated: 6 == 5
  Expression: BipartiteGraphs.nsrcs(structure.solvable_graph) == length(equations(sys))
   Evaluated: 6 == 5
Torn graph drops equations that are neither solved nor emitted |    2     2      1      5
# with it
Torn graph drops equations that are neither solved nor emitted |    5      5
# full ModelingToolkitTearing suite (CI recipe: develop StateSelection, Pkg.test())
     Testing ModelingToolkitTearing tests passed
```

ProcessSimulator `test/simple_cstr.jl`, using ModelingToolkit dev'd from this branch in a scratch env:

```text
simple_cstr   |    3      3  5m42.1s
```

With `ODEProblem(...; sparse = true)` on the same `flowsheet`:

```text
== master torn sparsity         graph rows = 9, equations = 8   W_sparsity: DimensionMismatch (9 vs 8)
== master + upstream fix        graph rows = 8, equations = 8   W_sparsity: (8, 8)   sparse=true retcode: Success
== this branch                  graph rows = 9, equations = 8   W_sparsity: (8, 8)   sparse=true retcode: Success
```

**ProcessSimulator may already be green without this PR.** Since ModelingToolkitBase 1.73.0 (`0ed8046c2e`, #5148), `ODEFunction` computes `W_sparsity` only when `sparse` or `sparsity` is requested. `simple_cstr.jl` requests neither. A fresh environment with only registered packages (ModelingToolkit 11.45.1, ModelingToolkitBase 1.77.0, ModelingToolkitTearing 1.20.6) passes: `simple_cstr | 3 3`. The red runs linked in the adoption request are from 2026-09-13 on ModelingToolkitBase 1.71.3 and have not been rerun since. The bug is still there for `sparse = true` users.

Also run: `runic --check` on both changed files: clean. `typos`: clean. `git diff --check`: clean. `GROUP=QA`: 39/40 (80 min). The one failure is `No unapproved public reexports` for `[:ap_var, :canonical_constraints, :constraint_residual]`. Those names come from `master` (`@public canonical_constraints, constraint_residual` was added in ModelingToolkitBase by #5189), and this PR touches no exports.

### Not verified

- ProcessSimulator's Downgrade job, which resolves to minimum compat versions, was not reproduced.
- I did not bisect when ModelingToolkitTearing started dropping the parameter-only equation, or whether it always has.
- `GROUP=InterfaceI` was not run in full. Only `test/reduction.jl`, which is part of it, was run.
- No docs build, because there are no docstring or public API changes. GPU, downstream, and allowed-to-fail jobs were not run.

## Risk assessment
- **Risk:** low
- **Blast radius:** `torn_system_jacobian_sparsity`, an internal method reached through `jacobian_sparsity`/`W_sparsity` for compiled systems. When the retained graph disagrees with the equation/unknown counts, it now returns `nothing` and the symbolic sparsity path is used instead. That path is slower to compute but correct. It is not a public API change.
- **Evidence:** failing-before/passing-after for both regression tests (above); `test/reduction.jl` 150/150; ProcessSimulator `simple_cstr` 3/3 with this branch; QA 39/40, with the one failure coming from master and unrelated to this PR.
- **Independent review:** pending
- **Merge:** needs human review. It depends on deciding between upstreaming the ModelingToolkitTearing fix and keeping the guard as the only fix.

🤖 Adoption by [Claude Code](https://claude.com/claude-code) (model: claude-opus-5-5[1m])

https://claude.ai/code/session_01J87aynrgK6d2bYwoKj4YAx

### DIFF
--- a/src/systems/codegen.jl
+++ b/src/systems/codegen.jl
@@ -5,6 +5,7 @@ function MTKBase.torn_system_jacobian_sparsity(sys::System)
     @unpack graph, var_to_diff = structure
 
     neqs = nsrcs(graph)
+    neqs == length(full_equations(sys)) == length(unknowns(sys)) || return nothing
     nsts = ndsts(graph)
     states_idxs = findall(!Base.Fix1(StateSelection.isdervar, structure), 1:nsts)
     var2idx = StructuralTransformations.uneven_invmap(nsts, states_idxs)
--- a/test/reduction.jl
+++ b/test/reduction.jl
@@ -298,6 +298,20 @@ Js = ModelingToolkitBase.jacobian_sparsity(sys)
 @test size(Js) == (3, 3)
 @test Js == Diagonal([0, 1, 1])
 
+@variables input_x(t)[1:4] output_x(t)[1:4] n(t)[1:4] algebraic(t)
+eqs = [
+    sum(input_x) ~ 1.0
+    sum(output_x) ~ 1.0
+    [output_x[i] ~ input_x[i] for i in 1:3]...
+    [D(n[i]) ~ output_x[i] for i in 1:4]...
+    0 ~ algebraic^2 - n[1]
+]
+@named model = System(eqs, t)
+sys = mtkcompile(model; inputs = collect(input_x))
+@test length(unknowns(sys)) == length(equations(sys)) == 5
+@test size(ModelingToolkitBase.jacobian_sparsity(sys)) == (5, 5)
+@test size(ModelingToolkitBase.W_sparsity(sys)) == (5, 5)
+
 # MWE for #1722
 vars = @variables a(t) w(t) phi(t)
 eqs = [


ChrisRackauckas:
This seems like it's covering some other bug?

AayushSabharwal (approved):

ChrisRackauckas-Claude:
> This seems like it's covering some other bug?

Yes. For the ProcessSimulator case the bug is in ModelingToolkitTearing, in `reorder_vars!`. `0 ~ 1 - sum(input_x)` becomes parameter-only after `inputs_to_parameters!`. It has no matched variable and belongs to no SCC, so `generate_system_equations!` never emits it. `reorder_vars!` still sizes the contracted graph as `nsrcs(graph) - nsolved_eq`, which here is 10 − 4 = 6, but only 5 equations are kept (`length(eq_ordering) = 5`). The retained `graph`/`solvable_graph` therefore get an empty sixth row, while `eq_to_diff` and the system have 5. ProcessSimulator's `flowsheet` shows the same defect with 9 graph rows for 8 equations.

Root-cause fix: size the graph by `nsrcs(graph) - length(eq_ordering)` and `ndsts(graph) - length(var_ordering)`. It is on https://github.com/ChrisRackauckas-Claude/StateSelection.jl/tree/fix-reorder-vars-dropped-eqs with a regression test that fails before the change (`nsrcs(graph) == 6` vs 5 equations) and passes after. The full ModelingToolkitTearing suite passes with it. With only that change and no MTK guard, the distilled test passes 3/3, and ProcessSimulator with `sparse = true` gets an 8x8 `W_sparsity`. That repo is JuliaComputing/StateSelection.jl, so I have not opened a PR there.

The guard still has its own job. `additional_passes` run after tearing, and on master a pass that adds an equation gives a 1x1 `jacobian_sparsity` for a 2-equation system, which `W_sparsity` silently broadcasts to a dense 2x2. I added that as a second regression test.

Also, current registered MTK (ModelingToolkitBase ≥ 1.73.0) no longer computes `W_sparsity` unless `sparse`/`sparsity` is requested. As a result ProcessSimulator's `simple_cstr` passes today without this PR (checked locally with MTK 11.45.1). The failure is still there for `sparse = true`. Details and before/after output are in the PR body.

🤖 Generated with [Claude Code](https://claude.com/claude-code) (model: claude-opus-5-5[1m])
https://claude.ai/code/session_01J87aynrgK6d2bYwoKj4YAx

PR #5060, opened 2026-09-03
Fixes to better support 32-bit builds

Drvi:
## Checklist

## Additional context

What I found that looked 64-bit specific. With this patch, I was able to precompile MTK on win32 CI (branch with a bunch of other fixes to other packages).

### DIFF
--- a/src/structural_transformation/symbolics_tearing.jl
+++ b/src/structural_transformation/symbolics_tearing.jl
@@ -26,7 +26,7 @@ end
 
 function safe_isinteger(@nospecialize(x::Number))
     if x isa Int64
-        return true
+        return typemin(Int) <= x <= typemax(Int)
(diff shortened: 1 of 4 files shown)


src/structural_transformation/symbolics_tearing.jl
```diff
@@ -26,7 +26,7 @@ end
 
 function safe_isinteger(@nospecialize(x::Number))
     if x isa Int64
-        return true
+        return typemin(Int) <= x <= typemax(Int)
```
ChrisRackauckas, line 29: what's this one for?

Drvi: This `safe_isinteger` effectively asserts that `x` fits into `Int` which might be smaller than `Int64`

Issue #5079, opened 2026-09-06
[bot-unreviewed] Implicit affect solve ignores the ODE solver's tolerance, so a solvable callback throws UnsolvableCallbackError


ChrisRackauckas:
> 🚧 **UNREVIEWED — awaiting review by @ChrisRackauckas.** Filed by an AI agent running as @chrisrackauckas; Chris has not seen this finding. Do not read it as his assessment, and please don't act on it until he confirms it's real and correctly filed.
> Harness: Claude Code 2.1.260 · Model: claude-opus-5[1m]
> Conversation: https://claude.ai/code/session_01UBTNFZ4A8Ciw27TmVHkpRD
> Triage accordingly: the reproducer below was machine-generated and may be wrong or incomplete.

## Summary

A symbolic affect on a DAE compiles to an `ImplicitAffect`, whose nonlinear solve is run at
**NonlinearSolve's default Float64 `abstol = 3e-13`**, with no relation to the tolerance the ODE
is being integrated at. `AffectSystem` appends all of `alg_equations(sys)` and `observed(sys)` to
the affect equations, so any DAE whose algebraic/observed residuals cannot reach `3e-13` in
floating point — i.e. any model with terms of magnitude ≳ 1e3, since the residual floor is
`eps * magnitude` — gets `UnsolvableCallbackError` thrown at its first event even though the
callback is perfectly solvable at the tolerance the user asked for.

The path is `ModelingToolkitBase/src/systems/callbacks.jl`, `(ia::ImplicitAffect)(integ)`:

```julia
affprob = remake(ia.affprob, tspan = (integ.t, integ.t))
affsol = init(affprob, IDSolve())
(check_error(affsol) === ReturnCode.InitialFailure) &&
    throw(UnsolvableCallbackError(all_equations(ia.aff)))
```

`IDSolve()` builds its nonlinear cache with `init(prob, alg.nlsolve)`
(`ImplicitDiscreteSolve/src/cache.jl`) and never sees a tolerance, so `NewtonRaphson` terminates
against `NonlinearSolveBase.get_tolerance(nothing, Float64) == 3e-13`. Nothing on this path
carries `integrator.opts.abstol` / `.reltol`.

Found on a battery model (concentrations O(1e4), `JuliaSimBatteries.jl`): the very first switch of
an ordinary charge protocol throws, while the same DAE integrates fine to `abstol = 1e-8`.

## MWE

[...]

PR #5080, opened 2026-09-06
[bot] Solve implicit affects at the ODE solver's tolerance

ChrisRackauckas:
> 🚧 **Filed by an AI agent running as @chrisrackauckas — Chris has not reviewed this PR.** Please ignore it until reviewed by @ChrisRackauckas.
> Harness: Claude Code 2.1.260 · Model: claude-opus-5[1m]
> Conversation: https://claude.ai/code/session_01UBTNFZ4A8Ciw27TmVHkpRD

## What and why

`(ia::ImplicitAffect)(integ)` solves the affect with `init(affprob, IDSolve())`, i.e. at
NonlinearSolve's default Float64 `abstol = 3e-13`, unrelated to the tolerance the ODE is being
integrated at. Since `AffectSystem` appends every `alg_equations(sys)` and `observed(sys)` to the
affect equations, any DAE whose residuals cannot reach `3e-13` in floating point — the floor is
`eps * magnitude`, so anything with terms ≳ 1e3 — throws `UnsolvableCallbackError` at its first
event even though the callback is solvable to the accuracy actually requested. Hit on a battery
model with O(1e4) concentrations at the first switch of an ordinary charge protocol.

This passes the triggering integrator's tolerances into the affect solve. Per @ChrisRackauckas:
*"It should go to the tolerance of the ODE solver."*

`affect_tolerance` is defensive because `integ` is not always an OrdinaryDiffEq integrator: no
`opts`, or no such field on `opts`, gives `nothing`, which is exactly the current behaviour
(NonlinearSolve's default). A per-component tolerance array is reduced to its tightest entry —
it is sized for the parent system's unknowns, not the affect's.

`check_error(affsol) === ReturnCode.InitialFailure` is left alone: with the tolerance fixed, a
failure there is a real failure, and the existing "impossible affect" test in `symbolic_events.jl`
(`x ~ Pre(x) + 2` against `x^2 + y^2 ~ 1`) still throws, at any tolerance.

**Blocked on a release**: the tolerance is only honoured by `ImplicitDiscreteSolve` ≥ 2.3.0
(https://github.com/SciML/OrdinaryDiffEq.jl/pull/4468 — `IDSolve` currently drops the tolerances
`init` was given). The `[compat]` bound here is bumped to `2.3`, so this cannot resolve until that
is registered.

## Verification

Issue with the MWE: https://github.com/SciML/ModelingToolkit.jl/issues/5079. Run on aarch64
Linux, Julia 1.10.12, with `lib/ModelingToolkitBase` and the ImplicitDiscreteSolve branch `dev`ed
into a scratch environment.

**Before** (`git stash push lib/ModelingToolkitBase/src/systems/callbacks.jl`, new testset run
standalone, ImplicitDiscreteSolve fix already in place — so this isolates the MTK half):

```
Implicit affect solves at the ODE solver's tolerance: Error During Test at mini_test.jl:7
  The callback defined by the following equations:

  0 ~ -g(t) + 1.0e7(-x(t) + y(t)^2)
  x(t) ~ Pre(x(t))
  w(t) ~ Pre(w(t))
  g(t) ~ 0.1 + Pre(g(t))

  is not solvable. Please check that the algebraic equations and affect equations are correct, ...

Test Summary:                                        | Error  Total   Time
Implicit affect solves at the ODE solver's tolerance |     1      1  44.3s
```

**After** (same testset):

```
Test Summary:                                        | Pass  Total   Time
Implicit affect solves at the ODE solver's tolerance |    3      3  51.4s
```

**Whole file** (`symbolic_events.jl`, the way `InterfaceI` runs it — `using ModelingToolkit;
import ModelingToolkitBase; include(...)`):

```
Test Summary:       | Pass  Broken  Total      Time
Symbolic Event Test |  382       2    384  15m49.7s
```

0 failures, 0 errors; the 2 broken are pre-existing. In particular the neighbouring
`Implicit affects with Pre` testset (11/11), which includes the genuinely unsolvable affect
`x ~ Pre(x) + 2` against `x^2 + y^2 ~ 1`, still passes — a real failure is still reported as one.

End to end, the issue's MWE now prints `solve returned Success, g = [0.4785, 0.5785]` instead of
throwing, and the state saved right after the affect satisfies the algebraic equation with
`|residual| = 3.2e-9` (its `Float64` floor is 1.2e-9), against the `981` the surrounding
integration leaves on this deliberately badly scaled equation.

Formatting: Runic clean on both changed files. `typos` is not installed on this machine, so the
spell check was not run locally.

## What I did not verify

- Only `symbolic_events.jl` was run, not the whole `InterfaceI` group (too large for the box).
- aarch64 Linux (Jetson AGX Orin), Julia 1.10.12 only. Not run on x86 or 1.12.
- The downstream battery model that motivated this was not re-run against this branch.

---
🤖 Posted by an AI agent — harness: Claude Code 2.1.260 · model: claude-opus-5[1m]
Conversation: https://claude.ai/code/session_01UBTNFZ4A8Ciw27TmVHkpRD

https://claude.ai/code/session_01UBTNFZ4A8Ciw27TmVHkpRD

### DIFF
--- a/lib/ModelingToolkitBase/Project.toml
+++ b/lib/ModelingToolkitBase/Project.toml
@@ -1,7 +1,7 @@
 name = "ModelingToolkitBase"
 uuid = "7771a370-6774-4173-bd38-47e70ca0b839"
 authors = ["Yingbo Ma <mayingbo5@gmail.com>", "Chris Rackauckas <accounts@chrisrackauckas.com> and contributors"]
-version = "1.68.2"
+version = "1.68.3"
 
 [deps]
 AbstractTrees = "1520ce14-60c1-5f80-bbc7-55ef81b5835c"
@@ -132,7 +132,7 @@ ForwardDiff = "1"
 FunctionWrappers = "1.1.2"
 FunctionWrappersWrappers = "1"
 Graphs = "1.5.2"
-ImplicitDiscreteSolve = "2.1.2"
+ImplicitDiscreteSolve = "2.3"
 InfiniteOpt = "0.6"
 InteractiveUtils = "1"
 IntervalSets = "0.7"
--- a/lib/ModelingToolkitBase/src/systems/callbacks.jl
+++ b/lib/ModelingToolkitBase/src/systems/callbacks.jl
@@ -910,12 +910,40 @@ struct ImplicitAffect{DVS, PS, AFFSYS, AFF, UG, AG, AUS, APS, US, PST, UGT, PG, 
     affprob::PROB
 end
 
+"""
+    affect_tolerance(integ, name::Symbol)
+
+The `abstol` or `reltol` an implicit affect's nonlinear solve should be run at, read off the
+integrator that triggered the callback. `nothing` (the nonlinear solver's own default) if the
+integrator does not expose that tolerance.
+
+The affect equations are the parent system's algebraic and observed equations, so their
+residuals floor at the same roundoff level the ODE solver already lives with; holding the
+affect solve to a tighter tolerance than the integration itself makes a solvable callback
+report as unsolvable. A per-component tolerance array is reduced to its tightest entry: it
+is sized for the parent system's unknowns, not the affect's.
+"""
+function affect_tolerance(integ, name::Symbol)
+    hasproperty(integ, :opts) || return nothing
+    opts = integ.opts
+    hasproperty(opts, name) || return nothing
+    tol = getproperty(opts, name)
+    # `false` is OrdinaryDiffEqCore's "no tolerance supplied".
+    tol isa Bool && return nothing
+    tol isa Number && return tol
+    tol isa AbstractArray && return isempty(tol) ? nothing : minimum(tol)
+    return nothing
+end
+
 function (ia::ImplicitAffect)(integ)
     ia.affu_setter!(ia.affprob, ia.affu_getter(integ))
     ia.affp_setter!(ia.affprob, ia.affp_getter(integ))
     # remake only updates tspan; result is a transient local, not stored back to struct
     affprob = remake(ia.affprob, tspan = (integ.t, integ.t))
-    affsol = init(affprob, IDSolve())
+    affsol = init(
+        affprob, IDSolve(); abstol = affect_tolerance(integ, :abstol),
+        reltol = affect_tolerance(integ, :reltol)
+    )
     (check_error(affsol) === ReturnCode.InitialFailure) &&
         throw(UnsolvableCallbackError(all_equations(ia.aff)))
     ia.u_setter!(integ, ia.u_getter(affsol))
--- a/lib/ModelingToolkitBase/src/systems/callbacks.jl
+++ b/lib/ModelingToolkitBase/src/systems/callbacks.jl
@@ -914,25 +914,33 @@ end
     affect_tolerance(integ, name::Symbol)
 
 The `abstol` or `reltol` an implicit affect's nonlinear solve should be run at, read off the
-integrator that triggered the callback. `nothing` (the nonlinear solver's own default) if the
-integrator does not expose that tolerance.
+integrator that triggered the callback.
 
 The affect equations are the parent system's algebraic and observed equations, so their
 residuals floor at the same roundoff level the ODE solver already lives with; holding the
 affect solve to a tighter tolerance than the integration itself makes a solvable callback
-report as unsolvable. A per-component tolerance array is reduced to its tightest entry: it
-is sized for the parent system's unknowns, not the affect's.
+report as unsolvable.
+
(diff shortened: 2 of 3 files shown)


ChrisRackauckas:
> 🤖 **Automated comment from an AI agent running as @chrisrackauckas — not written or reviewed by Chris.**

CI status on the first push, read rather than assumed: **39 jobs red, all from the same cause** —

```
ERROR: LoadError: Unsatisfiable requirements detected for package ImplicitDiscreteSolve [3263718b]
```

i.e. the `ImplicitDiscreteSolve = "2.3"` bound in `lib/ModelingToolkitBase/Project.toml`, which is
not registered yet. It becomes resolvable once https://github.com/SciML/OrdinaryDiffEq.jl/pull/4468
merges and 2.3.0 is tagged; nothing else can be learned from this run. Runic, runic-suggestions and
the spell check all pass.

---
🤖 Posted by an AI agent — harness: Claude Code 2.1.260 · model: claude-opus-5[1m]
Conversation: https://claude.ai/code/session_01UBTNFZ4A8Ciw27TmVHkpRD

ChrisRackauckas:
> 🤖 **Automated comment from an AI agent running as @chrisrackauckas — not written or reviewed by Chris.**

Pushed 2db9289420a10c70c2dc1ed20f2562188396f4af in response to review.

**Dropped**: both `hasproperty` guards. Every SciMLBase integrator has `.opts` with `abstol`/`reltol`, there is no official accessor to prefer over `getproperty`, and there is no integrator in hand that needed them. `affect_tolerance` is now four lines:

```julia
function affect_tolerance(integ, name::Symbol)
    tol = getproperty(integ.opts, name)
    tol isa Bool && return nothing
    tol isa AbstractArray && return isempty(tol) ? nothing : minimum(tol)
    return tol
end
```

**Kept, with the reasons now in the docstring and checked rather than assumed:**

*`Bool`* — `OrdinaryDiffEqCore/src/solve.jl:377-378` and `:389-390` (v4.16.0) store `false` for a tolerance not supplied under a discrete problem. `Bool <: Number`, and `NonlinearSolveBase.get_tolerance(u, false, Float64)` returns `0.0` (measured), i.e. a zero tolerance. Measured end to end on a solve whose residual floors at `eps`:

```
abstol = nothing  -> Success          u = 1.4142135623730951  resid = 4.44e-16
abstol = false    -> Success          u = 1.4142135623730951  resid = 4.44e-16
abstol = 0.0      -> InitialFailure   u = 1.4                 resid = -0.04
abstol = 3.0e-13  -> Success          u = 1.4142135623730951  resid = 4.44e-16
```

One correction to the record, since it changes how load-bearing this branch is: `false` reaching `IDSolve` is *already* mapped back to the default by SciML/OrdinaryDiffEq.jl#4468 (`nlsolve_tolerance(::Bool) = nothing`) — that is why the `false` row above passes. Since the affect solve always goes through `IDSolve` and this PR's compat floor is `ImplicitDiscreteSolve = "2.3"`, the branch here is the outer of two guards, not the one doing the work, and **no end-to-end test can discriminate it** — an affect firing under a discrete-problem integrator passes with the branch removed. I did not add a test that would pass either way and call it a regression test. What I added instead is a direct unit test of the two translations, which does discriminate: with both branches removed it fails 4 of its 5 assertions (`false === nothing`, twice, and the two array cases). The docstring now says explicitly that it is the outer guard.

(Attempting the end-to-end discrete construction also runs into something unrelated: a `Pre(g)` on a discrete parameter inside a shift-system affect fails in `mtkcompile` with `ArgumentError: gₜ₋₁(t) is present in the system but gₜ₋₁(t) is not an unknown`. Not chased here.)

*Array* — the nonlinear solve compares a scalar norm against the tolerance (`NonlinearSolveBase/src/termination_conditions.jl:246-248`, v2.49.3: `objective = apply_norm(mode.internalnorm, du); criteria = abstol`), so an array is not a supported input; and the ODE integrator's per-component array is sized for the parent system's unknowns while the affect's nonlinear problem has its own, so it would be the wrong length even if it were. `minimum` is flagged in the docstring as a deliberate choice — the tightest component, so the affect is never solved looser than any part of the integration.

**Tests** — `symbolic_events.jl`, run the way `InterfaceI` runs it:

```
Test Summary:       | Pass  Broken  Total      Time
Symbolic Event Test |  387       2    389  22m43.3s
```

0 failures, 0 errors; the 2 broken are pre-existing. Runic clean on both changed files. Nothing else in the PR was touched.

---
🤖 Posted by an AI agent — harness: Claude Code 2.1.260 · model: claude-opus-5[1m]
Conversation: https://claude.ai/code/session_01UBTNFZ4A8Ciw27TmVHkpRD

PR #5083, opened 2026-09-06
Add analytical parameter jacobian code generation

AJ0070:
MTK can already generate analytical jacobians with respect to unknowns (`jac = true`) and inputs (`calculate_control_jacobian`), but nothing with respect to parameters. SciMLBase already defines a `paramjac` field on `ODEFunction`, so this just populates it.

Adds `calculate_paramjac` and `generate_paramjac`, modelled on the existing control jacobian, plus a `paramjac = false` keyword on `ODEFunction`.

Column `j` of `pJ` is the derivative with respect to entry `j` of `SciMLStructures.canonicalize(SciMLStructures.Tunable(), p)[1]`, which is what SciMLSensitivity consumes. `tunable_parameters` is not usable as the column basis despite its docstring: it is a metadata filter, so it also returns `Initial(...)`, `::Int`, `FnType` and discrete parameters. The columns come from `reorder_parameters(sys, ps; flatten = false)[1]`, the same call `complete` uses to lay out `sys.ps`.

Notes:
- `paramjac` defaults to `false`, so nothing changes unless you opt in.
- `sparse` works on `calculate_paramjac`/`generate_paramjac` but is not forwarded from `ODEFunction`, since SciMLBase has no `paramjac_prototype`.
- A system with no tunable parameters throws instead of returning an empty function, because dense out of place codegen of an `n x 0` array fails inside SymbolicUtils.

Tests cover the column ordering, agreement with central differences, sparse and dense, in place and out of place, time independent and non split systems, and the `ODEFunction` wiring. Adjoint gradients through SciMLSensitivity match ForwardDiff to 3e-11.

AI Disclosure: Used Opus 5


lib/ModelingToolkitBase/src/systems/codegen.jl
```diff
@@ -1237,6 +1237,133 @@ function generate_control_jacobian(
...
+        sys::AbstractSystem;
+        sparse = false, simplify = false, ps = paramjac_parameters(sys)
+    )
+    check_symbolic_ad_allowed(sys)
+    rhs = SymbolicT[eq.rhs for eq in full_equations(sys)]
+
+    if sparse
+        return sparsejacobian(rhs, ps; simplify)
+    else
+        return jacobian(rhs, ps; simplify)
```
oxinabox, line 1333: is this really all there is to it?

AJ0070: Mostly yes: `full_equations` already inlines observed and torn equations, and `check_lhs` guarantees every lhs is `0` or `D(x)`, so `eq.rhs` is the `f` the solver integrates. Mass matrices are constant, so there is no `dM/dp` term.

Two things it did miss, both now fixed: a delay parameter got a silently zero column, since `Symbolics.jacobian` treats `x(t - tau)` as opaque, so `calculate_paramjac` throws for DDEs; and the sparse in-place function filled a mismatched buffer in the wrong cells, so it now asserts the pattern like `generate_jacobian` does.

ChrisRackauckas: yup the symbolic machinery takes care of most of it.

PR #5098, opened 2026-09-09
Sundials 6.7.1 + MCI 0.3 source for x86 InterfaceI

ChrisRackauckas-Claude:
## Summary
- Require **Sundials 6.7.1** (registered NVector Int32 fix).
- URL-source MonteCarloIntegration master (0.3) so BVDE→Integrals does not pull LatticeRules 0.0.1 on i686.
- Drop once MCI 0.3.0 is in General ([ranjanan#29](https://github.com/ranjanan/MonteCarloIntegration.jl/issues/29)).

## Test plan
- [ ] x86 InterfaceI green (or only failures also on x64)

PR #5073, opened 2026-09-05
Recognize copied symbolic missing bindings

ChrisRackauckas-Claude:
## What changed

Recognize symbolic missing bindings by `isequal`, not object identity. Deep-copying a problem can copy the symbolic constant wrapping `missing`; initialization then incorrectly treats a solvable parameter as a bound parameter and omits it from the initialization unknowns. This breaks parameter remakes used by ensembles, including constant states eliminated to solvable parameters.

The regression exercises both the original problem and its deep copy with the same parameter remake. No new public API or dependencies.

## Verification

Julia 1.11.9, local environments with identical dependency versions and only the ModelingToolkit/ModelingToolkitBase source paths changed:

```sh
GROUP=Initialization julia --project=EMA-deepcopy-main-env -e 'using Pkg; Pkg.test("ModelingToolkitBase")'
```

Unfixed production source at c4c04fe910a1d0509eb41395b8d6b40dc9f25068, with the new test added:

```text
Remake with copied missing parameter bindings: Error During Test
ArgumentError: Symbol N is not present in the system.
InitializationSystem Test | 663 pass, 1 error, 12 existing broken
Remake with copied missing parameter bindings | 2 pass, 1 error
```

With the fix before rebasing:

```text
Guess Propagation | 11/11
InitializationSystem Test | 665 pass, 12 existing broken
Initial Values Test | 65/65
Testing ModelingToolkitBase tests passed
```

`GROUP=QA` passed JET 54/54 and Aqua 21/21. Runic, `git diff --check`, and `typos` over the diff excluding Git index-hash metadata passed. No tests were skipped or weakened by this change.

After rebasing onto upstream 2b4f11a, the complete native Initialization group passed again:

```text
Guess Propagation | 11/11
InitializationSystem Test | 667 pass, 12 existing broken
Initial Values Test | 65/65
Testing ModelingToolkitBase tests passed
```

The corresponding QA rerun passed JET 54/54 and Aqua 21/21, exit 0. Upstream subsequently advanced to e76112cc0; the refreshed branch is 2e20892791e27e88bce578cbeff33395d3faff04.

Latest refresh verification at 2e20892791e27e88bce578cbeff33395d3faff04, on upstream e76112cc0:

```sh
GROUP=Initialization julia --project=EMA-deepcopy-current-env -e 'using Pkg; Pkg.test("ModelingToolkitBase")'
[...]

PR #5100, opened 2026-09-09
Preserve array differential equations with `mtkcompile(sys; scalarize_arrays = false)`

### LINKED ISSUE #5093: DAEProblem from a system with array-form differential equations fails with MethodError: no method matching vec(::Bool)
> 🚧 **UNREVIEWED — awaiting review by @ChrisRackauckas.** Filed by an AI agent running as @chrisrackauckas; Chris has not seen this finding. Do not read it as his assessment, and please don't act on it until he confirms it's real and correctly filed.
> Harness: Claude Code 2.1.263 · Model: claude-opus-5[1m]
> Session: https://claude.ai/code/session_01CTUg9M7Zn2HdFbFREPz2hi
> Triage accordingly: the reproducer below was machine-generated and may be wrong or incomplete. Do not read this as Chris's own assessment.

Building a `DAEProblem` from a system whose differential equations are in **array form**
(the shape a MethodOfLines.jl discretization produces: scalar unknowns, array-valued
equations relating them) dies inside the initialization system with an internal
`MethodError`.

## Reproducer

```julia
using ModelingToolkit, SciMLBase

@variables u(t)[1:3]
# The unknowns are the scalar elements; the equation relating them is array-valued.
sys = complete(System([D(u) ~ -u], t, ModelingToolkit.unwrap.(collect(u)), [];
    initial_conditions = Dict(u => [1.0, 2.0, 3.0]), name = :m))

DAEProblem(sys, Dict(), (0.0, 1.0))
```

```
ERROR: MethodError: no method matching vec(::Bool)
The function `vec` exists, but no method is defined for this combination of argument types.

Stacktrace:
  [1] scalar_index(val::Any, idx::Int64)
    @ SymbolicUtils ~/.julia/packages/SymbolicUtils/c9cTZ/src/symbolic_ops/getindex.jl:214
  [2] __stable_getindex(arr::BasicSymbolic{SymReal}, sidxs::SymbolicUtils.StableIndex{Int64})
    @ SymbolicUtils ~/.julia/packages/SymbolicUtils/c9cTZ/src/symbolic_ops/getindex.jl:248
  ...
 [16] discover_maybe_zeros(sys::System)
    @ ModelingToolkitBase .../src/systems/diffeqs/basic_transformations.jl:1328
 [17] __mtkcompile(sys::System; ...)
 [23] mtkcompile_initialization_system(isys::System, sys::System; ...)
    @ ModelingToolkitBase .../src/problems/initializationproblem.jl:203
 [24] InitializationProblem{true, SciMLBase.AutoDespecialize}(...)
 [25] maybe_build_initialization_problem(...)
 [29] (DAEProblem{true, SciMLBase.AutoDespecialize})(sys::System, op, tspan; ...)
```

`build_initializeprob = false` avoids it, which is what makes this diagnosable at all.

## Cause

The initialization system ends up with an **array-shaped key mapped to a scalar value**:

```julia
julia> isys = ModelingToolkit.ModelingToolkitBase.generate_initializesystem(sys);

julia> ModelingToolkit.ModelingToolkitBase.initial_conditions(isys)
AtomicArrayDict(uˍt(t) => false, u(t) => [1.0, 2.0, 3.0])
```

`uˍt(t)` is the 3-element dummy derivative, and its value is the scalar `false`.

It gets there in `generate_initializesystem_timevarying`
(`lib/ModelingToolkitBase/src/systems/nonlinear/initializesystem.jl`). `dd_guess_sym =
BSImpl.Const{VartypeT}(default_dd_guess)` is a scalar (`default_dd_guess = Bool(0)`), and it
is written with

```julia
write_possibly_indexed_array!(guesses, ttk, dd_guess_sym, COMMON_NOTHING)
```

where `ttk = default_toterm(eq.lhs)`. When the differential equation is scalarized, `ttk` is
scalar and this is correct. When the equation is in array form, `ttk` is the *whole array*
`uˍt(t)`, and `write_possibly_indexed_array!` (`src/atomic_array_dict.jl:118`) only builds a
shaped buffer for an **indexed** key — `split_indexed_var(k)` returns `isarr = false` for a
whole-array symbolic, so it takes the `else` branch and stores the scalar as-is.

Later `discover_maybe_zeros` (`basic_transformations.jl:1328`) reads that map elementwise:

```julia
if any(SU._iszero ∘ subber ∘ Base.Fix1(getindex, k), SU.stable_eachindex(k))
```

`subber` resolves `uˍt(t)[i]` through `get_possibly_indexed`, which indexes the stored value,
so `scalar_index(false, 1)` calls `vec(::Bool)`.

Two plausible fixes, but the choice is a maintainer's call:

1. Broadcast in `write_possibly_indexed_array!`: in the non-indexed branch, when `k` is
   array-shaped and `v` is not, store `Const(fill(v, size(k)))` the way the indexed branch
   already does. Fixes every caller.
2. Shape `dd_guess_sym` per key at the call site in `initializesystem.jl`.

## Wider context, in case it changes the priority

This is what stops a `DAEProblem` from being built from an un-`mtkcompile`d array-equation
system through ModelingToolkit's own initialization. MethodOfLines.jl v1.3.1 works around it
by passing `build_initializeprob = false` and `initializealg = BrownFullBasicInit()`
(`lib/../src/dae_discretization.jl`), which costs it the ability to honour initialization
equations — it carries a `brown_init_offenders` predicate to refuse the path when that would
matter. A related report, filed separately, is that the same shape reaches
`generate_initializesystem` from another direction and produces
`Encountered operator Differential(T, 1) which has different independent variable than the
one used in the system nothing` instead — I have not established whether that is the same
root cause, so it is not bundled here.

## Version info

```
ModelingToolkit v11.41.0
ModelingToolkitBase v1.69.0
SymbolicUtils v4.46.1
Symbolics v7.39.0
SciMLBase v3.53.1
```

```
Julia Version 1.12.6
Commit 15346901f00 (2026-04-09 19:20 UTC)
Build Info:
  Official https://julialang.org release
Platform Info:
  OS: Linux (aarch64-linux-gnu)
  CPU: 8 × Cortex-A78AE
  WORD_SIZE: 64
  LLVM: libLLVM-18.1.7 (ORCJIT, generic)
  GC: Built with stock GC
Threads: 1 default, 1 interactive, 1 GC (on 8 virtual cores)
```

Not checked on x86; this box is aarch64 Linux.

---
🤖 Posted by an AI agent — harness: Claude Code 2.1.263 · model: claude-opus-5[1m]
Conversation: https://claude.ai/code/session_01CTUg9M7Zn2HdFbFREPz2hi

### LINKED ISSUE #5097: Preserve array derivatives in implicit DAE residual code generation
> 🚧 **UNREVIEWED — awaiting review by @ChrisRackauckas.** Filed by an AI agent running as @chrisrackauckas; Chris has not seen this finding. Please do not read this as his assessment or act on it until he confirms it.
> Harness: Codex CLI 0.153.4 · Model: gpt-6-astra
> Session: local session `01a08018-78ff-7e42-a964-f5174f3f763b` (no shareable URL).

# Array derivative residual code grows with state dimension

On current ModelingToolkit master, a single array equation `D(x) ~ -x` survives `complete`, but implicit DAE code generation emits one scalar derivative read per element in each of its out-of-place and in-place functions. The derivative array is reconstructed with `SymbolicUtils.array_literal`. This prevents the residual function from having a fixed-size expression graph as the array dimension grows.

Verified on https://github.com/SciML/ModelingToolkit.jl/commit/fcda4b8c239f8bafcdf35a5fae03bbe22bdb8c50 with its local ModelingToolkitBase and SciCompDSL packages developed into an isolated environment, Symbolics 7.39.2, SciMLBase 3.53.1 and Julia 1.12.6/aarch64. No Dyad or battery package is imported by the reproducer.

## Reproducer

```julia
using ModelingToolkit, SciMLBase, InteractiveUtils
versioninfo()
println("ModelingToolkit ", pkgversion(ModelingToolkit), "; SciMLBase ", pkgversion(SciMLBase))
function astnodes(x)
    x isa Expr || return 1
    return 1 + sum(astnodes, x.args; init=0)
end
for n in (6, 24, 96)
    @variables x(t)[1:n]
    sys = complete(System([D(x) ~ -x], t, collect(x), []; name=:m))
    expr = DAEFunction(sys; expression=Val{true})
    code = string(expr)
    println((n=n, equations=length(equations(sys)), ast_nodes=astnodes(expr),
        scalar_du_reads=count(r"__mtk_arg_1\[\d+\]", code)))
end
```

Observed output (successful construction, no runtime exception):

```text
(n = 6, equations = 1, ast_nodes = 481, scalar_du_reads = 12)
(n = 24, equations = 1, ast_nodes = 623, scalar_du_reads = 48)
(n = 96, equations = 1, ast_nodes = 1631, scalar_du_reads = 192)
```

The expression contains both callable variants, hence `2n` scalar reads. The regex is a diagnostic for the current generated argument name, not a proposed stable API test. AST counts count an `Expr` recursively and other objects as one node; this is generated expression size, not a wall-clock benchmark or a count of quoted system metadata.

The generated code includes the following pattern, with one binding per element:

```julia
local c3 = __mtk_arg_1[1]
local c4 = __mtk_arg_1[2]
# ... one binding per derivative element ...
local derivative_array = SymbolicUtils.array_literal((n,), c3, c4, ...)
```

The source is `expand_array_derivatives!` in [ModelingToolkitBase codegen](https://github.com/SciML/ModelingToolkit.jl/blob/fcda4b8c239f8bafcdf35a5fae03bbe22bdb8c50/lib/ModelingToolkitBase/src/systems/codegen.jl#L38). It replaces each array derivative with an array literal of scalar derivatives. A bounded representation of a contiguous derivative block or slice would let the residual operate on that block directly, preserving the shape without materializing a scalar expression for every entry. Numeric storage and state-index metadata still need to scale with state count.

## Public-API control

The control below was run in the registered-package environment (ModelingToolkit 11.42.0, Symbolics 7.39.0). Direct `Symbolics.build_function(-x - dx, dx, x)` with separate array symbols generates an in-place expression containing **22 nodes** at all three sizes. Attaching that function to a `DAEFunction` solves the exponential-decay example and preserves `sol(t; idxs=x)` indexing at all three sizes. This establishes that the basic array residual has a compact code-generation path; it does not establish a drop-in replacement for MTK's initialization and callback handling.

```julia
using ModelingToolkit, Symbolics, OrdinaryDiffEqBDF, SciMLBase, Test
nodes(x) = x isa Expr ? 1 + sum(nodes, x.args; init=0) : 1
for n in (6,24,96)
    @variables x(t)[1:n] dx(t)[1:n]
    exprs = Symbolics.build_function(-x - dx, dx, x)
    println((n=n, nodes=nodes(exprs[2])))
    _, f! = Symbolics.build_function(-x - dx, dx, x; expression=Val{false})
    sys = complete(System([D(x) ~ -x],t,collect(x),[];name=:array))
    f = DAEFunction{true}((out,du,u,p,t)->f!(out,du,u);sys)
    prob = DAEProblem(f,-ones(n),ones(n),(0.0,1.0);differential_vars=trues(n))
    sol = solve(prob,DFBDF();reltol=1e-8,abstol=1e-10)
    @test sol(1.0;idxs=x) ≈ fill(exp(-1),n) rtol=1e-6
    println("direct solve/indexing pass ", n)
end
```

## Version information

```text
Julia Version 1.12.6
Commit 15346901f00 (2026-04-09 19:20 UTC)
Build Info:
  Official https://julialang.org release
Platform Info:
  OS: Linux (aarch64-linux-gnu)
  CPU: 12 × Cortex-A78AE
  WORD_SIZE: 64
  LLVM: libLLVM-18.1.7 (ORCJIT, generic)
  GC: Built with stock GC
Threads: 1 default, 1 interactive, 1 GC (on 12 virtual cores)
Environment:
  JULIA_PKG_PRECOMPILE_AUTO = 0
ModelingToolkit 11.42.0; SciMLBase 3.53.1
```

Related array initialization work: https://github.com/SciML/ModelingToolkit.jl/issues/5093 and https://github.com/SciML/ModelingToolkit.jl/issues/5095. No x86 measurement was made for this report.

---
🤖 Posted by an AI agent — harness: Codex CLI 0.153.4 · model: gpt-6-astra
Session: local session `01a08018-78ff-7e42-a964-f5174f3f763b` (no shareable URL).

ChrisRackauckas-Claude:
## Summary

`mtkcompile` scalarizes every array equation before structural simplification, so for a finite-difference discretization such as `D(u[2:(n - 1)]) ~ lap(u)` the compiled system, the generated code and the compile time all grow with the number of grid points. MethodOfLines currently gets O(1) scaling only by skipping `mtkcompile` entirely and building a `DAEProblem` from the completed system.

This PR adds `mtkcompile(sys; scalarize_arrays = false)`, which compiles the system without tearing/index reduction and keeps first-order array differential equations intact, so that `DAEProblem` (and, with #5101, `ODEProblem`) code generation is independent of the array length. It is the first of two ModelingToolkit PRs; the ODE code generation half is stacked on top of this one.

### What changes

- **`mtkcompile(sys; scalarize_arrays = false)`** (ModelingToolkitBase). Dispatches to the tearing-free compiler (`__mtkcompile_no_tearing`, the former `__mtkcompile` of ModelingToolkitBase; ModelingToolkit's `__mtkcompile` override returns early to it). That compiler now:
  - keeps `D(x) ~ rhs` where `x` is an array of unknowns or a slice of one, ordering the unknowns so the elements of each differentiated slice form a contiguous block matching the rows of the equation;
  - rewrites residual-form array equations `D(x) .- f ~ 0` (what MethodOfLines emits) to `D(x) ~ f`;
  - scalarizes array *algebraic* equations, so that trivially defined elements (boundary conditions) become observed as before;
  - counts an array equation as one row per element in the balance check;
  - records the choice under the `ScalarizeArraysCtx` metadata key (`arrays_scalarized(sys)`), which the initialization system inherits (mirrors `HomotopyCtx`).
  - Systems with brownians/jumps/poissonians/noise throw an `ArgumentError` with `scalarize_arrays = false`.
- **Code generation keeps the array form**:
  - `expand_array_derivatives!` binds a derivative of a slice to a `view` of `du` when its scalar derivatives are consecutive in `du`, instead of an `array_literal` of them (fixes #5097).
  - Arrays whose elements are partly unknowns and partly observed are reconstructed from `view`s of the argument buffers plus the observed elements (`partial_array_reconstruction`), instead of through an observed equation that lists every element. Guarded by `!arrays_scalarized(sys)`, so the default path is unchanged.
  - `default_toterm` handles derivatives of slices: `D(u[2:4])` becomes `uˍt(t)[2:4]`.
  - `full_equations` scalarizes preserved equations so jacobians/mass matrices/sparsity keep one row per scalar equation.
- **Initialization**:
  - `write_possibly_indexed_array!` broadcasts a scalar written to an array (or slice) key, so the scalar default derivative guess for an array `uˍt(t)` no longer produces `vec(::Bool)` (fixes #5093 — the reproducer from the issue is now a test).
  - The initialization system registers the rows of a preserved derivative equation so that it is not added a second time in scalarized form, and `get_initialization_problem_type` counts equation rows and copes with a `nothing` tearing state (no SCC problem without tearing).
- **Tests**: `test/array_equation_dae.jl` (previously not in `runtests.jl`) is wired in, plus new `test/array_equations_preserved.jl`.

### Example

```julia
using ModelingToolkit, OrdinaryDiffEqBDF

n = 96
@variables u(t)[1:n]
dx = 1 / (n - 1)
lap = (u[1:(n - 2)] .- 2 .* u[2:(n - 1)] .+ u[3:n]) ./ dx^2
@named heat = System([D(u[2:(n - 1)]) ~ lap, u[1] ~ 0, u[n] ~ 0], t, [u], [])

sys = mtkcompile(heat; scalarize_arrays = false)
equations(sys)   # 1 equation: D(u[2:95]) ~ ...
unknowns(sys)    # u[2], ..., u[95]
observed(sys)    # u[1] ~ 0, u[96] ~ 0

prob = DAEProblem(sys, [u => sinpi.(range(0, 1, length = n))], (0.0, 0.1))
sol = solve(prob, DFBDF())
```

### How to verify the O(1) scaling

`test/array_equations_preserved.jl` checks, for `n = 24, 48, 96`, that the compiled system has exactly one equation and that the generated in-place `DAEProblem` function (`generate_rhs(sys, GeneratedFunctionOptions(; expression = Val{true}); implicit_dae = true)`) has the same number of `Expr` nodes and no `array_literal`. The same file checks the solution against the scalarized compilation and the analytic solution of the heat equation, with and without an initialization problem, and a 2D slice.

### Limitations

- No tearing or index reduction: differential equations must be explicit (`D(x) ~ f` or `D(x) .- f ~ 0`), and algebraic equations that do not trivially define an unknown are kept as algebraic equations. Higher-order array equations are scalarized and order-reduced as before.
- `ODEProblem` from such a system still requires the follow-up #5101; here it keeps throwing the existing "array equations" error.

### Tests run locally

- ModelingToolkitBase groups `Initialization`, `SymbolicIndexingInterface`, `InterfaceI`, `InterfaceII`: pass, except the pre-existing `JumpSystem Test` error (`FieldError: type NamedTuple has no field abstol` from `affect_tolerance` in `callbacks.jl` on an `SSAIntegrator`), which also fails on `master` with JumpProcesses 9.32.2 and is untouched here.
- ModelingToolkit group `InterfaceI`: pass (1507 passed, 3 broken, as on master).
- MethodOfLines `test/Discretization/problem_construction.jl` against this branch (including the new ODE path from SciML/MethodOfLines.jl#688): pass.

### CI notes

- The integration tests (Catalyst, MethodOfLines, ...) and the benchmark job fail at resolution with "Unsatisfiable requirements ... ModelingToolkitBase restricted to versions 1.70.0" because the ModelingToolkit-side changes (`__mtkcompile` early return, `count_equation_rows`) need the ModelingToolkitBase in this PR. They will resolve once ModelingToolkitBase 1.70.0 is registered.
- `downgrade-mtkbase (InterfaceII)` and `sublibrary-ci InterfaceII` fail only on the pre-existing `JumpSystem Test` error above (the same job fails on #5098).
- `sublibrary-ci QA` fails only on JET: the QA environment now resolves JET 0.12.1, which reports ~245 "local variable may be undefined" findings in Moshi `@match` expansions across ModelingToolkitBase (files this PR does not touch); #5073 fails the same way with 234 findings. The ModelingToolkit `QA` job (JET 0.11.6) passes.
- `InterfaceI (x86)` fails to precompile Sundials/QuasiMonteCarlo on 32-bit, as on `master` (#5098).
- Documentation `build` fails on the pre-existing `unbound_inputs` `@ref` in `interactive_simulation.md`, fixed by #5099; the new `@docs` block here renders.

Related: #5093, #5097, SciML/MethodOfLines.jl (array-form `DAEProblem` path).

Made with [Cursor](https://cursor.com)


lib/ModelingToolkitBase/src/systems/codegen_utils.jl
```diff
@@ -454,17 +454,256 @@ function _compute_array_variable_buffer_idxs(args::Vector, ignore_vars, ignore_a
...
+observed, the size of the reconstruction is independent of the array length. For
+multi-dimensional arrays this applies when the present elements form a rectangular block
+laid out in column-major order in the buffer; otherwise they are read one at a time. The
+array is allocated with the `eltype` promoted from the buffers and the fallback elements, so
+that differentiating through the generated function with dual numbers works.
+
+Returns `nothing` if the array cannot be reconstructed: some element is neither in a buffer
+nor has a fallback, or the array is not indexed from one.
+"""
+function partial_array_reconstruction(
```
AayushSabharwal, line 502: I don't think this should be necessary. `array_variable_assignments` in `codegen_utils.jl` should handle this. If that generates suboptimal code (such as a large `array_literal`) we should fix it and make it use e.g. `ArrayMaker`.

AayushSabharwal: Oh I see that's exactly what this is doing.

lib/ModelingToolkitBase/src/systems/codegen_utils.jl
```diff
@@ -454,17 +454,256 @@ function _compute_array_variable_buffer_idxs(args::Vector, ignore_vars, ignore_a
...
+        name = get(fallback_names, i, nothing)
+        value = if name === nothing
+            buf, pos = idxs[i]
+            push!(used_buffers, buffer_name(buf))
+            Expr(:ref, buffer_name(buf), pos)
+        else
+            push!(element_exprs, name)
+            name
+        end
+        push!(writes, Expr(:(=), Expr(:ref, result, Tuple(carts[i])...), value))
```
AayushSabharwal, line 542: I guess the splat here is fine but it would be nice to just not have to build `carts` since it will never infer.

lib/ModelingToolkitBase/src/systems/codegen_utils.jl
```diff
@@ -454,17 +454,256 @@ function _compute_array_variable_buffer_idxs(args::Vector, ignore_vars, ignore_a
...
+        return nothing
+    end
+    function write_block!(region, buf, pos, len)
+        name = buffer_name(buf)
+        push!(used_buffers, name)
+        src = Expr(:call, view, name, pos:(pos + len - 1))
+        if N > 1
+            src = Expr(:call, reshape, src, map(length, region)...)
```
AayushSabharwal, line 550:
```suggestion
            src = Expr(:call, reshape, src)
            append!(src.args, Iterators.map(length, region))
```

lib/ModelingToolkitBase/src/systems/codegen_utils.jl
```diff
@@ -454,17 +454,256 @@ function _compute_array_variable_buffer_idxs(args::Vector, ignore_vars, ignore_a
...
+        end
+        dest = Expr(:call, view, result, region...)
```
AayushSabharwal, line 552:
```suggestion
        dest = Expr(:call, view, result)
        append!(dest.args, region)
```

lib/ModelingToolkitBase/src/systems/codegen_utils.jl
AayushSabharwal: This function should be checked using Cthulhu/`@code_typed` to make sure the closures don't cause excessive boxing from that one Julia bug. It would be nice to just use explicit `let` captures anyway.

lib/ModelingToolkitBase/src/systems/codegen_utils.jl
```diff
@@ -454,17 +454,256 @@ function _compute_array_variable_buffer_idxs(args::Vector, ignore_vars, ignore_a
...
+            end
+            i = j + 1
+        end
+    else
+        present = findall(!iszero ∘ first, idxs)
+        blockset = Set{Int}()
+        if !isempty(present)
+            lo = Tuple(carts[first(present)])
+            hi = lo
+            for i in present
```
AayushSabharwal, line 579: I have a gut feeling that this loop would really benefit from not using `CartesianIndices`, or at least turning `lo` and `hi` into `Vector{Int}`/`Memory{Int}` so that the broadcasts infer. We can even make it operate in-place.

lib/ModelingToolkitBase/src/systems/codegen_utils.jl
```diff
@@ -454,17 +454,256 @@ function _compute_array_variable_buffer_idxs(args::Vector, ignore_vars, ignore_a
...
+            end
+            region = ntuple(d -> lo[d]:hi[d], N)
+            lin = LinearIndices(sz)
+            block = [lin[c] for c in CartesianIndices(region)]
+            buf, pos = idxs[first(block)]
+            contiguous = length(block) == length(present) && !iszero(buf) && all(
+                enumerate(block)
+            ) do (k, i)
+                idxs[i] == (buf, pos + k - 1)
+            end
```
AayushSabharwal, line 592: Quick comment to describe what this is doing?

lib/ModelingToolkitBase/src/systems/codegen_utils.jl
```diff
@@ -454,17 +454,256 @@ function _compute_array_variable_buffer_idxs(args::Vector, ignore_vars, ignore_a
...
+            lo = Tuple(carts[first(present)])
+            hi = lo
+            for i in present
+                c = Tuple(carts[i])
+                lo = min.(lo, c)
+                hi = max.(hi, c)
+            end
+            region = ntuple(d -> lo[d]:hi[d], N)
+            lin = LinearIndices(sz)
+            block = [lin[c] for c in CartesianIndices(region)]
```
AayushSabharwal, line 586: `block`'s ndims won't infer. Does this need to be an `Array{T, N}` or can it just be a `Vector{T}`? If the latter, it's better to do `block = Vector{T}(undef, some_length)` and write to it in a loop.

lib/ModelingToolkitBase/src/systems/codegen_utils.jl
```diff
@@ -454,17 +454,256 @@ function _compute_array_variable_buffer_idxs(args::Vector, ignore_vars, ignore_a
...
+        push!(eltype_expr.args, Expr(:call, eltype, name))
+    end
+    for name in unique!(element_exprs)
+        push!(eltype_expr.args, Expr(:call, typeof, name))
+    end
+    alloc = if isempty(used_buffers)
+        Expr(:call, Expr(:curly, Array, eltype_expr, N), :undef, sz...)
+    else
+        Expr(:call, similar, first(used_buffers), eltype_expr, sz...)
+    end
```
AayushSabharwal, line 616:
```suggestion
    alloc = if isempty(used_buffers)
        Expr(:call, Expr(:curly, Array, eltype_expr, N), :undef)
    else
        Expr(:call, similar, first(used_buffers), eltype_expr)
    end
    append!(alloc.args, Iterators.map(length, sh))
```
No splat, no need for `sz`.

lib/ModelingToolkitBase/src/systems/codegen_utils.jl
AayushSabharwal: `CartesianIndices(region)` may not be 1-indexed, but `lin` is. Won't this cause problems? I guess what this is trying to do is `block = [prod(Tuple(c)) - prod(lo) for c in CartesianIndices(region)]`?

lib/ModelingToolkitBase/src/systems/codegen_utils.jl
```diff
@@ -454,17 +454,256 @@ function _compute_array_variable_buffer_idxs(args::Vector, ignore_vars, ignore_a
...
+    body = Expr(:block, Expr(:(=), result, alloc))
+    append!(body.args, writes)
+    push!(body.args, result)
+    push!(assignments, Assignment(arrvar, body))
+    return assignments
+end
```
AayushSabharwal, line 622: This function would be much easier to understand with some more comments.

ChrisRackauckas-Claude:
Closing this as an intermediate step that is not the direction we want.

`scalarize_arrays = false` gets O(1) code by switching tearing and index reduction off for the whole system, and then bolts a separate compiler path (`__mtkcompile_no_tearing`, `ScalarizeArraysCtx`, `arrays_scalarized`) onto every consumer. That is turning algorithms off rather than making them array-aware, and it would have to be kept in sync with the real `mtkcompile` forever. The goal is for tearing and index reduction themselves to keep array equations intact, which is a different piece of work and will come as its own PR.

What is useful here has been split out:
- #5101 has been rewritten from master (no longer stacked on this PR) and makes `ODEProblem(complete(sys), ...)` work for array differential equations directly, mirroring what `DAEProblem` already does, with generated code independent of the array length. No new keyword, no alternative compiler.
- The `default_toterm`, `write_possibly_indexed_array!` (#5093) and `du`-view (#5097) fixes are only needed once array equations survive `mtkcompile`, so they belong with that work.

lib/ModelingToolkitBase/src/systems/codegen_utils.jl
```diff
@@ -454,17 +454,256 @@ function _compute_array_variable_buffer_idxs(args::Vector, ignore_vars, ignore_a
...
+laid out in column-major order in the buffer; otherwise they are read one at a time. The
+array is allocated with the `eltype` promoted from the buffers and the fallback elements, so
+that differentiating through the generated function with dual numbers works.
+
+Returns `nothing` if the array cannot be reconstructed: some element is neither in a buffer
+nor has a fallback, or the array is not indexed from one.
+"""
+function partial_array_reconstruction(
+        arrvar::SymbolicT, idxs::Vector{Tuple{Int, Int}}, argument_name, buffer_offset::Int,
+        element_fallback
```
AayushSabharwal, line 504: Might want to `@nospecialize` the `argument_name` and `element_fallback`

lib/ModelingToolkitBase/src/systems/codegen_utils.jl
```diff
@@ -454,17 +454,256 @@ function _compute_array_variable_buffer_idxs(args::Vector, ignore_vars, ignore_a
...
+
+Assignments binding the slices `slices` of the array variable `arrvar` to `view`s of the
+argument buffers, for when `arrvar` itself cannot be reconstructed because elements outside
+the slices are not variables of the system (e.g. the interior of a discretized field whose
+boundary is observed). `idxs` is as in [`partial_array_reconstruction`](@ref). A slice
+whose elements are consecutive in one buffer is a `view` (reshaped for a multi-dimensional
+slice); any other slice is read one element at a time. Slices with an element in no buffer
+are skipped.
+"""
+function array_slice_assignments(
```
AayushSabharwal, line 635: The same general comments from above apply here.

ChrisRackauckas-Claude:
Superseded checkpoint pointers (this PR remains closed as the wrong intermediate):

- #5101 — `complete` → `ODEProblem`
- #5102 + https://github.com/JuliaComputing/StateSelection.jl/pull/158 — array groups through tearing/IR (draft)
- https://github.com/ChrisRackauckas-Claude/ModelingToolkit.jl/blob/array-aware-tearing/docs/src/internals/ARRAY_EQUATIONS_CHECKPOINT.md

PR #5107, opened 2026-09-11
`NonlinearSystem` init eqs

oscardssmith:
Rather than using `SteadyStateSystem`, this makes `NonlinearSystem` properly strip time dependence from initals which was preventing it from working before. Not fully sure this is doing what's intended, but it seems at least to be doing something reasonable.

### DIFF
--- a/lib/ModelingToolkitBase/src/systems/problem_utils.jl
+++ b/lib/ModelingToolkitBase/src/systems/problem_utils.jl
@@ -2251,6 +2251,20 @@ function __process_SciMLProblem(
     end
     add_observed_equations!(op, obs)
 
+    # For a nonlinear solve, guesses are the starting point of unknowns without a value
+    if !is_time_dependent(sys) && !is_initializeprob
+        all_guesses = merge(ModelingToolkitBase.guesses(sys), todict(guesses))
+        all_guesses = as_atomic_dict_with_defaults(
+            Dict{SymbolicT, SymbolicT}(all_guesses), COMMON_NOTHING
+        )
+        for v in dvs
+            get_possibly_indexed(op, v, COMMON_NOTHING) === COMMON_NOTHING || continue
+            guess = get_possibly_indexed(all_guesses, v, COMMON_NOTHING)
+            guess === COMMON_NOTHING && continue
+            write_possibly_indexed_array!(op, v, guess, COMMON_NOTHING)
+        end
+    end
+
     if warn_cyclic_dependency
         cycles = check_substitution_cycles(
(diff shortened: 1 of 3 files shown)


lib/ModelingToolkitBase/src/systems/system.jl
```diff
@@ -1369,33 +1386,130 @@ function NonlinearSystem(sys::System)
...
+    # flattening may list a variable both where it is declared and where it is referenced
+    dvs = unique!([get_unknowns(sys); map(eq -> eq.lhs, obs)])
+    # every derivative term of any order is zero at steady state
+    derivs = Set{SymbolicT}()
+    for eq in Iterators.flatten((eqs, get_initialization_eqs(sys)))
+        SU.search_variables!(derivs, eq; is_atomic = OnlyOperatorIsAtomic{Differential}())
+    end
+    subrules = Dict{SymbolicT, Float64}()
+    for dx in derivs
+        isoperator(dx, Differential) && (subrules[dx] = 0.0)
```
AayushSabharwal, line 1410: What if it's the derivative of an array variable?

lib/ModelingToolkitBase/src/systems/system.jl
```diff
@@ -1369,33 +1386,130 @@ function NonlinearSystem(sys::System)
...
+    function kind(ex)
+        empty!(vs)
+        SU.search_variables!(vs, ex; is_atomic = OperatorIsAtomic{Initial}())
+        iv in vs && return :temporal
+        return any(is_tvar, vs) ? :dynamic : :static
+    end
+    # `x = expr` at `t = 0` gives the starting point of `x` for the nonlinear solve: an
+    # initial condition if `expr` is known outright, else a guess to be resolved from the
+    # starting points of the other variables.
+    function add_start!(ics, guesses, x, expr)
```
AayushSabharwal, line 1481: Might as well pass the `kind` here instead of computing it twice?

lib/ModelingToolkitBase/src/systems/problem_utils.jl
```diff
@@ -2251,6 +2251,20 @@ function __process_SciMLProblem(
     end
     add_observed_equations!(op, obs)
 
+    # For a nonlinear solve, guesses are the starting point of unknowns without a value
+    if !is_time_dependent(sys) && !is_initializeprob
```
AayushSabharwal, line 2255: This seems like a separate change? Why is it necessary?

oscardssmith:
why did you close?

ChrisRackauckas:
Why would this direction be an improvement?

PR #5131, opened 2026-09-13
Support array unknowns on a `complete`d system for every problem type

ChrisRackauckas-Claude:
> **Note:** this PR should be ignored until reviewed by @ChrisRackauckas.

## What changed and why

A `complete`d `System` whose `unknowns` contain array symbolics now behaves exactly as if those entries were scalarized in place, for **every** problem type: `u` is the flat concatenation of the unknowns in `unknowns(sys)` order (an array unknown contributing `length(x)` consecutive entries — the layout `IndexCache` already assigns them), the generated code reconstructs each array unknown as a view into `u` through the existing `array_variable_assignments` machinery, and initialization, `du0`, `differential_vars`, `lb`/`ub` and symbolic indexing all use the same layout. `mtkcompile` is no longer required to scalarize them for any constructor.

Motivated by the NeuralPDE PDE pipeline in https://github.com/SciML/NeuralPDE.jl/pull/1165, whose unknowns are the neural-network parameter arrays and which currently has to call `mtkcompile` only to scalarize them.

**Implementation.** There is no trait and no opt-in keyword; the behaviour is unconditional.

- `check_array_equations_unknowns` → `check_array_equations`: it no longer rejects array *unknowns* at all. (Array *equations* still need `mtkcompile` outside the implicit-DAE path — see "Still open" below.)
- `flat_unknowns(sys) = scalarized_vars(unknowns(sys))` (`systems.jl`, next to `scalarized_vars`) replaces `unknowns(sys)` in every state-vector position: `__process_SciMLProblem`'s `dvs` (so `u0`, `du0` and `check_eqs_u0` are flat), `generate_rhs`/`generate_jacobian`/`generate_tgrad`/`generate_W`/`generate_dae_jacobian`/`jacobian_sparsity`/the cost and constraint generators/`build_explicit_observed_function`/`generate_custom_function`, and the `dvs` of the `ODE`/`DAE`/`SDE`/`Nonlinear`/`Discrete`/`ImplicitDiscrete`/`BVP`/`Jump`/`Optimization` problem constructors. For a system `mtkcompile` has already scalarized it returns an equal vector, so nothing else changes.
- **Initialization** was taught the same layout, which is where the two real bugs were:
  - `SetInitialUnknowns`/`GetUpdatedU0` built `Initial.(unknowns(sys))`; `parameter_index` returns a `ParameterIndex{Initials, UnitRange}` for an array unknown and the `::ParameterIndex{Initials, Int}` assert threw. They now use `flat_unknowns`, so each entry is the scalar `Initials` index, and `GetUpdatedU0`'s guess mask is per-element (it is used as `view(buffer, guessvars)` over the flat `u0`) and looks values up with `get_possibly_indexed`.
  - `generate_initializesystem_timevarying` stored the dummy-derivative guess with `write_possibly_indexed_array!(guesses, ttk, ...)`. An array-valued `ttk` (`xˍt(t)` for `@variables x(t)[1:3]`) is not an *indexed* symbolic, so a scalar `false` was stored under an array-shaped key and the next `get_possibly_indexed` on an element of it hit `MethodError: no method matching vec(::Bool)`. New `write_dd_guess!` stores `fill(guess, size(ttk))` in that case, mirroring what `add_initials!` already does.
  - `solved_unknowns` and `remake_initialization_data`'s `to_varmap(u0, dvs)` also take `flat_unknowns`, otherwise `initializeprobmap` rebuilt `u0` as a 2-element nested vector and `to_varmap` mismatched lengths.
- `DAEProblem`'s `differential_vars` now marks an array unknown's elements differential when `D(x)` over the whole array appears.

`ModelingToolkitBase` 1.71.1 → 1.72.0. No new public names. The docs change follows @ChrisRackauckas's review: the paragraph is now a single general statement in "The state vector and parameter object" rather than an "unlike the other problem types" note under optimization.

## What works now, per problem type

All with **default initialization settings** (no `build_initializeprob = false`), `@variables x(t)[1:3] y(t)`, `unknowns(sys) == [x, y]`:

| | before | after |
|---|---|---|
| `ODEProblem` | `ArgumentError: The system has array unknowns.` | `length(u0) == 4`, `f(u0, p, 0)` and the `Tsit5` solution match the analytic solution and `mtkcompile` |
| `DAEProblem` (scalar equations) | `TypeError: ... expected ParameterIndex{Initials, Int64}, got ParameterIndex{Initials, UnitRange{Int64}}` | `length(u0) == 4`, consistent `du0`, `all(differential_vars)`, `DFBDF` solution matches the analytic one |
| `DAEProblem` (array equation `D(x) ~ -a .* x`) | `MethodError: no method matching vec(::Bool)` | residual evaluates, `length(u0) == 4` |
| `NonlinearProblem` | `ArgumentError: The system has array unknowns.` | `length(u0) == 4`, `NewtonRaphson` solves, `sol[z] == [1, 2, 6]` |
| `SDEProblem` | `ArgumentError: The system has array unknowns.` | `length(u0) == 4`, `f`/`g` match `mtkcompile`, `SOSRI` solves |
| `OptimizationProblem` | `ArgumentError: The system has array unknowns.` | objective/gradient/hessian/constraints, `lb`/`ub` from `bounds` metadata, `AutoForwardDiff` solve |

Symbolic indexing is unchanged throughout: `prob[x]` returns the 3-vector, `prob[x[2]]` the element, `variable_index(sys, x) == 1:3`, and observed expressions over array unknowns (`sum(x) + y`, `x[3] * y`) read the same flat layout.

## Verification

Julia 1.11.9, scratch environment with `ModelingToolkitBase` `dev`ed from this branch (SciMLBase 3.54.0, Symbolics 7.39.2, SymbolicUtils 4.46.6, OrdinaryDiffEq 7.8.1, NonlinearSolve 4.30.0, StochasticDiffEq 7.2.0, Optimization 5.9.1).

### Failing before / passing after

The four new testsets, extracted verbatim into standalone files and run against an unmodified `origin/master` worktree (8875e771ab) and against this branch. Same files, same environment, only `ModelingToolkitBase` differs.

**Before** — all four error out of the `@testset`:

```
######## ode
Array unknowns on a `complete`d system: Error During Test at .../ode_arr_only.jl:4
  ArgumentError: The system has array unknowns. Call `mtkcompile` to handle this or scalarize them manually.
  Stacktrace:
    [1] check_array_equations_unknowns(eqs::Vector{Equation}, dvs::Vector{...})
      @ ModelingToolkitBase .../src/systems/abstractsystem.jl:3289
    [2] __process_SciMLProblem(...)
      @ ModelingToolkitBase .../src/systems/problem_utils.jl:2231
    [5] _ode_problem(::Type{ODEProblem{...}}, sys::System, op::Any, tspan::Tuple{Float64, Float64}; ...)
      @ ModelingToolkitBase .../src/problems/odeproblem.jl:166
Test Summary:                          | Error  Total  Time
Array unknowns on a `complete`d system |     1      1  3.9s
######## ode exit=1
######## nl exit=1      (same ArgumentError)
######## sde exit=1     (same ArgumentError)
######## opt exit=1     (same ArgumentError)
```

**After**:

```
######## ode
Array unknowns on a `complete`d system |   24     24  31.0s
######## ode exit=0
######## nl
Array unknowns on a `complete`d system |   11     11   9.7s
######## nl exit=0
######## sde
Array unknowns on a `complete`d system |   10     10  18.4s
######## sde exit=0
######## opt
Array unknowns on a `complete`d system |   27     27  40.7s
######## opt exit=0
```

The two initialization bugs are only reachable once the guard is gone, so they do not appear in the "before" column above. Forcing the guard off on master reproduces them exactly as quoted in the table (`SetInitialUnknowns` at `problem_utils.jl:1526`, `get_possibly_indexed` at `atomic_array_dict.jl:163` ← `add_observed_equations!` at `problem_utils.jl:196` ← the initialization problem's own `__process_SciMLProblem`).

### Test groups

Each file in its own `julia` process, `using Test, SciMLBase; include(file)`:

- **`Initialization`** — `guess_propagation.jl` exit=0, `initializationsystem.jl` exit=0, `initial_values.jl` exit=0. (3/3)
- **`InterfaceI`** — all 35 files exit=0: `abstractsystem`, `abstractsystem_interface`, `abstractcollocation`, `variable_scope`, `variable_parsing`, `linearity`, `binding_semantics`, `input_output_handling`, `system_building`, `simple_mtkcompile`, `mtkcomplete`, `odesystem`, `system_tspan`, `dq_units`, `mass_matrix`, `split_parameters`, `static_arrays`, `components`, `error_handling`, `basic_transformations`, `changeofvariables`, `symbolic_events`, `stream_connectors`, `domain_connectors`, `dep_graphs`, `function_registration`, `precompile_test`, `dae_jacobian`, `array_equation_dae`, `jacobiansparsity`, `modelingtoolkitize`, `constants`, `parameter_bindings`, `equation_type_accessors`, `accessor_functions`, `complex`.
- **Also run**: `nonlinearsystem.jl` exit=0, `sdesystem.jl` exit=0, `optimization/optimizationsystem.jl` exit=0, `array_equation_dae.jl` exit=0 (all pass: `array equations reach DAEProblem` 6/6, `other problem types still require scalarized equations` 1/1, `array-equation DAE solves to the analytic solution` 2/2, `array equations over a 2D slice keep their shape` 2/2, `` array equations written as `D(slice) ~ rhs` `` 4/4).
- **`QA`** — `JET Tests` 54/54 pass; `Aqua Tests` 20 pass / 1 fail. Identical on an unmodified `origin/master` worktree (I ran both): the failure is `Quality Assurance: JET-test failed at JET/src/JETBase.jl:1046` with `═════ 38 toplevel errors found ═════`, all `ArgumentError: Package StaticArraysCore not found in current path` — `JET.report_package` cannot resolve `ModelingToolkitBase`'s dependencies from the `qa` project in this local setup. Pre-existing, and this diff adds no imports and no dependencies.

`array_equation_dae.jl` existed in the repo but was not referenced from `runtests.jl`; since it is the regression guard for the DAE array-equation codegen this PR touches (`generate_rhs`'s `dvs`), and for the MethodOfLines shape (scalarized unknowns, array equations over slices), it is now wired into `InterfaceI`.

`Runic` and `typos` are clean on every changed file.

### Existing test updated

`odesystem.jl`'s `@testset "Issue #2597"` pinned the removed limitation — it asserted that `ODEProblem` throws for array *unknowns* as well as array *equations*. It now asserts the new contract: scalar equations over an array unknown build a problem with the right `length(u0)`, an array equation still throws.

## Still open

- **Array *equations* still require `mtkcompile` for every constructor except the implicit-DAE path**, and `check_array_equations` keeps that guard. Making them work is not contained: the explicit path in `generate_rhs` builds `rhss = [eq.rhs for eq in eqs]` with no `expand_array_derivatives!`/`array_residual_maker` treatment, and `calculate_jacobian`, `calculate_tgrad`, `calculate_hessian`, `jacobian_sparsity` and the mass-matrix construction all differentiate that vector against `dvs` assuming one scalar row per equation. That is a separate change with its own test surface, so I left it alone rather than half-doing it.
- **A system with array equations over *slices* cannot build its initialization problem**, with or without this PR. `DAEProblem(complete(sys), op, tspan)` on the MethodOfLines shape (`broadcast(-, D(u[2:5]), lap) ~ zeros(4)` with scalarized unknowns) fails at

  ```
  ArgumentError: Encountered operator `Differential(t, 1)` which has different independent variable
  than the one used in the system `nothing`.
  Context: broadcast(-, Differential(t, 1)((u(t))[2:5]), broadcast(/, broadcast(+, broadcast(-, ...
    at validate_operator utils.jl:803
    at collect_vars! utils.jl:916
    at discover_globalscoped abstractsystem.jl:624
    at __mtkcompile systems.jl:227
    at mtkcompile_initialization_system initializationproblem.jl:209
    at InitializationProblem initializationproblem.jl:56
    at maybe_build_initialization_problem problem_utils.jl:1919
  ```

  i.e. `mtkcompile` of the *initialization* system does not accept a derivative of a slice. I verified this reproduces byte-for-byte on unmodified `origin/master`, so it is pre-existing and orthogonal to array unknowns (that system's unknowns are already scalarized). It is why MethodOfLines passes `build_initializeprob = false` today. I did not document that keyword anywhere as a way to use array unknowns; this is reported here as the open problem it is. The whole-array form `D(x) ~ -a .* x` over an array unknown *does* work with default initialization, and is tested.

## Not verified

- The `InterfaceII`, `Extensions`, `Downstream`, `SymbolicIndexingInterface`, `FMI` and GPU groups were not run — this scratch environment does not resolve their dependency sets. `InterfaceII` in particular contains `code_generation.jl`, `index_cache.jl` and `discrete_system.jl`, which are the closest remaining neighbours of this diff; CI is what will cover them.
- The docs build was not run.
- `JumpProblem`, `BVProblem`, `DiscreteProblem`, `ImplicitDiscreteProblem` and `LinearProblem` got the same `flat_unknowns` treatment for consistency, but I did not write array-unknown tests for them — only ODE/DAE/SDE/Nonlinear/Optimization are covered.

## Worth pushing back on

- `flat_unknowns` allocates a fresh vector on every call, including for systems that have no array unknowns. It is noise next to code generation, but it is a real (small) regression on hot paths that call these generators repeatedly.
- Symbolic `grad = true`/`hess = true` over a *lazy* array cost is silently wrong, and this PR does not fix it: `Symbolics.gradient(sum(abs2, x .- a), [x[1], x[2], x[3], y])` returns `[0, 0, 0, …]` because the `mapreduce` term is not scalarized before differentiation, and `sum((x .- a) .^ 2)` throws `ArgumentError: invalid index: _1` inside `SymbolicUtils.scalarize`. This reproduces identically on master via `calculate_cost_gradient(mtkcompile(sys))`. The new optimization tests therefore exercise symbolic `grad`/`hess` with an explicitly indexed cost and the lazy array costs through an `adtype`. Happy to reduce it to a SciML-free MWE and file it upstream if you want.

## Rebase (2026-09-13)

Rebased onto master at ModelingToolkitBase 1.71.3 (conflict: version number only, kept 1.72.0). Re-ran on the rebased commit, each file in its own process: `test/optimization/optimizationsystem.jl`, `test/odesystem.jl`, `test/nonlinearsystem.jl`, `test/sdesystem.jl`, `test/array_equation_dae.jl`, `test/initializationsystem.jl`, `test/initial_values.jl`, `test/guess_propagation.jl`: all pass. (`odesystem.jl`'s "`full_equations` doesn't recurse infinitely" 180 s subprocess check failed once while eight other Julia processes were running on the machine and passed on a standalone re-run.) NeuralPDE's PDE test files and docs pages were re-run against the rebased branch: all pass except one parameter-estimation test whose BFGS trajectory diverges from an 8e-15 relative difference in gradient summation order between this branch and its first version; forward objective values are bit-identical. That is a fragility of that test, tracked on the NeuralPDE side.

🤖 Generated with [Claude Code](https://claude.com/claude-code) (model: claude-fable-5-1)

https://claude.ai/code/session_01WYPUXLxE9jVu2jXLFRd8ax


docs/src/API/problems.md
```diff
@@ -79,6 +79,22 @@ SciMLBase.ODEInputFunction
 ModelingToolkit.constraints_to_penalties
 ```
 
+Unlike the other problem types, `OptimizationProblem` and `OptimizationFunction` accept a
+system whose unknowns include array variables without requiring `mtkcompile` to scalarize
+them; `complete` is enough. The state vector `u` is the concatenation of the unknowns in
```
ChrisRackauckas: We should make all problem types handle this?
```suggestion
The state vector `u` is the concatenation of the unknowns in
```

ChrisRackauckas: DAEProblem already does.

ChrisRackauckas-Claude: Checked on this branch (the DAE path is untouched by the PR, so this is master behaviour): `DAEProblem(complete(sys), op, tspan)` with an array unknown does not work today, it only skips the `mtkcompile` check through `implicit_dae` and then fails further down.

- `D(x) ~ -a .* x` (array equation): `MethodError: no method matching vec(::Bool)` in `add_observed_equations!` → `get_possibly_indexed` (`atomic_array_dict.jl:163`).
- `D(x[1]) ~ ...` (scalar equations over `x[1:3]`): `TypeError: expected ParameterIndex{Initials, Int64}, got ParameterIndex{Initials, UnitRange{Int64}}` in `SetInitialUnknowns` (`problem_utils.jl:1526`).

With the trait forced `true` for every constructor, `ODEProblem`, `DAEProblem` and `NonlinearProblem` all stop at that same `SetInitialUnknowns` typeassert. So the gate exists because optimization problems have no initialization system; that is the only constructor where flattening `u` is sufficient as of now. The argument names are not shared with the DAE path either, DAE has no such flag.

Making it unconditional means teaching the initialization index handling (`Initials` ranges for array unknowns) and the observed-equation lookup to accept array unknowns. Working on that now, so the trait and the kwarg go away and `u` is the concatenation of the unknowns for every problem type.

ChrisRackauckas-Claude: Correction to my previous reply, after running MethodOfLines' own `DAEProblem(pdesys, disc)` (MethodOfLines 1.5.0, ModelingToolkitBase 1.71.1): its discretized system has *scalarized* unknowns (`unknowns(sys) == [u(t)[1], …, u(t)[6]]`) while the equations stay array-valued (`D(u(t)[2:5]) ~ …` over slices), and that builds and evaluates fine. So the DAE path already handles array **equations** over indexed unknowns, as you say.

What fails everywhere, DAE included, is an unscalarized array symbolic in `unknowns(sys)` itself (`@variables x(t)[1:3]`, `unknowns(sys) == [x, y]`): `u0` construction (`varmap_to_vars`, problem_utils.jl:459, `Cannot convert Vector{Float64} to Float64`) and the initialization setters (`SetInitialUnknowns`, `ParameterIndex{Initials, UnitRange}`). That is the NeuralPDE case (the network parameters are one array unknown) and what this PR flattens for optimization. The PR is being widened so every constructor treats `u` as the flat concatenation of `unknowns(sys)`, with initialization enabled, and the trait/kwarg go away.

ChrisRackauckas: But this is a temporary limitation, so we shouldn't document it like this is something we expect.

lib/ModelingToolkitBase/src/systems/nonlinear/initializesystem.jl
```diff
@@ -37,6 +37,22 @@ function generate_initializesystem(
...
+Guess `guess` for the dummy-derivative variable `ttk`. An array-valued `ttk` is not an
+indexed symbolic, so it needs an array of the guess rather than the scalar one
+`write_possibly_indexed_array!` would store under an array-shaped key.
+"""
+function write_dd_guess!(guesses::AtomicArrayDict{SymbolicT}, ttk::SymbolicT, guess::SymbolicT)
+    if Symbolics.isarraysymbolic(ttk)
+        guesses[ttk] = BSImpl.Const{VartypeT}(fill(unwrap_const(guess), size(ttk)))
+    else
+        write_possibly_indexed_array!(guesses, ttk, guess, COMMON_NOTHING)
+    end
```
ChrisRackauckas, line 55: We should probably make array-shaped values happen in `write_possibly_indexed_array!`, but I'll accept this workaround to get the greater NeuralPDE.jl change in. But this should be documented as a hack / workaround to be removed.

ChrisRackauckas-Claude: Documented as a workaround in the docstring, with the removal tracked in https://github.com/SciML/ModelingToolkit.jl/issues/5147.

PR #5049, opened 2026-08-30
feat: accept array equations on the NonlinearProblem path

utkuyilmaz1903:
## Summary

- Accept unscalarized array equations in `NonlinearProblem` and `NonlinearLeastSquaresProblem` after `complete`.
- Assemble mixed scalar and array equations into one residual, with one row per array element.
- Preserve out-of-place `ForwardDiff.Dual` values from unknowns and tunable parameters.
- Zero the derivatives of array unknowns and of slices in the steady-state conversion.
- Preserve `FunctionWrapperSpecialize` and non-square residual prototypes.

## Constraints

- Array residuals require `mtkcompile` before using `jac = true` or `sparse = true`.
- The out-of-place allocator is restricted to nonlinear codegen. `DAEProblem{false}` Dual promotion is outside this PR.

## Tests

- 1D and 2D array equations
- Mixed scalar and array equations
- Array unknowns carrying an array equation, on top of #5131
- In-place, out-of-place, expression, and specialization paths
- Non-square least-squares residuals
- Unknown and parameter AD
- Constant-size generated code up to `n = 101`
- Existing #4983 DAE regression suite

## Compat

- New tests import `unwrap` from Symbolics so they run on older Symbolics.
- DiffEqBase floor is `7.18.1` so downgrade CI accepts already-unwrapped parameters in nested AutoDespecialize.

Related: SciML/MethodOfLines.jl#623.

Issue #5140, opened 2026-09-13
Enzyme SCC initialization regresses at concretize_initializeprob on Julia 1.13


ChrisRackauckas-Claude:
Enzyme reverse differentiation through SCC initialization regresses at https://github.com/SciML/ModelingToolkit.jl/commit/07e618174820bbe7a2847cdbfd1ba1af6d854cbc (https://github.com/SciML/ModelingToolkit.jl/pull/5062) under Julia 1.13.0. A clean immediate-parent checkout passes the same reduced workload with the same dependency manifest; the commit raises `EnzymeNoTypeError` while calling a `CacheWriter{Any}` through SciMLBase's despecialized-parameter barrier.

This was discovered in downstream SciMLBase CI and reproduced with unchanged SciMLBase source. It is separate from the array-remake change in that downstream PR.

### Reproducer

Save this as `core8_scc_enzyme_probe.jl`. It is reduced from SciMLSensitivity's existing `test/Core8/desauty_dae_mwe.jl`, preserving the original imports while keeping only SCC initialization differentiation. No MethodOfLines package is involved.

```julia
using Test
using ModelingToolkit, OrdinaryDiffEq
using ModelingToolkit: t_nounits as t, D_nounits as D
import SciMLStructures as SS
import SciMLSensitivity
import SciMLBase
using SymbolicIndexingInterface
using FiniteDiff
using ForwardDiff
using Tracker
using Enzyme
using Mooncake
using NonlinearSolve: NewtonRaphson

using InteractiveUtils
versioninfo()
@parameters a b c
@variables x(t) y(t) z(t)

eqs = [
    D(x) ~ a * x + y + z,
    0 ~ y^3 + y - b * x,
    0 ~ z^3 + z - c * y,
]

@mtkbuild sys = ODESystem(eqs, t)


@testset "Enzyme SCC initialization" begin
    prob = ODEProblem(sys, [x => 1.0], (0.0, 0.1), [a => -0.5, b => 2.0, c => 1.5], guesses = [y => 1.0, z => 0.5]; use_scc = true)
    iprob = prob.f.initialization_data.initializeprob
    itunables, irepack, _ = SS.canonicalize(SS.Tunable(), parameter_values(iprob))
    function enzyme_init_loss(t, iprob_)
        _, irepack_, _ = SS.canonicalize(SS.Tunable(), parameter_values(iprob_))
        iprob2 = remake(iprob_, p = irepack_(t))
[...]

Issue #5125, opened 2026-09-12
Flaky: downgrade Initialization `use_scc = true` gradient test returns 1e153 (Dual solve goes non-finite at t=0)


ChrisRackauckas-Claude:
`downgrade-mtkbase (Initialization)` (Downgrade Sublibraries workflow, Julia lts) failed once and then passed on a retrigger of the **same commit** (11ac9fb, branch of https://github.com/SciML/ModelingToolkit.jl/pull/5117, which does not touch initialization or AD code). Recording it so the next person who sees it does not chase their own diff.

## Failure

https://github.com/SciML/ModelingToolkit.jl/actions/runs/34704969119/job/103584112015

```
Initialization problem type is shared across models, use_scc = true: Test Failed at lib/ModelingToolkitBase/test/initializationsystem.jl:2313
  Expression: ≈((ForwardDiff.gradient(loss, [1.5]))[1], fd, rtol = 0.0001)
   Evaluated: 1.3172268582515041e153 ≈ -1.153893572602982 (rtol=0.0001)
ERROR: LoadError: Some tests did not pass: 690 passed, 1 failed, 0 errored, 12 broken.
```

Earlier in the same log the Dual-valued solve inside `loss` aborts on the first step: `At t=0.0, dt was forced below floating point epsilon 5.0e-324 ... ForwardDiff.Dual{ForwardDiff.Tag{...}}` with `All 8 state variables are non-finite (NaN/Inf)`, so the 1e153 is the derivative of garbage. The `use_scc = false` iteration in the same process passed.

## Passes

- Retriggered run, same commit, same resolved floors: https://github.com/SciML/ModelingToolkit.jl/actions/runs/34712340782/job/103603372204
- master tip f43b262, same floors: https://github.com/SciML/ModelingToolkit.jl/actions/runs/34704570510
- Locally on 11ac9fb with the floors the action resolved (NonlinearSolveBase 2.49.4, SciMLBase 3.48.0, OrdinaryDiffEqCore 4.15.1, StochasticDiffEqCore 2.2.1, Symbolics 7.37.0), Julia 1.10.11: full group `InitializationSystem Test | 691 12 703`, and the testset extracted into a standalone script six times (three on 1.10.11, three on 1.10.12), all gradients within 1e-8 of the finite-difference value. The values are not bit-identical across processes (fd drifts ~2e-8, ad ~1e-9), so the generated problem/solve is already session-dependent even when correct.

[...]

Issue #5063, opened 2026-09-04
ModelingToolkitBase QA lane red since #4832: JET typo mode reports 263 errors, 234 of them from Moshi `@match` codegen


ChrisRackauckas-Claude:
## Summary

The `lib/ModelingToolkitBase [QA] / Julia 1` lane of Sublibrary CI has been red on every `master` run since #4832 (`92c27bb9fd`, 2026-08-11) introduced `SciMLTesting.run_qa(ModelingToolkitBase; jet = true)`, i.e. `JET.test_package(ModelingToolkitBase; target_modules = (ModelingToolkitBase,), mode = :typo)`. The check reports **263 possible errors** and fails; before that commit the QA lane ran only `test/qa/jet.jl` (54 `@test_call`s) and the `report_package` check did not exist, so there is no MTK commit to bisect to.

Latest master run: https://github.com/SciML/ModelingToolkit.jl/actions/runs/33444273161/job/99660311927 (`e91f13e4`, JET 0.12.1, Julia 1.12.7: `JET-test failed ... 263 possible errors found`). Every earlier run that executed the job is red as well, e.g. https://github.com/SciML/ModelingToolkit.jl/actions/runs/31478008118 (the #4832 PR itself, merged red), https://github.com/SciML/ModelingToolkit.jl/actions/runs/32557851438/job/96996767096, https://github.com/SciML/ModelingToolkit.jl/actions/runs/33305672292/job/99250653964.

The root `tests / QA (julia 1)` lane is green because `test/qa/Project.toml` pins `JET = "0.9,0.10,0.11"` and analyses only the `ModelingToolkit` facade; `lib/ModelingToolkitBase/test/qa/Project.toml` has no JET compat, but the JET version is not the cause (see below).

## Local reproduction (clean `master` worktree, `22cb5b0a21`)

```bash
export JULIA_DEPOT_PATH="$HOME/sandbox/initwrap/depot:$HOME/.julia"
cd lib/ModelingToolkitBase/test/qa
julia +1.12 --startup-file=no --project=. -e 'using Pkg; Pkg.instantiate(); include("aqua.jl")'
```

```
Quality Assurance: JET-test failed at ~/.julia/packages/JET/vOTu0/src/JETBase.jl:1312
  Expression: (JET.report_package)(ModelingToolkitBase; toplevel_logger = nothing, target_modules = (ModelingToolkitBase,), mode = typo)
  ═════ 263 possible errors found ═════
[...]

PR #5148, opened 2026-09-13
Unify array-equation support behind a capability trait; fold in #5101 ODE support and MultiObjectiveOptimizationFunction

ChrisRackauckas-Claude:
> **Ignore until reviewed by @ChrisRackauckas.**

## Summary

Stacked on the merged #5131 (array *unknowns* via `flat_unknowns`); folds in the closed #5049 (array equations on the nonlinear path) and the open #5101 (array differential equations on the ODE path), and, as the last commit, the multi-objective optimization work from `ChrisRackauckas-Claude:multiobjective-optimization`. If this lands, #5101 and that branch are subsumed.

"Accepts array equations" is a named, per-constructor capability: `accepts_array_equations(constructor)` gates `__process_SciMLProblem`, and `ODEFunction`, `DAEFunction` and `NonlinearFunction` opt in. All three lower an array equation to one scalar row per element through the shared `array_residual_maker`; `full_equations` scalarizes array equations, so everything downstream of it (symbolic Jacobians, sparsity, `tgrad`, the initialization system) sees one row per residual without further changes.

### Review of the earlier revision and what changed

The earlier revision of this PR (through d834c1a) was reviewed in full. Everything in it was correct as far as tests and probes could tell; the following was simplified or extended on top:

- **The residual-form rewrite is gone.** `explicit_array_derivative_form` pattern-matched `D(u[2:4]) .- f ~ 0` back into `D(u[2:4]) ~ f` in `generate_rhs`, `calculate_massmatrix`, `full_equations` and `build_explicit_observed_function`. The scalar path accepts no such thing: `D(x) - f ~ 0` is a DAE and must go through `mtkcompile` or `DAEProblem`. Array equations now behave the same way (the ODE path rejects them with the same "nondifferentiated variables" error, and `DAEProblem` keeps accepting them, as `array_equation_dae.jl` tests). That removes ~50 lines and five call sites.
- **`jac = true` / `sparse = true` work for array residuals on `NonlinearFunction` and `DAEFunction`** instead of throwing. Since `full_equations` scalarizes, `calculate_jacobian` and `jacobian_sparsity` already produce one row per residual; the guards only hid that. Tested against a ForwardDiff Jacobian (nonlinear, dense and sparse, including a solve) and against the analytic stencil with the `γ` term (DAE, dense). Consequence: #5136 is not needed for the Jacobian itself. `sparse = true` on a DAE in residual form still errors inside `calculate_massmatrix` ("Only semi-explicit constant mass matrices"); that is pre-existing and equally true for scalar residual-form DAEs, so it is out of scope here.
- **`array_equation_dae.jl` was included twice** (InterfaceI on master, and again in InterfaceII by this PR). The duplicate is removed.
- **The array-residual DAE tests solve with `BrownFullBasicInit` again** rather than a hand-computed consistent `du0` under `NoInit()`, so solver-driven DAE initialization is exercised through the array residual. #5101 had switched to `NoInit` because `BrownFullBasicInit` tripped a `ParameterDespecializationWrapper` assertion on the downgrade job's DiffEqBase 7.15; the `DiffEqBase = "7.18.1"` floor this PR already carries is what fixes that, verified by running the test against a pinned DiffEqBase 7.18.1 (see below).
- Docs: the "use `mtkcompile` before `jac`/`sparse`" notes are removed since they no longer apply; the nonlinear tutorial note describes what array equations look like instead.

### Multi-objective optimization (last commit)

`OptimizationProblem` accepts `f::MultiObjectiveOptimizationFunction` in SciMLBase but MTK never produced one (`generate_cost` always consolidates to a scalar).

- `costs(sys)`: the unconsolidated objective vector (own costs, then the consolidated cost of each subsystem, namespaced). Exported and documented next to `cost`.
- `MultiObjectiveOptimizationFunction(sys)`: vector-valued objective over `costs(sys)`; `jac` builds the objective Jacobian; `hess` is a **vector of generated functions, one per objective**, which is the shape `OptimizationBase.instantiate_function` iterates (`[... for h in f.hess]`). The original branch generated one function returning a vector of matrices, which that consumer cannot use.
- The six constraint fields (`cons`, `cons_j`, `cons_h`, their prototypes, `cons_expr`) are assembled by one `generate_constraint_fields` shared with `OptimizationFunction` instead of a duplicated 25-line block.
- `OptimizationProblem(sys, op; multiobjective = true)` routes through it; `weights` is rejected in combination with it.
- All new public names (`costs`, the `generate_multiobjective_*` / `calculate_multiobjective_*` functions, the SciMLBase constructor) have docstrings and `@docs` entries.

### Not in scope

`generate_rhs` residual codegen is unchanged from #5049; OOP `Dual` promotion for `DAEProblem{false}`; array `constraints` in `generate_cons`; sparse Jacobians for residual-form DAEs (see above).

## Verification

Julia 1.12.7, ModelingToolkitBase test environment on this machine.

All commands run from `lib/ModelingToolkitBase` unless noted; `GROUP=… julia --project -e 'using Pkg; Pkg.test()'` for the groups.

| What | Result |
|---|---|
| `GROUP=InterfaceI` | `InterfaceI \| 1732 pass \| 5 broken \| 1737 \| 43m52s` — `Testing ModelingToolkitBase tests passed`. The 5 broken are pre-existing `@test_broken`s; this PR adds none. |
| `GROUP=InterfaceII` | every top-level testset green, `tests passed`: `Array Equation ODEProblem Test \| 51 \| 51`, `Array-equation Nonlinear \| 172 \| 172`, `NonlinearSystem Test \| 113 \| 113`, `Code Generation Test \| 65 \| 65`, `JumpSystem Test \| 4207 \| 4207`, … (2 pre-existing broken in `Optimal Control + Constraints`). |
| `test/array_equation_dae.jl` (InterfaceI member, rerun alone after the last test edit) | `array_equation_dae.jl \| 32 \| 32 \| 1m02s` |
| `test/optimization/multiobjective.jl` (new, Optimization group) | `optimization/multiobjective.jl \| 35 \| 35 \| 8m38s` |
| `test/optimization/optimizationsystem.jl` (shares `generate_constraint_fields`) | `126 pass \| 1 broken (pre-existing) \| 127 \| 3m22s` |
| `GROUP=QA` | JET level tests `54 \| 54`; Aqua: 20 of 21 pass. The one failure is `JET.report_package(...; mode = :typo)` with **234 reports**, all `local variable ##And#…/##Call#… may be undefined` from `Moshi.@match` expansions, none in code this PR touches. That check has never been green on master since it was enabled (#4832); a separate agent reproduced 234 on clean master with the same JET/Moshi/SymbolicUtils versions, reduced it to a Moshi+JET-only MWE, and confirmed 0 reports with the upstream fix https://github.com/Roger-luo/Moshi.jl/pull/93. Tracked in https://github.com/SciML/ModelingToolkit.jl/issues/5063; no new issue opened. |
| `BrownFullBasicInit` on the array-residual DAE at the compat floor | scratch env with `DiffEqBase` pinned to **7.18.1** (and separately 7.21.1): `retcode Success`, max error `3.0e-3` against the analytic heat solution, same as the test asserts. |
| Root package `GROUP=QA` (repo root) | `QA \| 40 \| 40 \| 4m35s`, `Testing ModelingToolkit tests passed` after adding the six new reexported names to `test/qa/qa.jl`'s allow-list (the first CI run failed only that check). |
| Runic (`--check --diff`) and `typos` over every changed file | clean. |
| Docs (`julia --project=docs docs/make.jl`, repo root) | builds through `Doctest`, `ExpandTemplates` (all `@example`/`@docs` blocks), `CrossReferences`, `CheckDocument` (`checkdocs = :exports`) and `RenderDocument`; only pre-existing `size_threshold_warn` warnings. The external `linkcheck` step was disabled for the local rerun after GitHub answered `429` to the `NEWS.md` link on the first run. |

**Not verified:** Julia LTS / `pre`; the downgrade CI job itself (only the 7.18.1 pin above); `Optimization` group's `dynamic_optimization.jl` / `test_infiniteopt.jl` (the Pyomo/CasADi pixi install fails on this machine's network, unrelated); downstream packages; GPU. `sparse = true` on residual-form DAEs is a pre-existing `calculate_massmatrix` limitation and is intentionally not covered.

**Judgment calls to push back on:** (1) dropping the residual-form ODE rewrite from #5101 (array `D(u) .- f ~ 0` now needs `DAEProblem`/`mtkcompile`, exactly like scalar); (2) `hess` on the multi-objective function is a vector of per-objective functions rather than one function returning a vector of matrices, chosen to match `OptimizationBase`; (3) `jac`/`sparse` guards removed instead of extended.

### CI on ee92d65 (first push) and e4f7d29 (second push)

Everything is green except the following, all of which reproduce on master's own CI and are not touched by this PR:

- `sublibrary-ci / lib/ModelingToolkitBase [QA] / Julia 1` and the root `tests / QA` JET step: the 234 Moshi `@match` reports, tracked in https://github.com/SciML/ModelingToolkit.jl/issues/5063 (the root `tests / QA` job additionally flagged the six new reexports; fixed by the last commit).
- `tests / InterfaceI` and `tests / InterfaceII` on Julia 1 and lts, `downgrade-mtkbase (InterfaceI|InterfaceII)`, `Downgrade Tests - InterfaceI`: every failure is inside the "Array unknowns on a `complete`d system" / "DAEProblem" testsets that https://github.com/SciML/ModelingToolkit.jl/pull/5131 added to `odesystem.jl`, `sdesystem.jl` and `nonlinearsystem.jl` (raw `u0`/`f`/`g` order comparisons across `complete` vs `mtkcompile`, and `symbolic_container(::IndexCache)` missing under the downgrade dependency set). Master's run for the #5131 merge commit https://github.com/SciML/ModelingToolkit.jl/actions/runs/34779646543 fails the same jobs; the previous master commit was green. Fixed separately: https://github.com/SciML/ModelingToolkit.jl/pull/5153 raises the `SymbolicIndexingInterface` floor to 0.3.47 (the version whose `getsym` no longer recurses into `parameter_index(::IndexCache, ::Int)`), and https://github.com/SciML/ModelingToolkit.jl/pull/5154 makes those tests compare through symbolic indexing instead of assuming `mtkcompile` keeps the array unknown's element order (with ModelingToolkitTearing loaded it reverses it). This PR's own `array_equation_dae.jl` passes on the downgrade job, i.e. `BrownFullBasicInit` on the array residual works at the DiffEqBase 7.18.1 floor in CI as well.
- `sublibrary-ci / lib/ModelingToolkitBase [Initialization] / Julia lts` on the second push only: `initializationsystem.jl` "Initialization problem type is shared across models, use_scc = true", `ForwardDiff.gradient(loss, [1.5])[1]` evaluates to `9.898834314793689e154`. The same job with identical dependency versions passed on the first push, and master's run for df574699 (https://github.com/SciML/ModelingToolkit.jl/actions/runs/34766850677) fails with the identical value, so it is a pre-existing intermittent failure, tracked in https://github.com/SciML/ModelingToolkit.jl/issues/5125. Root cause is upstream: the `remake`d initialization problem has Dual `u0` but a non-Dual-array `p`, so NonlinearSolve's Dual hooks do not fire and `default_nlsolve`'s Broyden iterates in Dual arithmetic, letting the partials grow to ~1e155 while only the value residual is checked for convergence.
- Downstream `SciMLSensitivity.jl/Core8/1/`: `Enzyme through init` in SciMLSensitivity's own `desauty_dae_mwe.jl` raises `EnzymeNoTypeError`. Fails on every master `Downstream.yml` run since https://github.com/SciML/ModelingToolkit.jl/pull/5062 (first bad job https://github.com/SciML/ModelingToolkit.jl/actions/runs/34695771592/job/103558875060, identical dependency set as the last good one) and on SciMLSensitivity's own CI against released MTK 11.43.1; tracked in https://github.com/SciML/ModelingToolkit.jl/issues/5140.
- Docs `build`: `makedocs` stops at `[:linkcheck]` because `https://docs.sciml.ai/PDEBase/stable/interface/` (a link https://github.com/SciML/ModelingToolkit.jl/pull/5149 added on master) answers 403; master's own docs job for that commit fails the same way. Everything before linkcheck (doctests, all `@example`/`@docs` blocks, cross-references, `checkdocs`) passes on CI, and the full render passes locally with linkcheck off (see Verification).
- `downgrade-mtkbase (Optimization) / test`: cancelled after the 6-hour limit on both pushes, exactly as on master's run for 444ccfb0; the group is known to be slow (see `test_groups.toml`) and the downgrade dependency set makes it slower still.

## Links

- Subsumes https://github.com/SciML/ModelingToolkit.jl/pull/5101 and https://github.com/SciML/ModelingToolkit.jl/pull/5049 (closed); stacked on https://github.com/SciML/ModelingToolkit.jl/pull/5131 (merged)
- Makes https://github.com/SciML/ModelingToolkit.jl/pull/5136 unnecessary for dense DAE Jacobians
- Multi-objective source branch: https://github.com/ChrisRackauckas-Claude/ModelingToolkit.jl/tree/multiobjective-optimization
- Separate one-line fix found while building the docs locally: https://github.com/SciML/ModelingToolkit.jl/pull/5152
- Master regressions from #5131 found via this PR's CI, fixed separately: https://github.com/SciML/ModelingToolkit.jl/pull/5153 and https://github.com/SciML/ModelingToolkit.jl/pull/5154
- Pre-existing QA JET failure: https://github.com/SciML/ModelingToolkit.jl/issues/5063

🤖 Generated with [Claude Code](https://claude.com/claude-code) (model: claude-fable-5-1); the earlier revision of this PR was generated with Devin CLI (model: SWE-2 Max).

https://claude.ai/code/session_01VorhgJYhDasyVLpGNrviga


ChrisRackauckas:
I finally like it, and there's a lot riding on it downstream so I'm going for it, but @AayushSabharwal please take a look and if any clean ups are required we can modify post.

PR #5158, opened 2026-09-14
Fix ForwardDiff gradients through nonlinear initialization solves

### LINKED ISSUE #5125: Flaky: downgrade Initialization `use_scc = true` gradient test returns 1e153 (Dual solve goes non-finite at t=0)

ChrisRackauckas-Claude:
## Summary

Fixes https://github.com/SciML/ModelingToolkit.jl/issues/5125.

`ForwardDiff` gradients through ODE initialization could come back silently
corrupted (e.g. `1.3172268582515041e153` instead of `-1.1538935708`). Carrying
`Dual` values into the initialization nonlinear solve is unreliable: quasi-Newton
rank-one updates divide by `Dual`-valued quantities, so partials can grow without
bound while the value residuals still converge
(https://github.com/SciML/NonlinearSolve.jl/issues/1283), and nested-`Dual`
buffer writes can throw outright.

- `update_initializeprob!` now strips `Dual` values from `u0`, `p` and the
  residual prototypes so the nonlinear solver runs on the value problem with an
  honest retcode. The residual callable is marked with `InitSensitivityData`,
  which carries the dual-valued parameter object for later reconstruction.
- The initialization maps run `correct_initialsol_duals` on the solution before
  evaluating: it re-attaches the exact parameter sensitivities
  `-J_u \ ∂f/∂θ` via the implicit function theorem and rebuilds a `Dual`-valued
  solution and parameter object. The correction is injected inside the maps —
  the `initializeprobpmap` closures, the `InitializationMap` inner callable, and
  the FullSpecialize generated bodies — so `initializeprobmap` /
  `initializeprobpmap` keep their existing `RuntimeGeneratedFunction` /
  `InitializationMap` types.
- `SCCNonlinearProblem` is handled blockwise: each block's `explicitfuns!` cache
  writes are replayed on dual and primal parameter objects so sensitivities of
  upstream block solutions propagate into downstream blocks; nonlinear and
  homotopy blocks get the same IFT correction and linear blocks are re-solved
  exactly on the dual `A`/`b`.
- The ChainRulesCore extension applies the same strip in its
  `update_initializeprob!` method.

## Test evidence

New regression testset `"Initialization gradient through quasi-Newton solves,
use_scc = $use_scc"` in `lib/ModelingToolkitBase/test/initializationsystem.jl`
(`chain_dae_problem(3, 1.0)`, forced `Broyden()` initialization, three gradient
trials plus mechanism assertions on `update_initializeprob!` /
`correct_initialsol_duals`).

On the unfixed code the testset fails deterministically — `u0` and `p.tunable`
remain `Dual`-valued after `update_initializeprob!`, and the Dual-valued solve
throws a tag-mixing `MethodError` inside the nonlinear polyalgorithm:

```
Initialization gradient through quasi-Newton solves, use_scc = false: Test Failed
  Expression: eltype(state_values(updated)) <: Float64
Initialization gradient through quasi-Newton solves, use_scc = false: Test Failed
  Expression: eltype((parameter_values(updated)).tunable) <: Float64
  ... MethodError(Float64(::Dual)) in the initialization solve ...
Initialization gradient through quasi-Newton solves, use_scc = false | 3  2  1  6
```

With the fix:

```
Initialization gradient through quasi-Newton solves, use_scc = false | 9  9
Initialization gradient through quasi-Newton solves, use_scc = true  | 9  9
```

`ForwardDiff.gradient` now returns `-1.1538935709` vs a finite-difference
reference of `-1.1538935687` (issue expectation `-1.1538935708`).

#### Test plan

- [x] `GROUP=Initialization julia --project=lib/ModelingToolkitBase -e 'using Pkg; Pkg.test()'`:
      **753 pass, 0 fail, 12 broken** (includes the FullSpecialize map-type
      assertions; the `Broken` entries are pre-existing `@test_broken`s).
- [x] Same testset under full `ModelingToolkit` (`use_scc = true` produces a real
      `SCCNonlinearProblem`): 9/9, with blockwise sensitivities cross-checked
      against hand-computed IFT values including cross-block propagation.
- [x] Runic format check clean on all touched files; `typos` clean.
- [x] `GROUP=QA`: `all_qualified_accesses_are_public` passes with the new
      non-public accesses (`ForwardDiff` `Partials`/`jacobian`/`npartials`/
      `partials`/`value`, `NonlinearSolveBase.get_raw_f`, `SciMLBase.strip_solution`)
      registered in `test/qa/aqua.jl`'s `NONPUBLIC_QUALIFIED_ACCESSES` — none of
      these have a public spelling in the resolved dependency versions.
- [ ] `GROUP=QA` JET `report_package(mode = :typo)` fails with "234 possible
      errors" — **pre-existing on master**, already tracked as
      https://github.com/SciML/ModelingToolkit.jl/issues/5063 (all findings are
      Moshi `@match` codegen false positives, pending upstream Moshi.jl#93);
      none are in the code this PR adds.
- [ ] Docs build not run locally — the changes only add internal docstrings (no
      `@docs` entries; `checkdocs = :exports`).

🤖 Generated with [Devin](https://devin.ai) — harness: Devin CLI 3000.10.21, model: SWE-2 Max
Session (local transcript): /home/crackauc/.local/share/devin/cli/summaries/history_dd4c5ce4b64440e8.md


ChrisRackauckas-Claude:
Upstream update: https://github.com/SciML/NonlinearSolve.jl/pull/1293 fixes the solver-side root cause. `DualAbstractNonlinearProblem` only matched when `p` was literally a `Dual` or `AbstractArray{<:Dual}`, so duals inside structured parameter containers (like `DespecializedParameters`) bypassed the ForwardDiff path and the quasi-Newton internals ran in `Dual` arithmetic — which is how `Success` could come back with ~1e153 partials.

Validated: master MTK + the NonlinearSolve branch (dev'd) gives `ForwardDiff.gradient` through `initprob` agreeing with finite differences to ~1e-8 on **both** `use_scc = false` (AD `-1.0778858595152796` vs FD `-1.0778858785087664`) and `use_scc = true` (AD `-1.077885859515277` vs FD `-1.0778858833493388`), without this PR's workaround.

Once a NonlinearSolve release containing it is out and MTK's compat allows it, this PR is no longer needed for correctness — though the regression test added here may still be worth keeping (possibly rebased to drop the workaround).

Issue #169, opened 2019-08-09
Solving Lorenz system errors


jamblejoe:
Hi,
The Lorenz system, as on the main page of ModelingToolkit, errors if one tries to solve it with the standard solver:
```
@parameters t σ ρ β
@variables x(t) y(t) z(t)
@derivatives D'~t

eqs = [D(x) ~ σ*(y-x),
       D(y) ~ x*(ρ-z)-y,
       D(z) ~ x*y - β*z]

de = ODESystem(eqs)
f = ODEFunction(de, [x,y,z], [σ,ρ,β])
u0 = [19.,20.,50.]
ode_prob = ODEProblem(f, u0, (0., 10.))
```
works fine but
`sol = solve(ode_prob)`
 gives the following error

> BoundsError
Stacktrace:
 [1] getindex at ./number.jl:78 [inlined]
 [2] ##365(::Array{Float64,1}, ::Array{Float64,1}, ::Int64, ::Float64) at /home/goran/.julia/packages/ModelingToolkit/MDrkF/src/utils.jl:55
 [3] macro expansion at /home/goran/.julia/packages/ModelingToolkit/MDrkF/src/utils.jl:102 [inlined]
 [4] macro expansion at ./none:0 [inlined]
 [5] fast_invokelatest at ./none:0 [inlined]
 [6] out_f_safe at /home/goran/.julia/packages/ModelingToolkit/MDrkF/src/systems/diffeqs/diffeqsystem.jl:226 [inlined]
 [7] ODEFunction at /home/goran/.julia/packages/DiffEqBase/LCorD/src/diffeqfunction.jl:209 [inlined]
[...]

PR #5163, opened 2026-09-17
Fix `solve(::SteadyStateProblem)` regressions from the deferred SCC lowering

### LINKED ISSUE #5162: Regression in 11.44.0: `solve(::SteadyStateProblem)` fails in the new SCC lowering (#5155) — `bound_parameters` typeassert, and initialization equations carried into a time-independent system
> _This issue was written by Claude Code (an AI agent) on behalf of the account filing it. Both reproducers below were executed and verified in the environment listed under **Versions** — they are not untested sketches._

#5155 (`steady_state_sccprob`, released in v11.44.0) breaks `solve(::SteadyStateProblem)` for two common model shapes. Both reproduce with ModelingToolkit + NonlinearSolve only.

The lowering is documented as deferred — *"building it requires a full `mtkcompile` that most `SteadyStateProblem` consumers never need"* — but `SciMLBase`'s `NonlinearProblem(::SteadyStateProblem)` conversion (`nonlinear_problems.jl:256`) calls it on **every** solve, so every steady-state solve now runs it.

## MWE 1 — `bound_parameters` typeassert

Trigger: the model is `complete`d before it is simplified.

```julia
using ModelingToolkit, NonlinearSolve

@parameters α=1.3 β=0.9 γ=0.8 δ=1.8
@variables x(t)=3.1 y(t)=1.5

model = complete(System([D(x) ~ α * x - β * x * y,
                         D(y) ~ -δ * y + γ * x * y], t; name = :lotka))

sys = mtkcompile(model)
solve(SteadyStateProblem(sys, []))
```

```
ERROR: TypeError: in typeassert, expected ModelingToolkitBase.ParameterBindingsGraph, got a value of type Nothing
  [1] bound_parameters      @ ModelingToolkitBase/src/systems/abstractsystem.jl:1716 [inlined]
  [2] NonlinearSystem(sys::System)
                            @ ModelingToolkitBase/src/systems/system.jl:1386
  [3] _sccprob(prob)        @ ModelingToolkit/src/problems/sccnonlinearproblem.jl:830
  [4] NonlinearProblem      @ SciMLBase/src/problems/nonlinear_problems.jl:256 [inlined]
```

Dropping the `complete` makes it pass.

**Why:** `_sccprob` walks the parent chain to its root and calls `NonlinearSystem(ssys)`. That constructor calls `bound_parameters` whenever the system is complete (`system.jl:1385-1387`):

```julia
if iscomplete(sys)
    append!(new_ps, collect(bound_parameters(sys)))
end
```

and `bound_parameters` typeasserts the bindings graph (`abstractsystem.jl:1716`). But `complete` deliberately stores a parent that has no graph — `abstractsystem.jl:667-668` sets `parent = complete(sys; split = false, flatten = false)`, and the `flatten = false` branch at line 699 does `@set! sys.parameter_bindings_graph = nothing` ("reduce the potential for outdated information"). So for a model completed before `mtkcompile`, the walk lands on a system that is complete and has `parameter_bindings_graph === nothing` by construction — not an edge case.

## MWE 2 — initialization equations rejected

Trigger: the model carries `initialization_eqs` referring to an unknown. (Not `complete`d here, so the walk stops at an incomplete root, `bound_parameters` is skipped, and it gets one step further.)

```julia
using ModelingToolkit, NonlinearSolve

@parameters α=1.3 β=0.9 γ=0.8 δ=1.8
@variables x(t)=3.1 y(t)=1.5

model = System([D(x) ~ α * x - β * x * y,
                D(y) ~ -δ * y + γ * x * y], t;
    initialization_eqs = [x ~ 3.1], name = :lotka)

sys = mtkcompile(model)
solve(SteadyStateProblem(sys, []))
```

```
ERROR: ArgumentError: Initialization equations for time-independent systems can only contain
parameters. Found Set([x(t)]) in x(t) ~ 3.1. If the equations refer to the initial guess for
unknowns, use the `Initial` operator.
  [3] generate_initializesystem_timeindependent
        @ ModelingToolkitBase/src/systems/nonlinear/initializesystem.jl:374
  [7] InitializationProblem @ ModelingToolkitBase/src/problems/initializationproblem.jl:93
 [19] SCCNonlinearProblem   @ ModelingToolkit/src/problems/sccnonlinearproblem.jl:0
 [25] _sccprob(prob)        @ ModelingToolkit/src/problems/sccnonlinearproblem.jl:835
```

**Why:** `NonlinearSystem` copies initialization equations across verbatim (`system.jl:1392`):

```julia
initialization_eqs = initialization_equations(sys), name = nameof(sys),
```

`x ~ 3.1` is valid on the time-dependent model, but the lowering's target is time-independent, where only parameters are allowed. The equations need translating (to `Initial(x)`) or dropping for the residual system.

## Versions

Julia 1.13.0, ModelingToolkit v11.44.0, ModelingToolkitBase v1.74.0, NonlinearSolve v4.32.0, SciMLBase v3.55.0.

Both MWEs return `ReturnCode.Success` on ModelingToolkit v11.43.1 — **with ModelingToolkitBase v1.74.0 still resolved**. So the regression is the MTK-side `steady_state_sccprob` method, not the MTKBase ctor tagging: the MTKBase stub (`problems/odeproblem.jl:238`) returns `nothing` and nothing is lowered.

---
_Generated by [Claude Code](https://claude.ai/code)_

ChrisRackauckas-Claude:
## Summary

Fixes https://github.com/SciML/ModelingToolkit.jl/issues/5162 — the deferred SCC lowering stored on `SteadyStateProblem` by #5155 (released in 11.44.0) broke `solve(::SteadyStateProblem)` in several ways:

- **`NonlinearSystem(sys)` used the child-merging accessors** (`equations`, `unknowns`, `bindings`, …) while also recursively converting and keeping subsystems — double-counting everything on hierarchical systems (first failure: a duplicate `t => Inf` binding). It now reads local fields via `get_*` accessors, and `iv => Inf` is only injected at the top level since subsystems inherit it as a domain binding.
- **`bound_parameters` typeasserted** on complete systems whose `parameter_bindings_graph` was invalidated by `complete(; flatten = false)` (the parent snapshot `complete` records). Bound parameters remain in `ps` there, so the graph lookup is now guarded.
- **Initialization equations were copied untranslated** into the time-independent residual, so `[x ~ 3.1]` was rejected for missing `Initial`. The new `steady_state_initialization_eqs` wraps bare unknowns/observables in `Initial`, zeroes derivatives (including `Initial(D(x))`, stored as `Initial(xˍt)`), and drops the resulting trivial identities.
- **Bound parameters leaked into the operating point** in `calculate_op_from_u0_p`, colliding with the lowered system's own `bindings` ("Cannot merge without overriding"). They are now skipped.
- **Empty SCC schedules**: a residual that `mtkcompile` reduces entirely to `observed` produced `var_sccs == []`; the lowering is now a stateless `NonlinearProblem` (with `u0 = Float64[]`) instead of an invalid `SCCDecomposition` call.
- **Variable bindings rejected by the time-independent constructor**: a completed system's variable defaults (e.g. `z(t) => z0`) live in `bindings`, which `System` rejects when `iv === nothing`. `NonlinearSystem` now moves them to `initial_conditions` — mirroring `convert_bindings_for_time_independent_system`, but on the local `get_*` fields so the hierarchical fix is preserved. (Surfaced by SteadyStateDiffEq 2.17's `__solve_scc_lowering`, which invokes the deferred lowering on `DynamicSS` solves.)

**Companion PR**: the `solve(::SteadyStateProblem)` dispatch that routes a materialized `SCCNonlinearProblem` through its own `solve` (bare `solve` and `solve(prob, nlalg)` previously died on `prob.u0`) moved to SteadyStateDiffEq per review — SciML/SteadyStateDiffEq.jl#169. This PR's `solve(ssprob)` tests import SteadyStateDiffEq and need an SSD release containing that PR; everything else here is self-contained.

## Test plan

Regression coverage added in `test/scc_nonlinear_problem.jl` (new `SteadyStateProblem SCC lowering regressions` testset: `complete`-before-`mtkcompile`, `initialization_eqs` with bare unknowns, bound parameters, hierarchical systems, multi-SCC `solve`/`solve(alg)`, empty-schedule residual) and `lib/ModelingToolkitBase/test/nonlinearsystem.jl` (invalidated bindings graph, init-eq translation, hierarchical bindings, variable-bindings→initial-conditions).

Verified locally:
- `test/scc_nonlinear_problem.jl`: all pass with the SteadyStateDiffEq companion branch dev'd in. The new tests discriminate — on unfixed code they fail with the `bound_parameters` TypeError, verbatim untranslated init equations, and `FieldError: type SCCNonlinearProblem has no field u0`.
- `lib/ModelingToolkitBase/test/nonlinearsystem.jl`: 122/122 pass.
- The exact downstream `SteadyState` testset from `lib/ModelingToolkitBase/test/sciml_problem_inputs.jl` (the CI failure): 512/512 with SteadyStateDiffEq 2.17.0 + this fix.
- `test/qa/qa.jl`: Quality Assurance 20/20.
- `runic` and `typos` clean. Docs build not run — no docstring/`docs/`/public-API changes.

Known remaining gap (unchanged on master, not made worse here): `init(::SteadyStateProblem)` on an SCC-lowered system gets past conversion but then hits `prob.u0` — no `init(::SCCNonlinearProblem)` exists upstream (there is no SCC integrator type); needs follow-up in SCCNonlinearSolve/SciMLBase.

🤖 Generated with [Devin](https://devin.ai) (harness: Devin CLI 3000.10.31, model: SWE-2 Max); local session, no shareable URL — transcript summary at `/home/crackauc/.local/share/devin/cli/summaries/history_d2dbf511181844c4.md`.


ChrisRackauckas-Claude:
CI triage on the first run, plus the fix now pushed in `2f765c4`.

**Real failure (fixed): `tests/SymbolicIndexingInterface`**

SteadyStateDiffEq 2.17 added `__solve_scc_lowering`, which invokes the deferred `lowered_problem` closure inside `DynamicSS` solves as well. That exercised `NonlinearSystem(sys)` on a system carrying a *variable* binding (`Z(t) => Z0`), which the `System` constructor rejects when `iv === nothing`. This is a latent incompatibility on master — `NonlinearSystem(osys)` throws the same `ArgumentError` on `0e0bd078` — newly surfaced by the dependency bump.

Fix: `NonlinearSystem(sys)` now moves variable bindings into `initial_conditions` before constructing the time-independent system, mirroring `convert_bindings_for_time_independent_system` but on the local `get_*` fields (the recursive `bindings(sys)` would re-fold subsystem data, reintroducing the hierarchical double-counting fixed earlier in this PR).

Verified locally with SteadyStateDiffEq 2.17.0: the exact downstream `SteadyState` testset from `sciml_problem_inputs.jl` (512 `remake` + `DynamicSS` + `EnsembleProblem` variants) passes 512/512.

**Pre-existing / environmental (not from this diff):**
- `tests/QA` + `downgrade-mtkbase/QA`: Aqua `persistent_tasks` fails to locate `ChangesOfVariables` (a `LogExpFunctions` dep) — manifest/env issue on Julia 1.13.
- `sublibrary-ci MTKBase [QA]`: JET reports 234 `##And#NNN may be undefined` errors in Moshi-generated code — identical failure on master `52a3cf57`.
- `downgrade-mtkbase/Initialization`: ForwardDiff gradient returns `9.9e154` — the identical failure appears on the unrelated `struct-support` PR's downgrade run.
- `build` (docs): linkcheck got a 403 from docs.sciml.ai — external flake.
- `NeuralPDE/NNPDE2`, `SciMLSensitivity/Core8`: `EnzymeRuntimeActivityError` / `EnzymeNoTypeError` — unrelated AD issues.
- `SciMLBase` downstream: the `bound_parameters` typeassert this PR fixes — it runs against released MTK 11.44.0.

**Known remaining gap:** `init(::SteadyStateProblem)` on SCC-lowered systems. On master it dies at the `bound_parameters` typeassert during `NonlinearProblem` conversion; with this PR the conversion succeeds but `init` then hits `prob.u0` on the `SCCNonlinearProblem` — there is no `init(::SCCNonlinearProblem)` (no SCC integrator type) anywhere upstream, so this needs a follow-up in SCCNonlinearSolve/SciMLBase rather than a local workaround. `solve` paths (the subject of #5162) are all covered.

🤖 Generated with [Devin](https://devin.ai) (model: SWE-2 Max) — local session, no shareable URL; transcript summary at `/home/crackauc/.local/share/devin/cli/summaries/history_d2dbf511181844c4.md`

src/problems/sccnonlinearproblem.jl
```diff
@@ -835,3 +856,34 @@ function MTKBase.steady_state_sccprob(sys::System, op; kwargs...)
...
+    if alg !== nothing &&
+            !(alg isa Union{SciMLBase.AbstractNonlinearAlgorithm, SCCNonlinearSolve.SCCAlg})
+        return invoke(
+            solve, Tuple{SciMLBase.AbstractNonlinearProblem, Vararg{Any}},
+            prob, args...; kwargs...
+        )
+    end
+    # Concretize like `NonlinearSolveBase.solve_up` (evaluates `u0(p, Inf)`-style
+    # `u0`s, promotes integer `u0`s, applies `u0`/`p` solve kwargs).
+    _prob = SimpleNonlinearSolve.NonlinearSolveBase.get_concrete_problem(prob; kwargs...)
```
ChrisRackauckas: This should be in the solver SteadyStateDiffEq.

ChrisRackauckas-Claude: Moved — now in SciML/SteadyStateDiffEq.jl#169, pushed here as `a1cf8e0`. The interception forwards the materialized `SCCNonlinearProblem` to its own `solve` dispatch and `invoke`s the upstream path for everything else (`DynamicSS`, ODE algs, non-SCC lowerings), so it composes with 2.17's `__solve_scc_lowering` rather than duplicating it. SSD can't name `SCCAlg` (deliberately no SCCNonlinearSolve dep), so the gate is on `AbstractNonlinearAlgorithm` + the lowering's type.

One consequence worth noting: bare `solve(ssprob)` / `solve(ssprob, ::NewtonRaphson)` on an SCC-lowered problem now requires SteadyStateDiffEq loaded — `scc_nonlinear_problem.jl` gained the import (it's already a test dep). If you'd rather it live in `SCCNonlinearSolve` (owns `SCCNonlinearProblem`, fixes dispatch for all callers without SSD), that means `__solve(::SCCNonlinearProblem, ::Nothing)` plus intersection methods for each NLB algorithm family to stay unambiguous — doable, but the `solve`-level interception avoids the ambiguity surface entirely.

Sequencing: this branch's `solve(ssprob)` tests need the SSD release once #169 merges.

🤖 Generated with [Devin](https://devin.ai) (model: SWE-2 Max)

PR #5165, opened 2026-09-18
Export `ap_var` and document the property-access syntax for analysis points

baggepinnen:
## `ap_var`

`ap_var` returns the variable of an analysis-point endpoint. The endpoints of a point, `ap.input` and `ap.outputs`, are connector systems in a connector-based model and the connected variables themselves in a causal one, and `ap_var` is the only accessor that covers both cases. It is needed whenever the variable behind a point is required, for example when assembling the operating point of a linearization taken at that point:

```julia
ap = sys.plant_input
u = ap_var(only(ap.outputs))
```

It was previously internal, so downstream code had to reach it through `parentmodule(ModelingToolkit.AnalysisPoint).ap_var`. This PR exports it from ModelingToolkitBase, which re-exports it from ModelingToolkit, and rewrites its docstring to document both methods, the meaning of `ap.outputs` against `ap.input`, and the namespacing of the returned variable. It is added to the linearization and analysis section of the API reference.

## Analysis-point syntax in the documentation

The docstrings and documentation now refer to an analysis point by property access on the system in which it was created, `sys.plant_input`, rather than by its `Symbol` name, `:plant_input`. The `Symbol` form is retained for backwards compatibility and is documented as the old style that will be deprecated. The docstrings also state that the system must be the unsimplified and uncompleted model, since both `mtkcompile` and `complete` remove the analysis points, and that a point belonging to a subsystem is reached through the hierarchy, `sys.inner.plant_input`.

Updated: `AnalysisPoint`, `ap_var`, `open_loop`, `generate_control_function`, `get_sensitivity`, `get_comp_sensitivity`, `get_looptransfer`, the three corresponding `_function` variants, `get_linear_analysis_function`, `linearization_ap_transform`, `isolate_subsystem`, the shared `loop_openings` documentation fragment, the linear analysis tutorial and the `isolate_subsystem` example in the API reference.

The `Symbol` argument at the point of creation, `connect(C.output, :plant_input, P.input)`, is unchanged; the name is how the point is created.

## Verification

The documentation examples that were changed were executed against this branch and produce the same results as the `Symbol` form:

```
matrices_S.A = [-2.0;;]
matrices_T.D = [0.0;;]
matrices_L.A = [-1.0;;]
matrices_PS.A = [-2.0;;]
isolate_subsystem(closed_loop, closed_loop.plant_input, closed_loop.plant_output) -> (true, true)
```

No behavior is changed and no test is modified; the property-access form is already covered by the tests in `lib/ModelingToolkitBase/test/analysis_points.jl`.


AayushSabharwal (approved):

baggepinnen:
@AayushSabharwal would you mind merging if okay, I don't understand the wall of red CI here 😅

PR #5172, opened 2026-09-20
Reject unknowns left in the remainder by calculate_A_b

ChrisRackauckas-Claude:
## Summary

`calculate_A_b` builds `A`/`b` by running `Symbolics.linear_expansion` over each row's residual for every unknown in turn, then trusts the reported `islinear` to mean the leftover remainder is unknown-free. That trust is misplaced: `linear_expansion` only promises `a*x + b == t`, so a remainder can still *syntactically* contain an unknown whose coefficients cancelled during expansion — e.g. `x`-terms hidden inside an opaque call argument (`JuliaSymbolics/Symbolics.jl#1995`).

`b` is then compiled into `update_b!(b, p)`-style functions that receive parameters but **no unknowns**, so a remainder still referencing an unknown produces broken generated code — observed as `UndefVarError` on a bare namespaced array symbol (`pipe₊m`) when solving an `SCCNonlinearProblem` initialization whose `LinearProblem` sub-SCCs contain such a row.

This change validates the final remainder of each row against the full unknown set (extended to cover the parent array of element unknowns, so a bare whole-array reference is also caught). A contaminated remainder is reported through the existing `NotAffineError`/`nothing` path, which routes `LinearProblem`/`SCCNonlinearProblem` callers to a nonlinear solve instead of emitting bad code.

### Reproducer

```julia
using ModelingToolkitBase, Symbolics
import SymbolicUtils as SU
@variables x
@parameters h w c (f::Any)(..)
sys = complete(System(
    [0 ~ -10x + f(-h * x + ifelse(c > 0, h * x, h * x) + w)],
    [x], [h, w, c, f]; name = :remainder_unknowns))
ModelingToolkitBase.calculate_A_b(sys)
```

On current master this returns `(A, b)` with `b = f(ifelse(0 < c, h*x, h*x) + w - h*x)` — `x` leaks into `b`. With this change it throws `NotAffineError` (or `calculate_A_b(sys; throw = false)` returns `nothing`).

A regression test asserting the invariant (`nothing` or an `x`-free `b`) is included; it passes on this branch and would fail on master once `res !== nothing` reaches the second disjunct.

### Context

- Upstream Symbolics fix for the expansion itself: JuliaSymbolics/Symbolics.jl#1995. The two are complementary — this check is a defense-in-depth guard for a generated-code contract that `calculate_A_b` cannot currently verify.
- Real-world trigger: `InitializationProblem(...; use_scc = true)` on a system mixing whole-array `Initial(...)` parameters with scalarized unknowns — the linear SCC's symbolic `b` retained `(pipe₊m)[2,1]` inside an opaque call, and codegen emitted `pipe₊m` bare.

## Test plan

- [x] `test/linearproblem.jl` — all testsets pass locally (76/76), including the new `` `calculate_A_b` does not put unknowns in `b` `` testset
- [x] Verified the new check is not vacuous: without it, `b` contains `x`; with it, `NotAffineError`/`nothing`
- [x] End-to-end: the downstream `use_scc=true` initialization that previously threw `UndefVarError` now solves (`retcode=Success`) via the nonlinear fallback

🤖 Generated with Devin CLI (model: swe-2-max); session: local transcript /home/crackauc/.local/share/devin/cli/summaries/history_063c47f16952470b.md


AayushSabharwal:
No, just rely on the fix in https://github.com/JuliaSymbolics/Symbolics.jl/pull/1995

PR #5176, opened 2026-09-21
Mark the nonnumeric portion of MTKParameters as inactive/non-differentiable and move diffcache buffers in `caches`

SebastianM-C:
> **Stacked on #5178**, which moves the `add_diffcache` buffers into the `caches` portion. Only the commits from "Mark the nonnumeric portion of `MTKParameters` inactive for Enzyme" onwards belong to this PR; the layout commits underneath will disappear once #5178 is merged and this branch is rebased. Kept as a draft while an interaction between the inactive rule and multithreading is investigated.
>
> _Note added by Claude Code on Sebastian's behalf._

## Checklist

## Additional context

For systems that carry large non-differentiable buffers, like interpolation data, Enzyme would not be able to tell that it does not need to keep track of that and would shadow the non-numeric portion when not needed. Depending on how large the data structures are, the performance penalty can be negligible or a large portion.

This PR makes it clear that the non-numeric section of the `MTKParameters` is not intended to be differentiated, so Enzyme can safely ignore that.

In the process of writing this I found that the caches created via `add_diffcache` were stored in the non-numeric portion, so I got some help from Claude to move them to the `caches`, before the SCC buffers. In this way we preserve the correctness and also have a more intuitive location for those caches. (@AayushSabharwal was there a particular reason to have those caches in the non-numeric section?)

I also documented the new behavior to make the transition clear. The tests of the downstream consumers do not seem to show any regressions.


ChrisRackauckas-Claude (approved):

ChrisRackauckas (approved):

ChrisRackauckas:
>  (@AayushSabharwal was there a particular reason to have those caches in the non-numeric section?)

I think that was always a mistake? It was made exactly for these caches and separated for this reason, since caches need to be specifically handled for correct differentiation.

SebastianM-C:
We also need a compat bump for MTK since MTKBase would land with the new layout. Also found and fixed a SII regression

AayushSabharwal:
Yeah it diffcaches in nonnumerics is by accident. I just created the symbolic parameters and let `IndexCache` do its thing, which landed them there.

SebastianM-C:
It turns out that marking the non-numeric portion as inactive under Enzyme is not sound. This should probably be closed, I'll add a more detailed write-up before I close.

